PR #15, opened 2016-09-28
Fix/refactor

mspector:
adds proper error handling to OpenReviewPy

### DIFF
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -121,7 +121,6 @@ class Client(object):
         response = self.__handle_response(response)
 
         for g in response.json()['groups']:
-            g = self.get_group(g['id']).to_json()
             groups.append(Group.from_json(g))
 
         groups.sort(key=lambda x: x.id)
@@ -152,7 +151,7 @@ class Client(object):
         for i in response.json()['invitations']:
             invitation = Invitation.from_json(i)
             invitations.append(invitation)
-            invitations.sort(key=lambda x: x.id)
+        invitations.sort(key=lambda x: x.id)
         return invitations
 
 
@@ -183,7 +182,6 @@ class Client(object):
         for n in response.json()['notes']:
             note = Note.from_json(n)
             notes.append(note)
-        notes.sort(key=lambda x: x.forum)
         return notes
 
     def exists(self, groupid):
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -10,12 +10,6 @@ from Crypto.Hash import HMAC, SHA256
 class OpenReviewException(Exception):
     pass
 
-class GroupNotFound(OpenReviewException):
-    def __init__(self, value):
-        self.value = value
-    def __str__(self):
-        return repr(self.value)
-
 class Client(object):
 
     def __init__(self, baseurl=None, process_dir='../process/', webfield_dir='../webfield/',username=None,password=None):
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -80,19 +80,7 @@ class Client(object):
 
         g = response.json()['groups'][0]
         
-        group = Group(g['id'],
-            cdate = g.get('cdate'),
-            ddate = g.get('ddate'),
-            writers = g.get('writers'),
-            members = g.get('members'), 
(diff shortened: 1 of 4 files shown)


openreview/openreview.py
```diff
@@ -17,7 +18,7 @@ def __init__(self, baseurl=None, process_dir='../process/', webfield_dir='../web
             try:
                 self.baseurl = os.environ['OPENREVIEW_BASEURL']
             except KeyError:
-                self.baseurl = 'http://localhost:3000'
+                self.baseurl = 'http://openreview.net'
```
mspector: this is a significant change, from localhost --> openreview

I'm interested to hear your opinion on this change. It makes sense because most users of openreviewpy will only ever interact with the openreview.net instance. But it means that, when we are testing things, we'll have to make sure that we're using localhost

openreview/openreview.py
```diff
@@ -298,92 +249,65 @@ def get_notes(self, id=None, forum=None, invitation=None, replyto=None, tauthor=
             params['trash']=True
 
         response = requests.get(self.notes_url, params=params, headers=self.headers)
-        
+        response = self.__handle_response(response)
+
+        for n in response.json()['notes']:
+            note = Note(n['id'],
```
melisabok: Could you move this code to a class method from_json?
Something like this:

```
    @classmethod
    def from_json(json_string):
          
```

openreview/openreview.py
```diff
@@ -17,7 +18,7 @@ def __init__(self, baseurl=None, process_dir='../process/', webfield_dir='../web
             try:
                 self.baseurl = os.environ['OPENREVIEW_BASEURL']
             except KeyError:
-                self.baseurl = 'http://localhost:3000'
```
melisabok: I would keep the localhost by default, and force the user to specify the baseUrl.

openreview/openreview.py
mspector: yes, do you want it in this PR? or should I make another PR for this change?

melisabok: can you do it in this PR?

openreview/openreview.py
```diff
@@ -240,41 +147,14 @@ def get_invitations(self, id=None, invitee=None, replytoNote=None, replyForum=No
...
-                    return error
-            except KeyError:
-                for error in response.json()['error']:
-                    return error
+        response = self.__handle_response(response)
+
+        for i in response.json()['invitations']:
+            invitation = Invitation.from_json(i)
+            invitations.append(invitation)
+            invitations.sort(key=lambda x: x.id)
```
melisabok: The sort should be outside the loop

openreview/openreview.py
```diff
@@ -192,35 +118,16 @@ def get_groups(self, prefix=None, regex=None, member=None, host=None, signatory=
...
-                
-        except requests.exceptions.HTTPError as e:
-            try:
-                for error in response.json()['errors']:
-                    return error
-            except KeyError:
-                for error in response.json()['error']:
-                    return error
+        for g in response.json()['groups']:
+            g = self.get_group(g['id']).to_json()
```
melisabok: do we still need to call get_group for each group of the response?

openreview/openreview.py
```diff
@@ -298,92 +178,51 @@ def get_notes(self, id=None, forum=None, invitation=None, replyto=None, tauthor=
             params['trash']=True
 
         response = requests.get(self.notes_url, params=params, headers=self.headers)
-        
+        response = self.__handle_response(response)
+
+        for n in response.json()['notes']:
+            note = Note.from_json(n)
+            notes.append(note)
+        notes.sort(key=lambda x: x.forum)
```
melisabok: Sort outside the loop, btw the notes service returns the notes sorted by date. I don't why you want to sort by forum

mspector: hmm I'm not sure either. I'll take it out

openreview/openreview.py
mspector: i just tested it, and it looks like we don't

melisabok:
Excellent!!!! Library is getting better and better 💯

PR #16, opened 2016-10-05
added add and remove member functions

mspector:


openreview/openreview.py
```diff
@@ -222,6 +222,29 @@ def send_mail(self, subject, recipients, message):
 
         return response
 
+    def add_members(self, groupId, members):
```
melisabok: rename to: add_member_to_group(group, members)

openreview/openreview.py
```diff
@@ -222,29 +222,33 @@ def send_mail(self, subject, recipients, message):
...
-            response = requests.put(self.groups_url+'/members', json={'id':group,'members':members}, headers=self.headers)
-            response = self.__handle_response(response)
-            return self.get_group(response.json()['id'])
-
-        if type(members)==str:
-            return add_member(groupId,[members])
-        if type(members)==list:
-            return add_member(groupId,members)
+    def add_members_to_group(self, group, members):
+        if self.exists(group.id):
```
melisabok: it is not necessary the validation, because the API return an error if the groups does not exist.

openreview/openreview.py
```diff
@@ -222,29 +222,33 @@ def send_mail(self, subject, recipients, message):
...
+            def add_member(group,members):     
+                response = requests.put(self.groups_url+'/members', json={'id':group,'members':members}, headers=self.headers)
+                response = self.__handle_response(response)
```
melisabok: in the future you can wrap the request and the handle_response in a single method of the client. So you don't need to add the handle everytime that you do a request

openreview/openreview.py
mspector: I found that the error message was unclear - it just says "no match"

melisabok: So. let's change the API. 

The message should be 'Not Found'.

PR #28, opened 2017-01-24
added user functions

mspector:


openreview/openreview.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, baseurl=None, process_dir='../process/', webfield_dir='../web
         self.notes_url = self.baseurl+'/notes'
         self.profiles_url = self.baseurl+'/user/profile'
         self.token = self.__login_user(username,password)
+        self.secure_activation = secure_activation
```
melisabok: this is a setting of the server, why do you need to user it in the client?

mspector:
I moved the secure_activation option to the function. I need it there to determine whether or not to provide a real token or a email address

PR #30, opened 2017-02-14
add multiReply and taskCompletionCount to invitations

mspector:


openreview/openreview.py
```diff
@@ -397,9 +399,11 @@ def to_json(self):
             'invitees': self.invitees,
             'noninvitees': self.noninvitees,
             'signatures': self.signatures,
-            'reply':self.reply,
+            'multiReply': self.multiReply,
+            'taskCompletionCount': self.taskCompletionCount,
+            'reply': self.reply,
             'process': self.process,
-            'web': self.web
+            'web': self.web,
```
melisabok: is there an extra comma?

mspector: ah, interesting, you're right.

it's interesting that it didn't fail when I tested it. but i'll fix it.

PR #31, opened 2017-02-16
add tags functionality and fix bugs

mspector:


melisabok:
I think it is ok, @mzarozinski  do you want to check is something missing?

mzarozinski:
Tiny little thing,

@classmethod
def from_json(Note,n):
Says "Note" instead of "Tag".

Otherwise, looks good.

mspector:
good catch, thanks!

PR #34, opened 2017-02-23
add regex, tags parameters

melisabok:


mspector:
sorry, i approved prematurely - could you also increment the version in the setup file?

PR #42, opened 2017-04-11
Fix/post return values

mspector:


openreview/openreview.py
```diff
@@ -276,10 +280,11 @@ def post_group(self, group, overwrite = True):
         """
         if overwrite or not self.exists(group.id):
             if not group.signatures: group.signatures = [self.signature]
+            if not group.writers: group.writers = [self.signature]
             response = requests.post(self.groups_url, json=group.to_json(), headers=self.headers)
             response = self.__handle_response(response)
 
-        return Group.from_json(response.json())
+        return self.get_group(response.json()['id'])
```
melisabok: are you sure you want to get the group after the post?

openreview/openreview.py
```diff
@@ -301,14 +307,15 @@ def post_note(self, note):
         if not note.signatures: note.signatures = [self.signature]
         response = requests.post(self.notes_url, json = note.to_json(), headers = self.headers)
         response = self.__handle_response(response)
-        return Note.from_json(response.json())
+
+        return self.get_note(response.json()['id'])
```
melisabok: the post note is returning the full object, you don't need to call get_note

openreview/openreview.py
melisabok: may be we should change the API to return the full group/invitation in the response

openreview/openreview.py
```diff
@@ -419,26 +428,26 @@ def post(self, client):
...
 
         default_reply = {
-            'forum': None,
-            'replyto': None,
+            'forum': forum,
+            'replyto': replyto,
             'readers': {'values': ['everyone']},
-            'signatures': {'values-regex': '.*'},
-            'writers': {'values-regex': '.*'},
+            'signatures': {'values-regex': '~.*'},
```
melisabok: I'm not sure about setting default values here in the library.

openreview/openreview.py
mspector: one way or another, I think the post_group function should return the whole, up-to-date group object. If changing the API is a better solution than doing another get, then that's fine

openreview/openreview.py
mspector: ok. So I should keep the line:

return Note.from_json(response.json())

is that right?

openreview/openreview.py
mspector: this might require a longer conversation... I can take out the defaults for now, though.

The reason I wanted to add defaults is because I think there are certain group parameters that are much more common than others, and it can reduce the amount of boilerplate code. but yes, we have to do this very carefully

openreview/openreview.py
melisabok: I agree. We should change the API. Doing two network calls in the post_group is not efficient.

openreview/openreview.py
melisabok: yes, you can test it also.

openreview/openreview.py
melisabok: I agree about having default values, but I'm not sure if the library is the right place, I don't want to give it more responsibility that being a bridge between the API and the users. I believe the scripts repo should have this kind of helpers to create default configurations. We can have a conversation about this.

openreview/openreview.py
```diff
@@ -419,26 +428,26 @@ def post(self, client):
...
-            'writers': {'values-regex': '.*'},
-            'content':{}
-        },
+            'forum': forum,
+            'replyto': replyto,
+            'readers': {},
+            'signatures': {},
+            'writers': {},
+            'content': {}
+        }
```
mspector: I took out the defaults here, but I left the "default_reply" dictionary, because I think it's helpful to see which fields are expected. What do you think @melisabok ?

melisabok:
is @mzarozinski fix deployed?

mspector:
I'm not sure if it's deployed, but it is in master now

melisabok:
ok, I can merge it. But if you deploy the library you need to deploy OpenReview code first

PR #53, opened 2017-05-18
add paperhash parameter

melisabok:
@mzarozinski this is the way to call it:

```
openreview.get_notes(paperhash = 'kim|interpretable_3d_human_action_analysis_with_temporal_convolutional_networks')
```

Be sure that this index is in the database:

```
notesDb.ensureHashIndex('content.paperhash');
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup
 
 setup(name='openreview-py',
-      version='0.5.4.8',
+      version='0.5.5',
       description='OpenReview client library',
       url='https://github.com/iesl/openreview-py',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup
 
 setup(name='openreview-py',
-      version='0.5.4.7',
+      version='0.5.4.8',
       description='OpenReview client library',
```
mspector: let's upgrade to 0.5.5, since there are several unpublished changes

melisabok:
done!

PR #55, opened 2017-06-01
fixed error handling

mspector:
I have more changes coming today, so I'm not incrementing the version number for this one

### DIFF
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -67,7 +67,10 @@ class Client(object):
 
             return response
         except requests.exceptions.HTTPError as e:
-            raise OpenReviewException(response.json()['errors'])
+            if 'errors' in response.json():
+                raise OpenReviewException(response.json()['errors'])
+            else:
+                raise OpenReviewException(response.json())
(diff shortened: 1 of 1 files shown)


openreview/openreview.py
```diff
@@ -67,8 +67,7 @@ def __handle_response(self,response):
 
             return response
         except requests.exceptions.HTTPError as e:
-            for k,v in response.json().iteritems():
-                raise OpenReviewException(str(v))
+            raise OpenReviewException(response.json()['errors'])
```
melisabok: I'm not sure if always you will get an errors property, could you check if exists first? otherwise return the whole json.

melisabok:
When I login with a wrong password I should get an error "Invalid username or password" and I get this:
screen shot 2017-06-02 at 12 08 34 pm

mspector:
that's strange. I'm getting the correct error message when I try it. i'll investigate more

melisabok:
My bad, I didn't reload the library. I'm sorry!

PR #56, opened 2017-06-01
remove auto signature feature

mspector:


melisabok:
I would suggest also that the constructor of the object(Group, Note, Invitation) requires a collection of signatures and writers, don't make them optional.

PR #60, opened 2017-08-01
python 3 compatibility

mspector:


melisabok:
what about this?

mspector:
this is ready too, i just tested again to make sure

PR #64, opened 2017-11-02
fix get groups by regex, fix get references

mspector:


openreview/openreview.py
```diff
@@ -184,7 +184,7 @@ def get_groups(self, prefix = None, regex = None, member = None, host = None, si
         """Returns a list of Group objects based on the filters provided."""
         params = {}
         if prefix != None: params['regex'] = prefix+'.*'
-        if regex != None: params['regex'] = regex
+        if regex != None: params['id'] = regex
```
mspector: I wonder actually if the endpoint should accept the argument "regex" in the backend instead of "id" for this. What do you think?

melisabok: it accepts the parameter 'regex' also but it is not efficient.

melisabok: I think you need to keep both parameters, id and regex

melisabok: we use the regex parameter to get the groups where the use is signatory, to sign the notes. But this service doesn't use indices so we need use it with the signatory parameter also.

PR #69, opened 2017-12-08
format error messages

melisabok:


melisabok:
@mspector could you take a look?

PR #71, opened 2017-12-12
add overwriting parameter to note, and add new function to tools module

mspector:


openreview/tools.py
```diff
@@ -50,3 +50,29 @@ def create_profile(client, email, first, last, middle = None):
 		raise openreview.OpenReviewException('There is already a profile with this email: {0}'.format(email))
 
 
+def build_groups(conference_group_id):
```
mspector: add a default parameter for empty_params here

openreview/tools.py
```diff
@@ -50,3 +50,29 @@ def create_profile(client, email, first, last, middle = None):
...
+        'writers': [],
+        'signatures': [],
+        'signatories': [],
+        'members': []
+    }
+
+    groups = {p: openreview.Group(p, **empty_params) for p in paths}
+    groups[conference_group_id].writers = groups[conference_group_id].signatories = [conference_group_id]
+
+    admin_id = conference_group_id + '/Admin'
```
mspector: remove admin group

(maybe create another function that calls this function?)

openreview/tools.py
```diff
@@ -16,37 +16,61 @@ def get_profile(client, value):
...
+    Given a group ID, returns a list of empty groups that correspond to the given group's subpaths
+    (e.g. Test.com, Test.com/TestConference, Test.com/TestConference/2018)
+
+    '''
+
+    path_components = conference_group_id.split('/')
+    paths = ['/'.join(path_components[0:index+1]) for index, path in enumerate(path_components)]
+
+    if not default_params:
+        default_params = {
```
melisabok: is it useful to create an empty group?

May be we can send a user id as a parameter to set the signature and signatory. 

Are you using this function? where?

mspector: Good point. I'm not sure if this is the best way to do it.

Right now, it works like this: when the "post_group" function gets called by the client, if there is an empty signature or writer field, the client fills it with the username that it was initialized with. (I remember that we discussed whether or not this was a good idea - I like it, but we can change this if you think it's not a good design).

That means that the empty fields are useful, since it's essentially saying that the writers and signatures will be set to whoever posts the group, and that there are no signatories or members by default. 

I have used this function once so far, to set up the groups for the NAACL preprint server. I call this function with the super user in a very simply jupyter notebook script like this:

```
import openreview
client = openreview.Client()
naacl_groups = openreview.tools.build_groups('naacl.org/NAACL/2018/Preprint')
for n in naacl_groups:
    client.post_group(n)
```

I haven't committed this code because I would like to design the python library (and the "builder" module) in a way that encourages interactivity.

Are these good design decisions? let me know if you disagree, I don't want to go down a bad path

mspector:
I'm going to close this for right now, since there are two features being added in this PR, and I think they should be separated. I created a separate pull request for the "overwriting" feature here: https://github.com/iesl/openreview-py/pull/72

PR #79, opened 2018-02-01
add update profile function

mspector:
needed for researcher data import


melisabok:
Sounds good but be careful how you do the update, this endpoint was designed to be used by the UI.

mspector:
ok, thanks. I did test it very thoroughly. is it ok to merge?

mspector:
thanks!

PR #82, opened 2018-02-21
Feature/add tools

mspector:


melisabok:
It is great that we can add this code in the tools module 💯

mspector:
Yes, it's been long overdue

PR #84, opened 2018-03-08
Feature/add tools

mspector:


openreview/tools.py
```diff
@@ -51,6 +53,99 @@ def create_profile(client, email, first, last, middle = None):
         raise openreview.OpenReviewException('There is already a profile with this email: {0}'.format(email))
 
 
+def get_or_create_profile(client, email, first, last, allow_duplicates=False, verbose=False):
```
melisabok: I think create profile is doing this, why we need another funciton?

openreview/tools.py
```diff
@@ -194,3 +289,44 @@ def get_paper_conflicts(client, paper):
...
+            ids.append(member)
+
+    group.members = ids + emails
+    client.post_group(group)
+
+def get_all_notes(client, invitation, limit=1000):
+    done = False
+    notes = []
+    while not done:
+        batch = client.get_notes(invitation=invitation, limit=limit)
```
melisabok: you need to pass an offset otherwise you get always the same result.

openreview/tools.py
mspector: ah yes, you're right. I'm not sure why I missed that

melisabok: don't worry it is not urgent to fix this.

openreview/tools.py
mspector: huh, you are right, that's weird. the test I did must have been flawed somehow.

PR #92, opened 2018-03-19
Fix/conflicts

melisabok:


openreview/matching.py
```diff
@@ -15,35 +14,49 @@ def get_conflicts(author_profiles, user_profile):
...
-    		institution_domains += tools.subdomains(domain)
+        domain = h.get('institution', {}).get('domain', None)
+        if domain:
+            institution_domains += get_domains(domain, subdomains = True)
     domain_conflicts.update(institution_domains)
 
     domain_conflicts.update(institution_domains)
 
-    relation_conflicts.update([r['email'] for r in profile.content.get('relations', [])])
+    relation_conflicts.update([get_domains(r['email'], subdomains = False) for r in profile.content.get('relations', [])])
```
melisabok: I'm not sure if we need to compute the relation conflicts matching the domains or the full email address. I realized that is the last one, in this case, I will change the backend so we can get the email addresses not masked for the superuser.

openreview/matching.py
```diff
@@ -15,35 +14,49 @@ def get_conflicts(author_profiles, user_profile):
...
 
-	conflicts.update(authors_domain_conflicts.intersection(user_domain_conflicts))
-	conflicts.update(authors_relation_conflicts.intersection(user_relation_conflicts))
-
-	return list(conflicts)
+    return list(conflicts)
 
 def get_profile_conflicts(profile):
     domain_conflicts = set()
     relation_conflicts = set()
```
mspector: maybe this "relation_conflicts" set should start with one item: the profile's id

relation_conflicts = set([profile.id])

what do you think?

PR #93, opened 2018-03-20
fix overly restrictive conflicts in matching

mspector:
before this change, the relation conflicts were the result of the intersection between all author relations and all reviewer relations. So for example:

Andrew's relations = [Michael, Melisa, Zach]
Ari's relations = [Michael, Zach]
Intersection = [Michael, Zach]

this would result in a conflict between Andrew and Ari, because they share common relations. This is too restrictive.

The change in this PR makes it so that we take the intersection between the author emails and the reviewer relations, and between the reviewer emails and the author relations.

### DIFF
--- a/openreview/matching.py
+++ b/openreview/matching.py
@@ -5,55 +5,67 @@ import openreview
 def get_conflicts(author_profiles, user_profile):
 
     authors_domain_conflicts = set()
+    authors_email_conflicts = set()
     authors_relation_conflicts = set()
-    conflicts = set()
-
-    author_emails = author_profiles.keys()
-    user_emails = user_profile.content['emails']
 
     for author_email, profile in author_profiles.iteritems():
-        if profile:
-            current_domain_conflicts, current_relation_conflicts = get_profile_conflicts(profile)
-        else:
-            current_domain_conflicts = get_domains(author_email, subdomains = True)
-            current_relation_conflicts = [ author_email ]
 
-        authors_domain_conflicts.update(current_domain_conflicts)
-        authors_relation_conflicts.update(current_relation_conflicts)
+        author_conflicts = get_author_conflicts(profile, author_email)
 
-    user_domain_conflicts, user_relation_conflicts = get_profile_conflicts(user_profile)
+        authors_domain_conflicts.update(author_conflicts['domains'])
+        authors_email_conflicts.update(author_conflicts['emails'])
+        authors_relation_conflicts.update(author_conflicts['relations'])
 
-    conflicts.update(authors_domain_conflicts.intersection(user_domain_conflicts))
-    conflicts.update(authors_relation_conflicts.intersection(set(user_emails)))
-    conflicts.update(user_relation_conflicts.intersection(set(author_emails)))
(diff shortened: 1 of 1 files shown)


openreview/matching.py
```diff
@@ -8,16 +8,24 @@ def get_conflicts(author_profiles, user_profile):
...
 
-    for p in author_profiles.values():
-        author_domain_conflicts, author_relation_conflicts = get_profile_conflicts(p)
+    author_emails = author_profiles.keys()
+    user_emails = user_profile.content['emails']
 
-        authors_domain_conflicts.update(author_domain_conflicts)
-        authors_relation_conflicts.update(author_relation_conflicts)
+    for author_email, profile in author_profiles.iteritems():
+        if profile:
```
mspector: this is the first main change. Before, a dict of profiles keyed by emails was being passed in. Now, it can handle a dict keyed by emails where some of the values are profiles, and some are None (if, for example, the author doesn't have a profile).

openreview/matching.py
```diff
@@ -8,16 +8,24 @@ def get_conflicts(author_profiles, user_profile):
...
+
+        authors_domain_conflicts.update(current_domain_conflicts)
+        authors_relation_conflicts.update(current_relation_conflicts)
 
     user_domain_conflicts, user_relation_conflicts = get_profile_conflicts(user_profile)
 
     conflicts.update(authors_domain_conflicts.intersection(user_domain_conflicts))
-    conflicts.update(authors_relation_conflicts.intersection(user_relation_conflicts))
+    conflicts.update(authors_relation_conflicts.intersection(set(user_emails)))
+    conflicts.update(user_relation_conflicts.intersection(set(author_emails)))
```
mspector: this is the second main change. Instead of finding the intersection between all author relations and all reviewer relations, we find two intersections: one between the author relations and reviewer emails (to see if the authors have the reviewer as a relation), and one between the reviewer relations and the author emails (to see if the reviewer has any of the authors as a relation).

openreview/matching.py
```diff
@@ -4,67 +4,67 @@
...
+    user_info = get_profile_info(user_profile)
 
     conflicts = set()
-    conflicts.update(authors_domain_conflicts.intersection(user_conflicts['domains']))
-    conflicts.update(authors_relation_conflicts.intersection(user_conflicts['emails']))
-    conflicts.update(authors_email_conflicts.intersection(user_conflicts['relations']))
+    conflicts.update(author_domains.intersection(user_info['domains']))
+    conflicts.update(author_relations.intersection(user_info['emails']))
+    conflicts.update(author_emails.intersection(user_info['relations']))
+    conflicts.update(author_emails.intersection(user_info['emails']))
```
melisabok: why are you doing this? this is the same that domains conflicts

mspector: ah, you're right that in practice this will be the same thing.

I was trying to catch the case where the reviewer is an author, but if the reviewer is an author, then there is a domain conflict. i'll remove it.

melisabok:
Best pair programming and code review ever 💯

PR #94, opened 2018-03-23
major refactor of matching module

mspector:
this is a major refactor of the matching module.

Summary of changes:

- moved all the functionality of the "match()" function in the [openreview_matcher repository](https://github.com/iesl/openreview-matching/blob/master/openreview_matcher/matcher.py) into the openreview.matching module. 
- The gurobipy dependency is still outside of the openreview-py module. This is so we can easily replace it later with a better integer program solver (at some point I'll ask Ari to put together his new implementation)
- refactored the "match()" function into modular functions to make it easier to follow

It may be worth going over this in person (or over a video call).

### DIFF
--- a/openreview/matching.py
+++ b/openreview/matching.py
@@ -1,9 +1,11 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 import openreview
+import tools
+import numpy as np
+from collections import defaultdict
 
 def get_conflicts(author_profiles, user_profile):
-
     author_domains = set()
     author_emails = set()
     author_relations = set()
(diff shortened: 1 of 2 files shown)


melisabok:
I think this it is great! Let me take a closer look next week.

openreview/matching.py
```diff
@@ -80,3 +82,294 @@ def get_domains(entity, subdomains = False):
...
+        'invitation': assignment_invitation.id,
+        'readers': assignment_invitation.reply['readers']['values'],
+        'writers': assignment_invitation.reply['writers']['values'],
+        'signatures': assignment_invitation.reply['signatures']['values'],
+        'content': {
+            'label': label
+        }
+    }
+
+def create_or_update_config(client, label, params, number=None):
```
melisabok: I would call it save_config

openreview/matching.py
```diff
@@ -80,3 +82,294 @@ def get_domains(entity, subdomains = False):
...
+    '''
+    makes a network call to get any existing configuration notes that match the given label,
+    and updates the note with the given params.
+
+    Returns an openreview.Note object with the updated params.
+    '''
+    config_notes = tools.get_all_notes(client, params['invitation'])
```
melisabok: you can get all the notes by label:

/notes?content.label=label

and if number is present you can add number to the request:

/notes?content.label=label&number=number

openreview/matching.py
```diff
@@ -80,3 +82,294 @@ def get_domains(entity, subdomains = False):
...
+    if number:
+        config_notes = [c for c in config_notes if int(c.number) == int(number)]
+    config_by_label = {n.content['label']: n.to_json() for n in config_notes}
+    configuration_note_params = config_by_label.get(label, {})
+    user_constraints = configuration_note_params.get('content', {}).get('constraints', {})
```
melisabok: Why do you get the constraints to set the constraints again?

openreview/matching.py
```diff
@@ -80,3 +82,294 @@ def get_domains(entity, subdomains = False):
         return valid_domains
     else:
         return full_domain
+
+def match(client, configuration_note, Solver):
```
melisabok: I think this piece of code is too important to have it as a helper of the library, why don't we add it to the matching code?

We can keep a part that is agnostic to openreview, but we will need a layer to integrate both parts and [this](https://github.com/iesl/openreview-matching) repo should be the place.

openreview/matching.py
```diff
@@ -80,3 +82,294 @@ def get_domains(entity, subdomains = False):
...
+        user_id = user_by_index[user_index]
+        forum = forum_by_index[paper_index]
+        match = solution[var_name]
+
+        if match == 1:
+            assignments_by_forum[forum].append(user_id)
+
+    return assignments_by_forum
+
+def create_or_update_assignments(assignments, existing_assignments, assignment_invitation, configuration_note, scores_by_user_by_forum):
```
melisabok: I would call it save_assignments

openreview/matching.py
```diff
@@ -80,3 +82,294 @@ def get_domains(entity, subdomains = False):
...
+        'scores'
+        'userId'
+
+    '''
+    alternate_entries = []
+
+    for user, scores in scores_by_user.iteritems():
+        if user not in assigned_userids:
+            alternate_entries.append({
+                'finalScore': np.mean([val for feature, val in scores.iteritems() if val != '+inf' and val != '-inf']),
```
melisabok: you are re calculating the score again

openreview/matching.py
```diff
@@ -80,3 +82,294 @@ def get_domains(entity, subdomains = False):
...
+        'finalScore'
+        'scores'
+        'userId'
+    '''
+
+    assignment_entries = []
+    for user, scores in scores_by_user.iteritems():
+        if user in userids:
+            assignment_entries.append({
+                'finalScore': np.mean([val for feature, val in scores.iteritems() if val != '+inf' and val != '-inf']),
```
melisabok: Also here, you are calculating the score again. We should read the value from line L258

PR #96, opened 2018-03-27
Add timestamp_GMT

pmandler:
Do I need to change the version number too? Or is that only for changes to openreview.py?


mspector:
the version number just needs to be changed before it gets deployed. we can probably wait for some other changes to get merged in before we do a version update.

PR #98, opened 2018-03-31
add invitations template functions

mspector:
adds standard invitation contents and functions to create invitations from templates

### DIFF
--- a/openreview/invitations/content.py
+++ b/openreview/invitations/content.py
@@ -1,0 +1,56 @@
+comment = {
+    'title': {
+        'order': 0,
+        'value-regex': '.{1,500}',
+        'description': 'Brief summary of your comment.',
+        'required': True
+    },
+    'comment': {
+        'order': 1,
+        'value-regex': '[\\S\\s]{1,5000}',
+        'description': 'Your comment or reply (max 5000 characters).',
+        'required': True
+    }
+}
+
+submission = {
+    'title': {
+        'order': 1,
+        'value-regex': '.{0,500}',
+        'description': 'Brief summary of your review.',
+        'required': True
+    },
+    'review': {
+        'order': 2,
+        'value-regex': '[\\S\\s]{1,200000}',
+        'description': 'Please provide an evaluation of the quality, clarity, originality and significance of this work, including a list of its pros and cons (max 200000 characters).',
+        'required': True
+    },
+    'rating': {
+        'order': 3,
+        'value-dropdown': [
+            '10: Top 5% of accepted papers, seminal paper',
+            '9: Top 15% of accepted papers, strong accept',
+            '8: Top 50% of accepted papers, clear accept',
+            '7: Good paper, accept',
+            '6: Marginally above acceptance threshold',
+            '5: Marginally below acceptance threshold',
+            '4: Ok but not good enough - rejection',
+            '3: Clear rejection',
+            '2: Strong rejection',
+            '1: Trivial or wrong'
+        ],
+        'required': True
+    },
+    'confidence': {
+        'order': 4,
+        'value-radio': [
+            '5: The reviewer is absolutely certain that the evaluation is correct and very familiar with the relevant literature',
+            '4: The reviewer is confident but not absolutely certain that the evaluation is correct',
+            '3: The reviewer is fairly confident that the evaluation is correct',
+            '2: The reviewer is willing to defend the evaluation, but it is quite likely that the reviewer did not understand central parts of the paper',
+            '1: The reviewer\'s evaluation is an educated guess'
+        ],
+        'required': True
+    }
+}
--- a/openreview/invitations/content.py
+++ b/openreview/invitations/content.py
@@ -1,3 +1,16 @@
+bid = {
+    'tag': {
+        'description': 'Bid description',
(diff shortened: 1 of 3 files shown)


openreview/invitations/content.py
```diff
@@ -0,0 +1,56 @@
...
+    },
+    'comment': {
+        'order': 1,
+        'value-regex': '[\\S\\s]{1,5000}',
+        'description': 'Your comment or reply (max 5000 characters).',
+        'required': True
+    }
+}
+
+submission = {
```
melisabok: is this a submission invitation? it seems a review invitation

melisabok:
I'm not sure what are you trying to do this. We will discuss it later.

mspector:
@melisabok  can we merge this?

mspector:
sorry, found a bug, fixing now

(we can talk later about this)

openreview/invitations/templates.py
```diff
@@ -232,4 +161,42 @@ def __init__(self, name, conference_id, duedate = 0,
...
+    for match in matches:
+        discovered_field = re.sub('<|>', '', match)
+        template_string = template_string.replace(match, str(paper_params[discovered_field]))
+        print "    new value: ", template_string
+    return template_string
+
+def generate_invitation(invitation_template, paper):
+    params = copy.deepcopy(invitation_template)
+    reply = params.pop('reply')
+    paper_params = paper.to_json()
```
melisabok: Why do you need to convert a paper to json instead of using the python object?

openreview/invitations/templates.py
```diff
@@ -232,4 +161,42 @@ def __init__(self, name, conference_id, duedate = 0,
...
+            if field == 'content' or field == 'process' or field == 'webfield':
+                pass
+            elif type(field_map[field]) == dict:
+                for subfield in field_map[field]:
+                    if subfield != 'description':
+                        field_map[field][subfield] = fill_list_or_str(field_map[field][subfield])
+            else:
+                field_map[field] = fill_list_or_str(field_map[field])
+
+    return openreview.Invitation(**dict(params, **{'reply': reply}))
```
melisabok: I would enforce the definition of each parameter when we create the invitation, and avoid the deep copy, for example:
```
openreview.Invitation(
writers = invitation_template[writers],
invitees = invitation_template[invitees],
....
)
```
We are sure that we are sending the correct parameters. Sending a dict of params you won't have control of which parameters are sending.

openreview/invitations/templates.py
```diff
@@ -232,4 +161,42 @@ def __init__(self, name, conference_id, duedate = 0,
...
+    return template_string
+
+def generate_invitation(invitation_template, paper):
+    params = copy.deepcopy(invitation_template)
+    reply = params.pop('reply')
+    paper_params = paper.to_json()
+
+    def fill_list_or_str(value, paper_params=paper_params):
+        if type(value) == list:
+            return [fill_template(v, paper_params) for v in value]
```
melisabok: you are defining an inner function that is calling to an outer function, is this ok?

Why do you define an inner function?

openreview/invitations/templates.py
mspector: good point, I'm not sure why I originally chose to do that. I'll change it

openreview/invitations/templates.py
mspector: another good point... I'm not sure why this is an inner function, it could just as easily be an outer function

openreview/invitations/templates.py
mspector: actually, after thinking about this, I'm not sure if this is the right way to do it. invitations don't all have exactly the same set of fields, especially within the "reply" field. I think it should be the function caller's responsibility to set the parameters correctly - this function should only serve to iterate through the fields and replace the templates (e.g. <number>). This may suggest that this function should be named differently, like "fill_invitation_template()" instead of "generate_invitation()"

that being said, I'm interested to hear counter-arguments... what do you think?

melisabok: I disagree, we should define the contract and the function caller should respect it. This is the contract:

```
def __init__(self, id, writers=None, invitees=None, noninvitees=None, readers=None, nonreaders=None, reply=None, replyto=None, forum=None, invitation=None, web=None, process=None, signatures=None, duedate=None, cdate=None, rdate=None, ddate=None, multiReply=None, taskCompletionCount=None):

```

reply is a dictionary

openreview/invitations/templates.py
```diff
@@ -232,4 +161,71 @@ def __init__(self, name, conference_id, duedate = 0,
...
+    '''
+
+    exclusions = [
+        'web','process','duedate','cdate','rdate','ddate',
+        'multiReply','taskCompletionCount'
+    ]
+
+    new_params = _fill_template(invitation_template, paper, exclude = exclusions)
+
+    return openreview.Invitation(
```
mspector: @melisabok is this what you had in mind?

melisabok: I really like it! It is simpler to understand

mspector:
@melisabok ok, I removed the exclusions. I think you're right that they're unnecessary. We can add them back in if they later seem to be needed

PR #102, opened 2018-04-12
Fix/get notes

mspector:


openreview/openreview.py
```diff
@@ -256,36 +256,16 @@ def get_invitations(self, id = None, invitee = None, replytoNote = None, replyFo
...
-            params['number'] = number
-        if limit != None:
-            params['limit'] = limit
-        if offset != None:
-            params['offset'] = offset
-        if mintcdate != None:
-            params['mintcdate'] = mintcdate
-
+    def get_notes(self, **params):
+        '''
```
mspector: how will we know which parameters are ok for the API?

look into this - how does a user access the docstring?

PR #107, opened 2018-05-08
tools.py add post_submission_groups, templates add id to Comment invite

pmandler:


openreview/invitations/templates.py
```diff
@@ -115,14 +115,19 @@ def __init__(self, name, conference_id, duedate = 0,
         super(AddBid, self).__init__(**self.inv_params)
 
 class Comment(openreview.Invitation):
-    def __init__(self, name, conference_id, duedate = 0,
+    def __init__(self, name = None, conference_id = None, id = None, duedate = 0,
```
mspector: I think we actually don't even need the "name" variable anymore

openreview/tools.py
```diff
@@ -440,4 +440,53 @@ def add_assignment(user, parent_group, unassigned_individual_groups, individual_
 def timestamp_GMT(year, month, day, hour=0, minute=0, second=0):
     return int((datetime.datetime(year, month, day, hour, minute, second) - datetime.datetime(1970, 1, 1)).total_seconds() * 1000)
 
+''' Create paper group, authors group, reviewers group, review non-readers group
```
mspector: comments like these are called "docstrings", and they should be inside the function body

PR #109, opened 2018-05-08
add recruitment template

mspector:
adds a recruitment invitation template and content


openreview/invitations/templates.py
```diff
@@ -167,6 +167,54 @@ def __init__(self, name, conference_id, duedate = 0,
...
+                'values': ['~Super_User1']
+            },
+            'signatures': {
+                'values-regex': '\\(anonymous\\)'
+            },
+            'writers': {
+                'values-regex': '\\(anonymous\\)'
+            }
+        }
+
```
pmandler: Why use anonymous? Only a Super User can see them.

mspector: this is so that they can sign the note without being logged in. any value other than anonymous requires a user token

PR #117, opened 2018-05-24
Split assign into add_assignment() and remove_assignment() and pull o…

pmandler:
…ut common code helper functions.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -17,14 +17,6 @@ def get_profile(client, value):
             return e
     return profile
 
-def get_profile_id(client, value):
-    profile = get_profile(client, value)
-    if profile:
-        user = profile.id
-    else:
-        user = value
-    return user
-
 def create_profile(client, email, first, last, middle = None, allow_duplicates = False):
 
     profile = get_profile(client, email)
@@ -379,7 +371,8 @@ def add_assignment(client, paper_number, conference, reviewer,
 
     Prints the results to the console.
     '''
-    user = get_profile_id(client, reviewer)
+    profile = get_profile(client, reviewer)
+    user = profile.id if profile else reviewer
 
     if user not in parent_group.members:
         client.add_members_to_group(parent_group, user)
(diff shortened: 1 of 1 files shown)


openreview/tools.py
```diff
@@ -360,85 +346,164 @@ def assign(client, paper_number, conference,
...
+    @paper_number: the number of the paper to assign
+    @conference: the ID of the conference being assigned
+    @reviewer: may be an email address or a tilde ID;
+        adds the given user to the parent and individual groups defined by
+        the paper number, conference, and labels
+    @parent_group_params: optional parameter that overrides the default
+    @individual_group_params: optional parameter that overrides the default
+     '''
+
+    parent_group = get_or_create_reviewer_group(client, paper_number, conference, parent_group_params, parent_label)
```
melisabok: it seems these 4 lines are repeated in both functions: add_assignment and remove_assignment. Could we put then together in a common function?

openreview/tools.py
```diff
@@ -360,85 +346,164 @@ def assign(client, paper_number, conference,
...
+    Removes the given user from the parent group,
+        and any assigned individual groups.
+    '''
+
+    user_groups = [g.id for g in client.get_groups(member=user)]
+
+    for user_entity in user_groups:
+        if user_entity in parent_group.members:
+            client.remove_members_from_group(parent_group, user_entity)
+            print "{:40s} xxx {}".format(user_entity, parent_group.id)
```
melisabok: can we result a result, same that we did in the add_assignment

openreview/tools.py
melisabok: may be we can do together in a single function:

- get or create reviewer group
- get individual group

no need to define them separately

openreview/tools.py
```diff
@@ -17,6 +17,14 @@ def get_profile(client, value):
             return e
     return profile
 
+def get_profile_id(client, value):
```
melisabok: I don't think that we need a function for this, get_profile_id doesn't return always an id. So the name is confusing. 

I would replace the function call with these 2 lines:

```
profile = tools.get_profile(client, user)
user = profile.id if profile else user
```

pmandler:
Can we merge this in with the other change for 0.8.0?

PR #119, opened 2018-06-11
export invitations folder

melisabok:
@pmandler 

Could you check if your localhost works?

you should do:

```
pip install .
```

without the -e option

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,8 @@ setup(name='openreview-py',
       author_email='spector@cs.umass.edu, mbok@cs.umass.edu',
       license='MIT',
       packages=[
-          'openreview'
+          'openreview',
+          'openreview/invitations'
       ],
       install_requires=[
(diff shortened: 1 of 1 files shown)


pmandler:
How do I test it before merging it? Pip install . on its own did not work.

melisabok:
you need to checkout this branch, could you?

pmandler:
Yeah I tried:
   git checkout fix/invitation-export 
from the openreview-py directory where setup.py resides and I got:
   error: pathspec 'fix/invitation-export' did not match any file(s) known to git.
Then tried:
  git checkout --track fix/invitation-export
and got:
  fatal: Cannot update paths and switch to branch at the same time

melisabok:
git fetch first?

melisabok:
New version uploaded: https://pypi.org/project/openreview-py/0.8.0/

PR #121, opened 2018-06-26
add kwargs to get notes function

mspector:
adding **kwargs allows us to add arbitrary arguments, without breaking backwards compatibility (e.g. both trash and includeTrash arguments will work)

some advantages to **kwargs over explicitly naming the arguments:
- the API can evolve without us having to update the python library
- it's more readable

some disadvantages:
- you can't see what arguments are allowed just by looking at the function contract. but this isn't as much of a problem if we have central documentation for valid arguments.

### DIFF
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -256,7 +256,7 @@ class Client(object):
         invitations.sort(key = lambda x: x.id)
         return invitations
 
-    def get_notes(self, id = None, paperhash = None, forum = None, invitation = None, replyto = None, tauthor = None, signature = None, writer = None, includeTrash = None, number = None, limit = None, offset = None, mintcdate = None):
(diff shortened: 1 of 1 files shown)


mspector:
@melisabok what do you think of this? we have talked about something related before, and at the time we decided not to use **kwargs because of the disadvantage I listed in the above comment. but I think it will not be a problem when we have some central documentation (in fact, it might be a good thing, so that comments don't go out of date in the python library.)

(One counter-argument to this could be that we should just put in the effort to maintain both the documentation and the python library contract)

melisabok:
I would keep the name of the parameters. I think it is ok to keep the library synced with the API otherwise you won't know what is happening when you get an invalid argument error. 

I think we have to add more value to the library instead of the trying to make it more generic. 

Implementing readocs.org will help us to read the parameters of each function. This is a good start.

mspector:
ok, sounds good.

another question: should I remove includeTrash? it might break some old scripts, but we rarely reuse scripts anyway

melisabok:
ok, I don't remember using includeTrash in the scripts but we could fix the scripts.

mspector:
ok, I don't think we need to worry about it. we can fix scripts later if we need to (sounds like we probably won't).

the changes have been made

PR #123, opened 2018-07-05
define a new Profile entity

melisabok:
- new Profile class
- create, update and add references using Profile class
- replace __str__ with __repr__, it is easier to debug the objects, use object vars instead of serializing to json. 

@pmandler , how to add a new reference to a profile:
```
client.update_profile(openreview.Profile(referent='~Melisa_TestBok1', 
                                         invitation = '~/-/profiles', 
                                         content= { 'dblp': 'https://dblp.org/mbok'}, 
                                         signatures = ['~Melisa_TestBok1']))
```

this example add the url for dblp site.


openreview/openreview.py
```diff
@@ -461,9 +459,9 @@ def __init__(self, id, cdate = None, ddate = None, writers = None, members = Non
             with open(web) as f:
                 self.web = f.read()
 
-    def __str__(self):
```
melisabok: @mspector please take a look at this, I'm not sure if it is the right solution. 

Using __repr__ is easier to read the results, no need to use print. Avoid to_json calls.

melisabok:
It depends on https://github.com/iesl/openreview/pull/1021

openreview/openreview.py
mspector: Thanks Melisa, I think this is good.

A quick search revealed this useful stack overflow post: https://stackoverflow.com/questions/1436703/difference-between-str-and-repr

The relevant part:
> [...] in simple terms: almost every object you implement should have a functional __repr__ that’s usable for understanding the object. Implementing __str__ is optional: do that if you need a "pretty print" functionality (for example, used by a report generator).

So according to this, it would mean that we should have __repr__ return self.to_json(), and have __str__ return the pretty printed version.

openreview/openreview.py
```diff
@@ -677,6 +676,7 @@ def to_json(self):
             'ddate': self.ddate,
             'number': self.number,
             'content': self.content,
+            'metaContent': self.metaContent,
```
mspector: what is metaContent?

openreview/openreview.py
melisabok: Ok, I implemented both.

melisabok:
This can be merged too

melisabok:
I'm going to add a few tests before merging

PR #124, opened 2018-07-10
add docstrings, update for python 2/3 compatibility

mspector:
mostly just adds docstrings to functions without them (or updates existing docstrings).

updates a few lines (mostly dealing with the handling of unicode literals) for python 2/3 cross compatibility

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -91,6 +91,7 @@ def build_groups(conference_group_id, default_params=None):
 
     return sorted(groups.values(), key=lambda x: len(x.id))
 
+@deprecated
 def post_group_parents(client, group, overwrite_parents=False):
     '''
     Helper function for posting groups created by build_groups function.
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -1,29 +1,29 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
 import openreview
 import re
 import datetime
 
 super_user_id = 'OpenReview.net'
 
-def get_profile(client, value):
-    profile = None
-    try:
-        profile = client.get_profile(value)
-    except openreview.OpenReviewException as e:
-        # throw an error if it is something other than "not found"
-        if e[0][0] != 'Profile not found':
-            print "OpenReviewException: {0}".format(e)
-            return e
-    return profile
-
 def create_profile(client, email, first, last, middle = None, allow_duplicates = False):
+
+    '''
+    Given email, first name, last name, and middle name (optional), creates and returns
+    a user profile.
+
+    If a profile with the same name exists, and allow_duplicates is False, an exception is raised.
+
+    If a profile with the same name exists and allow_duplicates is True, a profile is created with
+    the next largest number (e.g. if ~Michael_Spector1 exists, ~Michael_Spector2 will be created)
+    '''
 
     profile = get_profile(client, email)
 
     if not profile:
         response = client.get_tildeusername(first, last, middle)
-        tilde_id = response['username'].encode('utf-8')
+        tilde_id = response['username']
 
         if tilde_id.endswith(last + '1') or allow_duplicates:
 
@@ -58,6 +58,9 @@ def build_groups(conference_group_id, default_params=None):
     Given a group ID, returns a list of empty groups that correspond to the given group's subpaths
     (e.g. Test.com, Test.com/TestConference, Test.com/TestConference/2018)
 
+    >>> [group.id for group in build_groups('ICML.cc/2019/Conference')]
+    [u'ICML.cc', u'ICML.cc/2019', u'ICML.cc/2019/Conference']
+
     '''
 
     path_components = conference_group_id.split('/')
@@ -78,6 +81,11 @@ def build_groups(conference_group_id, default_params=None):
     return sorted(groups.values(), key=lambda x: len(x.id))
 
 def post_group_parents(client, group, overwrite_parents=False):
+    '''
+    Helper function for posting groups created by build_groups function.
+
+    Recommended that this function be deprecated.
+    '''
     groups = build_groups(group.id)
 
     posted_groups = []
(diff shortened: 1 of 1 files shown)


openreview/tools.py
```diff
@@ -1,29 +1,29 @@
...
-    profile = None
-    try:
-        profile = client.get_profile(value)
-    except openreview.OpenReviewException as e:
-        # throw an error if it is something other than "not found"
-        if e[0][0] != 'Profile not found':
-            print "OpenReviewException: {0}".format(e)
-            return e
-    return profile
-
```
pmandler: Why remove this?  I've used it.

openreview/tools.py
```diff
@@ -1,29 +1,29 @@
...
+
+    If a profile with the same name exists and allow_duplicates is True, a profile is created with
+    the next largest number (e.g. if ~Michael_Spector1 exists, ~Michael_Spector2 will be created)
+    '''
+
     profile = get_profile(client, email)
 
     if not profile:
         response = client.get_tildeusername(first, last, middle)
-        tilde_id = response['username'].encode('utf-8')
```
pmandler: Can we get rid of the encode() now? It seems it's there for a reason.

openreview/tools.py
mspector: it seems redundant with the client's get_profile function. The only difference seems to be that this function returns the exception, rather than just raising it, which I think is unconventional behavior.

openreview/tools.py
mspector: I've imported `unicode_literals` at the top of this file, which should solve this problem. It makes it so that this code is compatible with either python 2 or 3. 

A slightly more detailed but still brief explanation, based on my understanding: python 2 will interpret `response['username']` as a byte string, not a unicode string (byte strings are just sequences of bytes, whereas unicode is a translation of bytes to characters. This allows unicode to represent characters used by some non-english names.) Since we want `tilde_id` to account for foreign characters, we have to "encode" it to unicode (utf-8 is a type of unicode) before assigning it to the variable.

In python 3, strings are interpreted as unicode by default, so explicit encoding isn't needed. By calling `from __future__ import unicode_literals`, we're telling the python 2 interpreter to treat strings as unicode by default, as python 3 does.

openreview/tools.py
pmandler: The difference is that it doesn't raise exceptions if the profile isn't found, it only raises exceptions if there is something wrong with the profile. It really should raise the exception if it's not the "not found" error.

melisabok: I agree with Pam, we should change the original get_profile function, for Not Found errors return None instead of an exception. 

It easier to handle the None values

openreview/tools.py
```diff
@@ -1,29 +1,29 @@
...
-    except openreview.OpenReviewException as e:
-        # throw an error if it is something other than "not found"
-        if e[0][0] != 'Profile not found':
-            print "OpenReviewException: {0}".format(e)
-            return e
-    return profile
-
 def create_profile(client, email, first, last, middle = None, allow_duplicates = False):
 
+    '''
```
melisabok: if the email exists return an error too

openreview/tools.py
mspector: ok, I think we're all in agreement then. I just didn't think it belonged in the tools module.

openreview/tools.py
```diff
@@ -78,6 +81,11 @@ def build_groups(conference_group_id, default_params=None):
     return sorted(groups.values(), key=lambda x: len(x.id))
 
 def post_group_parents(client, group, overwrite_parents=False):
+    '''
+    Helper function for posting groups created by build_groups function.
+
+    Recommended that this function be deprecated.
```
melisabok: is there a way to indicate deprecated functions in the doc? an annotation? @mohituniyal ?

melisabok:
I'm not sure about change the code in the same PR that we add documentation, I'm not sure if removing all the encode calls will raise some errors. Is there a way to test all these changes?

mspector:
@melisabok you're right, I'll remove the python 3 changes for now, that will be simpler for now while we're working on documentation.

melisabok:
Ok, may be we could create another branch with the python 3 compatibility and run some tests before merging the code, what do you think?

mspector:
yes, I agree. I have a branch I'm working on with python 3 compatibility, but no tests yet. I've been waiting for the profile and documentation changes to be merged in first

mspector:
I reverted the python 3 compatibility changes, except the changes to the print calls, because I think that's simple enough to include without testing

openreview/tools.py
```diff
@@ -1,24 +1,24 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
+from __future__ import print_function
```
melisabok: what does it do this import?

mspector: in python 2, print is a statement:
>>> print "hello world"

in python 3, print is a function:
>>> print("hello world")

mspector: I can change this back if you want. But there aren't really any consequences to keeping it in, and we will soon change it back anyway

melisabok: ok, if you say that this works fine, we can leave it.

melisabok:
Do we merge this code?

openreview/tools.py
melisabok: @mspector could you add @deprecated before the function name?

openreview/tools.py
```diff
@@ -1,24 +1,24 @@
...
+    Given email, first name, last name, and middle name (optional), creates and returns
+    a user profile.
+
+    If a profile with the same name exists, and allow_duplicates is False, an exception is raised.
+
+    If a profile with the same name exists and allow_duplicates is True, a profile is created with
+    the next largest number (e.g. if ~Michael_Spector1 exists, ~Michael_Spector2 will be created)
+    '''
+
     profile = get_profile(client, email)
```
melisabok: we can remove the get_profile function because it is being used here, could you revert that change too?

mspector: do you mean "can't" remove? so you want me to get_profile back into tools?

melisabok: I'm sorry, can't.

PR #125, opened 2018-07-10
Feature/documentation

mohituniyal:


docs/source/About OpenReview.rst
```diff
@@ -0,0 +1,38 @@
+Read more about OpenReview
```
mspector: can the name of this file be changed to something without spaces?

melisabok:
Add a link to the documentation in the README file.

docs/source/About OpenReview.rst
mohituniyal: Done. Changed name of "About OpenReview.rst" to "about.rst" and "autodoc.rst" to "api.rst".

melisabok:
It looks good to me, @mspector  do you want to do a final check?

openreview/tools.py
```diff
@@ -91,7 +90,6 @@ def build_groups(conference_group_id, default_params=None):
 
     return sorted(groups.values(), key=lambda x: len(x.id))
 
-@deprecated
```
mspector: are we giving up on this for now?

openreview/tools.py
```diff
@@ -626,17 +615,12 @@ def recruit_reviewer(client, email, first,
 
     hash_seed: a random number for seeding the hash.
     recruit_message: a formattable string containing the following string variables:
-        name
-        accept_url
-        decline_url
+    name
```
mspector: I think these tabs should stay, or there should be some other way of differentiating `name`, `accept_url`, and `decline_url` from the function variables (these three are variables to the formattable string, not the function)

openreview/tools.py
melisabok: Yes, it means to add a new library to the setup.py, I don't know if it is worth. When we have more deprecated functions we could add it.

PR #133, opened 2018-07-12
Clean up code

melisabok:
- Remove unused process_dir and webfield_dir parameters
- Don't set a cdate, let's leave that to ther server, remove epoche_time function
- Remove result sorting for groups and invitations. The response is already sorted by date, if you need another sorting we should implement it in the backend.
- Make mandatories fields for Group, Invitation, Tag and Notes
- Remove forum, reply, invitation from the Invitation constructor, they should be in the reply dictionary
- Add tcdate and tmdate that were missing in the classes.

### DIFF
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -8,16 +8,19 @@ import re
 import datetime
 import builtins
 
-def epoch_time():
-    return int((datetime.datetime.utcnow() - datetime.datetime(1970, 1, 1)).total_seconds()*1000)
-
 class OpenReviewException(Exception):
     pass
 
 class Client(object):
 
-    def __init__(self, baseurl = None, process_dir = '../process/', webfield_dir = '../webfield/', username = None, password = None):
-        """CONSTRUCTOR DOCSTRING"""
+    def __init__(self, baseurl = None, username = None, password = None):
+        """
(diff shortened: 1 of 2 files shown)


melisabok:
Yes, I think we should fix the scripts. We need to have a solid library. Let me update the version number

openreview/openreview.py
```diff
@@ -8,16 +8,19 @@
...
-
 class OpenReviewException(Exception):
     pass
 
 class Client(object):
 
-    def __init__(self, baseurl = None, process_dir = '../process/', webfield_dir = '../webfield/', username = None, password = None):
-        """CONSTRUCTOR DOCSTRING"""
+    def __init__(self, baseurl = None, username = None, password = None):
+        """
```
mohituniyal: Just a note: I do not think this doc-string will show up in the documentation.

melisabok: Do you know why?

I copied the example from Elastic Search:

https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/__init__.py#L174

mohituniyal: Sorry, looks like it does show up. I made an incorrect assumption that any function with a leading underscore does not show up in the documentation.

PR #135, opened 2018-07-13
Feature/get all venues

### LINKED ISSUE #128: Tools: write a method to get all the venues
It could be all the members of the host groups or all the groups that contains a submission invitation?

mohituniyal:
This branch is created to resolve #128 

Please review this and let me know if anything more needs to be done here.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -708,7 +708,4 @@ def get_all_venues(client):
         Returns a list of all the venues
         :arg client: Object of :class:`~openreview.Client` class
         '''
-        try:
-            return client.get_group("host").members
-        except openreview.OpenReviewException as e:
-            return "Could not retrieve venues. Exceptions observed: \"" + e[0][0] + "\""
+        return client.get_group("host").members
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -703,3 +703,9 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
             readers=[conference_id, chairs],
             signatories=[]))
 
+def get_all_venues(client):
+        '''
+        Returns a list of all the venues
+        :arg client: Object of :class:`~openreview.Client` class
+        '''
+        return client.get_group("host").to_json()['members']
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -708,4 +708,7 @@ def get_all_venues(client):
         Returns a list of all the venues
         :arg client: Object of :class:`~openreview.Client` class
         '''
-        return client.get_group("host").to_json()['members']
+        try:
+            return client.get_group("host").members
+        except openreview.OpenReviewException as e:
+            return "Could not retrieve venues. Exceptions observed: \"" + e[0][0] + "\""
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .pyc
 .DS_Store
 ./dist
+__pycache__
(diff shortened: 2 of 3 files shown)


openreview/tools.py
```diff
@@ -703,3 +703,12 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
             readers=[conference_id, chairs],
             signatories=[]))
 
+def get_all_venues(client):
```
melisabok: Could you add a test for this?
It should get the venues from the dev site and shouldn't be an empty result or an error

mohituniyal: Test added in a new file tests/test_tools.py

melisabok:
@mspector do you want to take a look?

openreview/tools.py
```diff
@@ -703,3 +703,12 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
...
+        '''
+        Returns a list of all the venues
+        :arg client: Object of :class:`~openreview.Client` class
+        '''
+        try:
+            return client.get_group("host").members
+        except openreview.OpenReviewException as e:
+            return "Could not retrieve venues. Exceptions observed: \"" + e[0][0] + "\""
```
mspector: What's the intended use case for this function? 

as it currently stands I'm not sure why you wouldn't just call client.get_group('host').members directly. It looks like the only difference with this function is that it handles exceptions.

By the way, I don't know if returning the string with the exception is the behavior we want. Why shouldn't the exception be raised if something goes wrong? if there is some reason, I would recommend that the value returned in case of exception should evaluate to False (e.g. None, or an empty list)

melisabok: I think this is good example to show to the new users that want to use the library. They don't know how to start. This is like a helper function that can give the user the list of active venues without knowing about what a group and a host group is. 

I was planing to show this function in the documentation, the same as tools.get_submission_invitations, as a tutorial of how to start to get data from OpenReview. These are the same features that we have implemented as static files for the crawlers. 

I agree that we should return None or an empty list but when there is a network error I would like to return an error instead. I'm not convinced about throwing an Exception but may be that is the way to go.

mspector: Ok, fair enough. If it's for tutorial purposes then it makes sense.

When you say that you would like to return an error, as opposed to throwing an exception, what is the difference?

melisabok: 1. I think it is shortcut for the users to get all the current venues that we support. Yes, it is a tutorial purpose but it is functional too. 

2. I don't want to do try/catch for every call that I do, I would like to get an error or a list of groups in the return call. I don't know if that exists in Python.

melisabok: May be I'm wrong, we should throw the exception.

mohituniyal: So, the options we have are:
1. Let the call raise an exception
2. Capture the exception and return None
3. Capture the exception and return an empty list
4. Capture the exception and return the exception's message
Currently, I have implemented the fourth one.
Whichever option we choose, maybe we can also make that a part of our coding practice.

So, what is the verdict on this @melisabok @mspector ?

melisabok: I think option 1, raise the exception,  is fine for now.

mspector: @melisabok @mohituniyal sounds like this is ready to be merged. any objections?

mohituniyal: Removed exception handing from def get_all_venues.
Note: If an exception (most probably a "Group not found" exception) is encountered it will also fail the circleCI test.

Issue #127, opened 2018-07-10
Tools: write a method to get all submission invitations


melisabok:
We can filter by open/closed submission. This could call to the get invitation by regex:

```
get_invitations(regex='.*/-/.*submission')
```

mohituniyal:
Resolved

PR #136, opened 2018-07-13
Feature/get submission invitations

mohituniyal:
Please review

Refer PR #127

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -700,7 +700,7 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
 
 def get_submission_invitations(client, open_only=False):
     '''
-    Returns a list of invitations ids visible to the client caccording to the parameter "open_only".
+    Returns a list of invitation ids visible to the client according to the value of parameter "open_only".
     
     :arg client: Object of :class:`~openreview.Client` class
     :arg open_only: Default value is False. This is a boolean param with value True implying that the results would be invitations having a future due date.
@@ -710,24 +710,15 @@ def get_submission_invitations(client, open_only=False):
     [u'machineintelligence.cc/MIC/2018/Conference/-/Submission', u'machineintelligence.cc/MIC/2018/Abstract/-/Submission', u'ISMIR.net/2018/WoRMS/-/Submission', u'OpenReview.net/Anonymous_Preprint/-/Submission']
     '''
     
-    try:
-        #Calculate the epoch for current timestamp
-        now = int(time.time()*1000)
-        
-        if open_only == True:    
-            invitations = client.get_invitations(regex='.*/-/.*[sS]ubmission.*',minduedate=now)
-        else:
-            invitations = client.get_invitations(regex='.*/-/.*[sS]ubmission.*')
-        
-        # For each group in the list, append the invitation id to a list
-        invitation_ids = []
-        
-        for inv in invitations :
-            invitation_ids.append(inv.id)
-        
-        return invitation_ids
-    except openreview.OpenReviewException as e:
-        return None
+    #Calculate the epoch for current timestamp
+    now = int(time.time()*1000)
+    duedate = now if open_only==True else None
+    invitations = client.get_invitations(regex='.*/-/.*[sS]ubmission.*',minduedate=duedate)
+    
+    # For each group in the list, append the invitation id to a list
+    invitation_ids = [inv.id for inv in invitations]
+    
+    return invitation_ids
 
 def get_all_venues(client):
         '''
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 import openreview
 import re
 import datetime
+import time
 
 super_user_id = 'OpenReview.net'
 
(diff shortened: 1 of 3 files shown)


openreview/tools.py
```diff
@@ -703,3 +698,39 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
...
+        
+        for g in groups :
+            
+            #Calculate the epoch for current timestamp
+            now = int(time.time()*1000)
+
+            # Assume the invitation is closed unless proven open
+            due = 0
+
+            if g.to_json()['duedate'] != None and now < g.to_json()['duedate']:
```
melisabok: no need to convert to json to get the duedate

openreview/tools.py
```diff
@@ -703,3 +698,39 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
...
+                due = 1
+            
+            if status == "open" and due == 1:
```
melisabok: @mspector I need your opinion about how to filter the invitations.

melisabok: I think we don't need to filter the result locally, we can add the filter to the get_invitations call.

mspector: do you think it should be filtered in the backend, or in the python library?

melisabok: the homepage of openreview gets the open submissions using these parameters:

```
regex: .*/-/[Ss]ubmission
minduedate: 1531505712680
```

I'm not sure if minduedate is implemented in the library.

openreview/tools.py
```diff
@@ -703,3 +698,39 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
...
+    '''
+    Returns a list of invitations ids according to the status.
+    
+    :arg client: Object of :class:`~openreview.Client` class
+    :arg status: Status filter for invitations. Default value is "all". This is a string with following possible value: all, open and closed.
+    '''
+    
+    try:
+        # Get a list of all group with invitations satisfying the regex
+        groups = client.get_invitations(regex='.*/-/.*submission')
```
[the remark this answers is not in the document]
mohituniyal: Corrected

openreview/tools.py
mohituniyal: Corrected

openreview/tools.py
```diff
@@ -703,3 +698,39 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
...
+    '''
+    Returns a list of invitations ids according to the status.
+    
+    :arg client: Object of :class:`~openreview.Client` class
+    :arg status: Status filter for invitations. Default value is "all". This is a string with following possible value: all, open and closed.
+    '''
+    
+    try:
+        # Get a list of all invitations satisfying the regex
+        invitations = client.get_invitations(regex='.*/-/.*[sS]ubmission.*')
```
melisabok: We can filter the invitation in the server using the parameters: minduedate.
 
In this case the status parameter can be renamed to open_only: [True, False]

if open_only is true then send minduedate = now
if open_only is false or None then don't send minduedate parameter. 

Does it make sense?

mohituniyal: Suggested changes have been implemented.

openreview/tools.py
```diff
@@ -703,6 +698,37 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
...
+            invitations = client.get_invitations(regex='.*/-/.*[sS]ubmission.*')
+        
+        # For each group in the list, append the invitation id to a list
+        invitation_ids = []
+        
+        for inv in invitations :
+            invitation_ids.append(inv.id)
+        
+        return invitation_ids
+    except openreview.OpenReviewException as e:
```
melisabok: I would follow the same rules that get_all_venues and don't catch the exception

openreview/tools.py
```diff
@@ -703,6 +698,37 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
...
+    Example Usage:
+    >>> get_submission_invitations(c,True)
+    [u'machineintelligence.cc/MIC/2018/Conference/-/Submission', u'machineintelligence.cc/MIC/2018/Abstract/-/Submission', u'ISMIR.net/2018/WoRMS/-/Submission', u'OpenReview.net/Anonymous_Preprint/-/Submission']
+    '''
+    
+    try:
+        #Calculate the epoch for current timestamp
+        now = int(time.time()*1000)
+        
+        if open_only == True:
```
melisabok: I would set the minduedate value in the if and do the call get_invitation once so we reduced the repeated code:

```
duedate = now if open_only else None
invitations = client.get_invitations(regex='.*/-/.*[sS]ubmission.*',minduedate=duedate)

```

openreview/tools.py
```diff
@@ -703,6 +698,37 @@ def post_submission_groups(client, conference_id, submission_invite, chairs):
...
+        
+        if open_only == True:    
+            invitations = client.get_invitations(regex='.*/-/.*[sS]ubmission.*',minduedate=now)
+        else:
+            invitations = client.get_invitations(regex='.*/-/.*[sS]ubmission.*')
+        
+        # For each group in the list, append the invitation id to a list
+        invitation_ids = []
+        
+        for inv in invitations :
```
melisabok: There is a nice way to iterate a list in a single line with Python called [for comprehension](https://docs.python.org/2/tutorial/datastructures.html)

```
invitation_ids = [i.id for i un invitations]
```

mohituniyal: All 3 changes have been applied.

Issue #131, opened 2018-07-12
Expose the login method as a public method


melisabok:
Shouldn't be a private method:

POST /login 

is part of the API and should be public.

mohituniyal:
Just to confirm, we want **Client** class's private method **__login_user(self.username, self.password)** to be publicly accessible?

melisabok:
Yes, I think it should be public. 

If the user doesn't send the username/password when it creates the Client instance then it can login later. 

This is issue is related to #130

mohituniyal:
Resolved with #137

Issue #130, opened 2018-07-12
Create an instance of Client class without username/password


melisabok:
This will be a guest user, no token needed. 

If not parameter set and not environment variable then create the client without a token.

mspector:
This may deserve its own issue, but it might also be a good idea to start thinking about more conventional ways of storing user credentials. Is storing them in environment variables secure enough? I'm not sure. 

I've also been thinking lately that the current way of logging in might leave the user's password too exposed (if not stored in environment variables, the user has to type their credentials in plain text in the python code).

melisabok:
I think we should remove the environment variables.

mohituniyal:
Resolved with #137

PR #137, opened 2018-07-17
Feature/guest login

mohituniyal:
With this change the Client class can be instantiated for a guest user i.e. without a user-id and password. 
Also, the login_user function has been made publicly accessible now as a member of Client class.

Refer #130 #131

### DIFF
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -55,7 +55,7 @@ class Client(object):
         self.profiles_url = self.baseurl + '/user/profile'
         self.reference_url = self.baseurl + '/references'
         self.tilde_url = self.baseurl + '/tildeusername'
-        self.token = self.__login_user(self.username, self.password)
+        self.token = self.login_user(self.username, self.password)
         self.headers = {'Authorization': 'Bearer ' + self.token, 'User-Agent': 'test-create-script'}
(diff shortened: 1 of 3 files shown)


openreview/openreview.py
```diff
@@ -76,28 +81,32 @@ def __handle_response(self,response):
...
             except KeyError:
-                password = getpass.getpass("Please provide your OpenReview password: ")
-
-        self.user = {'id':username,'password':password}
+                pass
 
-        response = requests.post(self.login_url, json=self.user)
+        user = {'id':username,'password':password}
+        header = {'User-Agent': 'test-create-script'}
+        response = requests.post(self.login_url, headers=header, json=user)
```
mspector: referencing our conversation about this from the standup earlier today: so, if the user and password are None, I guess the server returns a guest token?

tests/test_client.py
```diff
@@ -1,12 +1,15 @@
-import openreview
+from openreview import *
```
mspector: why change this? you're using the openreview namespace below anyway, which is good practice (e.g. line 8)

is it so that you can avoid doing `openreview.openreview.OpenReviewException` on line 31? If so, we should fix the package structure.

PR #141, opened 2018-07-23
add iterget function/generator

mspector:
this is useful if you want to retrieve a very large number of notes/references/invitations/groups but don't want to make all the network calls upfront or store them in memory.

(it's been useful to me as I've been working with the DBLP data)

This should make the _get_all() function sort of useless. If you like this change we can remove that function.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -67,9 +67,9 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
 def build_groups(conference_group_id, default_params=None):
     '''
     Given a group ID, returns a list of empty groups that correspond to the given group's subpaths
-    
+
     (e.g. Test.com, Test.com/TestConference, Test.com/TestConference/2018)
-    
+
     >>> [group.id for group in build_groups('ICML.cc/2019/Conference')]
     [u'ICML.cc', u'ICML.cc/2019', u'ICML.cc/2019/Conference']
     '''
@@ -179,7 +179,7 @@ def get_bibtex(note, venue_fullname, year, url_forum=None, accepted=False, anony
 def subdomains(domain):
     '''
     Given an email address, returns a list with the domains and subdomains.
-    
+
     >>> subdomains('johnsmith@iesl.cs.umass.edu')
     [u'iesl.cs.umass.edu', u'cs.umass.edu', u'umass.edu']
     '''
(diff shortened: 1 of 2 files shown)


melisabok:
So, let's remove _get_all

openreview/tools.py
```diff
@@ -345,13 +367,13 @@ def get_all_tags(client, invitation, limit=1000):
     '''
     Given an invitation, returns all Tags that respond to it, ignoring API limit.
     '''
-    return _get_all(client.get_tags, invitation, limit)
+    return [t for t in iterget(client.get_tags, invitation=invitation, limit=limit)]
```
melisabok: can you do?

```
return list(iterget(client.get_tags, invitation=invitation, limit=limit))
```

Based on:
https://stackoverflow.com/questions/3790848/fastest-way-to-convert-an-iterator-to-a-list

mspector: oh! I didn't realize you could do that. i will change it

openreview/tools.py
```diff
@@ -324,38 +324,43 @@ def replace_members_with_ids(client, group):
...
+        if len(batch) < params['limit']:
             done = True
-    return objects
 
 def get_all_tags(client, invitation, limit=1000):
     '''
     Given an invitation, returns all Tags that respond to it, ignoring API limit.
     '''
-    return _get_all(client.get_tags, invitation, limit)
+    return list(iterget(client.get_tags, invitation=invitation, limit=limit))
```
melisabok: Does it work in Python? I haven't tried it.

melisabok: May be you can add a test :)

mspector: yeah it does. speaking of which, should I be adding a test for this? I'm not sure how we've been doing it

mspector:
@melisabok @mohituniyal I added a test, but I'm not sure if I'm setting it up correctly. Maybe we can discuss at our meeting?

tests/test_tools.py
```diff
@@ -15,3 +16,20 @@ def test_get_submission_invitations(self):
     def test_get_all_venues(self):
         venues = openreview.get_all_venues(self.client)
         assert venues, "Venues could not be retrieved"
+
+    def test_iterget(self):
+        data_size = 1000
```
melisabok: I would write the test doing a get against the server that is now dev.openreview.net.

mspector:
@melisabok @mohituniyal is this ok to merge?

PR #143, opened 2018-07-24
add hash key generation to recruit_reviewer function

mspector:
instead of adding a function get_hash, I'm inserting the hash key generation directly into the recruit_reviewers function. My reasoning is that this hash manipulation isn't happening anywhere other than here, and in the corresponding recruitment process functions.


melisabok:
Sounds good to me

PR #144, opened 2018-07-25
Feature/get pdf

mohituniyal:
Added methods get_pdf and put_pdf

get_pdf takes the argument note-id to get the associated pdf.
put_pdf takes the path of the pdf to be uploaded

Note that no checks have been placed in put_pdf method to avoid a guest user uploading files.

Please review.


openreview/openreview.py
```diff
@@ -207,6 +208,44 @@ def get_profiles(self, email_or_id_list):
...
+        params = {}
+        params['id'] = id
+
+        headers = self.headers.copy()
+        headers['content-type'] = 'application/pdf'
+
+        with open(fname) as f:
+            response = requests.put(self.pdf_url, files={'data': f}, headers = headers)
+
+        return response.content.strip("{}").split(':')[1].strip('\"')
```
mspector: I can see why this works, but it might be a little clearer if instead you convert the response content to a dict (like so: `response_dict = json.loads(response.content)`) and then return `response_dict['url']`

@melisabok what do you think?

mohituniyal: Thanks for the suggestion Michael. Implemented this.

openreview/openreview.py
```diff
@@ -207,6 +208,45 @@ def get_profiles(self, email_or_id_list):
...
+        '''
+        params = {}
+        params['id'] = id
+
+        headers = self.headers.copy()
+        headers['content-type'] = 'application/pdf'
+
+        response = requests.get(self.pdf_url, params = params, headers = headers)
+
+        return response.content
```
melisabok: I think you need to handle the response here, like the other calls, to keep consistency.

openreview/openreview.py
```diff
@@ -207,6 +208,45 @@ def get_profiles(self, email_or_id_list):
...
+        params = {}
+        params['id'] = id
+
+        headers = self.headers.copy()
+        headers['content-type'] = 'application/pdf'
+
+        with open(fname) as f:
+            response = requests.put(self.pdf_url, files={'data': f}, headers = headers)
+
+        response_dict = json.loads(response.content)
```
melisabok: handle the response here too.

mohituniyal: Fixed this. Please take a look at the changes to __handle_response

melisabok:
Thanks!

PR #145, opened 2018-07-27
get all references tool

mspector:
should I write a new test for this function? the "iterget" function is being tested, and the other "get all" functions don't have tests. but I can add some if people think I should.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -325,20 +325,18 @@ def replace_members_with_ids(client, group):
     group.members = ids + emails
     client.post_group(group)
 
-def iterget(get_function, **kwargs):
+def iterget(get_function, batch_size=1000, **params):
     '''
     Iterator over a given get function from the client.
     '''
     done = False
     offset = 0
-    params = {
-        'limit': 1000 if 'limit' not in kwargs else kwargs['limit']
-    }
 
     while not done:
-        params['offset'] = offset
-
-        params.update(kwargs)
+        params.update({
+            'offset': offset,
+            'limit': batch_size
+        })
         batch = get_function(**params)
         offset += params['limit']
         for obj in batch:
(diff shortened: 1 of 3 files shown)


melisabok:
Yes, please add a test that runs 'get_all_references' and return more than the limit.

openreview/tools.py
```diff
@@ -359,6 +359,12 @@ def get_all_notes(client, invitation, limit=1000):
     '''
     return list(iterget(client.get_notes, invitation=invitation, limit=limit))
 
+def get_all_references(client, invitation, limit=1000):
```
melisabok: Why do you need to pass a limit to get all the references?

melisabok: May be the limit should be set in the get_all_references implementation. The user that tried to get all the references doesn't want to know about limit it just want to get all the references. It would be nice to return the iterator instead.

melisabok: Good catch! Mongo is failing because it is trying to sort without an index, I will take a look!

mspector:
@melisabok I lost the comment where you said this, but in reference to the suggestion that we return an iterator instead of the whole list:

that's ok with me, but why not just use the iterget function then? or is it useful to have "iterget_references" or something?

mspector:
no, we can just call it "get_references" (or get_all_references). My question is just about whether or not we need the functions at all. The current functions do two things: they specify the iterget function's target object type (references, notes, or tags), and they wrap the result in a list. If we have the "get_all" functions return just an iterator, then the functions are not really doing that much. It's fine with me, but I'm not sure how useful it is.

(I really don't have a strong opinion about this)

melisabok:
I have a question, with an iterator the next request call is being executed when the iterator loop reaches the size limit, right?

In that case, I would return the iterator instead of a list, but call it get_references or get_references_iter or get_iter_references?

I would keep a function per type of entity so the user doesn't need to send the client function in the get_iter call.

mspector:
Yes, that's right. Ok, sounds good: so by default, the "get all" functions will return iterators.

mspector:
thanks @melisabok ! don't merge quite yet please, I want to think a little more about the function names (should the function name make it clear that it returns an iterator?) 

I also need to update the comments

mspector:
@melisabok @mohituniyal I changed the names and comments, so this should be ready now

tests/test_tools.py
```diff
@@ -37,3 +41,22 @@ def mock_get(limit=100, offset=0):
         notes_iterator = openreview.tools.iterget(self.client.get_notes)
         notes_list = list(notes_iterator)
         assert notes_list is not None, "Notes iterator failed"
+
+    def test_iterget_notes(self):
+        batch_size = 1000
+        notes_iterator = openreview.tools.iterget_notes(self.client, None, batch_size=batch_size)
```
melisabok: I'm not sure about passing the batch_size in the function, this should be an internal setting.

melisabok: Can you do?

```
openreview.tools.iterget_notes(self.client)
```

With no  invitation parameter

mspector: 1. We can make batch_size internal if you want, but it is optional right now, so the user doesn't have to touch it if they don't want. I just set the variable in the tests to be explicit. But if you think it's better to not have it available at all then I can take it out.

2. You can't call the iterget_notes function without the invitation, but I think you're right that you should be able to. I'll remove the invitation argument so that this is possible (which will mean that any invitation setting will be passed through the **params argument), but it'll break some old code

mohituniyal: This might be slightly off topic, but what do you guys think about some documentation explaining how to use **params for "iterget" ?

mspector: I can add some explanatory comments. **kwargs is a feature of the language, so I don't think we should explain it in too much depth, but I agree it could use some comments

melisabok: **params can not be any value, it should be only the parameters accepted in the get_references call, may be add a link to the doc of the get_references.

mspector: ah, right. so maybe the iterget_notes/references/tags functions shouldn't have **params, and should instead have the same arguments as their non-iterator equivalents.

iterget should still have **params though, in my opinion.

I'm going to come back to this after I finish setting up dummy sites for ICLR and NIPS, so melisa and zach can test

mspector:
@mohituniyal @melisabok I made the arguments explicit, and added some comments. if you like the comments, I can add them to the openreview.py functions too

PR #146, opened 2018-07-27
Feature/get preferred name

mspector:


mspector:
@melisabok @mohituniyal any thoughts about this? does this function belong here? I've found it useful during my data collection but I can put it elsewhere if you think I should (e.g. in the openreview-evidence repo)

openreview/tools.py
```diff
@@ -65,6 +65,32 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
...
+    profile = client.get_profile(group_id)
+    names = profile.content['names']
+    preferred_names = [n for n in names if n.get('preferred', False)]
+    if preferred_names:
+        primary_preferred_name = preferred_names[0]
+    else:
+        primary_preferred_name = names[0]
+
+    name_parts = []
+    if primary_preferred_name.get('first'):
```
melisabok: you could do:

```
primary_preferred_name.get('first', '')
```

and no need to use an if.

melisabok:
I'm not sure why are we going to need this function. But we can leave it here.

openreview/tools.py
mspector: if you do this, and there's no middle name, the list that gets passed into the join() function will be ["First", "", "Last"], which will put two spaces between "First" and "Last". This was the best way I could come up with to get around that

PR #150, opened 2018-08-15
python 3 compatibility changes

mspector:
tests are passing and this is ready to go! works with both python 2 and python 3.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
-from __future__ import print_function
+from __future__ import absolute_import, division, print_function, unicode_literals
 import openreview
 import re
 import datetime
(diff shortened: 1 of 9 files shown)


openreview/tools.py
```diff
@@ -351,39 +351,142 @@ def replace_members_with_ids(client, group):
...
+        else:
+            next_obj = self.current_batch[self.obj_index]
+            if (self.obj_index + 1) == len(self.current_batch):
+                self.update_batch()
+                self.obj_index = 0
+            else:
+                self.obj_index += 1
+            return next_obj
+
+    next = __next__
```
mspector: this line is the reason I had to turn iterget into an iterator class object, instead of a generator function. in python 2, iterables use the function `next()` to get the next item in the series, while in python 3, the function is called `__next__`. There isn't a way to reassign this in a generator function, as far as I know, so I had to rewrite it as an iterator class and reassign the function like this.

melisabok:
It looks good to me, do you want to change the version number?

mspector:
@melisabok I made one more change, which is to add the --universal flag "python setup.py bdist_wheel". this indicates that the package is usable with python 2 or 3:

https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels

PR #153, opened 2018-08-23
remove middlename validation in create_profile

mspector:


openreview/tools.py
```diff
@@ -35,7 +35,7 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
     profile = get_profile(client, email)
 
     if not profile:
-        response = client.get_tildeusername(first, last, middle)
+        response = client.get_tildeusername(first, last, None)
```
mohituniyal: Do we not want users to have middle names?

mspector: the idea here is to validate the name as if it's just firstname and lastname.

for example, if there is "Michael Spector", and I try to create "Michael Li Spector", it throws an error.

that being said, your question made me realize that this affects more than just validation. I need to do two calls for tilde IDs: one to validate first and last, and another to get the proper id for first, middle, and last.

melisabok: I wonder if we can do the validation doing client.get_profile(first, last) instead of get_tilderusername

openreview/tools.py
```diff
@@ -35,10 +35,20 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
     profile = get_profile(client, email)
 
     if not profile:
-        response = client.get_tildeusername(first, last, middle)
-        tilde_id = response['username']
 
-        if tilde_id.endswith(last + '1') or allow_duplicates:
+        try:
+            client.get_profile("~{first}_{last}1".format(first=first, last=last))
```
melisabok: We should implement the parameters, first, middle and last in the library:

https://openreview.net/user/profile?first=Melisa&last=Bok

This returns an array of profiles, if the result is empty then the profile doesn't exist

openreview/tools.py
```diff
@@ -35,10 +35,20 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
...
+            profile_exists = True
+        except openreview.OpenReviewException as e:
+            if any('Profile not found' in arg for arg in e.args):
+                profile_exists = False
+            else:
+                raise(e)
+
+        id_response = client.get_tildeusername(first, last, middle)
```
melisabok: This line can be inside the if (not profile_exists)...

openreview/tools.py
```diff
@@ -61,9 +71,9 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
             return profile
 
         else:
-            raise openreview.OpenReviewException('There is already a profile with this first: {0}, middle: {1}, last name: {2}'.format(first, middle, last))
+            raise openreview.OpenReviewException('There is already a profile with this first and last name: \"{first} {last}\"'.format(first=first, last=last))
     else:
-        return profile
+        raise openreview.OpenReviewException('There is already a profile with this email address: {}'.format(email))
```
melisabok: You are right, we should raise the exception.

openreview/tools.py
melisabok: I'm sorry I was wrong, get profile by first and last name doesn't return an exact match. I think we should use the previous validation:

```
client.gettildeusername(first , last)
```

and check the number

mspector:
@melisabok I think this is ready. let me know if you have any other suggestions

PR #157, opened 2018-08-28
Feature/update profile functions

mspector:
2nd attempt after passing tests:

simple change to remove the dependency on the client. I think it's better for the function to do less

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -70,7 +70,7 @@ def get_preferred_name(profile):
     Returns a string representing the user's preferred name, if available,
     or the first listed name if not available.
 
-    Accepts emails or tilde ids.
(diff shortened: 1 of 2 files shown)


openreview/tools.py
```diff
@@ -65,15 +65,14 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
     else:
         return profile
 
-def get_preferred_name(client, group_id):
+def get_preferred_name(profile):
     '''
     Returns a string representing the user's preferred name, if available,
     or the first listed name if not available.
 
     Accepts emails or tilde ids.
```
mohituniyal: I guess now this should say "Accepts profile object"

PR #159, opened 2018-09-07
Fix/validate middle in create profile

mspector:
this fixes a bug where duplicates could make it through the validation even when `allow_duplicates` is set to `False`, if a profile exists with first, middle, and last name, but not first and last.

example:

- Profile called ~Samira_Ebrahimi_Kahou1 exists.
- We try to create a new profile with first=Samira, middle=Ebrahimi, last=Kahou
- Validator checks for ~Samira_Kahou1 and finds nothing, so allows us to create ~Samira_Ebrahimi_Kahou2 (this is bad)

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -36,20 +36,35 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
 
     if not profile:
 
-        # validate only first and last names.
-        # this is so that we catch more potential collisions.
+        # validate the name with just first and last names,
+        # and also with first, middle, and last.
+        # this is so that we catch more potential collisions;
         # let the caller decide what to do with false positives.
-        validation_response = client.get_tildeusername(first, last, None)
 
-        # the username in validation_response will end with 1 if no other profiles with this first and last name exist.
-        if validation_response['username'].endswith('1'):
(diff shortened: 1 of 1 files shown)


openreview/tools.py
```diff
@@ -36,19 +36,34 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
...
+        username_full_unclaimed = username_response_full['username'].endswith('1')
+
+        if all([username_FL_unclaimed, username_full_unclaimed]):
             profile_exists = False
         else:
             profile_exists = True
 
-        id_response = client.get_tildeusername(first, last, middle)
-        tilde_id = id_response['username']
+        tilde_id = username_response_full['username']
```
mohituniyal: If a profile with name fname=A, lname=C (tilde = ~A_C_1) already exists, and the user tries to create a new profile with fname=A, mname=B, and lname=C, and say allow_duplicates=True, then the tilde id that the user gets is ~A_B_C_1 and not ~A_C_2. Is this ok ?

mspector: I think that's the expected behavior. If a middle name is passed into create_profiles, it should try to create a profile with the middle name included.

PR #160, opened 2018-09-17
move template filling functions into tools.py and remove invitation-s…

mspector:
…pecific functionality.

Doesn't change anything about the way the function works.


mspector:
tests are failing because the superuser password on dev has changed

mohituniyal:
I think we should add some comments to these functions while we are placing them in tools.py because these functions will be visible in the ReadTheDocs documentation.

mspector:
@mohituniyal what do we do about "private" functions? (I say "private" because python doesn't have any concept of truly private functions like Java, but the convention is to name functions with a leading underscore (e.g. `_fill_str`) when the function is not intended to be used by an outside caller)

mohituniyal:
> @mohituniyal what do we do about "private" functions? (I say "private" because python doesn't have any concept of truly private functions like Java, but the convention is to name functions with a leading underscore (e.g. `_fill_str`) when the function is not intended to be used by an outside caller)

By default Sphinx does not generate documentation for "private" functions. So, we can choose not to annotate them. I would still request a 1 liner for each function just for future reference.

PR #165, opened 2018-09-21
allow get profiles pagination

mspector:


openreview/openreview.py
```diff
@@ -201,21 +201,48 @@ def get_profile(self, email_or_id):
         profile = response.json()['profile']
         return Profile.from_json(profile)
 
-    def get_profiles(self, email_or_id_list):
+    def get_profiles(self, email_or_id_list, limit=1000):
```
melisabok: the limit should be internal because if you send a list of 2k ids then we need to return 2k profiles no matter the value of the limit

PR #167, opened 2018-09-21
Feature/allow lists in templates

mspector:
UPDATE: improved since last time.

this change allows users to generate invitations with templates that refer to fields which are lists.

Before the change, these fields would return a string representation of a list, which would cause problems down the line. See examples:

```python
# Before the change
>> template = {'readers': '<readers>'}
>> openreview.tools.fill_template(template, paper)
{
  'readers': "['reader1', 'reader2', 'reader3']"
}
```

```python
# After the change
>> template = {'readers': '<readers>'}
>> openreview.tools.fill_template(template, paper)
{
  'readers': ['reader1', 'reader2', 'reader3']
}
```

I did a check of all our current usages of `fill_templates` and it looks like we're basically just using it to template the `number` and `forum` fields. This change is backwards compatible in that respect.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -10,6 +10,7 @@ import openreview
 import re
 import datetime
 import time
+import ast
 from Crypto.Hash import HMAC, SHA256
 
 super_user_id = 'OpenReview.net'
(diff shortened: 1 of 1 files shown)


mspector:
@mohituniyal @melisabok this is no longer needed for ICLR. it may be useful in the future, but then again, it's probably not a good idea to change functionality when it's not needed, so I'm going to close it.

PR #168, opened 2018-09-24
Change create_profile to only look for endswith '1' because endswith(…

pmandler:
…) doesn't like special characters.  Add function create_profile_ref() for creating profile references.


openreview/tools.py
```diff
@@ -37,8 +37,7 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
     if not profile:
         response = client.get_tildeusername(first, last, middle)
         tilde_id = response['username']
-
-        if tilde_id.endswith(last + '1') or allow_duplicates:
+        if tilde_id.endswith('1') or allow_duplicates:
```
mspector: can you describe the nature of the issue this function and special characters? what kind of error are you getting? is it an encoding issue?

this isn't great because it won't catch the 11'th duplicate (e.g. ~Michael_Spector11). This may seem ridiculous until you consider that there could easily be 11 "Wei Li"s in the CS community.

openreview/tools.py
```diff
@@ -65,6 +64,28 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
...
+                                 invitation=existing_profile.invitation,
+                                 signatures=[source_id],
+                                 writers=[source_id],
+                                 content=new_profile_content)
+    try:
+        profile = client.update_profile(profile)
+        if debug:
+            # test if fails to get original
+            db_profile = client.get_profile(id)
+    except Exception as e:
```
mspector: you should use openreview.OpenReviewException here instead of just "Exception". it's not great to except just anything

openreview/tools.py
```diff
@@ -65,6 +64,28 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
...
+        print "Exception {}".format(e)
+        print id
+    if debug:
+        super_group = client.get_group('OpenReview.net')
+        if len(super_group.members) > 1:
+            print super_group.members
```
mspector: you should call `print` as a function in all these cases, because we're trying to keep things "bilingual" between python 3 and 2. one of the ways we do that is by importing the print_function from __future__ (see line 3).

openreview/tools.py
pmandler: I'm trying to reproduce the error.  I'll let you know when I do.

mspector: also, I just realized that this commit is on a branch that is quite old - you may want to do a merge locally before you go further, so that you don't need to merge things after the fact

pmandler: For ~Antonio_Cubero_Fernández1 it produces this error:
 'ascii' codec can't decode byte 0xc3 in position 4: ordinal not in range(128)

pmandler:
Can I just close this pull request to get rid of it?  Then I'll get up to the current version and add create_profile_ref() to that.

mspector:
Yes, that sounds like a good idea. Make sure to delete both this remote branch and the local version (but don't forget to save any code you want to keep before you do)

PR #172, opened 2018-10-03
Added few examples of using get_notes

mohituniyal:


melisabok:
We will need to upgrade the version number to upload the library to pip

melisabok:
I had to comment two tests that we causing errors in the server because we have another database in dev. I hope I can fix the build and run these tests in another environment. This works for now. If we merge it the library will automatically published in pip.

docs/source/workshop.rst
```diff
@@ -427,4 +426,22 @@ Users with appropriate access can comment on a submission and reply to other's c
...
+	>>>iclr18_submissions = client.get_notes(invitation="ICLR.cc/2018/Conference/-/Submission")
+
+Extracting comments made on a forum
+--------------------------------------
+
+Comments, just like submissions, are saved as notes. So they are also accessible using get_notes().
+Note that the invitation argument in get_notes() accepts regex.
+Consider the following example to get all the "Public Comments" made on submissions in ICLR 2019 conference.::
+
+	>>>iclr19_public_comments = client.get_notes(invitation="ICLR.cc/2019/Conference/-/Paper.*/Public_Comment")
```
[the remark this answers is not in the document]
mspector: if we're going to use python 3, it should be print() with parentheses.

but if this requires you to go back and change a lot of things, we can do it later

docs/source/workshop.rst
```diff
@@ -427,4 +426,44 @@ Users with appropriate access can comment on a submission and reply to other's c
...
+
+Extracting comments made on a forum
+----------------------------------------
+
+All comments made on a particular forum/submission can be extracted like this::
+
+	>>>iclr19_forum_comments = client.get_notes(forum="<forum-id>")
+
+Also, the public comments on a particular forum can be extracted like this::
```
mspector: You may want to point out that "Public_Comment" is an arbitrary string - it can be anything. It just so happens that we use this naming convention in ICLR.

docs/source/workshop.rst
```diff
@@ -427,4 +426,44 @@ Users with appropriate access can comment on a submission and reply to other's c
...
+Also, the public comments on a particular forum can be extracted like this:: 
+
+	>>>iclr19_forum_public_comments = client.get_notes(forum="<forum-id>", invitation="ICLR.cc/2019/Conference/-/Paper.*/Public_Comment")
+
+Accessing data in comments
+------------------------------
+
+The data in a comment, or basically Notes objects, can be accessed like this::
+
+	>>>print(iclr19_forum_public_comments[0].content["title"])
```
melisabok: @mspector Mohit did the right thing :)

docs/source/workshop.rst
mspector: I'm not 100% sure about this advice, though. It's true that it's an arbitrary string, but maybe we should be treating it like a convention, and encouraging (forcing?) other conferences to use it.

mohituniyal: @mspector I just added something to this end. Let me know if I should write more or maybe in a different way to convey this idea.

PR #178, opened 2018-10-11
Fixed add-assignments to correct readership of AnonReviewer groups

mohituniyal:


openreview/tools.py
```diff
@@ -648,23 +647,37 @@ def add_assignment(client, paper_number, conference, reviewer,
...
+        signatories = individual_group_params.get('signatories', default_signatories)[:]
+        signatories.append(anonreviewer_id)
+
+        signatures = individual_group_params.get('signatures', default_signatures)[:]
 
         individual_group = openreview.Group(
             id = anonreviewer_id,
-            **individual_group_params)
-
-        individual_group.readers.append(anonreviewer_id)
```
mohituniyal: Appending to individual_group.readers was causing modifications to individual_group_params, a parameter of this function and thus the change was getting propagated to the code calling this function.

mspector: ahh I see. nice job!

PR #182, opened 2018-10-15
Add & Remove assignment functions return affected groups now

mohituniyal:
Instead of printing the groups names, the functions remove_assignment & add_assignment now return the tuple (user_profile_id, affected_group_ids), except for the scenario when add_assign finds that the user is already in the target groups. This can be resolved by including status of the add/remove operation but I think we can add that change later and get this out now for ICLR.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -698,10 +698,13 @@ def add_assignment(client, paper_number, conference, reviewer,
     '''
     profile = get_profile(client, reviewer)
     user = profile.id if profile else reviewer
+    affected_groups = []
 
     if user not in parent_group.members:
         client.add_members_to_group(parent_group, user)
-        print("{:40s} --> {}".format(user, parent_group.id))
+        affected_groups.append(parent_group.id)
+    else:
+        print("{:40s} Already present in Group {}".format(user, parent_group.id))
 
     assigned_individual_groups = [a for a in individual_groups if user in a.members]
     if not assigned_individual_groups:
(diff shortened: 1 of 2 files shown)


openreview/tools.py
```diff
@@ -826,16 +836,21 @@ def assign(client, paper_number, conference,
...
+    user = ""
+
     if reviewer_to_remove:
-        remove_assignment(client, paper_number, conference, reviewer_to_remove,
+        user, changed_groups = remove_assignment(client, paper_number, conference, reviewer_to_remove,
                           parent_group_params, parent_label, individual_label)
+
     if reviewer_to_add:
-        add_assignment(client, paper_number, conference, reviewer_to_add,
+        user, changed_groups = add_assignment(client, paper_number, conference, reviewer_to_add,
```
mspector: If you have a `reviewer_to_remove` and a `reviewer_to_add`, I think `changed_groups` will only contain the result of `add_assignment`, right?

mohituniyal: Did you mean to say "If you have a reviewer_to_remove OR a reviewer_to_add..."? I am asking because we should not get both at the same time.
Also, "user" will contain the profile id, if available, while reviewer_to_add or reviewer_to_remove may contain email ids.

melisabok: I would like to have two different methods, one to add and another to remove. It is confusing to have both functionalities in a single place.

mohituniyal: @mspector Sorry, somehow I was under the assumption that this script does either reviewer addition or deletion at a time. I will fix this.

melisabok: We could mark 'assign' function as deprecated and use add_assignment and remove_assignment from the scripts.

mspector: You're right, that docstring is ambiguous. I like Melisa's suggestion.

mohituniyal: Done. Please take a look @mspector and @melisabok.

PR #183, opened 2018-10-17
pycrypto-->pycryptodome

mspector:
inspired by Jeshuran's troubles. See: https://stackoverflow.com/questions/41813030/problems-with-installation-pycrypto-in-python-3-6


melisabok:
I don't think we are using this library anymore, could we remove it?

pmandler:
fine with me

PR #188, opened 2018-11-02
Feature/get profiles by name

mspector:
I have found these functions to be useful during my data collection and uploading... what do you guys think?

### DIFF
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -201,7 +201,15 @@ class Client(object):
         profile = response.json()['profile']
         return Profile.from_json(profile)
 
-    def get_profiles(self, email_or_id_list, limit=1000):
+    def get_profiles_by_name(self, first, last):
+        '''
+        Returns a single profile by first and last name
+        '''
+        response = requests.get(self.profiles_url, params = {'first': first, 'last': last}, headers = self.headers)
+        response = self.__handle_response(response)
+        return [Profile.from_json(p) for p in response.json()['profiles']]
+
+    def get_profiles(self, email_or_id_list):
         """
         |  If the list is tilde_ids, returns an array of profiles
         |  If the list is emails, returns an array of dictionaries with 'email' and 'profile'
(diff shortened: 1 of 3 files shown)


openreview/openreview.py
```diff
@@ -1055,6 +1064,23 @@ def to_json(self):
             body['tauthor'] = self.tauthor
         return body
 
+    def get_names(self):
```
melisabok: I like this!

tests/test_client.py
```diff
@@ -113,6 +113,9 @@ def test_get_profiles(self, client):
         assert '~Melisa_Bok1' in profiles[1].id
         assert '~Andrew_McCallum1' in profiles[0].id
 
+        assert 'Andrew McCallum' in profiles[0].get_names()
```
melisabok: is it possible to add a test to get the profiles by first and last name?

openreview/openreview.py
```diff
@@ -200,7 +200,15 @@ def get_profile(self, email_or_id):
         profile = response.json()['profile']
         return Profile.from_json(profile)
 
-    def get_profiles(self, email_or_id_list, limit=1000):
+    def get_profiles_by_name(self, first, last):
```
melisabok: Could it be possible to use the same function signature? I mean to have a single 'get_profiles' call?

openreview/openreview.py
```diff
@@ -200,7 +200,15 @@ def get_profile(self, email_or_id):
...
+        '''
+        Returns a single profile by first and last name
```
mohituniyal: I think the doc-string needs to be corrected, since the function is returning a list and is called get_profiles_by_name

openreview/openreview.py
mspector: I considered calling it just "names()", which do you like better?

openreview/openreview.py
mspector: I wrote this a while ago, and I think there was a reason I didn't do it that way, but I'm not certain. Let me try to make it work again, maybe it the problem will reveal itself (or maybe I'll be able to do it)

melisabok:
I think I'm going to close this PR and create a new one with the new profiles API integration.

melisabok:
I'm closing this PR, new endpoint implemented here: https://github.com/iesl/openreview-py/pull/206

PR #191, opened 2018-11-15
Feature/conference builder

melisabok:
First version of the conference builder (WORK IN PROGRESS):

- Create a conference with all the hierarchy of groups and landing pages.
- Setup a home page.
- Open submissions.
- Close submissions.
- Recruit reviewers and area chairs.
- Set program chairs.

Only working for double blind submission type like ICLR or AKBC. 

BONUS TRACK:

- Documentation.
- Unit tests against an openreview clean instance. 
- UI tests with Selenium 💯 .


openreview/conference/builder.py
```diff
@@ -0,0 +1,349 @@
...
+        options = {
+            'reviewers_id': reviewers_id,
+            'reviewers_declined_id': reviewers_declined_id,
+            'hash_seed': hash_seed
+        }
+        invitation = self.invitation_builder.set_reviewer_recruiter_invitation(self.id, options)
+        invitation = self.webfield_builder.set_recruit_page(self.id, invitation, self.get_homepage_options())
+        recruit_message = '''Dear {name},
+
+        You have been nominated by the program chair committeee of ''' + self.id + ''' to serve as a reviewer.  As a respected researcher in the area, we hope you will accept and help us make the conference a success.
```
mspector: in the past, we've used a variable called something like `short_phrase` for this kind of thing. So instead of using the ID `AKBC.ws/2019/Conference`, you could set `self.short_phrase = 'AKBC 2019'`, and that could be used here instead.

openreview/conference/builder.py
```diff
@@ -0,0 +1,349 @@
...
+        if subject_areas:
+            content['subject areas'] = {
+                'order' : 5,
+                'description' : "Select or type subject area",
+                'values-dropdown': subject_areas,
+                'required': True
+            }
+
+        order = 10
+        for field in additional_fields:
```
mspector: I'm not certain about this, but:

It might be nice if the structure of the objects in `additional_fields` was the same as the structure of the objects in "content". That might also open up the possibility of inserting additional fields elsewhere in the list, rather than only allowing them to appear at the end.

But again, I'm not certain if this is a good idea

mspector: instead of a counter you could do this:

`for order, field in enumerate(additional_fields, start=10):`

openreview/conference/invitation.py
```diff
@@ -0,0 +1,101 @@
...
+            },
+            'content': invitations.submission
+        }
+
+        reply = self.__build_options(default_reply, options.get('reply', {}))
+
+
+        with open(os.path.join(os.path.dirname(__file__), 'templates/submissionProcess.js')) as f:
+            content = f.read()
+            content = content.replace("var SHORT_PHRASE = '';", "var SHORT_PHRASE = '" + conference_id + "';")
```
mspector: another place where `conference.short_phrase` could be used

openreview/conference/builder.py
```diff
@@ -0,0 +1,349 @@
...
+
+    @classmethod
+    def homepage_webfield_template(cls):
+        return "noBlindConferenceWebfield.js"
+
+class DoubleBlindConferenceType(ConferenceType):
+
+    @classmethod
+    def homepage_webfield_template(cls):
+        return "homepage.js"
```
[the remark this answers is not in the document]
melisabok: Do you mean the name of the template file? Yes, I should name it more informative.

openreview/conference/builder.py
```diff
@@ -0,0 +1,349 @@
...
+            'pdf': {
+                'description': 'Upload a PDF file that ends with .pdf',
+                'order': 9,
+                'value-regex': 'upload',
+                'required':True
+            }
+        }
+
+        if subject_areas:
+            content['subject areas'] = {
```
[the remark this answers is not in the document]
melisabok: I don't remember that we defined that. Technically is possible. But this is a good time to make a decision.

openreview/conference/builder.py
melisabok: I will add the short phrase field, thanks!

openreview/conference/builder.py
melisabok: Yes, I thought about that. I like to iterate a list instead of an object because with the list you keep the order. Other option is to have the structure of the content and validate the order. It can be collisions between them and that requires to change all the content. Let me think about it.

openreview/conference/builder.py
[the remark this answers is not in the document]
mspector: Yeah. My personal opinion is that field names should have one of the following patterns:

1) "Human Readable": fields are optimized for readability by formatting the string like a title, and can contain spaces.
2) "machine_readable": fields are optimized for compatibility by formatting the string in all lowercase and using underscores instead of spaces

right now it feels like we're somewhere in between

melisabok:
I replied all @mspector comments, I prefer to wait an merge all this PR together. Thanks!

zbialecki:
@melisabok I have a few changes to the webfield templates I'm working on too, we should hold off merging until those are ready too

melisabok:
Sure,  I'm still working on this, we may have conflicts but it is fine! 

There are 2 conferences that are using this code now: COLT and MIDL.

docs/source/conference_builder.srt
```diff
@@ -0,0 +1,101 @@
...
+This call will setup an invitation to submit papers following the conference type, in this case is double blind, and with the submission deadline for Friday, October 5, 2019
+
+    >>> invitation = conference.open_submissions(due_date = datetime.datetime(2019, 10, 5, 23, 59))
+
+
+
+Close Submissions
+-----------------
+
+The submission will be closed when the current date passes the due date specified in open submissions call. In this case it will force the close of the submissions and it will disable the edition of the papers by the authors. They can not edit the submission anymore
```
pmandler: addition, not edition

openreview/conference/webfield.py
```diff
@@ -0,0 +1,138 @@
...
+
+        with open(os.path.join(os.path.dirname(__file__), 'templates/recruitResponseWebfield.js')) as f:
+            content = f.read()
+            content = content.replace("var CONFERENCE_ID = '';", "var CONFERENCE_ID = '" + conference_id + "';")
+            content = content.replace("var HEADER = {};", "var HEADER = " + json.dumps(header) + ";")
+            invitation.web = content
+            return self.client.post_invitation(invitation)
+
+
+    def set_author_page(self, conference_id, group, options = {}):
```
pmandler: What is the purpose of the author page?

openreview/conference/builder.py
```diff
@@ -0,0 +1,400 @@
...
+        invitation = self.client.get_invitation(id = self.id + '/-/Submission')
+        if invitation.reply['writers'] != self.id:
+            invitation.reply['writers'] = self.id
+
+        now = round(time.time() * 1000)
+        if invitation.duedate > now:
+            invitation.duedate = now
+        self.client.post_invitation(invitation)
+
+        # if freeze submissions then remove writers
```
pmandler: Should check the value of freeze_submissions.

openreview/conference/webfield.py
melisabok: It is like an Author Console to see all their assigned tasks for that conference

openreview/conference/builder.py
melisabok: You are right! Thanks!

openreview/conference/templates/authorWebfield.js
```diff
@@ -39,10 +41,11 @@ function load() {
   var invitationsP;
   var tagInvitationsP;
 
-  if (!user || _.startsWith(user.id, 'guest_')) {
```
melisabok: I have to revert this change, because it doesn't work. user is 'guest_1234839'

openreview/conference/templates/commentProcess.js
```diff
@@ -0,0 +1,90 @@
+function(){
```
pmandler: This deserves a description comment - sends emails to authors, PCs, ACs and reviewers depending on the readers of the note (or of the parent note?).

PR #192, opened 2018-11-28
make fields optionals and implement super field

melisabok:


mspector:
so, you want to add the "super" property to master? I don't mind, but why?

melisabok:
it was already implemented in master.

PR #194, opened 2018-12-10
Feature/parallel execution

melisabok:
Helper function to run various requests in parallel. 

If you run the function from a python notebook, you can use a progress bar. 

screen shot 2018-12-10 at 3 45 16 pm

@marshall62 it can be helpful to create the metadata notes.


mspector:
really cool!!

setup.py
```diff
@@ -17,7 +17,8 @@
           'requests>=2.18.4',
           'future',
           'nbsphinx',
-          'jupyter_client'
+          'jupyter_client',
```
mspector: what's this for?

melisabok: I removed it, I was trying to use a fancy progress bar.

PR #198, opened 2018-12-14
fix freeze submission invitation and add the conference to active venues

melisabok:


openreview/conference/builder.py
```diff
@@ -138,23 +138,30 @@ def close_submissions(self, freeze_submissions = True):
...
+
             notes_iterator = tools.iterget_notes(self.client, invitation = invitation.id)
             for note in notes_iterator:
                 if note.writers != [self.id]:
                     note.writers = [self.id]
                     self.client.post_note(note)
 
+        # Add venue to active venues
+        active_venues_group = self.client.get_group(id = 'active_venues')
+        self.client.add_members_to_group(active_venues_group, [self.get_id()])
```
mspector: I wonder if this should be done more explicitly..? what do you think?

melisabok: Don't we want to add the venue to active venues once the submission is closed, or do you think it should be optional?

mspector: maybe it should be optional. on the other hand, it's easy to change, and we'll find out if it's necessary to make it optional later.

PR #207, opened 2019-01-28
close comment invitations

melisabok:


melisabok:
I'm adding more tests but you can take a look in the meantime. 

This only implements the basic single blind review.

PR #213, opened 2019-02-04
Fix/iterget and details improvements

mspector:
some small improvements I made over the weekend.

1. if "details" param isn't provided, default to noDetails=True
2. allow the caller of iterget_notes to set a limit on the batch size


mspector:
just realized I should add some tests. Will do that soon.

melisabok:
I'm going to close this, an open an issue to finish it later.

melisabok:
Issue here: https://github.com/iesl/openreview/issues/1385

PR #216, opened 2019-02-04
Feature/double blind support

melisabok:
Double blind support is finally here.


openreview/conference/builder.py
```diff
@@ -109,6 +134,9 @@ def get_homepage_options(self):
             options['deadline'] = self.header.get('deadline')
         return options
 
+    def get_submissions(self, details = None):
```
pmandler: should this be get_blind_submissions()? If used on a single-blind, it will fail.

openreview/conference/builder.py
```diff
@@ -164,9 +192,58 @@ def close_submissions(self, freeze_submissions = True):
...
+                signatures= [self.id],
+                writers= [self.id],
+                readers= [self.id],
+                content= {
+                    "authors": ['Anonymous'],
+                    "authorids": [self.id],
+                    "_bibtex": None
+                })
+
+            posted_blind_note = self.client.post_note(blind_note)
```
pmandler: Why is the note posted twice?

openreview/conference/builder.py
```diff
@@ -200,11 +277,14 @@ def set_reviewers(self, emails):
...
-        notes_iterator = tools.iterget_notes(self.client, invitation = self.get_submission_id())
+        notes_iterator = self.get_submissions(details = 'original')
 
         for n in notes_iterator:
             group = self.__create_group('{conference_id}/Paper{number}'.format(conference_id = self.id, number = n.number), self.id)
-            self.__create_group('{number_group}/{author_name}'.format(number_group = group.id, author_name = self.authors_name), self.id, n.content.get('authorids'))
+            authorids = n.content.get('authorids')
+            if n.details and n.details['original']:
+                authorids = n.details['original']['content']['authorids']
+            self.__create_group('{number_group}/{author_name}'.format(number_group = group.id, author_name = self.authors_name), self.id, authorids)
```
pmandler: This section could use a comment.  I'm not sure what's going on - how is authorids used?

openreview/conference/builder.py
mohituniyal: 'number' is generated when it is posted the first time, using which we populate the correct per-paper author, reviewer, and area chair groups (as can be seen at line 227) which get added as readers for the blind note. Subsequent post is needed to save these changes.

openreview/conference/builder.py
melisabok: it gets all submissions not matter the conference configuration. The trick is in get_blind_invitation_id.

openreview/conference/builder.py
melisabok: In there future we need to figure out a way to this without posting twice. May be we can pass the number value to the backend.

openreview/conference/builder.py
melisabok: It is creating the author groups with the real author emails as members. The authorids could come from a single blind submission or double blind, that is what the if is doing.

PR #222, opened 2019-02-13
show blind submissions if exists otherwise shows original notes

melisabok:
@pmandler this PR shows the original notes in the Program Chairs console before the submission deadline.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -299,6 +299,8 @@ class Conference(object):
             posted_blind_note = self.client.post_note(posted_blind_note)
             blinded_notes.append(posted_blind_note)
 
+        # Update page with double blind submissions
+        self.__set_program_chair_page()
         return blinded_notes
 
     def open_bids(self, due_date, request_count = 50, with_area_chairs = False):
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -45,22 +45,22 @@ class Conference(object):
             return self.client.add_members_to_group(group, members)
 
     def __set_author_page(self):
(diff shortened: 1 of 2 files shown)


melisabok:
it is what we usually do to allow them to see the blinded notes.

melisabok:
it is not admin permissions in all openreview.

pmandler:
How is it not admin permissions in all openreview? What am I missing?

melisabok:
basically I'm adding the pc group as a member of the conference group. 
for example:

```
id: 'ICLR.cc/2019/Conference',
members: 'ICLR.cc/2019/Conference/Program_Chairs'
```

the conference group is like the 'admin' group. We create invitations and groups based on this id.

pmandler:
If it was adding it to get_conference_id() I'd understand, but adding it to ~Super_User1 doesn't make sense to me.

melisabok:
I'm not adding anything to the super user group, where do you see that?

pmandler:
Line 343 of builder.py

openreview/conference/builder.py
```diff
@@ -337,6 +339,8 @@ def open_meta_reviews(self, name, due_date = None, public = False):
 
     def set_program_chairs(self, emails):
         self.__create_group(self.get_program_chairs_id(), self.id, emails)
+        ## Give program chairs admin permissions
+        self.__create_group(self.id, '~Super_User1', [self.get_program_chairs_id()])
```
melisabok: This is saying that the signature of the Program Chairs group is the super user. Check the function parameters: 

```
def __create_group(group_id, group_owner_id, members)
```

pmandler:
I knew I was missing something.  Thanks.

PR #227, opened 2019-02-19
fix bid invitation and webfield

melisabok:


openreview/conference/templates/areachairWebfield.js
```diff
@@ -78,11 +79,17 @@ var loadData = function(result) {
   }
 
   var invitationsP = Webfield.getAll('/invitations', {
-    invitation: WILDCARD_INVITATION, invitee: true,
+    regex: WILDCARD_INVITATION, invitee: true,
```
zbialecki: Just out of curiosity, what is the difference between these 2? I know it's possible to do some basic regex queries using the id or invitation params.

openreview/conference/templates/tabsConferenceWebfield.js
```diff
@@ -153,7 +153,7 @@ function renderContent(notes, userGroups, activityNotes, authorNotes) {
       $('#your-consoles .submissions-list').append([
         '<li class="note invitation-link">',
           '<a href="/group?id=' + AREA_CHAIRS_ID + '" >',
-          AREA_CHAIRS_NAME.replace('_', ' ') + ' Console',
+          AREA_CHAIRS_NAME.replace(/_/g, ' ') + ' Console',
```
zbialecki: It seems like in some of the webfields you are using the variable `AREA_CHAIRS_NAME` and other you're using `AREA_CHAIR_NAME`. I'd rather just use 1 variable, or if they need to be different make the names more different.

openreview/conference/templates/areachairWebfield.js
melisabok: invitation parameter in the get invitation call doesn't exist(the backend ignores it), it is only supported in the get notes call. I guess this is a bug.

openreview/conference/templates/tabsConferenceWebfield.js
melisabok: AREA_CHAIR_NAME is the name of the group id, like: Senior_Programm_Committee. I think I have to use the prettyId here.

openreview/conference/templates/areachairWebfield.js
zbialecki: Oh right, I didn't notice that. But what about the id param? Could you use that instead of regex?

melisabok: We can use id or regex. 

If we use id, we need to set '.*' in the id parameter.
If we use regex, we can send any regex.

zbialecki: Good to know, thanks!

PR #229, opened 2019-02-21
Fix/set assignments

melisabok:
- Deploy assignments into groups, reviewers or area chairs. (Copied from previous scripts)
- Create reviewer recommendation invitations based on the results of the first reviewer matching. 
- Create a recommendation webfield to make recommendations by areas chairs. @zbialecki please review. 
- Set subject areas array through the builder at the beginning. We don't need to pass the array every time we need to use it.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -70,6 +70,9 @@ class Conference(object):
         if bid_invitation:
             return self.webfield_builder.set_bid_page(self, bid_invitation)
 
+    def __set_recommendation_page(self):
+        return True
+
     def set_id(self, id):
         self.id = id
 
@@ -312,6 +315,13 @@ class Conference(object):
         invitation.invitees = []
         return self.client.post_invitation(invitation)
 
+    def open_recommendations(self, due_date, reviewer_assingment_title):
+        notes_iterator = self.get_submissions()
+        assingment_notes_iterator = tools.iterget_notes(self.client, invitation = self.id + '/-/Paper_Assignment', content = { 'title': reviewer_assingment_title })
+        self.invitation_builder.set_recommendation_invitation(self, due_date, notes_iterator, assingment_notes_iterator)
+        return self.__set_recommendation_page()
+
+
     def open_comments(self, name, public, anonymous):
         ## Create comment invitations per paper
         notes_iterator = self.get_submissions()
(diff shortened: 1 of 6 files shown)


openreview/conference/builder.py
```diff
@@ -182,6 +188,12 @@ def set_homepage_layout(self, layout):
     def set_double_blind(self, double_blind):
         self.double_blind = double_blind
 
+    def set_subject_areas(self, subject_areas):
+        self.subject_areas = subject_areas
+
+    def get_subject_areas(self):
```
mspector: my opinion is that we don't need getters and setters like this, unless we want to do some kind of validation -- we can just set the subject_areas attribute directly

melisabok: I'm not sure what I'm going to do with getters and setters. In some cases I need them, in other cases I don't. I want to keep the same line.

openreview/conference/invitation.py
```diff
@@ -516,3 +516,83 @@ def set_reviewer_recruiter_invitation(self, conference_id, options = {}):
...
+                        profileId = group.get('userId'),
+                        bid = group.get('scores').get('bid'),
+                        tpms = group.get('scores').get('affinity'))
+                    )
+            else:
+                raise openreview.OpenReviewException('Assignment note not found for ' + note.id)
+            paper_recommendation_invitation = openreview.Invitation(
+                id = conference.get_id() + '/-/Paper{number}/Recommendation'.format(number = note.number),
+                super = recommendation_invitation.id,
+                invitees = [conference.get_program_chairs_id(), conference.get_id() + '/Paper{number}/Area_Chairs'.format(number = note.number)],
```
mohituniyal: If possible, I think we should avoid using 'Area_Chairs' here.

melisabok: As I talked with Michael a few days ago, we are doing the assignments using 'Area_Chairs' and 'AnonReviewers' strings, even in UAI. Otherwise I need to parametrize these strings everywhere in the builder and I decided to do this later.

zbialecki:
I added some changes to the recommendation webfield template, mainly removed some redundant stuff and refactored it a bit. I wasn't able to fully test it though, since I didn't have the right invitation.

melisabok:
Thanks Zach, I tested it and it is working perfectly. I renamed the tab.

mohituniyal:
Should I merge this or are we waiting for @mspector ?

PR #230, opened 2019-02-25
Fix/readers refactor

melisabok:
- Use the builder to set if the submission paper is public or not, instead of using open_submissions method. Because we use the public settings in two different places: single blind submissions invitation and double blind submission invitation. 
- When setting the conference double blind: you can specify who can read the original notes. By default is only authors and pcs, but you could add reviewers or area chairs (for example: COLT.)

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -195,7 +195,6 @@ class Conference(object):
 
     def get_submission_readers(self):
         return [
-            self.get_program_chairs_id(),
             self.get_area_chairs_id(),
             self.get_reviewers_id()
         ]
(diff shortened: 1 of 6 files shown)


openreview/conference/builder.py
```diff
@@ -188,6 +190,21 @@ def set_homepage_layout(self, layout):
...
         self.double_blind = double_blind
 
+    def set_submission_public(self, submission_public):
+        self.submission_public = submission_public
+
+    def get_submission_readers(self):
+        return [
+            self.get_area_chairs_id(),
+            self.get_reviewers_id()
+        ]
```
pmandler: Why doesn't this include PCs?

melisabok: We are including the conference group id in all the places. Given that PCs are members of conference group, I thought that wasn't necessary. But I can change it.

pmandler: A note to that effect somewhere would be useful, especially if we remove PCs from the conference group for some reason down the line.

melisabok: In that case, I will add the PCs here so it is more clear.

melisabok: ok, I added it.

PR #236, opened 2019-03-01
set a default profile to compute conflict of interest using at least the email address

melisabok:
Fixes https://github.com/iesl/openreview-py/issues/234

Michael, please review carefully, I may miss something. I'm creating a default profile for the missing one to use for the conflict of interest computation.


mspector:
looks good to me!

PR #238, opened 2019-03-05
Fix/conference builder minor bugs

marshall62:
some minor alterations to fix a bug about the submission invitation and to add a method to get the assignment notes.


openreview/conference/matching.py
```diff
@@ -338,14 +338,20 @@ def setup(self, affinity_score_file):
         reviewers_group = openreview.tools.replace_members_with_ids(client,reviewers_group)
         if not all(['~' in member for member in reviewers_group.members]):
             print('WARNING: not all reviewers have been converted to profile IDs. Members without profiles will not have metadata created.')
+        valid_reviewers_ids = [r for r in reviewers_group.members if '~' in r]
```
melisabok: I removed these changes on this PR: https://github.com/iesl/openreview-py/pull/236/files

I don't want to filter out the users with no profile.

marshall62: Made the changes to matching.py so that it has what is on master branch

PR #243, opened 2019-03-08
Fix/double blind optional

melisabok:
You can run this test to verify how it works:

tests/test_optional_double_blind.py

I want to keep the same field name for all the conferences and the builder will check the existence of this field to decide if overrides the author identities or not. 

Depending the setting set_submission_public is True or False, the field represents to reveal the identity to everyone or to the review committee. We could try to customize the field value to make this distinction.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -314,7 +314,7 @@ class Conference(object):
                 '_bibtex': None
             }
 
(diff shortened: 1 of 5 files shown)


melisabok:
Please, don't merge this if we see that is not need in the workshops.

PR #246, opened 2019-03-13
Fix/matching scores

melisabok:
Now the builder supports the following scores:

- Affinity score through an input file name
- TPMS score through an input file name
- Bids
- Recommendations
- Subject areas

There is an unit test for UAI 🥂 

NOTE: the recommendation interpolation is not working in python 2.7, I believe is the index function. I will fix it next week. No need to merge now, it is a reference to @mohituniyal for UAI matching.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -181,7 +181,7 @@ class Conference(object):
             recommendation_id = recommendation_id + '/Paper' + str(number)
 
         recommendation_id = recommendation_id + '/-/' + self.recommendation_name
-        return recommendation_id       
+        return recommendation_id   
 
     def get_registration_id(self):
         return self.id + '/-/' + self.registation_name        
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -408,9 +408,9 @@ class Conference(object):
                 authorids = n.details['original']['content']['authorids']
             self.__create_group('{number_group}/{author_name}'.format(number_group = group.id, author_name = self.authors_name), self.id, authorids)
 
-    def setup_matching(self, affinity_score_file = None):
+    def setup_matching(self, affinity_score_file = None, tpms_score_file = None):
         conference_matching = matching.Matching(self)
-        return conference_matching.setup(affinity_score_file)
+        return conference_matching.setup(affinity_score_file, tpms_score_file)
 
     def set_assignment(self, user, number, is_area_chair = False):
 
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -34,7 +34,7 @@ class Conference(object):
         self.submission_name = 'Submission'
         self.bid_name = 'Bid'
         self.recommendation_name = 'Recommendation'
-        self.registation_name = 'Registration'
+        self.registration_name = 'Registration'
         self.layout = 'tabs'
 
     def __create_group(self, group_id, group_owner_id, members = []):
@@ -184,7 +184,7 @@ class Conference(object):
         return recommendation_id   
 
     def get_registration_id(self):
-        return self.id + '/-/' + self.registation_name        
+        return self.id + '/-/' + self.registration_name        
 
     def set_conference_groups(self, groups):
         self.groups = groups
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -364,12 +364,12 @@ class Conference(object):
 
     def open_recommendations(self, start_date = None, due_date = None, reviewer_assingment_title = None):
         notes_iterator = self.get_submissions()
-        assingment_notes_iterator = None
+        assignment_notes_iterator = None
         
         if reviewer_assingment_title:
-            assingment_notes_iterator = tools.iterget_notes(self.client, invitation = self.id + '/-/Paper_Assignment', content = { 'title': reviewer_assingment_title })
+            assignment_notes_iterator = tools.iterget_notes(self.client, invitation = self.id + '/-/Paper_Assignment', content = { 'title': reviewer_assingment_title })
         
-        self.invitation_builder.set_recommendation_invitation(self, start_date, due_date, notes_iterator, assingment_notes_iterator)
+        self.invitation_builder.set_recommendation_invitation(self, start_date, due_date, notes_iterator, assignment_notes_iterator)
         return self.__set_recommendation_page()
 
     def open_registration(self, start_date = None, due_date = None, with_area_chairs = False):
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -33,6 +33,7 @@ class Conference(object):
         self.program_chairs_name = 'Program_Chairs'
         self.submission_name = 'Submission'
         self.bid_name = 'Bid'
+        self.recommendation_name = 'Recommendation'
         self.registation_name = 'Registration'
         self.layout = 'tabs'
 
@@ -173,6 +174,14 @@ class Conference(object):
 
     def get_bid_id(self):
         return self.id + '/-/' + self.bid_name
+
+    def get_recommendation_id(self, number = None):
+        recommendation_id = self.id + '/'
+        if number:
+            recommendation_id = recommendation_id + 'Paper' + str(number) + '/'
+
+        recommendation_id = recommendation_id + self.recommendation_name
+        return recommendation_id       
 
     def get_registration_id(self):
         return self.id + '/-/' + self.registation_name        
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -33,6 +33,7 @@ class Conference(object):
         self.program_chairs_name = 'Program_Chairs'
         self.submission_name = 'Submission'
         self.bid_name = 'Bid'
+        self.registation_name = 'Registration'
         self.layout = 'tabs'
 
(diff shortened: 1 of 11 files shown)


openreview/conference/invitation.py
```diff
@@ -547,14 +547,16 @@ def set_reviewer_recruiter_invitation(self, conference_id, options = {}):
     def set_recommendation_invitation(self, conference, start_date, due_date, notes_iterator, assingment_notes_iterator):
 
         assignment_note_by_forum = {}
-        for assignment_note in assingment_notes_iterator:
-            assignment_note_by_forum[assignment_note.forum] = assignment_note.content
+        if assingment_notes_iterator:
```
mspector: small typo here:

**assingment**_notes_iterator --> assignment_notes_iterator

openreview/conference/invitation.py
```diff
@@ -600,27 +605,144 @@ def set_recommendation_invitation(self, conference, start_date, due_date, notes_
...
+
+        subj_desc = 'To properly assign papers to reviewers, we ask that reviewers provide their areas of expertise from among the provided list of subject areas. Please submit your areas of expertise by selecting the appropriate options from the "Subject Areas" list.\n\n'
+
+        coi_desc = 'In order to avoid conflicts of interest in reviewing, we ask that all reviewers take a moment to update their OpenReview profiles with their latest information regarding work history and professional relationships. After you have updated your profile, please confirm that your OpenReview profile is up-to-date by selecting yes in the "Profile Confirmed" section.\n\n'
+
+        tpms_desc = 'In addition to subject areas, we will be using the Toronto Paper Matching System (TPMS) to compute paper-reviewer affinity scores. Please take a moment to sign up for TPMS and/or update your TPMS account with your latest papers. Then, please ensure that the email address that is affiliated with your TPMS account is linked to your OpenReview profile. After you have done this, please confirm that your TPMS account is up-to-date by selecting yes in the "TPMS Account Confirmed" section.\n\n'
+
```
mspector: what do you think about adding the data consent question here, too?

there may be good reasons to leave it out, I'm not sure. but it would be optional.

openreview/conference/invitation.py
```diff
@@ -600,27 +605,144 @@ def set_recommendation_invitation(self, conference, start_date, due_date, notes_
...
+        subj_desc = 'To properly assign papers to reviewers, we ask that reviewers provide their areas of expertise from among the provided list of subject areas. Please submit your areas of expertise by selecting the appropriate options from the "Subject Areas" list.\n\n'
+
+        coi_desc = 'In order to avoid conflicts of interest in reviewing, we ask that all reviewers take a moment to update their OpenReview profiles with their latest information regarding work history and professional relationships. After you have updated your profile, please confirm that your OpenReview profile is up-to-date by selecting yes in the "Profile Confirmed" section.\n\n'
+
+        tpms_desc = 'In addition to subject areas, we will be using the Toronto Paper Matching System (TPMS) to compute paper-reviewer affinity scores. Please take a moment to sign up for TPMS and/or update your TPMS account with your latest papers. Then, please ensure that the email address that is affiliated with your TPMS account is linked to your OpenReview profile. After you have done this, please confirm that your TPMS account is up-to-date by selecting yes in the "TPMS Account Confirmed" section.\n\n'
```
mspector: I wonder if we should include instructions for uploading papers to OpenReview, instead of instructing them to go to TPMS by default?

openreview/conference/builder.py
```diff
@@ -33,6 +33,8 @@ def __init__(self, client):
         self.program_chairs_name = 'Program_Chairs'
         self.submission_name = 'Submission'
         self.bid_name = 'Bid'
+        self.recommendation_name = 'Recommendation'
+        self.registation_name = 'Registration'
```
mspector: small typo: registation --> registration

openreview/conference/invitation.py
mspector: small typo here:

**assingment**_notes_iterator --> assignment_notes_iterator

openreview/conference/invitation.py
```diff
@@ -600,27 +605,144 @@ def set_recommendation_invitation(self, conference, start_date, due_date, notes_
...
+
+
+        # Create super invitation with a webfield
+        registration_parent_invitation = openreview.Invitation(
+            id = conference.get_id() + '/-/Parent/Registration',
```
mspector: I think this is just a matter of personal taste, but what do you think about naming this "Registration/Super" instead? or "Super/Registration"?

No strong argument about it, I think it mostly doesn't matter. But "Parent" doesn't feel descriptive enough to me.

openreview/conference/invitation.py
melisabok: I didn't do anything optional here. It is prepared for a conference like UAI. I will create an issue to make the registration fields configurable.

openreview/conference/invitation.py
melisabok: Good idea, this implies to create more groups/invitations?

I will create an issue to add this too.

openreview/conference/invitation.py
melisabok: Any name is fine for me. It is the actual UAI configuration. 

I think we shouldn't have to create a fake forum to submit the reviewer registration.

openreview/conference/invitation.py
mspector: sorry, what I mean is that consenting to have their data collected would be optional. the question itself would be there by default, so that we're always asking if we can use/publish their data.

openreview/conference/invitation.py
mspector: it may not necessarily imply that. can't users just upload papers to the OpenReview Archive page?

openreview/conference/invitation.py
melisabok: Yes, we can do that but in another PR. I had to add this registration phase just to calculate the subject areas scores for the matching.

openreview/conference/invitation.py
melisabok: Yes, don't they need to be added to the invitees list?

openreview/conference/invitation.py
mspector: Yeah, I agree, ideally we wouldn't do that

openreview/conference/invitation.py
melisabok: I created an issue here: https://github.com/iesl/openreview-py/issues/250

We can add more detail and I will implement it soon.

openreview/conference/invitation.py
melisabok: Ok, I renamed the invitation.

melisabok:
@mspector thanks for your review. Please check if the way to calculating the subjectArea and recommendation scores are ok and merge it please 🙏

openreview/conference/invitation.py
mspector: Just checked on this -- it looks like the OpenReview direct upload invitation is open to "~":

https://openreview.net/invitations?id=OpenReview.net/Archive/-/Direct_Upload

openreview/conference/matching.py
```diff
@@ -12,6 +12,11 @@ def clear(self, client, invitation):
         for note in note_list:
             client.delete_note(note)
 
+    def _jaccard_similarity(self, list1, list2):
```
mspector: I think this calculation is correct. but it might be easier to read if you use the set `union` function instead of calculating it manually (also, you can call pass a set into `len()`, you don't need to convert to a list)

example:
```
set1 = set(list1)
set2 = set(list2)
intersection = set1.intersection(set2)
union = set1.union(set2)
return len(intersection) / len(union)
```

openreview/conference/matching.py
```diff
@@ -61,6 +66,15 @@ def _build_entries(self, author_profiles, reviewer_profiles, paper_bid_jsons, sc
                     if bid_score != 0.0:
                         user_entry['scores']['bid'] = bid_score
 
+            reviewer_recommendations = [ t['tag'] for t in sorted(paper_recommendation_jsons, key=lambda x: x['cdate'])]
+            if reviewer_recommendations:
+                ## Set value between High(0.5) and Very High(1)
```
mspector: this looks like it's calculating the score based on the default bid score mapping (the idea being that recommendations should be valued the same as "High" or "Very High" bids). This makes sense to me. But we might run into problems in the future if the user wants to change the bid score mapping.

I'm not sure if there's a good solution for this right now, so I'm not recommending any change. But we should keep an eye out for this in the future.

openreview/conference/matching.py
mspector: also, python 3 doesn't require converting the final division to a float (division in python 3 does this automatically), but python 2.7 does. 

if you want to maintain backwards compatibility, you can either explicitly call `float()`, or you can add this import statement: `from __future__ import division`

openreview/conference/invitation.py
melisabok: Great, I will implement this in the issue #250

openreview/conference/matching.py
melisabok: Thanks! I made both changes.

openreview/conference/matching.py
melisabok: I agree, both conversions functions should come from each conference matching configuration. We don't have that right now. I believe once we remove the metadata notes we can translate this the matching repo.

PR #248, opened 2019-03-18
Fix two typos

pmandler:


melisabok:
@pmandler could you fix the test that is failing too?

It is in the file: test_double_blind_conference

PR #258, opened 2019-03-27
Fix/pc console

melisabok:
- Allow reviewer de-anonymization when creating the review invitations
- PC Console supports tilde signatures to display reviewer progress.

### DIFF
--- a/openreview/conference/templates/programchairWebfield.js
+++ b/openreview/conference/templates/programchairWebfield.js
@@ -4,6 +4,7 @@ var CONFERENCE_ID = '';
 var SUBMISSION_ID = '';
 var BLIND_SUBMISSION_ID = '';
 var HEADER = {};
+var SHOW_AC_TAB = false;
 
 var OFFICIAL_REVIEW_INVITATION = CONFERENCE_ID + '/-/Paper.*/Official_Review';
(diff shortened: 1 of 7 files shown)


openreview/conference/templates/programchairWebfield.js
```diff
@@ -49,8 +49,13 @@ var getOfficialReviews = function(noteNumbers) {
     var ratingExp = /^(\d+): .*/;
 
     _.forEach(notes, function(n) {
-      var num = getNumberfromGroup(n.signatures[0], 'Paper');
-      var index = getNumberfromGroup(n.signatures[0], 'AnonReviewer');
+      if (_.startsWith(n.signatures[0], '~')) {
+        var num = getNumberfromGroup(n.invitation, 'Paper');
+        var index = n.signatures[0];
```
melisabok: I have to use the signature id as the index because the review can be signed as ~Melisa_Bok1.

melisabok:
@pmandler I added another fix: show AC tab only when use_area_chairs is true.

pmandler:
With the new changes, the checks have failed.  I did test code before the latest checkin and it did fix the issue with the KGB PC console.

pmandler:
The fix worked on the Paper Status tab, but there are still problems on the Reviewers tab.

PR #263, opened 2019-04-01
Fix/check writable permissions

melisabok:


pmandler:
Still can't run recruit_reviewers().  
File "/Users/mandler/projects/openreview-py/openreview/conference/webfield.py", line 46, in set_landing_page
...
openreview.openreview.OpenReviewException: [{'type': 'forbidden', 'path': 'id', 'value': 'adai.ai', 'user': 'mandler@cs.umass.edu'}]
Where mandler@cs.umass.edu is a member of Program_Chairs.

pmandler:
Somehow I'm not getting these changes when I checkout fix/check-writable-permissions.

PR #265, opened 2019-04-05
Fix/get profile

melisabok:
Get the profile for the logged user using the token.

### DIFF
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -42,6 +42,7 @@ class Client(object):
             password = os.environ.get('OPENREVIEW_PASSWORD')
 
         self.token = token
+        self.profile = None
         self.groups_url = self.baseurl + '/groups'
         self.login_url = self.baseurl + '/login'
         self.register_url = self.baseurl + '/register'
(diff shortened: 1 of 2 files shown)


openreview/openreview.py
```diff
@@ -195,16 +197,19 @@ def get_tag(self, id):
...
-        response = requests.get(self.profiles_url, params = {att: email_or_id}, headers = self.headers)
+        params = {}
+        if email_or_id:
+            tildematch = re.compile('~.+')
+            if tildematch.match(email_or_id):
+                att = 'id'
+            else:
+                att = 'email'
+            params[att] = email_or_id
+        response = requests.get(self.profiles_url, params = params, headers = self.headers)
```
pmandler: What happens if the params are empty? I don't understand the purpose of the change.

melisabok: the backend returns the profile of the logged user, the one that is represented in the token. 

I need to get the user profile when the client is created using the token and not the username/password.

```
client = openreview.Client(token = 'sdfsdfjskd589sdflkj')
```

PR #269, opened 2019-04-05
Fix/review notifications

mohituniyal:
1. With this change the regex will use "AnonReviewer" only.
2. A confirmation email just for the writer of the review
3. Email for others in the submitted group
4. Added the review note's link at the end of the url being sent in the emails

### DIFF
--- a/openreview/conference/templates/reviewProcess.js
+++ b/openreview/conference/templates/reviewProcess.js
@@ -27,7 +27,7 @@ function(){
       if (note.readers.includes('everyone') || note.readers.includes(PAPER_AREACHAIRS)) {
         var areachair_mail = {
           groups: [PAPER_AREACHAIRS],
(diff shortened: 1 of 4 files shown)


openreview/conference/templates/reviewProcess.js
```diff
@@ -14,41 +12,50 @@ function(){
...
         };
         promises.push(or3client.or3request( or3client.mailUrl, author_mail, 'POST', token ));
       }
 
-      promises.push(or3client.addGroupMember(REVIEWERS_SUBMITTED, note.signatures[0], token));
-
-      return Promise.all(promises);
+      return Promise.all(promises)
+      .then(function(result){
+        return or3client.addGroupMember(REVIEWERS_SUBMITTED, note.signatures[0], token);
```
melisabok: I had to remove the submitted reviewer notification because this group can be empty and the /messages endpoint throws an error when the group is empty.

PR #272, opened 2019-04-09
Fix/ac optional comments

melisabok:
This PR fixes HSDIP process function errors, handling the AC groups as optional. 

Also the meta reviews can be posted by program chairs when there are no ACs set.

### DIFF
--- a/openreview/conference/invitation.py
+++ b/openreview/conference/invitation.py
@@ -417,12 +417,17 @@ class MetaReviewInvitation(openreview.Invitation):
         content = invitations.meta_review.copy()
 
         readers = ['everyone']
+        regex = conference.get_program_chairs_id()
+        invitees = [conference.get_program_chairs_id()]
+        private_readers = [conference.get_program_chairs_id()]
+
+        if conference.use_area_chairs:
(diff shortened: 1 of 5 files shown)


openreview/conference/invitation.py
```diff
@@ -415,12 +417,17 @@ def __init__(self, conference, name, note, start_date, due_date, public):
         content = invitations.meta_review.copy()
 
         readers = ['everyone']
+        regex = conference.get_program_chairs_id()
+        invitees = [conference.get_program_chairs_id()]
+        private_readers = [conference.get_program_chairs_id()]
+
+        if conference.use_area_chairs:
+            regex = conference.get_area_chairs_id(note.number)[:-1] + '[0-9]+'
```
pmandler: What if there are 10 or more ACs?  Too unlikely?

melisabok: + means one or more characters, so it is unlimited.

pmandler: I meant the [:-1] part only strips off one character.

melisabok: That is removing the 's' from the name, like Area_Chairs --> Area_Chair1

PR #280, opened 2019-04-29
allow for overrides to the request form and for options not available…

mspector:
… there.

This PR is needed for https://github.com/iesl/openreview-scripts/pull/1082

This change adds an optional parameter to the `get_conference` function that lets the caller pass in variables to override the content in the request form note.

It also checks for the presence of a new field in the request note, called "Custom Submission Fields", that lets the user add new fields to the submission form. It's not available in the request form invitation yet, but we could think about adding it in the future. For now, it's helpful for us to add custom fields with a simple script, without having to repeat the logic that checks for new dates and updates the webfield with them.

### DIFF
--- a/openreview/conference/helpers.py
+++ b/openreview/conference/helpers.py
@@ -1,9 +1,10 @@
 import openreview
 import datetime
 
-def get_conference(client, request_form_id):
+def get_conference(client, request_form_id, content_override_args={}):
 
     note = client.get_note(request_form_id)
+    note.content.update(content_override_args)
 
     if note.invitation not in 'OpenReview.net/Support/-/Request_Form':
         raise openreview.OpenReviewException('Invalid request form invitation')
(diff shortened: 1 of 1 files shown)


openreview/conference/helpers.py
```diff
@@ -61,9 +62,16 @@ def get_conference(client, request_form_id):
         'Submissions should be public, but reviews should be private.' == note.content.get('Open Reviewing Policy', '') :
         builder.set_submission_public(True)
 
+    # 'Custom Submission Fields' doesn't exist in the request form yet,
+    # but it can be added through a script via the
+    # `content_override_args` parameter.
+    additional_fields = note.content.get('Custom Submission Fields', {})
```
melisabok: We should add this field to the form

openreview/conference/helpers.py
```diff
@@ -1,9 +1,10 @@
 import openreview
 import datetime
 
-def get_conference(client, request_form_id):
+def get_conference(client, request_form_id, content_override_args={}):
```
melisabok: I would prefer to add a new revision to the note that overrides the values so we don't loose track of all the changes.

openreview/conference/helpers.py
mspector: I agree. I think to make this work, we'll need a special widget that lets users easily create fields that are formatted for `content.reply`, since they have a special structure (e.g. some set of options that lets them choose between "values", "values-regex", "values-dropdown", etc.)

melisabok: Yeah, that is still in progress by Zach. In the meantime, I would use value-json, the same widget that we use for Custom Loads in the matching that accepts a json string.

mohituniyal:
Covered this in https://github.com/iesl/openreview-py/pull/291 . Closing this PR.

PR #291, opened 2019-05-07
Feature/submission additional options

mohituniyal:
This is needed for https://github.com/iesl/openreview-scripts/pull/1087

### DIFF
--- a/openreview/conference/helpers.py
+++ b/openreview/conference/helpers.py
@@ -64,6 +64,8 @@ def get_conference(client, request_form_id):
 
     builder.set_override_homepage(True)
     conference = builder.get_result()
-    conference.open_submissions(start_date = submission_start_date, due_date = submission_due_date)
+
+    submission_additional_options = note.content.get('Additional Submission Options', {})
+    conference.open_submissions(start_date = submission_start_date, due_date = submission_due_date, additional_fields = submission_additional_options)
     conference.set_program_chairs(emails = note.content['Contact Emails'])
     return conference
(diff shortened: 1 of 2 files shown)


openreview/conference/invitation.py
```diff
@@ -27,6 +28,9 @@ def __init__(self, conference, start_date, due_date, readers, additional_fields,
         for field in remove_fields:
             del content[field]
 
+        if isinstance(additional_fields, str):
+            additional_fields = json.loads(additional_fields)
```
melisabok: I would prefer to do this in the helpers function and the SubmissionInvitation should expect always a dictionary.

openreview/conference/helpers.py
```diff
@@ -64,6 +65,11 @@ def get_conference(client, request_form_id):
 
     builder.set_override_homepage(True)
     conference = builder.get_result()
-    conference.open_submissions(start_date = submission_start_date, due_date = submission_due_date)
+
+    submission_additional_options = note.content.get('Additional Submission Options', {})
+    if submission_additional_options and isinstance(submission_additional_options, str):
```
melisabok: you don't need to check string instance, the note.content.get can return an None if there is no additional options.

mohituniyal: Removed the check for data type of _note.content.['Additional Submission Options']_ as it is always a string.
As discussed, we can not pass None in additional options due to the way we are processing additional options as of now. I will change that in a separate PR.

PR #295, opened 2019-05-09
Feature/edges bid interface

zbialecki:
This includes the Melisa's code from the features/edges branch


zbialecki:
The tests will pass once my feature/edges-bid-interface branch is merged into openreview

PR #297, opened 2019-05-13
pre-create reviewer and area chair groups (/invited and /declined groups)

mohituniyal:


mohituniyal:
Resolves https://github.com/iesl/openreview-py/issues/287

PR #305, opened 2019-05-29
Update bidWebfield template to use edges API

zbialecki:
Also update bid invitation builder to be an edge invitation. 

This branch replaces the previous feature/edges-bid-interface branch that had some issues with it's git history.

### DIFF
--- a/openreview/conference/templates/bidWebfield.js
+++ b/openreview/conference/templates/bidWebfield.js
@@ -10,8 +10,6 @@ var SUBMISSION_ID = '';
 var BID_ID = '';
 var SUBJECT_AREAS = '';
 
-var USER_SCORES_INVITATION_ID = CONFERENCE_ID + '/-/User_Scores';
-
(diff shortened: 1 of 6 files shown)


openreview/conference/templates/bidWebfield.js
```diff
@@ -10,8 +10,6 @@ var SUBMISSION_ID = '';
 var BID_ID = '';
 var SUBJECT_AREAS = '';
 
-var USER_SCORES_INVITATION_ID = CONFERENCE_ID + '/-/User_Scores';
```
melisabok: We can remove this for now but we need to remember that we may need to sort the papers by a type of edge.

melisabok:
I fixed the test, it was failing because the same venue was being used in another test and the webfield wasn't updated.

zbialecki:
> I fixed the test, it was failing because the same venue was being used in another test and the webfield wasn't updated.

Thanks!

PR #308, opened 2019-06-03
Fix/consoles

melisabok:
Console improvements:

- Show the submissions tab 'always' in the Author console.
- Set the submissions tab 'active' in the Author console.
- Create the conference author group and add all the paper author groups as members. It will be used to notify authors. 
- Show the meta-review column 'only' where there are area chairs in the PC Console.
- Fix tests.

### DIFF
--- a/openreview/conference/templates/authorWebfield.js
+++ b/openreview/conference/templates/authorWebfield.js
@@ -40,7 +40,7 @@ function load() {
   var invitationsP;
   var tagInvitationsP;
 
-  if (!user) {
+  if (!user || _.startsWith(user.id, 'guest_')) {
     authorNotesP = $.Deferred().resolve([]);
     invitationsP = $.Deferred().resolve([]);
     tagInvitationsP = $.Deferred().resolve([]);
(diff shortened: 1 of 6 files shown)


openreview/conference/templates/authorWebfield.js
```diff
@@ -113,21 +114,14 @@ function renderContent(authorNotes, invitations, tagInvitations) {
...
-      container: '#your-submissions',
-      search: { enabled: false },
-      displayOptions: paperDisplayOptions
-    });
-
-    $('.tabs-container a[href="#your-submissions"]').parent().show();
-  } else {
-    $('.tabs-container a[href="#your-submissions"]').parent().hide();
-  }
+  Webfield.ui.submissionList(authorNotes, {
```
zbialecki: Were you able to figure out why the other tabs had a big block of empty space at the top?

melisabok: Nop, I decided to keep the same order and set the submission tab active as default.

zbialecki: OK

PR #318, opened 2019-06-10
Add latex formatting to bibtex.  Also reverse author names by request.

pmandler:


openreview/tools.py
```diff
@@ -216,7 +217,16 @@ def capitalize_title(title):
         authors = 'Anonymous'
     else:
         first_author_last_name = note.content['authors'][0].split(' ')[-1].lower()
-        authors = ' and '.join(note.content['authors'])
+        if names_reversed:
```
melisabok: Pam, could you validate this output in one of our unit test?

For example, after we create a paper here:

https://github.com/iesl/openreview-py/blob/master/tests/test_single_blind_conference.py#L175

melisabok:
The build is failing and I guess because you need to include the new library:

https://circleci.com/gh/iesl/openreview-py/1752?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

pmandler:
How do I pip install wherever it is that pylatexenc needs to go?

melisabok:
I believe you have to add the dependency in the setup.py file:

https://github.com/iesl/openreview-py/blob/master/setup.py#L15

as an installation requirement.

melisabok:
Now it is failing because the bibtext test is executed before the [client test.](https://github.com/iesl/openreview-py/blob/master/tests/test_client.py) where expects to have 0 notes. @pmandler could you remove the test that is in the test client?

pmandler:
To run all of the tests do I simply run: "pytest tests/"  ?
Should I change the name of test_bibtex to test_xlatex.py so it is run later in the sequence?

melisabok:
No, I think there are other problems here. I will take a look.

melisabok:
Tests are passing now but I think we need to add another test to evaluate if names_reverse is true.

PR #322, opened 2019-06-12
Update basic conference homepage template to use pagination instead of infinite scroll

zbialecki:
Requires https://github.com/iesl/openreview/pull/1458


zbialecki:
The tabsConferenceWebfield.js template has now been updated to default to using the pagination widget. Because this depends on iesl/openreview#1458 the tests are failing for now, but once that is merged we can re-run them.

PR #324, opened 2019-06-14
Bug fixes for recruit reviewers and decision stage constructor

mohituniyal:
1. Fixed error in recruit_reviewers method. Error was that the function was not considering that the len of names can be different than the length of emails.
2. Fixed and error in decision stage default options. Error was that if the called passes None or empty list in options, then the constructor was not using the defaults. This was not the intended behavior. 
3. Minor changes to helpers.py in the way it reads forum's content fields.


openreview/conference/builder.py
```diff
@@ -650,7 +650,7 @@ def recruit_reviewers(self, emails = [], title = None, message = None, reviewers
 
         for index, email in enumerate(emails):
             if email not in set(reviewers_invited_group.members):
-                name = invitee_names[index] if invitee_names else None
+                name = invitee_names[index] if (invitee_names and index < len(invitee_names)) else None
```
pmandler: Can we throw an error if the invitee_names isn't the same length as emails so the user has a chance to correct the input?

mohituniyal: This is being called from a process function in the request form. Even if this raises an exception, the user would not find out about that and we will get an email about the exception. I think it is better to just send the emails out using the emails and the provided names. 
Another way of accepting emails and names would be to have the input be in csv format with (email, name) in each line. User can leave out the name corresponding to an email. 
Let me know if you feel that would work better.

pmandler: Just to be a pain - can we have it as an option?  I imagine it would be useful to upload a csv file for the first run of reviewers, and to put in individual email/name pairs to recruit additional reviewers.

mohituniyal: @pmandler these changes have been made in https://github.com/iesl/openreview-scripts/pull/1102

PR #325, opened 2019-06-14
Feature/improve docs

carlosmondra:
Improve docs in openreview.py (Client, Group, Tag, Invitation, and Profile classes)
Improve docs in tools.py

### DIFF
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -160,7 +160,7 @@ class Client(object):
         :return: Dictionary containing user information and the authentication token
         :rtype: dict
 
-        Example Usage:
+        Example:
 
         >>> res = client.activate_user('new@user.com', {
             'names': [
@@ -204,9 +204,9 @@ class Client(object):
         :type id: str
 
         :return: Dictionary with the group information
-        :rtype: dict
+        :rtype: Group
 
-        Example Usage:
+        Example:
 
         >>> group = client.get_group('your-email@domain.com')
         """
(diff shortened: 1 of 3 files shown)


openreview/openreview.py
```diff
@@ -20,17 +21,17 @@ class OpenReviewException(Exception):
     pass
 
 class Client(object):
-
+    """
+    :param baseurl: URL to the host, example: https://openreview.net (should be replaced by 'host' name)
+    :type baseurl: str, optional
```
pmandler: defaults to environmental var OPENREVIEW_BASEURL

openreview/openreview.py
```diff
@@ -20,17 +21,17 @@ class OpenReviewException(Exception):
...
+    :param username: OpenReview username
+    :type username: str, optional
+    :param password: OpenReview password
+    :type password: str, optional
+    :param token: Session token
+    :type token: str, optional
```
pmandler: What is a token in this case?  What does it do?

openreview/openreview.py
carlosmondra: Ok, I can add this

openreview/openreview.py
carlosmondra: I believe this is the Authentication token. I can be more specific and add this

carlosmondra:
So I added the Deprecated library in the setup.py file like so:
install_requires=[
          'pycryptodome',
          'requests>=2.18.4',
          'future',
          'nbsphinx',
          'tqdm',
          "Deprecated"
      ]
Just want to make sure that this is right

PR #339, opened 2019-07-10
Feature/groupby edges query

marshall62:
A new API method to support groupBy edge queries.    
Also adds "alternates" field to the config invitation built by the matching module.  This holds a number which is the percentage of top scores for each paper that should be generated as aggregate_score edges.   These top N% of each paper's reviewers forms the set of alternates reviewers.

-- It was necessary to create the get_edges_group method because when I tried creating a similar function in the matcher rather than in openreview.Client, the results coming back from openreview were status of 200 but no JSON.  The headers were correct (taken from the openreview.Client), so I couldn't figure out what it didn't like and, ultimately, we're going to want an API method to do this anyway.

### DIFF
--- a/openreview/conference/matching.py
+++ b/openreview/conference/matching.py
@@ -206,6 +206,12 @@ class Matching(object):
                         'description': 'Min number of reviews a person should do',
                         'order': 5
                     },
+                    'alternates': {
+                        'value-regex': '[0-9]+',
+                        'required': True,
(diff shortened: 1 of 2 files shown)


zbialecki:
Does this "top n%" method include a minimum? For very small conferences I think we need to default to at least 10 alternates

zbialecki:
@marshall62 Did you see my last question? What is the minimum number of aggregate score edges this generates?

PR #341, opened 2019-07-11
Feature - reviewer reassignment

mohituniyal:
Required changes for reviewer recruitment for ACs.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -46,8 +46,8 @@ class Conference(object):
         if group is None:
             return self.client.post_group(openreview.Group(
                 id = group_id,
-                readers = [self.id, group_owner_id, group_id].extend(additional_readers),
-                writers = [self.id].extend(additional_writers),
+                readers = [self.id, group_owner_id, group_id] + additional_readers,
+                writers = [self.id] + additional_writers,
                 signatures = [self.id],
                 signatories = [group_id] if is_signatory else [self.id],
                 members = members))
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -372,39 +372,40 @@ class Conference(object):
         if not self.submission_stage.double_blind:
             raise openreview.OpenReviewException('Conference is not double blind')
 
-        submissions_by_original = { note.original: note.id for note in self.get_submissions() }
+        submissions_by_original = { note.original: note for note in self.get_submissions() }
 
         self.invitation_builder.set_blind_submission_invitation(self)
         blinded_notes = []
 
         for note in tools.iterget_notes(self.client, invitation = self.get_submission_id(), sort = 'number:asc'):
-            note_id = submissions_by_original.get(note.id)
-            blind_note = openreview.Note(
-                id = note_id,
-                original= note.id,
-                invitation= self.get_blind_submission_id(),
-                forum=None,
-                signatures= [self.id],
-                writers= [self.id],
-                readers= [self.id],
-                content= {
-                    "authors": ['Anonymous'],
-                    "authorids": [self.id],
-                    "_bibtex": None
-                })
+            blind_note = submissions_by_original.get(note.id)
+            if not blind_note:
+                blind_note = openreview.Note(
+                    id = None,
+                    original= note.id,
+                    invitation= self.get_blind_submission_id(),
+                    forum=None,
+                    signatures= [self.id],
+                    writers= [self.id],
+                    readers= [self.id],
+                    content= {
+                        "authors": ['Anonymous'],
+                        "authorids": [self.id],
+                        "_bibtex": None
+                    })
 
-            posted_blind_note = self.client.post_note(blind_note)
+                blind_note = self.client.post_note(blind_note)
 
-            posted_blind_note.readers = self.submission_stage.get_blind_readers(self, posted_blind_note.number)
+                blind_note.readers = self.submission_stage.get_blind_readers(self, blind_note.number)
 
-            posted_blind_note.content = {
-                'authorids': [self.get_authors_id(number = posted_blind_note.number)],
-                'authors': ['Anonymous'],
-                '_bibtex': None #Create bibtext automatically
-            }
+                blind_note.content = {
+                    'authorids': [self.get_authors_id(number = blind_note.number)],
+                    'authors': ['Anonymous'],
+                    '_bibtex': None #Create bibtext automatically
+                }
 
-            posted_blind_note = self.client.post_note(posted_blind_note)
-            blinded_notes.append(posted_blind_note)
+                blind_note = self.client.post_note(blind_note)
+            blinded_notes.append(blind_note)
 
         # Update page with double blind submissions
         self.__set_program_chair_page()
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -13,6 +13,7 @@ class Conference(object):
 
     def __init__(self, client):
         self.client = client
+        self.request_form_id = None
         self.new = False
         self.use_area_chairs = False
         self.legacy_invitation_id = False
@@ -797,6 +798,7 @@ class ReviewStage(object):
 class CommentStage(object):
 
     def __init__(self, start_date = None, allow_public_comments = False, anonymous = False, unsubmitted_reviewers = False, reader_selection = False, email_pcs = False):
+        self.name = 'Comment'
         self.start_date = start_date
         self.allow_public_comments = allow_public_comments
         self.anonymous = anonymous
@@ -969,6 +971,9 @@ class ConferenceBuilder(object):
 
     def use_legacy_invitation_id(self, legacy_invitation_id):
         self.conference.legacy_invitation_id = legacy_invitation_id
+
+    def set_request_form_id(self, id):
+        self.conference.request_form_id = id
 
     def get_result(self):
 
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -41,13 +41,16 @@ class Conference(object):
         self.decision_stage = DecisionStage()
         self.layout = 'tabs'
 
-    def __create_group(self, group_id, group_owner_id, members = [], is_signatory = True):
-
+    def __create_group(self, group_id, group_owner_id, members = [], is_signatory = True, additional_readers = None, additional_writers = None):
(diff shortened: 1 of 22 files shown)


openreview/conference/builder.py
```diff
@@ -515,7 +515,13 @@ def set_reviewers(self, emails = []):
         notes_iterator = self.get_submissions()
 
         for n in notes_iterator:
-            self.__create_group(self.get_reviewers_id(number = n.number), self.id)
+            self.__create_group(
+                self.get_reviewers_id(number = n.number),
+                self.id,
+                is_signatory = True,
+                additional_readers = ['{conference_id}/Paper{number}/Area_Chairs'.format(conference_id = self.id, number = n.number)] if self.use_area_chairs else [],
```
melisabok: you can use:

```
self.get_area_chairs_id(number = number)
```

instead of building the string

openreview/conference/templates/areachairWebfield.js
```diff
@@ -600,6 +610,17 @@ var buildTableRow = function(note, reviewerIds, completedReviews, metaReview) {
   return [cellCheck, cell0, cell1, cell2, cell3, cell4, cell5];
 };
 
+var findNextAnonGroupNumber = function(paperNumber){
+  nextAnonNumber = 1;
+  paperReviewerNums = Object.keys(reviewerSummaryMap[paperNumber].reviewers);
+  for (var i = 1;; i++) {
```
melisabok: may be using a find here is better.

openreview/conference/templates/areachairWebfield.js
```diff
@@ -651,12 +672,94 @@ var registerEventHandlers = function() {
   });
 
   $('#group-container').on('click', 'a.collapse-btn', function(e) {
-    $(this).next().slideToggle();
+    // $(this).next().slideToggle();
```
melisabok: is this commented code ok?

openreview/conference/templates/areachairWebfield.js
```diff
@@ -651,12 +672,94 @@ var registerEventHandlers = function() {
...
     if ($(this).text() === 'Show reviewers') {
       $(this).text('Hide reviewers');
     } else {
       $(this).text('Show reviewers');
     }
+    // return false;
```
melisabok: same here, don't we need to return false?

openreview/conference/templates/areachairWebfield.js
```diff
@@ -651,12 +672,94 @@ var registerEventHandlers = function() {
...
+  });
+
+  $('#group-container').on('click', 'button.btn.assign-reviewer-button', function(e) {
+    var $link = $(this);
+    var paperNumber = $link.data('paperNumber');
+    var paperForum = $link.data('paperForum');
+    console.log(paperNumber);
```
melisabok: please remove the console logs

openreview/conference/templates/areachairWebfield.js
```diff
@@ -651,12 +672,94 @@ var registerEventHandlers = function() {
...
+        signatories: [CONFERENCE_ID + '/Paper' + paperNumber + '/AnonReviewer' + nextAnonNumber]
+      }),
+      promptMessage('Reviewer ' + view.prettyId(userToAdd) + ' has been assigned to paper ' + paperNumber)
+    );
+    var forumUrl = 'https://openreview.net/forum?' + $.param({
+        id: paperForum,
+        noteId: paperForum,
+        invitationId: CONFERENCE_ID + '/-/Paper' + paperNumber + '/Official_Review'
+      });
+    var lastReminderSent = localStorage.getItem(forumUrl + '|' + userToAdd);
```
melisabok: if the user is new why do we need to get the last reminder?

openreview/conference/templates/areachairWebfield.js
```diff
@@ -651,12 +672,94 @@ var registerEventHandlers = function() {
...
+          CONFERENCE_ID + '/Program_Chairs',
+          CONFERENCE_ID + '/Paper' + paperNumber + '/Area_Chairs',
+          CONFERENCE_ID + '/Paper' + paperNumber + '/AnonReviewer' + nextAnonNumber],
+        writers: [
+          CONFERENCE_ID + '/Program_Chairs',
+          CONFERENCE_ID + '/Paper' + paperNumber + '/Area_Chairs'],
+        signatures: [CONFERENCE_ID + '/Paper' + paperNumber + '/Area_Chairs'],
+        signatories: [CONFERENCE_ID + '/Paper' + paperNumber + '/AnonReviewer' + nextAnonNumber]
+      }),
+      promptMessage('Reviewer ' + view.prettyId(userToAdd) + ' has been assigned to paper ' + paperNumber)
```
melisabok: you need to wait until both promises are finished to prompt the message, also we need to catch the errors. For example:

```
$.when(
   Webfield.post(...),
   Webfield.post(...)
)
.then(function(results) {
   promptMessage
})
```

I think the errors are being catch in the Webfield.js code but I'm not sure.

openreview/conference/templates/areachairWebfield.js
```diff
@@ -651,12 +672,94 @@ var registerEventHandlers = function() {
...
+     $.when(
+      Webfield.delete('/groups/members', {
+        id: CONFERENCE_ID + '/Paper' + paperNumber + '/Reviewers',
+        members: [userId]
+      }),
+      Webfield.delete('/groups/members', {
+        id: CONFERENCE_ID + '/Paper' + paperNumber + '/AnonReviewer' + reviewerNumber,
+        members: [userId]
+      }),
+      promptMessage('Reviewer ' + view.prettyId(userId) + ' has been unassigned for paper ' + paperNumber)
```
melisabok: same here

openreview/conference/templates/areachairWebfield.js
mohituniyal: We do not need to do a toggle here as the widget does it by itself.

openreview/conference/templates/areachairWebfield.js
mohituniyal: Done. Sorry about that.

openreview/conference/templates/areachairWebfield.js
mohituniyal: This was mostly because I was not sure of the return for Storage.getItem() in case the item is missing. I have replaced it with null now.

openreview/conference/templates/areachairWebfield.js
zbialecki: The errors are being caught by the Webfield.post function. It will automatically show the error prompt if one of the requests fails

zbialecki: But yes I agree with Melisa, the promptMessage should be in the `.then` callback

openreview/conference/templates/areachairWebfield.js
```diff
@@ -651,12 +675,89 @@ var registerEventHandlers = function() {
...
       $(this).text('Show reviewers');
     }
+  });
+
+  $('#group-container').on('click', 'button.btn.btn-assign-reviewer', function(e) {
+    var $link = $(this);
+    var paperNumber = $link.data('paperNumber');
+    var paperForum = $link.data('paperForum');
+    var $currDiv = $('#' + paperForum + '-add-reviewer');
+    userToAdd = $currDiv.find('input').val().trim();
```
zbialecki: Is this var defined somewhere else or is it missing the `var`?

openreview/conference/templates/areachairWebfield.js
```diff
@@ -651,12 +675,89 @@ var registerEventHandlers = function() {
...
+    reviewerSummaryMap[paperNumber].expandReviewerList = true;
+    reviewerSummaryMap[paperNumber].sendReminder = true;
+    reviewerSummaryMap[paperNumber].enableReviewerReassignment = ENABLE_REVIEWER_REASSIGNMENT;
+    var $revProgressDiv = $('#' + paperForum + '-reviewer-progress');
+    $revProgressDiv.html(Handlebars.templates.noteReviewers(reviewerSummaryMap[paperNumber]));
+    return false;
+  });
+
+   $('#group-container').on('click', 'a.unassign-reviewer-link', function(e) {
+    var $link = $(this);
```
zbialecki: It looks like the indentation of this function is a bit off. Seems like only 1 space?

openreview/conference/templates/areachairWebfield.js
```diff
@@ -600,6 +612,18 @@ var buildTableRow = function(note, reviewerIds, completedReviews, metaReview) {
   return [cellCheck, cell0, cell1, cell2, cell3, cell4, cell5];
 };
 
+var findNextAnonGroupNumber = function(paperNumber){
```
zbialecki: This function is seems overly complex... basically you just want to find the next available reviewer number from the reviewers map right when there might be gaps in the keys? So instead of searching the whole list each time why don't you just check if `i === paperReviewerNums[i]` and if not `return i`? If you get through the whole list `return i + 1`

openreview/conference/templates/areachairWebfield.js
mohituniyal: It was missing the var. Corrected.

openreview/conference/templates/areachairWebfield.js
mohituniyal: Agreed. The function has been refactored as per suggestion.

openreview/conference/templates/areachairWebfield.js
mohituniyal: Agreed. This has been corrected.

openreview/conference/templates/areachairWebfield.js
```diff
@@ -651,12 +777,119 @@ var registerEventHandlers = function() {
...
+      id: CONFERENCE_ID + '/Paper' + paperNumber + '/Reviewers',
+      members: [userId]
+    })
+    .then(function(result) {
+      return Webfield.delete('/groups/members', {
+        id: CONFERENCE_ID + '/Paper' + paperNumber + '/AnonReviewer' + reviewerNumber,
+        members: [userId]
+      });
+    })
+    .then(() => promptMessage('Reviewer ' + view.prettyId(userId) + ' has been unassigned for paper ' + paperNumber));
```
melisabok: arrow functions are not supported in all the browsers

PR #351, opened 2019-07-16
Feature/change edge invitation names

marshall62:
Adds /Reviewing into the invitation names related to the matcher.   Made names match what Zach needs for edge browser.

### DIFF
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -13,7 +13,7 @@ import os
 import getpass
 import re
 import datetime
-
+from collections import defaultdict
 
 
 class OpenReviewException(Exception):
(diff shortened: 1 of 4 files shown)


openreview/openreview.py
```diff
@@ -566,6 +566,32 @@ def get_edges(self, id = None, invitation = None, head = None, tail = None, labe
 
         return [Edge.from_json(t) for t in response.json()['edges']]
 
+    def get_edges_group (self, invitation, groupby='head', select='tail', limit=None, offset=None):
```
melisabok: can we call it 'get_grouped_edges'?

PR #353, opened 2019-07-17
add delete_group function

mspector:


melisabok:
you may need to fix the tests

PR #360, opened 2019-07-18
add python-Levenshtein as a dependency

mspector:


mspector:
this works on my machine, and I don't fully understand the errors that CircleCI is running into ... I can pick this up again at some point, but should I leave it open?

PR #363, opened 2019-07-23
Add query parameter to the edge invitations

melisabok:


melisabok:
I'm going to set the reviewers group id for now to demo ICLR, but we will need to discuss how to reuse the same invitation between acs and reviewers, or create different invitations.

mspector:
we can talk about it later, but my first reaction is that creating multiple invitations for different groups is ok

openreview/conference/matching.py
```diff
@@ -28,10 +28,16 @@ def _create_edge_invitation(self, id):
                 },
                 'content': {
                     'head': {
-                        'type': 'Note'
+                        'type': 'Note',
+                        "query" : {
```
zbialecki: Should be single quotes

melisabok: thanks, fixed.

PR #366, opened 2019-07-25
Feature/expertise selection interface

mohituniyal:
This interface is an invitation webfield. The invitation will be of the form https://openreview.net/invitation?id=ICLR.cc/2020/Conference/-/Expertise_Selection . The duedate will be the same as the submission deadline and the invitation is configured when a request form note is deployed. The invitation's duedate is updated whenever the request form is revised.

The expertise selection interface lists all of the available papers for the profile of the person logged in (has to be an area chair or reviewer for the venue to be able to access this invitation).
We show only the original notes (i.e. ones with note.original=null) ordered by note.cdate.

Screen Shot 2019-07-25 at 12 29 52 PM


zbialecki:
Nice work! My only real issue is that I feel like the way the buttons are styled it's a little confusing. To me, it looks like exclude is already selected for all the papers. I would rather have the buttons be more like the bid radio buttons with options for Include and Exclude, with Include selected by default. It feels more immediately obvious which choice was made.

Other small things:
* there's a lot of empty space below the instructions. Can that be reduced? 
* there seems to be extra padding between the papers in the list. Maybe due to some extra margin on the exclude button?
* I would say the first 2 lines of the instructions should be 1 line. There doesn't need to be a line break.

openreview/conference/webfield.py
```diff
@@ -88,6 +88,37 @@ def set_home_page(self, group, layout, options = {}):
             group.signatures = [group.id]
             return self.client.post_group(group)
 
+    def set_expertise_selection_page(self, conference, invitation):
+
+        default_header = {
+            'title': conference.get_short_name() + ' Expertise Selection',
+            'instructions': '<p class="dark">Listed below are the papers we have for you. <b>By default, we consider \
+                all of these papers to formulate your expertise.\
+                <br>Please click on \"Exclude\" for papers that should NOT be used to represent your expertise.</b><br>\
```
zbialecki: What is the `</b>` tag for?

melisabok:
We tried to do what you say: showing 'Include' selected as default but it seems that the 'default' field in the edge invitation doesn't work, do you know if there is default field for the tag widgets?

melisabok:
Now I remember another issue we had: the user can unselect any button leaving that paper in a weird state. 

I would rename the tab: 'All Papers' to 'All My Papers' and show two other tabs saying "My selected expertise" and "Excluded papers"

openreview/conference/webfield.py
mohituniyal: It was a stray bold tag. Thanks for noticing that.

mohituniyal:
>there's a lot of empty space below the instructions. Can that be reduced?
>>> There was an extra <br> tag in the header, which has been removed now. The padding below the header now is 1.5rem, which is the same as any conference home page with a submission button.

>there seems to be extra padding between the papers in the list. Maybe due to some extra margin on the exclude button?
>>> I found that the bottom padding between the papers in the list is the same as that on the bidding console page, i.e. 0.5rem. If needed, I can reduce this further.

>I would say the first 2 lines of the instructions should be 1 line. There doesn't need to be a line break.
>>> Done

zbialecki:
Actually, 1 more thought I had: we should add a link, "Update your profile" that links to /profile?mode=edit to the last bullet of the instructions to make it easy for people to update their profile right then.

mohituniyal:
> Actually, 1 more thought I had: we should add a link, "Update your profile" that links to /profile?mode=edit to the last bullet of the instructions to make it easy for people to update their profile right then.

Great suggestion @zbialecki. I made changes for that.

mohituniyal:
PR https://github.com/iesl/openreview/pull/1529 has been created for suggested styling changes.

openreview/conference/templates/expertiseBidWebfield.js
```diff
@@ -0,0 +1,289 @@
...
+    sort: 'cdate'
+  })
+  .then(function(notes){
+    return notes.filter(function(note){
+      return !(note.original);
+    });
+  });
+
+  var edgesMapP = Webfield.getAll('/edges', {
+    invitation: EXPERTISE_BID_ID
```
melisabok: I noticed the same in the bid webfield (I'm fixing it in my branch), could you add the parameter to the query ```tail = user.profile.id```? we don't have to merge the edge results later.

mohituniyal: Done.

PR #368, opened 2019-07-30
Feature/pc console rev reassign

mohituniyal:
PC Console:
> Added reviewer reassignment feature

Correction to AC Console:
> Add reviewer to Venue Reviewer pool (e.g. ICLR.cc/2020/Conference/Reviewers group) if the user types in a new reviewer email instead of selecting 1 from the drop down.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -75,7 +75,7 @@ class Conference(object):
     def __set_program_chair_page(self):
         program_chairs_group = tools.get_group(self.client, self.get_program_chairs_id())
         if program_chairs_group:
-            return self.webfield_builder.set_program_chair_page(self, program_chairs_group)
+            return self.webfield_builder.set_program_chair_page(self, program_chairs_group, self.enable_reviewer_reassignment)
 
     def __set_expertise_selection_page(self):
         expertise_selection_invitation = self.client.get_invitation(self.get_expertise_selection_id())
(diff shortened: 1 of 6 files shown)


openreview/openreview.py
```diff
@@ -673,6 +673,21 @@ def get_notes(self, id = None,
 
         return [Note.from_json(n) for n in response.json()['notes']]
 
+    def get_reference(self, id):
```
melisabok: I think you need to update your branch because this is already in master

openreview/conference/builder.py
```diff
@@ -493,13 +493,14 @@ def open_revise_reviews(self, name = 'Revision', start_date = None, due_date = N
     def close_revise_submissions(self, name):
         return self.__expire_invitations(name)
 
-    def set_program_chairs(self, emails):
+    def set_program_chairs(self, emails = [], enable_reviewer_reassignment = True):
         self.__create_group(self.get_program_chairs_id(), self.id, emails)
         ## Give program chairs admin permissions
         self.__create_group(self.id, '~Super_User1', [self.get_program_chairs_id()])
+        self.enable_reviewer_reassignment = enable_reviewer_reassignment
```
melisabok: are we using the same property to enable reassignment in both consoles? in that case we prefer to set the property using the builder directly instead of setting it through set_program_chairs or set_area_chairs

mohituniyal:
This PR is to merge these changes into feature/edges which is wrong as these changes should be going into master. Closing this PR. The next PR will also have changes based on the comments received here.

PR #369, opened 2019-07-31
Feature/edges patch add matching group

mspector:
This request is to merge into feature/edges.

This PR transfers "ownership" of the matching pipeline from the conference group to a group with id = {conference_id}/Matching (e.g. ICLR.cc/2019/Conference/Matching).

This is useful because top-level matching privileges can be given to users (e.g. Program_Chairs group) by adding them as members to ICLR.cc/2019/Conference/Matching. Prior to this, the PC group would need to be a member of the top level conference group, which can have unintended consequences. I argue that this way of doing things aligns better with our semantics and naming conventions for things.

corresponds with this PR: https://github.com/iesl/openreview/pull/1534

### DIFF
--- a/openreview/conference/matching.py
+++ b/openreview/conference/matching.py
@@ -8,20 +8,24 @@ class Matching(object):
     def __init__(self, conference):
         self.conference = conference
         self.client = conference.client
+        self.matching_group_id = self.conference.get_id() + '/Matching'
+
+    def get_matching_invitation_id(self, name):
+        return '{}/-/{}'.format(self.matching_group_id, name)
 
     def _create_edge_invitation(self, id):
         invitation = openreview.Invitation(
             id = id,
-            invitees = [self.conference.get_id()],
-            readers = [self.conference.get_id()],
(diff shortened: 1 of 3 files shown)


melisabok:
This requires more discussion. For example: do we need an edge assignment invitation per reviewer/ac group? How are we going to call it? Who can edit the assignment edges? Who can read the rest of the edges? Etc.

mspector:
I agree that the issues that you're describing are very important, and there are complex considerations. But I think it's separate from what I'm proposing here.

This is a small change that does only one thing: it separates the matching-specific permissions from the conference group. Right now, the conference group -- and *only* the conference group -- can edit edges that are created by the setup process. With this change, we can selectively give that permission to other groups if we want to (including the conference group itself, which gives us exactly the functionality that we had before).

I don't believe that these changes interact with the issues you're describing.

melisabok:
I think we can close this PR? @mspector ?

mspector:
Yes, I agree

PR #370, opened 2019-08-01
Fix/conflict subdomains

mspector:
`tld` is a module for working with and detecting top-level domains.

previously, we were using a faulty heuristic to detect and exclude top-level domains from conflicts. this change corrects that fault.


openreview/tools.py
```diff
@@ -409,9 +410,10 @@ def subdomains(domain):
         full_domain = domain
     domain_components = full_domain.split('.')
     domains = ['.'.join(domain_components[index:len(domain_components)]) for index, path in enumerate(domain_components)]
-    valid_domains = [d for d in domains if '.' in d]
+    valid_domains = [d for d in domains if not tld.is_tld(d)]
```
mspector: this is the central change

openreview/tools.py
```diff
@@ -407,11 +408,12 @@ def subdomains(domain):
...
     else:
         full_domain = domain
-    domain_components = full_domain.split('.')
+    domain_components = [c for c in full_domain.split('.') if c and not c.isspace()]
     domains = ['.'.join(domain_components[index:len(domain_components)]) for index, path in enumerate(domain_components)]
-    valid_domains = [d for d in domains if '.' in d]
+    valid_domains = [d for d in domains if not tld.is_tld(d)]
     return valid_domains
 
+@deprecated(version='0.9.24')
```
melisabok: is it possible to remove these deprecated functions?

mspector: I don't see why not - it might break something but we'll find out!

PR #372, opened 2019-08-01
Fix/matching profiles error

mspector:
should be merged after https://github.com/iesl/openreview-py/pull/370

fixes a bug that resulted in not all profiles being considered during conflict of interest calculation. this has caused many conflicts to go uncaught.


mspector:
closing this, will reopen without extraneous commits

PR #374, opened 2019-08-02
retrieve profiles in batches

mspector:
I actually think that this is just a short-term solution. It seems like this functionality should be available at a higher level than the matching submodule.

@mohituniyal any suggestion about how to best test this?

### DIFF
--- a/openreview/conference/matching.py
+++ b/openreview/conference/matching.py
@@ -279,6 +279,23 @@ class Matching(object):
         union = set1.union(set2)
         return len(intersection) / len(union)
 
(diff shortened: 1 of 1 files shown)


mspector:
@mohituniyal suggested that I move this functionality to `client.search_profiles`. This would be difficult because this fix only really works for the `ids` and `emails` arguments.

the api for `client.search_profiles` is a mess. it needs to be fixed.

melisabok:
@mohituniyal if you agree with this PR we can merge it.

PR #380, opened 2019-08-05
add a non-required field to the metadata and assignment invitation sp…

mspector:
…ec, so that they can be posted.

This change fixes the failing tests in the master branch of openreview-matcher.

I'm not sure if this is the best way to do it -- Ideally we'd specify the correct fields for these invitations, but these are special cases where we may not have the right kind of values (for example, metadata notes have one field in content called `entries`, and it's a list of dict values). 

personally I think this is ok, because we'll soon stop needing metadata and assignment notes anyway...

### DIFF
--- a/openreview/conference/matching.py
+++ b/openreview/conference/matching.py
@@ -225,7 +225,9 @@ class Matching(object):
                 },
                 'signatures': {
                     'values': [CONFERENCE_ID]},
-                'content': {}
+                'content': {
+                    'title': {'required': False, 'value': 'metadata note'}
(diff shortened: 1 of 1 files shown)


carlosmondra:
So I believe that this will work for now, but I'll make another pull request where I check that all the fields in the note are in the invitation. So if the note contains fields that the invitation.reply does not contain, it will also fail.

mspector:
closing because this is basically a duplicate of https://github.com/iesl/openreview-py/pull/377/files

PR #386, opened 2019-08-07
Feature/match by group

mspector:


openreview/conference/matching.py
```diff
@@ -1,22 +1,91 @@
...
+        e.g. "Affinity_Score"
+        '''
         invitation = openreview.Invitation(
-            id = id,
-            invitees = [self.conference.get_id()],
-            readers = [self.conference.get_id()],
-            writers = [self.conference.get_id()],
-            signatures = [self.conference.get_id()],
-            reply = {
+            id=self._get_edge_invitation_id(edge_name),
```
melisabok: Is this a pylint convention? I particularly prefer have spaces.

openreview/conference/matching.py
```diff
@@ -238,86 +323,51 @@ def _build_config_invitation(self, scores_specification):
...
+                            'No Solution',
+                            'Complete',
+                            'Deployed'
+                        ]
                     }
                 }
             })
         self.client.post_invitation(config_inv)
 
-    def _jaccard_similarity(self, list1, list2):
```
melisabok: why do you move these functions outside of the class?

tests/test_matching.py
```diff
@@ -177,37 +189,68 @@ def test_setup_matching(self, client, test_client, helpers):
...
+        assert client.get_invitation(id=reviewer_matching_ids['custom_load'])
+        assert client.get_invitation(id=reviewer_matching_ids['conflict'])
+        assert client.get_invitation(id=reviewer_matching_ids['aggregate_score'])
+        assert client.get_invitation(id=reviewer_matching_ids['paper_assignment'])
+
+        # Set up AC matching
+        conference.setup_matching(
+            conference.client.get_group(conference.get_area_chairs_id()))
 
-        assert client.get_invitation(id = 'auai.org/UAI/2019/Conference/-/Reviewing/Assignment_Configuration')
```
melisabok: I would prefer to be more explicit when I do an assert. Using another function to compute the expected value doesn't assure if the code we are testing is right and/or the test function is right.

openreview/conference/matching.py
melisabok: I guess we are following these conventions: https://www.python.org/dev/peps/pep-0008/

I didn't know about them, I like to add spaces everywhere :/

mspector: Yeah, I have just been following the default pylint settings. But I don't have a strong opinion about it, as long as we all use the same thing.

openreview/conference/matching.py
mspector: Just trying to follow pylint's suggestions. It's because these functions don't use `self` for anything

tests/test_matching.py
mspector: Yes, good point. I was trying not to change anything that was not either (1) directly related to the group-specific matching change, or (2) suggested by pylint. But I could try to improve the tests in general, if you'd like. Alternatively, I could make another PR.

openreview/conference/matching.py
melisabok: Ok, let's follow the pylint conventions.

tests/test_matching.py
melisabok: Having repeated code and assertion using strings is good. The test should be simple, it also clear that the invitation that you expect is the string that you are reading, it helps you to understand the behavior of the code. 
I suggest to go back to the original assertion:

```
assert client.get_invitation(id = 'auai.org/UAI/2019/Conference/Reviewers/-/Reviewing/Custom_Load')
assert client.get_invitation(id = 'auai.org/UAI/2019/Conference/Area_Chairs/-/Reviewing/Custom_Load')
```

mspector: Ah, I see what you're saying now. I thought you didn't like the use of client.get_invitation(...).

I'll change this back

tests/test_matching.py
```diff
@@ -12,6 +12,18 @@
 from selenium.common.exceptions import TimeoutException
 from selenium.common.exceptions import NoSuchElementException
 
+def _matching_invitation_ids(client, group_id):
```
melisabok: We don't need this function anymore, right?

mspector: did you want me to convert *all* cases where this is used into explicit strings? or just the assertion lines?

tests/test_matching.py
```diff
@@ -252,37 +295,68 @@ def test_setup_matching_with_tpms(self, client, test_client, helpers):
...
+        print(conference.get_reviewers_id())
+        reviewer_matching_ids = _matching_invitation_ids(
+            client, conference.get_reviewers_id())
+
+        assert client.get_invitation(id='auai.org/UAI/2019/Conference/Program_Committee/-/Reviewing/Assignment_Configuration')
+        assert client.get_invitation(id='auai.org/UAI/2019/Conference/Program_Committee/-/Reviewing/Custom_Load')
+        assert client.get_invitation(id='auai.org/UAI/2019/Conference/Program_Committee/-/Reviewing/Conflict')
+        assert client.get_invitation(id='auai.org/UAI/2019/Conference/Program_Committee/-/Reviewing/Aggregate_Score')
+        assert client.get_invitation(id='auai.org/UAI/2019/Conference/Program_Committee/-/Reviewing/Paper_Assignment')
+        assert client.get_invitation(id='auai.org/UAI/2019/Conference/Program_Committee/-/Reviewing/TPMS_Score')
```
melisabok: do you think it is better to have an affinity/tpms score invitation per matching group?

I'm realizing the bid invitation is being used to both groups: area chairs and reviewers and the invitation.reply.content is something like this:

```
                'content': {
                    'head': {
                        'type': 'Note'
                    },
                    'tail': {
                        'type': 'Group'
                    },
                    'label': {
                        'value-radio': ['Very High', 'High', 'Neutral', 'Low', 'Very Low'],
                        'required': True
                    }
                }
```

where the tail can be any group, do we need to split the bid invitation too? or should we keep using one invitation for all matching groups? 

@zbialecki thoughts?

tests/test_matching.py
melisabok: I think you are using this in all the assertion lines.

tests/test_matching.py
```diff
@@ -177,37 +189,68 @@ def test_setup_matching(self, client, test_client, helpers):
...
+        assert client.get_invitation(id='auai.org/UAI/2019/Conference/Senior_Program_Committee/-/Reviewing/Aggregate_Score')
+        assert client.get_invitation(id='auai.org/UAI/2019/Conference/Senior_Program_Committee/-/Reviewing/Paper_Assignment')
 
         bids = client.get_edges(invitation = conference.get_bid_id())
         assert bids
         assert 6 == len(bids)
 
-        custom_loads = client.get_edges(invitation = 'auai.org/UAI/2019/Conference/-/Reviewing/Custom_Load')
+        custom_loads = []
+        custom_loads.extend(
```
melisabok: why do you need to append both results into an array? why don't you assert each result independently?

mspector: I don't need to, this was just my attempt to keep things as similar as possible to the way it was before. Should I change it?

tests/test_matching.py
mspector: I don't have an opinion about bids. But I think it's slightly simpler to have TPMS/affinity scores be separated by matching group, just because of the way scores get generated. Our expertise modeling framework is based on groups, and by default, it generates one score file per group. But that can be changed, if there's a good reason to keep them all together.

tests/test_matching.py
mspector: I changed it

mspector:
@melisabok is this ok to be merged?

melisabok:
I'm running the matching tests and they are failing: 

```
TypeError: setup_matching() missing 1 required positional argument: 'match_group'
```

Could you check this before I merge it? I believe there are some unit tests using the builder.

tests/test_matching.py
zbialecki: I'm OK with having a separate affinity score invitation for each matching group, but it will add some inefficiencies because for example Area Chairs and Reviewers overlap so much. So we'll have a bunch of duplicated edges.

mspector: Do ACs and Reviewers typically overlap? if I remember correctly, ICLR doesn't do that, but maybe other conferences do

melisabok: I also keep them separated is a good decision, in terms of performance we won't have all the affinity scores edges in a single invitation.

PR #391, opened 2019-08-09
Fix/improve bid webfield

melisabok:
- Create one bid invitation per matching group: Reviewers, Area Chairs.
- Refactor the bid webfield to sort the All Papers tab by affinity score. This sorting can be optional, it the venue doesn't request the affinity scores then papers without sorting are rendered.
- If papers are sorting by score, the score is being displayed below each note panel.
- Papers are also filtered by conflicts edges (includes authored papers). If the venue doesn't have conflicts edges previously created then it doesn't apply any filter.
- No Bid was removed.
- Each bid type tab is rendered when the user clicks on it. It loads all the information from the server. It doesn't paginate the result of each tab, I assume there is no going to be more than 1k bids with the same type. But if that happens we can implement pagination here too.
- Header wording was changes, @mspector please review. 

This new UI should support unlimited amount of papers to review. 

Screen Shot 2019-08-07 at 4 38 02 PM

Screen Shot 2019-08-07 at 4 38 34 PM

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -34,6 +34,7 @@ class Conference(object):
         self.program_chairs_name = 'Program_Chairs'
         self.recommendation_name = 'Recommendation'
         self.registration_name = 'Registration'
+        self.affinity_score_name = None
         self.submission_stage = SubmissionStage()
         self.bid_stage = BidStage()
         self.expertise_selection_stage = ExpertiseSelectionStage()
@@ -285,6 +286,9 @@ class Conference(object):
 
     def get_registration_id(self):
         return self.get_invitation_id(self.registration_name)
+
+    def get_affinity_score_id(self):
+        return self.get_invitation_id(self.affinity_score_name)
 
     def get_invitation_id(self, name, number = None):
         invitation_id = self.id
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -85,12 +85,12 @@ class Conference(object):
     def __set_bid_page(self):
         bid_invitation = self.client.get_invitation(self.get_bid_id(group_id=self.get_reviewers_id()))
         if bid_invitation:
-            self.webfield_builder.set_bid_page(self, bid_invitation)
+            self.webfield_builder.set_bid_page(self, bid_invitation, self.get_reviewers_id())
 
         if self.use_area_chairs:
             bid_invitation = self.client.get_invitation(self.get_bid_id(group_id=self.get_area_chairs_id()))
             if bid_invitation:
-                self.webfield_builder.set_bid_page(self, bid_invitation)
+                self.webfield_builder.set_bid_page(self, bid_invitation, self.get_area_chairs_id())
 
     def __set_recommendation_page(self):
         recommendation_invitation = self.client.get_invitation(self.get_recommendation_id())
@@ -308,8 +308,8 @@ class Conference(object):
     def get_paper_assignment_id(self):
         return self.get_invitation_id('Reviewing/Paper_Assignment')
 
-    def get_affinity_score_id(self):
-        return self.get_invitation_id('Reviewing/Affinity_Score')
+    def get_affinity_score_id(self, group_id):
+        return self.get_invitation_id('Reviewing/Affinity_Score', prefix=group_id)
 
     def set_homepage_header(self, header):
         self.homepage_header = header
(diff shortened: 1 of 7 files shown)


openreview/conference/templates/bidWebfield.js
```diff
@@ -21,55 +23,116 @@ function main() {
...
+        //Keep affinity score order
+        var mapById = result.notes.reduce(function(mapById, note) {
+          mapById[note.id] = note;
+          return mapById;
+        }, {});
+        return noteIds.map(function(id) {
+          return mapById[id];
+        });
+      })
+      .then(function(notes) {
```
zbialecki: Since the previous `.then` doesn't perform any async actions is it possible to merge these 2 thens?

openreview/conference/builder.py
```diff
@@ -83,9 +83,14 @@ def __set_expertise_selection_page(self):
             return self.webfield_builder.set_expertise_selection_page(self, expertise_selection_invitation)
 
     def __set_bid_page(self):
```
mspector: For clarity, I suggest either: 

(1) changing the name of this function to `__set_bid_pages` (plural) and add a docstring that explains that this function sets the bid pages of both the reviewers page and the AC page (if applicable). 

or (2): parameterizing this function with the group ID, and calling it separately for reviewers and ACs wherever it is called.

I'm not sure if there are other page-setting functions that follow this pattern, but in my opinion it would be clearer one of these two ways

openreview/conference/builder.py
```diff
@@ -436,7 +446,9 @@ def open_bids(self):
         return self.__create_bid_stage()
 
     def close_bids(self):
-        return self.__expire_invitation(self.get_bid_id())
+        self.__expire_invitation(self.get_bid_id(self.get_reviewers_id()))
```
mspector: ^ this is a good example of what I meant in my previous comment. here, `__expire_invitation` has an invitation ID as a parameter, which makes sense to me

openreview/conference/builder.py
```diff
@@ -1051,26 +1065,11 @@ def get_result(self):
         if self.submission_stage:
             self.conference.set_submission_stage(self.submission_stage)
 
-        ## Create paper and author groups before any other stage that requires paper specific invitations
+        ## Create comittee groups before any other stage that requires them to create groups and/or invitations
```
mspector: tiny change: should be "committee"

mspector:
I have a question about our PR process: should we approve if we think our own comments have been addressed, or should we only approve if we think everyone's comments have been addressed (and that it's fully ready to be merged)?

e.g. In this PR I had a few comments, but I'm not insisting that they be changed. But I also see that @zbialecki  has some comments that @melisabok  will probably want to fix. Should I approve in this case?

melisabok:
I will wait until all of you approve it. If you think you are asking for changes you shouldn't approve it and wait until we make an agreement about the changes.

mspector:
I realized shortly after my comment that I really was asking for changes, and that I should probably have just waited for you to respond. I initially approved because the parts I commented on are possibly just matters of opinion (or a minor spelling mistake in a comment)

zbialecki:
My feeling on reviewing PRs is that different people will end up focusing on different parts of the code, so if you review and everything looks OK, then you should approve and let the other reviewers approve when they are ready.

openreview/conference/templates/bidWebfield.js
```diff
@@ -21,55 +23,116 @@ function main() {
...
+      return Webfield.post('/notes/search', {
+        ids: noteIds
+      })
+      .then(function(result) {
+        //Keep affinity score order
+        var mapById = result.notes.reduce(function(mapById, note) {
+          mapById[note.id] = note;
+          return mapById;
+        }, {});
+        return noteIds.map(function(id) {
```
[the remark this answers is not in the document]
melisabok: noteIds has the order of the notes, when I get the notes by ids the backend doesn't return them in the same order.

openreview/conference/templates/bidWebfield.js
melisabok: you are right! thanks

openreview/conference/templates/bidWebfield.js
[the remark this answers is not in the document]
zbialecki: I see, makes sense.

openreview/conference/builder.py
melisabok: I'm not sure if need to change the name to plural, I want to hide the implementation. The method that set a page per group id is in the invitation builder: https://github.com/iesl/openreview-py/blob/fix/improve-bid-webfield/openreview/conference/webfield.py#L120. 

I added a docstring to the function.

openreview/conference/templates/bidWebfield.js
```diff
@@ -21,55 +23,111 @@ function main() {
...
+
+      return Webfield.post('/notes/search', {
+        ids: noteIds
+      })
+      .then(function(result) {
+        //Keep affinity score order
+        var notesById = _.keyBy(result.notes, function(note) {
+          return note.id;
+        });
+        notes = noteIds.map(function(id) {
```
zbialecki: Is notes declared somewhere else? Or maybe you could just do `return noteIds.map...`

melisabok: oops, my bad. I fixed it.

PR #394, opened 2019-08-12
Fix/update docs

mspector:
this PR makes two main changes:
1. it moves the examples related to data retrieval to its own file, outside of "Creating a conference", and adds a full example script for retrieving ICLR 19 metadata
2. it adds a submodule to openreview-py that can be called from the command line

I think the first change is necessary, but we don't have to do the second if people don't think this is a good way of distributing this kind of thing. 

I personally think it will be useful to keep conference-specific data retrieval scripts somewhere, instead of trying to generalize the examples too much. There's also a good argument to be made that conference-specific retrieval scripts should be kept in the openreview-scripts repo, but including it with openreview-py means that users can call it directly from the command line.


docs/source/getting_data.rst
```diff
@@ -0,0 +1,242 @@
...
+
+	import argparse
+	import json
+	import os
+	from collections import defaultdict
+	from tqdm import tqdm
+	import openreview
+
+
+	def download_iclr19(client, outdir='./', get_pdfs=False):
```
melisabok: it seems to be a script here, we could put in the openreview-scripts repo.

melisabok:
I like the refactoring. 

I think the script should go in the openreview-scripts repo. The library shouldn't distribute scripts. We need to put more effort in writing better examples and/or document the builder that contain function to retrieve papers, decisions, etc.

docs/source/getting_data.rst
mspector: yeah, I had that thought as well. I though I would propose putting it in `openreview.scripts` so that it could be called from the command line:

```
>>> python -m openreview.scripts.download_iclr19
```

mspector:
I can move the script to openreview-scripts. 

Here's a question: should we remove the old scripts? I personally think that we should.

melisabok:
do you want to get rid of all the previous scripts?

mspector:
That's what I was thinking, but I'm open to other suggestions.

My thought was that they would be mostly out of date, and not very useful to anyone.

melisabok:
Was there a reason to merge your own PR?

mspector:
just trying to clean up. Mohit had approved and it seemed like your only issue was the script, which I removed...

melisabok:
We shouldn't approve our own PR unless there is an emergency and nobody can review it before sending it to live. 

I was also a reviewer of this PR and I haven't approved it yet.

mspector:
Sorry about that, my mistake. Should I revert it so that you can give it another look?

PR #396, opened 2019-08-13
Fix/bid webfield bug fixing

melisabok:
I forgot about this typo and I want to remove this test file from the folder, we could create it every time the test is run.


melisabok:
this should be ready to merge now

PR #397, opened 2019-08-14
Add error message to configuration invitation and configure the readers

melisabok:
This should fix the matching system / UI creating paper assignment edges.


melisabok:
Yes, that is right.

PR #403, opened 2019-08-27
Changes for buddy AC console link on conf homepage

mohituniyal:
Change needed on the conference homepage to recognize the buddy ac group.

Screen Shot 2019-09-11 at 3 29 41 PM


melisabok:
I'm going to put this on hold because it can be a more generic way to get all the consoles.

mohituniyal:
The consoles and their links on the conference homepages have been now named "Reviewers Console", "Authors Console", "Program Chairs Console", etc. Let me know if we need these to be "Reviewer Console", "Author Console", etc (i.e. with singulars) and I can splice off the last "s" from group names.

openreview/conference/templates/tabsConferenceWebfield.js
```diff
@@ -144,41 +159,18 @@ function renderContent(notesResponse, userGroups, activityNotes, authorNotes) {
...
     if (authorNotes.length) {
-      $('#your-consoles .submissions-list').append([
-        '<li class="note invitation-link">',
-          '<a href="/group?id=' + AUTHORS_ID + '">Author Console</a>',
-        '</li>'
-      ].join(''));
+      allConsoles.push(AUTHORS_ID);
     }
+    userGroups.forEach(function(group) {
+      if (!(allConsoles.includes(group))){
```
zbialecki: Unfortunately .includes is ES6 only. You should use indexOf in client js.

openreview/conference/templates/tabsConferenceWebfield.js
```diff
@@ -72,13 +71,13 @@ function load() {
...
     });
 
-    userGroupsP = Webfield.getAll('/groups', { member: user.id, web: true })
-      .then(function(groups) {
-        return _.filter(
-          _.map(groups, function(g) { return g.id; }),
-          function(id) { return _.startsWith(id, CONFERENCE_ID); }
-        );
-      });
+    userGroupsP = Webfield.getAll('/groups', { regex: CONFERENCE_ID + '/.*', member: user.id, web: true })
```
zbialecki: Is adding the regex param just a performance optimization? I see that we are filtering out groups that don't start with the conference ID below. Also, if a user is a member of CONFERENCE_ID, will this regex exclude the group CONFERENCE_ID since it has a '/' on the end?

openreview/conference/templates/programchairWebfield.js
```diff
@@ -149,8 +149,11 @@ var getAreaChairGroups = function(noteNumbers) {
         var num = getNumberfromGroup(g.id, 'Paper');
         var index = getNumberfromGroup(g.id, 'Area_Chair');
         if (num) {
-          if (g.members.length) {
-            var areaChair = g.members[0];
+          var areaChair;
+          areaChair = g.members.find(function(member){
```
zbialecki: `.find` is ES6 only. You should switch to `_.find`

openreview/conference/templates/tabsConferenceWebfield.js
melisabok: this userGroupP is being used 'only' to render the console so we should find all the group that exist after CONFERENCE_ID, I will remove the filtering because it doesn't make sense now.

melisabok:
Do you mind to check my last commit? I tried to write everything supported by all the browsers.

openreview/conference/templates/tabsConferenceWebfield.js
```diff
@@ -71,13 +71,7 @@ function load() {
...
     });
 
-    userGroupsP = Webfield.getAll('/groups', { regex: CONFERENCE_ID + '/.*', member: user.id, web: true })
-    .then(function(groups) {
-      return _.filter(
-        _.map(groups, function(g) { return g.id; }),
-        function(id) { return _.startsWith(id, CONFERENCE_ID); }
-      );
-    });
+    userGroupsP = Webfield.getAll('/groups', { regex: CONFERENCE_ID + '/.*', member: user.id, web: true });
```
zbialecki: This still should have a `then` with a map that just returns group ids though, no?

melisabok: I'm doing that later

melisabok: line 165

zbialecki: I see

melisabok: Is it ok? Can I merge this?

zbialecki: Yeah, looks good to me

PR #404, opened 2019-09-02
PC console paper assignment

mohituniyal:
This lets Program Chairs assign papers among themselves. 
They can also sort the papers to see their own assignments at the top.

Requires https://github.com/iesl/openreview/pull/1568


openreview/conference/templates/programchairWebfield.js
```diff
@@ -26,6 +26,9 @@ var reviewerSummaryMap = {};
 var allReviewers = [];
 var conferenceStatusData = {};
 
+var PC_PAPER_ASSIGNMENT = true;
```
melisabok: instead of using a boolean, I would show the tags if the invitation exits.

openreview/conference/templates/programchairWebfield.js
```diff
@@ -61,9 +64,15 @@ var getInvitationId = function(name, number) {
 }
 
 var getBlindedNotes = function() {
-  return Webfield.getAll('/notes', {
-    invitation: BLIND_SUBMISSION_ID, noDetails: true
-  });
+  if (PC_PAPER_ASSIGNMENT) {
```
melisabok: Can we get the tags separately from the notes? Each note may have more than one tag from other invitation, and it can be too expensive the call. 

Also, you could do the get tags call only if the tag invitation exists.

openreview/conference/templates/programchairWebfield.js
```diff
@@ -429,6 +446,47 @@ var displaySortPanel = function(container, sortOptions, sortResults) {
   });
 };
 
+var addTagsToPaperSummaryCell = function(data, pcAssignmentTagInvitations) {
+
+  if (pcAssignmentTagInvitations.length) {
```
zbialecki: I think condition this should be `pcAssignmentTagInvitations && pcAssignmentTagInvitations.length` since `getPcAssignmentTagInvitations()` can return undefined.

zbialecki: I also think this is a situation where you want to check the inverse of the condition and return, rather than wrapping the whole function body in an if.

mohituniyal: Agreed. I have made changes for this.

openreview/conference/templates/programchairWebfield.js
mohituniyal: This change has been made.

openreview/conference/templates/programchairWebfield.js
mohituniyal: Agreed. I have made the changes for this.

openreview/conference/templates/programchairWebfield.js
```diff
@@ -252,6 +255,24 @@ var getInvitations = function() {
...
+var getPcAssignmentTagInvitations = function() {
+  return Webfield.getAll('/invitations', { regex: PC_PAPER_TAG_INVITATION, tags: true});
+}
+
+var getPcAssignmentTags = function() {
+  Webfield.getAll('/tags', { invitation: PC_PAPER_TAG_INVITATION})
+  .then(function(results) {
+    if (results && results.length) {
+      results.forEach(function(tag){
+        if (! (tag.forum in pcTags)) {
```
melisabok: do you have an extra space here?

openreview/conference/templates/programchairWebfield.js
```diff
@@ -252,6 +255,24 @@ var getInvitations = function() {
...
+          pcTags[tag.forum] = [];
+        }
+        pcTags[tag.forum].push(tag);
```
melisabok: is there more than one tag per forum?

openreview/conference/templates/programchairWebfield.js
mohituniyal: Yes, I'll fix that.

openreview/conference/templates/programchairWebfield.js
mohituniyal: No, there should be only one. I remembered that the tag widget in the bid web-field created multiple tags when one clicks quickly on the options for one paper. Although that does not apply here, I just wanted to account for that. Also, the value for pcTags[tag.forum] has to be an array as that is what the tagWidget expects.

mohituniyal:
No, the tag invitations for this can not be created through the builder right now. You could use the script in https://github.com/iesl/openreview-scripts/pull/1119 for this.

openreview/conference/templates/programchairWebfield.js
```diff
@@ -1227,6 +1310,10 @@ controller.addHandler('areachairs', {
 
       var uniqueIds = _.union(uniqueReviewerIds, uniqueAreaChairIds);
 
+      if (pcAssignmentTagInvitations && pcAssignmentTagInvitations.length) {
+        getPcAssignmentTags();
```
melisabok: this is an async function and it should be chained with the other promises.

melisabok: @zbialecki I made some improvements  using the network calls, do you mind to take a look?

PR #409, opened 2019-09-09
Conf be able to read/write anon reviewer groups

mohituniyal:
Add conference group as reader and writer for each new Anon-Reviewer group.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -559,39 +559,31 @@ class Conference(object):
             parent_label = 'Area_Chairs',
             individual_label = 'Area_Chair')
         else:
-            result = tools.add_assignment(self.client,
-            number,
-            self.get_id(),
-            user,
-            parent_label = 'Reviewers',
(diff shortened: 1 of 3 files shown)


melisabok:
We should do the same the builder is doing to assign reviewers to avoid inconsistencies.

Here is an example:

Screen Shot 2019-09-09 at 8 45 35 AM

mohituniyal:
> We should do the same the builder is doing to assign reviewers to avoid inconsistencies.
> 
> Here is an example:
> 
> Screen Shot 2019-09-09 at 8 49 46 AM

melisabok:
It seems the builder is setting up the AC as writers of Anon groups:

https://github.com/iesl/openreview-py/blob/master/openreview/conference/builder.py#L568

so, it should work, right?

mohituniyal:
I made some changes to the set_assignment function in the builder. @melisabok please take a look, I think the groups (created by the builder and by the Ac/Pc consoles) are consistent now.

PR #413, opened 2019-09-12
Fix bug where content in second tab would be pushed down by content in first tab

zbialecki:


melisabok:
@pmandler could you check if this fixes it and then merge it?

pmandler:
The problem doesn't show up in the regular Author Console, it only showed up in the copy that I modified for the Reproducibility Challenge.  I can test that it doesn't mess things up, but I can't really test if it fixes anything.

melisabok:
what is the difference between one console and the other?

zbialecki:
The difference was in the order of tabs. When the "Your Submissions" tab is first, this problem becomes visible. I wanted to fix it here so that if other conferences modify this base file the problem doesn't show up.

pmandler:
Zach - cool, I didn't know that was the issue and I do thing Your Submissions should be the first tab because it's likely to be the most useful.
Melisa - Reproducibility shows both the Claim and the Report.

melisabok:
Thank you both, I'm trying to do a quick test in my localhost and then I will merge it.

melisabok:
The issue is fixed and I moved the submissions tab to the first position.

PR #417, opened 2019-09-18
Check if the content field exist

melisabok:
Some venues require a different review schema. I'm not sure if we should force 'rating, confidence and review' to be mandatory fields or not. This PR is allowing the webfield to be displayed without js errors.

### DIFF
--- a/openreview/conference/templates/programchairWebfield.js
+++ b/openreview/conference/templates/programchairWebfield.js
@@ -94,9 +94,9 @@ var getOfficialReviews = function(noteNumbers) {
       }
       if (num) {
         if (num in noteMap) {
-          ratingMatch = n.content.rating.match(ratingExp);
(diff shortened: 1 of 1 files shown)


openreview/conference/templates/programchairWebfield.js
```diff
@@ -1049,7 +1049,7 @@ var buildPaperTableRow = function(note, reviewerIds, completedReviews, metaRevie
         note: reviewObj.id,
         rating: reviewObj.rating,
         confidence: reviewObj.confidence,
-        reviewLength: reviewObj.content.review.length,
+        reviewLength: reviewObj.content && reviewObj.content.review.length,
```
zbialecki: I think this line should be `reviewObj.content.review && reviewObj.content.review.length`
We know the content object exists, but want to check that review is there.

melisabok: oops, sorry, you are right

PR #422, opened 2019-09-23
Feature/merge profiles

celestemartinez:
Implements the new function merge_profiles. 

It depends on https://github.com/iesl/openreview/pull/1605.

### DIFF
--- a/openreview/openreview.py
+++ b/openreview/openreview.py
@@ -32,18 +32,10 @@ class Client(object):
     :type token: str, optional
     """
     def __init__(self, baseurl = None, username = None, password = None, token= None):
+
         self.baseurl = baseurl
         if not self.baseurl:
-            self.baseurl = os.environ.get('OPENREVIEW_BASEURL', 'http://localhost:3000')
-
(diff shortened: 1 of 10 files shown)


openreview/openreview.py
```diff
@@ -475,6 +476,30 @@ def post_profile(self, profile):
         response = self.__handle_response(response)
         return Profile.from_json(response.json())
 
+    def merge_profiles(self, profileTo, profileFrom):
+        """
+        Merges two Profiles
+
+        :param profileTo: Profile object to merge to
+        :type profileTo: Profile
+        :parm profileFrom: Profile object to merge from
```
mspector: looks good. my one suggestion is that we add just a little more description here: maybe something like "Profile to merge from (this profile will be deleted)".

celestemartinez: Fixed

melisabok:
that is true, @celestemartinez could you create a test in test_client?

celestemartinez:
Created test in tes_client

PR #423, opened 2019-09-23
Feature/iclr20 registration

mohituniyal:
ICLR 2020 Registration Task for reviewers

Screen Shot 2019-09-23 at 4 01 31 PM (2)

### DIFF
--- a/openreview/conference/invitation.py
+++ b/openreview/conference/invitation.py
@@ -753,12 +753,11 @@ class InvitationBuilder(object):
             invitees.append(conference.get_area_chairs_id())
         invitees.append(conference.get_reviewers_id())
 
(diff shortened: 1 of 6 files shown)


openreview/conference/invitation.py
```diff
@@ -801,17 +788,97 @@ def set_registration_invitation(self, conference, start_date, due_date, with_are
...
+                "title": "Questionnaire for Reviewers"
             }
         ))
 
+        registration_content = {
+            'title': {
+                'value': conference.get_short_name() + ' Registration',
+                'order': 1
+            },
+            'profile confirmed': {
```
melisabok: I'm not vey sure about having two fields to confirm profile and expertise. I think they are the same but @mohituniyal  is right saying there are two different UIs to check this, @mspector what is your opinion?

PR #438, opened 2019-09-26
Add pagination to all tabs of PC console

zbialecki:
This greatly improves the speed of the console for large groups like ICLR.


melisabok:
This is awesome!

It looks really food. I like the pagination and how fast is to go to one page and the other, sorting, etc. 

I couldn't test the other tabs that I guess must also be paginated? or with review data. 

I noticed something weird when I go to a page, at the beginning it doesn't let me to scroll the page down. I hove to press the scroll bar to go the bottom.

zbialecki:
Cool, glad you like the change. You are correct that all the tabs are paginated in the same way. Also each tab remembers the last visited page so it can display the correct one when you go back to a certain tab.

I'm not sure about the scrolling issue... Maybe because there's a delay between page load and when the table is rendered? I haven't really be able to recreate what you're describing.

zbialecki:
I think this is ready to merge. @mohituniyal could you take a look when you get a chance?

PR #441, opened 2019-09-27
Get blind submissions when there are no scores available

melisabok:


melisabok:
Yes, I'm sorry for the messy code. I promise to refactor this later :(

PR #445, opened 2019-10-02
changes for enabling desk rejects

mohituniyal:


openreview/conference/builder.py
```diff
@@ -424,6 +424,11 @@ def create_withdraw_invitations(self):
 
         withdraw_invitations = self.invitation_builder.set_withdraw_invitation(self)
 
+    def create_desk_reject_invitations(self):
+        if not self.submission_stage.allow_desk_reject:
+            raise openreview.OpenReviewException('Conference does not allow desk rejects')
+
+        desk_reject_invitations = self.invitation_builder.set_desk_reject_invitation(self)
```
mspector: I don't think you need to pass `self` as an argument if you're calling it from the `self` object

mspector: whoops - thanks @mohituniyal for pointing out that I was wrong here.

PR #458, opened 2019-10-08
Add tabs for withdrawn and desk rejected papers

mohituniyal:


openreview/conference/templates/tabsConferenceDecisionsWebfield.js
```diff
@@ -8,55 +8,103 @@
...
   load().then(renderContent).then(Webfield.ui.done);
 }
 
 // Load makes all the API calls needed to get the data to render the page
 // It returns a jQuery deferred object: https://api.jquery.com/category/deferred-object/
 function load() {
-  var notesP = Webfield.getAll('/notes', { invitation: BLIND_SUBMISSION_ID, details: 'replyCount,original' });
 
-  var withdrawnNotesP = WITHDRAWN_INVITATION ? Webfield.getAll('/notes', { invitation: WITHDRAWN_INVITATION, noDetails: true }) : Promise.resolve({ notes: []});
+  var notesP = Webfield.api.getSubmissions(BLIND_SUBMISSION_ID, {
```
melisabok: it should be getAll instead of getSubmissions, we can not paginate the results from the server because we need to distribute them between the decisions tabs.

openreview/conference/templates/tabsConferenceDecisionsWebfield.js
```diff
@@ -68,61 +116,86 @@ function renderContent(notes, withdrawnNotes, decisionsNotes) {
...
+      if (activeTab < Object.keys(DECISION_HEADING_MAP).length) {
+        Webfield.ui.searchResults(
+          papersByDecision[containerId],
+          _.assign({}, paperDisplayOptions, {showTags: false, container: '#' + containerId})
+        );
+      }
     }, 100);
   });
 
+  var withdrawnNotesCount = withdrawnNotes.count || 0;
```
melisabok: adding the notes to the map doesn't work?

mohituniyal: I think we should keep these 2 tabs separate from the map as the map i.e. DECISION_HEADING_MAP contains notes for which decisions are available. I could still merge all results in one map if you feel strongly about it.

PR #462, opened 2019-10-10
Enable search in other tabs of PC console

zbialecki:
Search in areachair and reviewer status tabs searches over the names only.

### DIFF
--- a/openreview/conference/templates/programchairWebfield.js
+++ b/openreview/conference/templates/programchairWebfield.js
@@ -607,7 +607,7 @@ var displayPaperStatusTable = function() {
     // Currently only searching on note number and note title
     var filterFunc = function(row) {
       return (
-        row.note.number + '' === searchText ||
+        (row.note.number + '').indexOf(searchText) === 0 ||
         row.note.content.title.toLowerCase().indexOf(searchText) !== -1
       );
     };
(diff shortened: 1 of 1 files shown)


melisabok:
Great! Thanks Zach. 

It looks very good to me. 

Some comments:

1. I don't think we can sort the search results, so when the user tries to select a sort criteria we should clear the search textbox. Unless we can sort the search results :)
2. It seems the sorting in the Reviewer Status tab stopped working.
3. It would be nice to keep the search result when the user switches tabs and the come back, but this is a 'nice to have'.

zbialecki:
Points 1 and 2 are now fixed and I think we should save 3 for a future update. This should be ready  to be merged now!

mohituniyal:
Minor improvements: In case the search does not yield any result, we can show the usual text: "No papers to display" or something similar, instead of a blank table.

Screen Shot 2019-10-11 at 1 36 53 PM

> Just an idea: while searching by paper number, a search for, say, 33 should yield all these results: 33, 133, 233, ..., 1033, and so on. Would that be more helpful to the PCs?

melisabok:
@zbialecki what do you think?

I would prefer to deploy this today but I can wait for Mohit's improvements. Or do it in the next iteration.

zbialecki:
Matching multiple paper numbers would be easy to do and I guess it might be more intuitive.

Displaying an empty message is something I think we should save for a future release.

melisabok:
Should I merge it?

PR #467, opened 2019-10-16
Feature/recruitment notes visible to pcs

mohituniyal:
Fixes issue https://github.com/openreview/openreview-py/issues/455


openreview/conference/invitation.py
```diff
@@ -917,7 +917,7 @@ def set_reviewer_recruiter_invitation(self, conference, options = {}):
             'forum': None,
             'replyto': None,
             'readers': {
-                'values': ['~Super_User1']
+                'values': ['~Super_User1', conference.get_program_chairs_id()]
```
melisabok: use conference_id as readers should be enough

melisabok: @mohituniyal do you mind to put conference_id only as readers?

Issue #473, opened 2019-10-20
Builder: pre create Reviewers Submitted groups when there is manual assignments


melisabok:
These groups are being created when we deploy the matching result. We should create these groups no matter if the assignments are manual or through the matching system.

PR #475, opened 2019-10-21
Fix/create submitted groups

melisabok:


mohituniyal:
https://github.com/openreview/openreview-py/issues/473

PR #485, opened 2019-10-30
look for reviewer names in invitee_names

mohituniyal:
While reminding a reviewer, we will first check if the reviewer's id start with a '\~'. If so, we parse the user's name out of that. If it does not start with a '~', we look for the reviewer id in the email list that was passed to the recruit reviewer function and we check if invitee names were also provided. If both these conditions hold, we get the reviewer's name from invitee names list by looking at the same index as that held by the reviewer id in reviewer email list.
If nothing matches, we use "invitee" as the name in the email.
This allows us to supply names for reviewers (and ACs) to recruit reviewers function that can then be used in the emails.


openreview/conference/builder.py
```diff
@@ -740,9 +740,13 @@ def recruit_reviewers(self, emails = [], title = None, message = None, reviewers
 
         if remind:
             remind_reviewers = list(set(reviewers_invited_group.members) - set(reviewers_declined_group.members) - set(reviewers_accepted_group.members))
-            for reviewer in remind_reviewers:
-                name =  re.sub('[0-9]+', '', reviewer.replace('~', '').replace('_', ' ')) if reviewer.startswith('~') else 'invitee'
-                tools.recruit_reviewer(self.client, reviewer, name,
+            for reviewer_id in remind_reviewers:
```
melisabok: is there a way to reproduce this issue with a test?

It is difficult to see what you are fixing.

mohituniyal: @melisabok I have added a test and a description of the change to the PR.

PR #487, opened 2019-10-31
Lowercase input emails for reviewer reassignment

mohituniyal:
Linked to https://github.com/openreview/openreview-py/issues/486

### DIFF
--- a/openreview/conference/templates/areachairWebfield.js
+++ b/openreview/conference/templates/areachairWebfield.js
@@ -813,7 +813,8 @@ var registerEventHandlers = function() {
     var $currDiv = $('#' + paperNumber + '-add-reviewer');
     var userToAdd = $currDiv.find('input').attr('value_id').trim();
 
-    if (!userToAdd) {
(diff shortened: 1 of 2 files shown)


openreview/conference/templates/areachairWebfield.js
```diff
@@ -813,7 +813,8 @@ var registerEventHandlers = function() {
     var $currDiv = $('#' + paperNumber + '-add-reviewer');
     var userToAdd = $currDiv.find('input').attr('value_id').trim();
 
-    if (!userToAdd) {
+    if (!userToAdd || !((userToAdd.indexOf('@') > 0) || userToAdd.startsWith('~'))) {
```
melisabok: Are you missing ! before userToAdd.startsWith('~')?

mohituniyal: This is what I am checking here:
(No input Given) 
OR
NOT ( ( input has the char '@' but does not start at '@' ) OR (input starts with char '~' (e.g. when AC/PC select a name from the dropdown)) )

melisabok: you are right! I missed the parenthesis.

PR #489, opened 2019-11-02
Refactor decisions webfield and fix bug with consoles tab

zbialecki:
Fixes issue with loading spinner not being removed from the Your Consoles tab. Also fixes some other edge cases and makes code a bit cleaner.


melisabok:
It looks very good to me. Can I ask you another change?

We need to hide a decision tab if it doesn't have any note to display. We want this because sometimes rejected papers are hidden to the general public.

zbialecki:
The decision tabs to show are defined by the constant DECISION_HEADING_MAP so now you can safely remove any field from that object if you don't want the tabs showing up in the UI. Withdrawn submissions and Desk Rejected submissions are separate, and if you don't want those to appear you can just set the constants WITHDRAWN_SUBMISSION_ID and DESK_REJECTED_SUBMISSION_ID to null.

openreview/conference/templates/tabsConferenceDecisionsWebfield.js
```diff
@@ -225,6 +225,13 @@ function renderContent(notes, decisionNotes, withdrawnNotes, deskRejectedNotes,
     $('.tabs-container a[href="#desk-rejected-submissions"]').parent().hide();
   }
 
+  //Hide empty decision tabs
+  for (decision in papersByDecision) {
```
melisabok: @zbialecki please review :)

zbialecki: I would prefer not to do this. If a decision tab should be hidden then it can be removed from the DECISION_HEADING_MAP and it won't be rendered at all. Otherwise, I would think that all decision options should be visible otherwise people won't know they exist.

melisabok: We need to get the notes first in order to know if the result is empty or not :(

melisabok: Ok, I'm going to remove the reject key from the DECISION_HEADING_MAP

zbialecki: I'm not sure what you mean... at this point in the code all the notes have been fetched and categorized by decision.

melisabok: We first render the tabs and then the content.

zbialecki: Ah OK yes, but if the decision isn't in the HEADING_MAP then the tab won't be rendered either

melisabok: BTW I'm going to revert this commit.

PR #496, opened 2019-11-12
Feature/improve get pdf

carlosmondra:
This pull requests adds:
- A way to retrieve PDF revisions using the same `get_pdf` method by passing a new argument `is_reference`. If `is_reference` is True, then the passed id is assumed to be a Reference Id. I kept the argument as optional and False to keep backward compatibility
- A new section in read the docs that I called Mental Models. I am not sure if this is something that we want, so I could remove it, but it looks like this:

image


melisabok:
Do you mind to upgrade the version number so we can share it with the users?

carlosmondra:
@melisabok Can we merge this so that I can reply to the students?

PR #505, opened 2019-11-23
Improvements in methods for add/remove members from group

mohituniyal:


openreview/openreview.py
```diff
@@ -1084,21 +1084,26 @@ def add_members_to_group(self, group, members):
...
+        :param group: Group (or Group's id) to which the members will be added
+        :type group: Group or str
         :param members: Members that will be added to the group. Members should be in a string, unicode or a list format
         :type members: str, list, unicode
 
         :return: Group with the members added
         :rtype: Group
         """
         def add_member(group, members):
+            group_id = ''
```
melisabok: `group_id = [group if type(group) is string_types else group.id]`

mohituniyal: Done

PR #512, opened 2019-11-27
fix for withdraw and desk-reject invitations

mohituniyal:


openreview/conference/templates/withdraw_process.py
```diff
@@ -13,9 +13,7 @@ def process(client, note, invitation):
     forum_note.invitation = WITHDRAWN_SUBMISSION_ID
     if REVEAL_AUTHORS_ON_WITHDRAW:
         # REVEAL_AUTHORS_ON_WITHDRAW will only be True if this is a double blind conference
-        original_note = client.get_note(id = forum_note.original)
-        forum_note.content['authors'] = original_note.content['authors']
-        forum_note.content['authorids'] = original_note.content['authorids']
+        forum_note.content = {}
```
melisabok: Given the email we received this morning, we should save a bibtex entry for withdrawn papers.

melisabok: you can use this function:

https://github.com/openreview/openreview-py/blob/master/openreview/tools.py#L324

Issue #511, opened 2019-11-27
Withdrawn invitation should not break the link between the blind and original note


melisabok:
Also for desk reject invitation. 

We should keep accesible the original pdfs of all these papers.

mohituniyal:
Resolved.

PR #515, opened 2019-12-03
Fix for withdraw and desk-reject invitations + bibtex update

mohituniyal:
https://github.com/openreview/openreview-py/issues/511


mohituniyal:
@melisabok some help needed here. Any ideas on why the Py2 tests are not able to find the templates/withdraw_process.py file?

melisabok:
My tests are passing in localhost so I believe is something related to the cache in circleci, we need to merge this #516 first and we could try again.

Issue #514, opened 2019-12-02
Builder: create conference author groups members sorted by paper number


melisabok:

mohituniyal:
Closing this issue with mergehttps://github.com/openreview/openreview-py/pull/517.

PR #517, opened 2019-12-04
Add members in order to conference author group

mohituniyal:
https://github.com/openreview/openreview-py/issues/514

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -607,7 +607,7 @@ class Conference(object):
         return self.__set_reviewer_page()
 
     def set_authors(self):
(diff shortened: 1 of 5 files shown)


openreview/conference/builder.py
```diff
@@ -397,9 +397,9 @@ def get_homepage_options(self):
             options['deadline'] = self.homepage_header.get('deadline')
         return options
 
-    def get_submissions(self, accepted = False, details = None):
+    def get_submissions(self, accepted = False, details = None, sort = 'number:asc'):
```
melisabok: I prefer to have `sort=None` as a default parameter and specify the sort only when we need it.

PR #520, opened 2019-12-05
Fix/pc console mail endpoint

zbialecki:
Change /mail to /messages. Also replace some ES6 functions with ES5 versions.

Fixes https://github.com/openreview/openreview/issues/1773


zbialecki:
@mohituniyal when you have a minute can you review and merge this?

PR #524, opened 2019-12-10
Fix/upgrade python build

melisabok:


melisabok:
Yes, we should upgrade the instances soon. We need to plan this change carefully. 

In the meantime, I was wrong with the fix so I'm reverting my changes and adding the new fix:

https://github.com/openreview/openreview-py/pull/525

PR #529, opened 2019-12-16
Improvement - Reviewer registration task

mohituniyal:
Removed ICLR specific content from reviewer registration invitation. Also, the invitation's content can now be configured by passing in a content dict.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -301,10 +301,7 @@ class Conference(object):
         return self.get_invitation_id(self.recommendation_name, number)
 
     def get_registration_id(self, committee_id = None):
-        if committee_id == self.get_area_chairs_id():
-            return self.get_invitation_id(name = self.registration_name, prefix = self.id + '/' + self.area_chairs_name)
-        else:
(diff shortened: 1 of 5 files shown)


openreview/conference/builder.py
```diff
@@ -300,8 +300,8 @@ def get_bid_id(self, group_id):
     def get_recommendation_id(self, number = None):
         return self.get_invitation_id(self.recommendation_name, number)
 
-    def get_registration_id(self):
-        return self.get_invitation_id(self.registration_name)
+    def get_registration_id(self, committee_id = None):
```
melisabok: committee_id should be a mandatory parameter, you are using always as part of the invitation id

openreview/conference/matching.py
```diff
@@ -250,7 +250,13 @@ def _build_subject_area_scores(self, submissions):
         edges = []
         user_subject_areas = list(openreview.tools.iterget_notes(
             self.client,
-            invitation=self.conference.get_registration_id()))
+            invitation=self.conference.get_registration_id(self.conference.get_reviewers_id())))
```
melisabok: this has to be done once and it depends on `match_group` property of the matching class:

```
invitation=self.conference.get_registration_id(self.match_group.id)
```

PR #530, opened 2019-12-16
Only create the PC console webfield the first time through, after tha…

pmandler:
…t update necessary global vars.  This allows the console to be customized and not overwritten accidentally.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -168,15 +168,11 @@ class Conference(object):
         self.enable_reviewer_reassignment = enabled
 
         # Update PC & AC homepages
-        # edit enable_reviewer_reassignment
         pc_group = self.client.get_group(self.get_program_chairs_id())
         if enabled:
-            pc_group.web = pc_group.web.replace("ENABLE_REVIEWER_REASSIGNMENT = false;",
-                                                "ENABLE_REVIEWER_REASSIGNMENT = true;")
+            self.webfield_builder.edit_global_value(pc_group, "ENABLE_REVIEWER_REASSIGNMENT", "false", "true")
         else:
-            pc_group.web = pc_group.web.replace("ENABLE_REVIEWER_REASSIGNMENT = true;",
-                                                "ENABLE_REVIEWER_REASSIGNMENT = false;")
-        self.client.post_group(pc_group)
+            self.webfield_builder.edit_global_value(pc_group, "ENABLE_REVIEWER_REASSIGNMENT", "true", "false")
 
         if self.use_area_chairs:
             self.__set_area_chair_page()
@@ -499,11 +495,9 @@ class Conference(object):
 
         # Update PC console with double blind submissions
         pc_group = self.client.get_group(self.get_program_chairs_id())
-        submission_id = self.get_submission_id()
-        blind_id = self.get_blind_submission_id()
-        pc_group.web = pc_group.web.replace("var BLIND_SUBMISSION_ID = '" + submission_id + "';",
-                                            "var BLIND_SUBMISSION_ID = '" + blind_id + "';")
-        self.client.post_group(pc_group)
+        self.webfield_builder.edit_global_string_value(pc_group, "BLIND_SUBMISSION_ID",
+                                                self.get_submission_id(), self.get_blind_submission_id())
+
         return blinded_notes
 
     ## Deprecated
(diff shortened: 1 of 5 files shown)


openreview/conference/builder.py
```diff
@@ -163,11 +164,20 @@ def __create_decision_stage(self):
...
-    def __set_reviewer_reassignment(self, enabled = True):
+    def set_reviewer_reassignment(self, enabled = True):
         self.enable_reviewer_reassignment = enabled
 
         # Update PC & AC homepages
-        self.__set_program_chair_page()
+        # edit enable_reviewer_reassignment
+        pc_group = self.client.get_group(self.get_program_chairs_id())
+        if enabled:
+            pc_group.web = pc_group.web.replace("ENABLE_REVIEWER_REASSIGNMENT = false;",
```
melisabok: I prefer to do all the webfield operations in the Webfield Builder.

openreview/conference/builder.py
```diff
@@ -483,8 +497,13 @@ def create_blind_submissions(self, force=False):
                 blind_note = self.client.post_note(blind_note)
             blinded_notes.append(blind_note)
 
-        # Update page with double blind submissions
-        self.__set_program_chair_page()
+        # Update PC console with double blind submissions
+        pc_group = self.client.get_group(self.get_program_chairs_id())
+        submission_id = self.get_submission_id()
+        blind_id = self.get_blind_submission_id()
+        pc_group.web = pc_group.web.replace("var BLIND_SUBMISSION_ID = '" + submission_id + "';",
```
melisabok: same here

openreview/conference/builder.py
```diff
@@ -163,11 +164,20 @@ def __create_decision_stage(self):
         notes = list(self.get_submissions())
         return self.invitation_builder.set_decision_invitation(self, notes)
 
-    def __set_reviewer_reassignment(self, enabled = True):
```
melisabok: why did you remove the underscores here?

openreview/conference/builder.py
melisabok: We can create a method to reset the settings of the webfield template:

```
self.webfield_builder.set_program_chair_constant('ENABLE_REVIEWER_REASSIGNMENT', 'true')
```

openreview/conference/builder.py
melisabok: you can do:

```
self.webfield_builder.set_program_chair_constant('BLIND_SUBMISSION_ID', blind_id)
```

openreview/conference/builder.py
```diff
@@ -670,14 +689,14 @@ def set_assignment(self, user, number, is_area_chair = False):
             )
             return result
 
-    def set_assignments(self, assingment_title, is_area_chair=False):
+    def set_assignments(self, assignment_title, is_area_chair=False):
```
melisabok: thanks for fixing the typo

openreview/conference/builder.py
```diff
@@ -1189,7 +1208,11 @@ def get_result(self):
             self.conference.set_submission_stage(self.submission_stage)
 
         ## Create committee groups before any other stage that requires them to create groups and/or invitations
-        self.conference.set_program_chairs()
+        program_chairs_group = self.conference.set_program_chairs()
+        # if first time, add PC console
+        if not program_chairs_group.web:
```
melisabok: may be we can move this if to the `set_program_chairs` method

openreview/conference/builder.py
pmandler: Because I couldn't call it from set_assignments().  And I expect we'll want to remove the ability to reassign after reviews are in and would want to call it from conference.

openreview/conference/webfield.py
```diff
@@ -400,3 +400,11 @@ def set_program_chair_page(self, conference, group):
             group.web = content
             return self.client.post_group(group)
 
+    def edit_global_value(self, group, global_name, old_value, new_value):
+        group.web = group.web.replace("var "+global_name+" = "+old_value+";", "var "+global_name+" = "+new_value+";")
+        return self.client.post_group(group)
+
+    def edit_global_string_value(self, group, global_name, old_value, new_value):
```
mohituniyal: It seems edit_global_value and edit_global_string_value are doing the same thing. Is there a difference that I am missing @pmandler ?

mohituniyal: Also, with a regex search, I think, we can make it work without requiring the old value.

PR #542, opened 2020-01-13
define assignment config for solver class

mohituniyal:
Adds a radio button selection to let the user choose between MinMax and FairFlow solvers


melisabok:
We need to make sure to deploy this change with the matching system. Unless there is a way to keep backward compatibility.

PR #543, opened 2020-01-15
pull in information about different stages if submission stage is com…

pmandler:
…plete

This needs to be tested.

### DIFF
--- a/openreview/conference/helpers.py
+++ b/openreview/conference/helpers.py
@@ -96,6 +96,16 @@ def get_conference(client, request_form_id):
     if 'Organizers will assign papers manually' in paper_matching_options:
         builder.enable_reviewer_reassignment(enable = True)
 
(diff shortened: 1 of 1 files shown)


openreview/conference/helpers.py
```diff
@@ -96,6 +96,16 @@ def get_conference(client, request_form_id):
     if 'Organizers will assign papers manually' in paper_matching_options:
         builder.enable_reviewer_reassignment(enable = True)
 
+    if submission_due_date < datetime.datetime.now():
+        # if submissions have expired, check for creating bidding, reviews and meta-reviews
+        if 'bid_due_date' in note.content.keys():
```
melisabok: I think we don't need to get 'keys()':

```
if 'bid_due_date' in note.content
```

melisabok:
I'm going to close this PR and get back to this later.

PR #556, opened 2020-02-01
Render a status table for author console

melisabok:
Similar to the AC console, currently it is hiding review and meta review status.

Zach, please check the css, the author names and details aren't correct.

### DIFF
--- a/openreview/conference/templates/authorWebfield.js
+++ b/openreview/conference/templates/authorWebfield.js
@@ -48,9 +48,33 @@ function load() {
   } else {
     authorNotesP = Webfield.get('/notes', {
       'content.authorids': user.profile.id,
-      invitation: SUBMISSION_ID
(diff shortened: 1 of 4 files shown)


zbialecki:
I think the table looks good, and there aren't too many things I would change about the design, but I'm wondering if you plan on showing the review and meta review columns in the future? If not, I think we should remove that code from this webfield as a lot of it is unnecessary.

melisabok:
I think it would be nice to show the status of the reviews to the authors. 

I have another problem here, I need to get the blind submissions of only the papers where the user is an author.

melisabok:
I need to fix some CSS, author list and show details are not showed correctly:

Screen Shot 2020-02-14 at 6 27 35 PM

zbialecki:
I'm reviewing this now, I can fix the CSS.

PR #557, opened 2020-02-02
Parametrize the readers of the submission after withdraw or desk reject

### LINKED ISSUE #551: Withdraw and Reject invitation: make notes visible to the readers of the blinded note
Currently the readers are 'everyone', we should parametrize the readers of these notes based on their parent readers.

melisabok:
Fixes #551.

- By default withdrawn and desk-rejected notes are NOT revealed to the public. 
- Added tests for ECCV.

### DIFF
--- a/openreview/conference/invitation.py
+++ b/openreview/conference/invitation.py
@@ -546,7 +546,7 @@ class PaperDeskRejectInvitation(openreview.Invitation):
                 file_content = file_content.replace(
                     'REVEAL_AUTHORS_ON_DESK_REJECT = False',
                     'REVEAL_AUTHORS_ON_DESK_REJECT = True')
-            if conference.submission_stage.reveal_submissions_on_withdraw:
+            if conference.submission_stage.reveal_submissions_on_desk_reject:
                 file_content = file_content.replace(
                     'REVEAL_SUBMISSIONS_ON_DESK_REJECT = False',
                     'REVEAL_SUBMISSIONS_ON_DESK_REJECT = True')
--- a/openreview/conference/invitation.py
+++ b/openreview/conference/invitation.py
@@ -290,7 +290,7 @@ class CommentInvitation(openreview.Invitation):
 
 class WithdrawnSubmissionInvitation(openreview.Invitation):
 
-    def __init__(self, conference):
+    def __init__(self, conference, reveal_authors, reveal_submission):
 
         content = {
             'authorids': {
@@ -305,7 +305,7 @@ class WithdrawnSubmissionInvitation(openreview.Invitation):
             }
         }
 
-        if (conference.submission_stage.double_blind and not conference.submission_stage.reveal_authors_on_withdraw):
+        if (conference.submission_stage.double_blind and not reveal_authors):
             content['authors'] = {
                 'values': ['Anonymous']
             }
@@ -313,7 +313,7 @@ class WithdrawnSubmissionInvitation(openreview.Invitation):
                 'values-regex': '.*'
             }
 
-        if conference.submission_stage.reveal_submissions_on_withdraw:
+        if reveal_submission:
             readers = {
                 'values': ['everyone']
             }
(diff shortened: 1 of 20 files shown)


openreview/conference/invitation.py
```diff
@@ -350,6 +354,17 @@ def __init__(self, conference, note):
...
 
+
+        if conference.submission_stage.reveal_submissions_on_withdraw:
+            readers = {
+                'description': 'User groups that will be able to read this withdraw note.',
+                'values': ['everyone']
+            }
+        else:
+            readers = {
+                'values-copied': [conference.id, '{signatures}']
```
mohituniyal: Maybe we can include the paper author group here, instead of just the signature of the withdraw comment note.

melisabok: you are right, because this is an invitation per paper, right?

mohituniyal: Correct. This invitation creates a reply note on the forum and changes the invitation of the forum to a conference level invitation for all withdrawn papers.
So, it would be apt to show this reply note to all the authors of this paper.

mohituniyal:
I found that this change is failing the pytests because of a very strange reason.
https://github.com/openreview/openreview-py/blob/8236137be0162fe375fff40ab0064d139fadfccd/openreview/conference/invitation.py#L37 -> At this line, somehow a list that was passed into subject areas is now available in additional_fields which is supposed to be a dict.

melisabok:
@mohituniyal I can take a look at the error, thanks.

melisabok:
Please don't merge it yet, I'm adding a test for ECCV

melisabok:
Now,  it should be ready to merge.

openreview/conference/invitation.py
mohituniyal: The paper authors are still not able to see the desk reject note.

Screen Shot 2020-02-13 at 10 35 29 AM

melisabok: ahhh good catch, I'll fix it.

melisabok: do you think we need to add all the members of the committee: assigned AC, assigned Reviewer

mohituniyal: I'm not certain about that. Ideally, we should list them as readers. 
But, Reviewers and ACs would not have a direct link to this paper on their consoles anymore, since all the consoles get assignments by querying for notes responding to `Submission` or `Blind_Submission` invitation.
So, I don't think adding them as readers would have an impact.

melisabok: I'm setting the readers of the desk reject note with the same value than the blinded note.

melisabok: same for withdraw

mohituniyal: Sounds good.

celestemartinez:
One comment: if we call get_conference we would be resetting these:
`allow_withdraw = True`
`reveal_authors_on_withdraw = True`
`allow_desk_reject = True`
`reveal_authors_on_desk_reject = True`
Should we set these to False as default?
Maybe we should eventually get these from the request form as well.

melisabok:
Ohhh, you are right, good catch! I will fix it.

melisabok:
> Maybe we should eventually get these from the request form as well.

I agree, the best option here is to see the two different stages or buttons to setup Withdraw and Desk-reject invitations.

celestemartinez:
Last question: `allow_withdraw` and `allow_desk_reject` are set to False in the builder, why do we set them as True in get_conference_builder?

melisabok:
that is an interesting question :)

I don't know why we need to set a boolean to enable the withdraw and desk reject. I think that calling to `create_withdraw_invitations` or `create_desk_reject_invitations` should enable this automatically. 

I think I can remove this. @mohituniyal what do you think?

```
def create_withdraw_invitations(
          reveal_authors_on_withdraw=False, 
          reveal_submissions_on_withdraw=False
)

def create_desk_reject_invitations(
            reveal_authors_on_desk_reject=False,
            reveal_submissions_on_desk_reject=False
)
```

melisabok:
Please see my last commit.

PR #558, opened 2020-02-03
check if tpms scores in the input file are floats

mohituniyal:
Doing a regex check seemed like the easiest way to do this. Type casting to float did not work because it would still permit values like "-inf".


melisabok:
we could do this:

```
string = "abcd"
try:
    i = float(string)
    print i
except ValueError:
    #Handle the exception
    print 'error'
```

PR #566, opened 2020-02-11
Fix/PC console refactor

zbialecki:
This improves PC console performance by attempting to load all data in parallel (profiles still must be loaded after other data loads). In addition the configuration table is now loaded independently so that it shows almost immediately. Other tabs are not rendered until actually clicked on.

This PR also fixes a bug where selecting all papers in a filtered list actually selected all submitted papers.

Related to https://github.com/openreview/openreview/issues/1888


zbialecki:
Whew tests are finally passing! This is ready to review.

PR #567, opened 2020-02-11
Fix/reassign reviewer with conflicts

xkopenreview:
**1.changed the way that reviewers dropdown is rendered.** 
Existing code is rendering reviewers dropdown for all papers with the same dropdown options(all reviewers) on page load. Changed to render in click event of "Show reviewers" anchor tag by loading edges to filter out the reviewers which have conflict with the clicked paper.
**2. add a test case in test_double_blind_conference.py**
the setup in this test case has in total 3 reviewers and 1 paper for the area chair ac@mail.com. 1 of the reviewers has conflict with the paper. The test verifies that the reassign reviewers dropdown for the area chair has 2 options instead of 3 (1 reviewer with conflict is filtered out).

*please note that updateReviewerContainer method in areachairWebfield.js is also invoked when area chair reassign and unassign reviewers.

### DIFF
--- a/openreview/conference/templates/areachairWebfield.js
+++ b/openreview/conference/templates/areachairWebfield.js
@@ -1,4 +1,3 @@
-
 // Constants
 var CONFERENCE_ID = '';
 var SHORT_PHRASE = '';
--- a/openreview/conference/templates/areachairWebfield.js
+++ b/openreview/conference/templates/areachairWebfield.js
@@ -16,7 +16,7 @@ var WILDCARD_INVITATION = CONFERENCE_ID + '.*';
 var ANONREVIEWER_WILDCARD = CONFERENCE_ID + '/Paper.*/AnonReviewer.*';
 var AREACHAIR_WILDCARD = CONFERENCE_ID + '/Paper.*/Area_Chair[0-9]+$';
 var REVIEWER_GROUP = CONFERENCE_ID + '/' + REVIEWER_NAME;
-var REVIEWER_GROUP_WITH_CONFLICT=`${REVIEWER_GROUP}/-/Conflict`;
+var REVIEWER_GROUP_WITH_CONFLICT = `${REVIEWER_GROUP}/-/Conflict`;
 
 var reviewerSummaryMap = {};
 var allReviewers = [];
--- a/openreview/conference/templates/areachairWebfield.js
+++ b/openreview/conference/templates/areachairWebfield.js
@@ -10,15 +10,17 @@ var OFFICIAL_REVIEW_NAME = '';
 var OFFICIAL_META_REVIEW_NAME = '';
 var LEGACY_INVITATION_ID = false;
 var ENABLE_REVIEWER_REASSIGNMENT = false;
+var ENABLE_REVIEWER_REASSIGNMENT_TO_OUTSIDE_REVIEWERS = false;
 
 var WILDCARD_INVITATION = CONFERENCE_ID + '.*';
 var ANONREVIEWER_WILDCARD = CONFERENCE_ID + '/Paper.*/AnonReviewer.*';
 var AREACHAIR_WILDCARD = CONFERENCE_ID + '/Paper.*/Area_Chair[0-9]+$';
 var REVIEWER_GROUP = CONFERENCE_ID + '/' + REVIEWER_NAME;
-var REVIEWER_GROUP_WITH_CONFLICT = `${REVIEWER_GROUP}/-/Conflict`;
+var REVIEWER_GROUP_WITH_CONFLICT = REVIEWER_GROUP+'/-/Conflict';
 
 var reviewerSummaryMap = {};
 var allReviewers = [];
+var paperAndReviewersWithConflict ={};
 
 // Main function is the entry point to the webfield code
 var main = function() {
@@ -506,16 +508,21 @@ var renderStatusTable = function(profiles, notes, allInvitations, completedRevie
   }
 };
 
-var updateReviewerContainer = function (paperNumber, renderEmptyDropdown, reviewerWithConflictForThisPaper) {
+var updateReviewerContainer = function (paperNumber, renderEmptyDropdown) {
   renderEmptyDropdown = renderEmptyDropdown || false;
   $addReviewerContainer = $('#' + paperNumber + '-add-reviewer');
   $reviewerProgressContainer = $('#' + paperNumber + '-reviewer-progress');
   paperForum = $reviewerProgressContainer.data('paperForum');
 
-  var paperNoteId=reviewerSummaryMap[paperNumber].noteId;
-  var dropdownOptions=[];
-  if(renderEmptyDropdown===false){
-    dropdownOptions = _.map(!reviewerWithConflictForThisPaper ? allReviewers : allReviewers.filter(reviewer => reviewerWithConflictForThisPaper.indexOf(reviewer) == -1), function(member) {
+  var paperNoteId = reviewerSummaryMap[paperNumber].noteId;
+  var dropdownOptions = [];
+  if (!renderEmptyDropdown) {
+    var reviewersToRender = allReviewers;
+    var reviewerWithConflictForThisPaper = paperAndReviewersWithConflict[paperNumber];
+    if (reviewerWithConflictForThisPaper) reviewersToRender = allReviewers.filter(function (reviewer) {
+      return reviewerWithConflictForThisPaper.indexOf(reviewer) === -1;
+    })
+    dropdownOptions = _.map(reviewersToRender, function (member) {
       return {
         id: member,
         description: view.prettyId(member)
@@ -805,19 +812,27 @@ var registerEventHandlers = function() {
     var paperIdArray = $(this).attr('href').split("-");//href attribute is in format of #{paperId}-reviewers
     paperIdArray.pop(); //
     var paperId = paperIdArray.join('-').substring(1);
-    var paperNumber = reviewerSummaryMap[Object.keys(reviewerSummaryMap).filter(p => reviewerSummaryMap[p].noteId === paperId)].paperNumber;
+    var paperIndex = Object.keys(reviewerSummaryMap).filter(function (p) {
+      return reviewerSummaryMap[p].noteId === paperId;
+    })
+    var paperNumber = reviewerSummaryMap[paperIndex].paperNumber;
     if ($(this).text() === 'Show reviewers') {
       $(this).text('Hide reviewers');
       updateReviewerContainer(paperNumber, true);//render empty dropdown as placeholder for visual consistency
-      Webfield.get('/edges', { head: paperId, invitation: REVIEWER_GROUP_WITH_CONFLICT })
-        .then(result => {
-          var reviewerWithConflictForThisPaper = result.edges.map(p => p.tail);
-          $addReviewerContainer.children().remove()//remove empty dropdown when actual data arrive
-          updateReviewerContainer(paperNumber, false, reviewerWithConflictForThisPaper)//render dropdown with filtered value
+      Webfield.get('/edges', {
+        head: paperId,
+        invitation: REVIEWER_GROUP_WITH_CONFLICT
+      }).then(function (result) {
+        var reviewerWithConflictForThisPaper = result.edges.map(function (p) {
+          return p.tail;
         });
+        paperAndReviewersWithConflict[paperNumber]=reviewerWithConflictForThisPaper;
+        $addReviewerContainer.children().remove(); //remove empty dropdown when actual data arrive
+        updateReviewerContainer(paperNumber, false); //render dropdown with filtered value
+      });
     } else {
       $(this).text('Show reviewers');
-      $(`#${paperNumber}-add-reviewer`).children().remove()
+      $(paperNumber + '-add-reviewer').children().remove()
     }
   });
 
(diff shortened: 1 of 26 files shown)


openreview/conference/templates/areachairWebfield.js
```diff
@@ -16,9 +16,11 @@ var WILDCARD_INVITATION = CONFERENCE_ID + '.*';
 var ANONREVIEWER_WILDCARD = CONFERENCE_ID + '/Paper.*/AnonReviewer.*';
 var AREACHAIR_WILDCARD = CONFERENCE_ID + '/Paper.*/Area_Chair[0-9]+$';
 var REVIEWER_GROUP = CONFERENCE_ID + '/' + REVIEWER_NAME;
+var REVIEWER_GROUP_WITH_CONFLICT=`${REVIEWER_GROUP}/-/Conflict`;
 
 var reviewerSummaryMap = {};
 var allReviewers = [];
+var reviewerWithConflictForThisPaper=[];
```
melisabok: please add spaces between '='

openreview/conference/templates/areachairWebfield.js
```diff
@@ -505,17 +508,23 @@ var renderStatusTable = function(profiles, notes, allInvitations, completedRevie
   }
 };
 
-var updateReviewerContainer = function(paperNumber) {
+var updateReviewerContainer = function(paperNumber,renderEmptyDropdown=false) {
```
melisabok: same here, space after ','

melisabok: do we support default values in js? I didn't know it supports that.

openreview/conference/templates/areachairWebfield.js
```diff
@@ -803,8 +806,18 @@ var registerEventHandlers = function() {
   $('#group-container').on('click', 'a.collapse-btn', function(e) {
     if ($(this).text() === 'Show reviewers') {
       $(this).text('Hide reviewers');
+      var paperId=$(this).attr('href').split("-")[0].replace("#","");
+      var paperNumber=reviewerSummaryMap[Object.keys(reviewerSummaryMap).filter(p=>reviewerSummaryMap[p].noteId===paperId)].paperNumber;
+      updateReviewerContainer(paperNumber,true);//render empty dropdown as placeholder for visual consistency
+      var allReviewersWithConflictP=Webfield.get('/edges',{head:paperId,invitation:REVIEWER_GROUP_WITH_CONFLICT})
```
melisabok: you don't need to define this variable `var allReviewersWithConflictP1`, you are not using it anywhere.

openreview/conference/templates/areachairWebfield.js
```diff
@@ -803,8 +806,18 @@ var registerEventHandlers = function() {
...
+      .then(result=>{
+        reviewerWithConflictForThisPaper=result.edges.map(p=>p.tail);
```
melisabok: I think instead of being a global variable, this can be passed to `updateReviewerContainer`

openreview/conference/templates/areachairWebfield.js
```diff
@@ -505,17 +508,23 @@ var renderStatusTable = function(profiles, notes, allInvitations, completedRevie
...
   paperForum = $reviewerProgressContainer.data('paperForum');
 
-  var dropdownOptions = _.map(allReviewers, function(member) {
-    return {
-      id: member,
-      description: view.prettyId(member)
-    };
-  });
+  var paperNoteId=reviewerSummaryMap[paperNumber].noteId;
+  //var reviewerWithConflictForThisPaper=allReviewerWithConflict.filter(reviewerWithConflict=>reviewerWithConflict.head===paperNoteId).map(p=>p.tail);
```
melisabok: you can remove this line

openreview/conference/templates/areachairWebfield.js
zbialecki: This is a feature of ES6. Since this code is running on the client side and needs to support IE11 we cannot use any ES6 features in webfield templates.

openreview/conference/templates/areachairWebfield.js
```diff
@@ -505,17 +508,23 @@ var renderStatusTable = function(profiles, notes, allInvitations, completedRevie
...
+  var dropdownOptions=[];
+  if(renderEmptyDropdown===false){
+    dropdownOptions = _.map(allReviewers.filter(reviewer=>reviewerWithConflictForThisPaper.indexOf(reviewer)==-1), function(member) {
```
zbialecki: Arrow functions (=>) are also ES6 and not supported.

tests/test_double_blind_conference.py
```diff
@@ -1334,6 +1334,36 @@ def test_consoles(self, client, test_client, selenium, request_page):
...
+        ac_client=openreview.Client(baseurl = 'http://localhost:3000', username='ac@mail.com', password='1234')
+        request_page(selenium,"http://localhost:3000/group?id=AKBC.ws/2019/Conference/Area_Chairs",ac_client.token)
+
+        show_Reviewers_AnchorTag=selenium.find_element_by_xpath('//*[@id="1-reviewer-progress"]/a')
+        selenium.execute_script("arguments[0].click()",show_Reviewers_AnchorTag) #click on "show reviewers anchor tag"
+        time.sleep(5) #leave some time for the API and js to complete
+
+        #get the reviewers textbox and click on it to show dropdown
+        assign_Reviewer_Textbox_Present=EC.presence_of_element_located((By.XPATH,'//*[@id="1-add-reviewer"]/div/input'))
+        WebDriverWait(selenium,5).until(assign_Reviewer_Textbox_Present)
```
zbialecki: What is the difference between WebDriverWait and time.sleep? When should we use one vs. another?

openreview/conference/templates/areachairWebfield.js
xkopenreview: looks like we stuck at es5 as long as we have the user defined webfields

tests/test_double_blind_conference.py
xkopenreview: i think the timeout defined in webdriverwait is the max time that it will wait and the actual waiting time might be shorter if the specified element is generated in dom quicker.

xkopenreview:
found during testing that when reviewers section of multiple papers are expanded and collapsed, reassign reviewer textbox/dropdown/button can be removed from dom mistakenly.

This error happened because the value of $addReviewerContainer object is updated only assigned when updateReviewerContainer() method is called so it can be out of sync with actual user behavior.

Will try $addReviewerContainer or change the behavior of children removal of $addReviewerContainer

xkopenreview:
found another issue during fixing the $addReviewerContainer issue above.
It looks like the noteId can contain - so when splitting by dash instead of taking the 1st element, need to do a pop and then join with -. also change the removal of # by substring instead of replacement.

xkopenreview:
above errors are corrected in 27a9b7b

openreview/conference/templates/areachairWebfield.js
```diff
@@ -16,6 +16,7 @@ var WILDCARD_INVITATION = CONFERENCE_ID + '.*';
 var ANONREVIEWER_WILDCARD = CONFERENCE_ID + '/Paper.*/AnonReviewer.*';
 var AREACHAIR_WILDCARD = CONFERENCE_ID + '/Paper.*/Area_Chair[0-9]+$';
 var REVIEWER_GROUP = CONFERENCE_ID + '/' + REVIEWER_NAME;
+var REVIEWER_GROUP_WITH_CONFLICT=`${REVIEWER_GROUP}/-/Conflict`;
```
melisabok: please add a space between '='

zbialecki:
I still see a number of places where you are using `=>` functions, as well as template strings (also an ES6 feature). Please make sure you are using proper spacing around operators also.

melisabok:
@xkopenreview you need to remove the arrow functions before I can merge this PR. They are not supported by all the browsers.

xkopenreview:
changed fat arrow to function calls.
will try if it's possible to have ie to run es6

zbialecki:
Maybe you didn't push your changes because I'm still seeing the arrow functions.

openreview/conference/templates/areachairWebfield.js
```diff
@@ -16,6 +16,7 @@ var WILDCARD_INVITATION = CONFERENCE_ID + '.*';
 var ANONREVIEWER_WILDCARD = CONFERENCE_ID + '/Paper.*/AnonReviewer.*';
 var AREACHAIR_WILDCARD = CONFERENCE_ID + '/Paper.*/Area_Chair[0-9]+$';
 var REVIEWER_GROUP = CONFERENCE_ID + '/' + REVIEWER_NAME;
+var REVIEWER_GROUP_WITH_CONFLICT = `${REVIEWER_GROUP}/-/Conflict`;
```
zbialecki: All template strings need to be replaced with regular strings

openreview/conference/templates/areachairWebfield.js
```diff
@@ -801,10 +805,28 @@ var registerEventHandlers = function() {
   });
 
   $('#group-container').on('click', 'a.collapse-btn', function(e) {
+    var paperIdArray = $(this).attr('href').split("-");//href attribute is in format of #{paperId}-reviewers
+    paperIdArray.pop(); //
+    var paperId = paperIdArray.join('-').substring(1);
+    var paperNumber = reviewerSummaryMap[Object.keys(reviewerSummaryMap).filter(function (p) {
```
zbialecki: This would be better separated into 2 statements with an intermediate var for `Object.keys...`

xkopenreview:
> Maybe you didn't push your changes because I'm still seeing the arrow functions.

arrow functions are removed in commit 34195a4.
which lines do you see arrow functions? I did a search in areachairwebfield.js and can't find any

openreview/conference/templates/areachairWebfield.js
```diff
@@ -505,17 +507,25 @@ var renderStatusTable = function(profiles, notes, allInvitations, completedRevie
...
-  var dropdownOptions = _.map(allReviewers, function(member) {
-    return {
-      id: member,
-      description: view.prettyId(member)
-    };
-  });
+  var paperNoteId=reviewerSummaryMap[paperNumber].noteId;
+  var dropdownOptions=[];
+  if (renderEmptyDropdown === false) {
+    dropdownOptions = _.map(!reviewerWithConflictForThisPaper ? allReviewers : allReviewers.filter(function (reviewer) {
```
zbialecki: This should also be separated into at least 2 statements

openreview/conference/templates/areachairWebfield.js
```diff
@@ -505,17 +507,25 @@ var renderStatusTable = function(profiles, notes, allInvitations, completedRevie
...
 
-  var dropdownOptions = _.map(allReviewers, function(member) {
-    return {
-      id: member,
-      description: view.prettyId(member)
-    };
-  });
+  var paperNoteId=reviewerSummaryMap[paperNumber].noteId;
+  var dropdownOptions=[];
+  if (renderEmptyDropdown === false) {
```
zbialecki: It would be more idiomatic to say `!renderEmptyDropdown`

openreview/conference/templates/areachairWebfield.js
```diff
@@ -505,17 +507,25 @@ var renderStatusTable = function(profiles, notes, allInvitations, completedRevie
...
+    dropdownOptions = _.map(!reviewerWithConflictForThisPaper ? allReviewers : allReviewers.filter(function (reviewer) {
+      return reviewerWithConflictForThisPaper.indexOf(reviewer) == -1;
```
zbialecki: Should use `===` here

openreview/conference/templates/areachairWebfield.js
```diff
@@ -505,17 +507,25 @@ var renderStatusTable = function(profiles, notes, allInvitations, completedRevie
...
   $reviewerProgressContainer = $('#' + paperNumber + '-reviewer-progress');
   paperForum = $reviewerProgressContainer.data('paperForum');
 
-  var dropdownOptions = _.map(allReviewers, function(member) {
-    return {
-      id: member,
-      description: view.prettyId(member)
-    };
-  });
+  var paperNoteId=reviewerSummaryMap[paperNumber].noteId;
```
zbialecki: Please add spaces around `=`

xkopenreview:
comments are addressed in commit cffdf25

openreview/conference/templates/areachairWebfield.js
```diff
@@ -827,6 +856,17 @@ var registerEventHandlers = function() {
       promptError('Reviewer ' + view.prettyId(userToAdd) + ' has already been assigned to Paper ' + paperNumber.toString());
       return false;
     }
+
+    var reviewersWithConflict = paperAndReviewersWithConflict[paperNumber];
+    if (!ENABLE_REVIEWER_REASSIGNMENT_TO_OUTSIDE_REVIEWERS && //if enabled then skip checking
+      (_.includes(allReviewers, userToAdd) == false) || //input is not in dropdown or
```
zbialecki: Please change this to `=== false` or `!_.includes`. We try to avoid using double equals.

openreview/conference/templates/areachairWebfield.js
```diff
@@ -826,6 +855,25 @@ var registerEventHandlers = function() {
...
+    if(reviewersWithConflict && _.includes(reviewersWithConflict, userToAdd)){
+      promptError('The reviewer entered is invalid');
+      $currDiv.find('input').val('');
+      $currDiv.find('input').attr('value_id', '');
+      return false;
+    }
+
+    if (!ENABLE_REVIEWER_REASSIGNMENT_TO_OUTSIDE_REVIEWERS && //check only if reassign to outside is disabled
+      (_.includes(allReviewers, userToAdd) === false)) { // not in allreviewers means is a outside reviewer
+      promptError('Please choose only from the reviewers from the dropdown');
```
melisabok:
```suggestion
      promptError('Please choose only reviewers from the dropdown');
```

PR #572, opened 2020-02-18
Fix reveal authors = False

melisabok:


celestemartinez:
Both desk reject and withdraw process functions fail when `reveal_authors=True` and `reveal_submission=False`

PR #578, opened 2020-02-26
Add from and ignoreGroups parameters

melisabok:


openreview/openreview.py
```diff
@@ -1043,7 +1044,7 @@ def delete_group(self, group_id):
         return response.json()
 
 
-    def post_message(self, subject, recipients, message):
+    def post_message(self, subject, recipients, message, ignoreRecipients=None, sender=None):
```
mohituniyal: Is this param `sender` implemented in the backend ?

mohituniyal: Is this param `sender` implemented in the backend ?

melisabok: https://github.com/openreview/openreview/pull/1932

melisabok: https://github.com/openreview/openreview/pull/1926

melisabok: these are the two PRs already merged in master, I've been using this to send the reminders.

PR #581, opened 2020-03-02
Show count of bids and link to edge browser in reviewer summary cell

zbialecki:
Fixes https://github.com/openreview/openreview/issues/1916

### DIFF
--- a/openreview/conference/templates/programchairWebfield.js
+++ b/openreview/conference/templates/programchairWebfield.js
@@ -12,6 +12,7 @@ var DECISION_NAME = '';
 var BID_NAME = '';
 var RECOMMENDATION_NAME = '';
 var COMMENT_NAME = '';
+var SCORES_NAME = '';
 var AUTHORS_ID = '';
 var REVIEWERS_ID = '';
 var AREA_CHAIRS_ID = '';
@@ -137,10 +138,10 @@ var loadConfigurationTab = function() {
 };
 
 var getAllReviewers = function() {
-  return Webfield.get('/groups', { id: REVIEWERS_ID })
+  return Webfield.get('/groups', { id: REVIEWERS_ID, limit: 1 })
   .then(function(result) {
     var members = _.get(result, 'groups[0].members', []);
-    return members.sort()
+    return members.sort();
   });
 };
 
@@ -149,10 +150,10 @@ var getAllAreaChairs = function() {
     return $.Deferred().resolve([]);
   }
 
-  return Webfield.getAll('/groups', { id: AREA_CHAIRS_ID })
+  return Webfield.get('/groups', { id: AREA_CHAIRS_ID, limit: 1 })
   .then(function(result) {
     var members = _.get(result, 'groups[0].members', []);
-    return members.sort()
+    return members.sort();
   });
 };
 
@@ -188,11 +189,11 @@ var getUserProfiles = function(userIds, reviewerBidCounts, areaChairRecommendati
 
   var idSearch = ids.length ?
     Webfield.post('/profiles/search', { ids: ids }) :
-    $.Deferred().resolve([])
+    $.Deferred().resolve([]);
 
   var emailSearch = emails.length ?
     Webfield.post('/profiles/search', { emails: emails }) :
-    $.Deferred().resolve([])
+    $.Deferred().resolve([]);
 
   return $.when(idSearch, emailSearch)
     .then(function(idResults, emailResults) {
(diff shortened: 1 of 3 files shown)


melisabok:
Running test_eccv_conference.py the reviewers/ACs are not being shown because there are no assignment groups, we should list them anyway, it is very important to see the status of the bids before the assignment.

Screen Shot 2020-03-03 at 1 28 13 PM

melisabok:
This line is iterating through all the area chair group ids, we can start the for using the area chair members and then search the ids:

https://github.com/openreview/openreview-py/blob/master/openreview/conference/templates/programchairWebfield.js#L1046

This will fix another issue that is when an AC/Reviewer doesn't have assigned papers is not being shown.

zbialecki:
OK, I'll make that change.

zbialecki:
Now the reviewer tab and the AC tab will be populated regardless if they have been assigned papers. In addition the AC tab will also now show counts of completed recommendations similar to the counted bids in the reviewers tab.

openreview/conference/templates/programchairWebfield.js
```diff
@@ -1309,6 +1336,21 @@ var updateReviewerContainer = function(paperNumber) {
...
+  var invitationId = invGroup + '/-/' + invName;
+  var referrerUrl = '/group' + location.search + location.hash;
+
+  // Right now this is only showing bids, elmo scores, and conflicts as the
+  // other scores invitations + labels are not available in the PC console
+  return '/edge/browse' +
+    '?start=' + invitationId + ',tail:' + profileId +
+    '&traverse=' + invitationId +
+    '&browse=' + invitationId +
+    ';' + invGroup + '/-/ELMo_Score' +
```
melisabok:
```suggestion
    ';' + invGroup + '/-/Affinity_Score' +
```

melisabok: We should parametrize this, ECCV will use Affinity_Score and TPMS_Score.

melisabok:
Running test_eccv_conference.py I can see the reviewers but not the area chairs.

melisabok:
When I try to see the bids for for Reviewer ECCV_Three, it has 3 bids, I only see two papers in the edge browser:

Screen Shot 2020-03-03 at 5 51 14 PM


Now I understand your change in the openreview repo to show the Bid in the first column.

zbialecki:
> When I try to see the bids for for Reviewer ECCV_Three, it has 3 bids, I only see two papers in the edge browser:

That means that the paper did not match the invitation query specified by the edge invitation (usually because it was withdrawn).

zbialecki:
The ACs tab should be working correctly now. It also seems like the Bid invitation was set up with the wrong `reply.content.tail.type` so I fixed that too.

zbialecki:
Note: to see the AC recommendations count and edge browser links you have to add values to 2 of the  constants at the top of the file:
```
var RECOMMENDATION_NAME = 'Recommendation';
var SCORES_NAME = 'Affinity_Score';
```

openreview/conference/templates/programchairWebfield.js
```diff
@@ -266,6 +273,49 @@ var postReviewerEmails = function(postData) {
...
+};
+
+var getAreaChairRecommendationCounts = function() {
+  if (!RECOMMENDATION_NAME) {
+    return $.Deferred().resolve({});
+  }
+
+  // Can't perform a groupBy query on signatures, so we have to get the full list
+  return Webfield.getAll('/edges', {
+    invitation: REVIEWERS_ID + '/-/' + RECOMMENDATION_NAME,
```
melisabok:
```suggestion
    invitation: AREA_CHAIRS_ID + '/-/' + RECOMMENDATION_NAME,
```

openreview/conference/templates/programchairWebfield.js
```diff
@@ -266,6 +273,49 @@ var postReviewerEmails = function(postData) {
   });
 };
 
+var getReviewerBidCounts = function() {
+  if (!BID_NAME) {
+    return $.Deferred().resolve({});
+  }
+
+  return Webfield.getAll('/edges', {
+    invitation: REVIEWERS_ID + '/-/' + BID_NAME,
```
melisabok: area chairs have a different invitation to get the edges, it starts with the AREA_CHAIRS_ID

zbialecki: Ah ok, right now I am not showing the area chair bids, but I could get those here as well

PR #584, opened 2020-03-03
Fix/eccv feedback

melisabok:
Some feedback that we got from the last demo:

- Parametrize the name of the official comment
- Fix rating missing field in the AC console
- Add referrer parameter to all the consoles
- Add description to the recommendation invitation


openreview/conference/builder.py
```diff
@@ -1016,8 +1016,9 @@ def get_signatures(self, conference, number):
 
 class CommentStage(object):
 
-    def __init__(self, start_date = None, allow_public_comments = False, anonymous = False, unsubmitted_reviewers = False, reader_selection = False, email_pcs = False, authors=False):
-        self.name = 'Comment'
+    def __init__(self, official_name = None, start_date = None, allow_public_comments = False, anonymous = False, unsubmitted_reviewers = False, reader_selection = False, email_pcs = False, authors=False):
```
mohituniyal: Can we call the parameter `official_comment_name` instead?

openreview/conference/builder.py
```diff
@@ -1233,8 +1234,8 @@ def set_bid_stage(self, start_date = None, due_date = None, request_count = 50,
     def set_review_stage(self, start_date = None, due_date = None, name = None, allow_de_anonymization = False, public = False, release_to_authors = False, release_to_reviewers = False, email_pcs = False, additional_fields = {}, remove_fields = []):
         self.review_stage = ReviewStage(start_date, due_date, name, allow_de_anonymization, public, release_to_authors, release_to_reviewers, email_pcs, additional_fields, remove_fields)
 
-    def set_comment_stage(self, start_date = None, allow_public_comments = False, anonymous = False, unsubmitted_reviewers = False, reader_selection = False, email_pcs = False, authors=False ):
-        self.comment_stage = CommentStage(start_date, allow_public_comments, anonymous, unsubmitted_reviewers, reader_selection, email_pcs, authors)
+    def set_comment_stage(self, name=None, start_date = None, allow_public_comments = False, anonymous = False, unsubmitted_reviewers = False, reader_selection = False, email_pcs = False, authors=False ):
```
mohituniyal: Needs spaces around the assignment operator.

openreview/conference/invitation.py
```diff
@@ -830,7 +830,7 @@ def __init__(self, conference, note):
             writers = [conference.id],
             signatures = [conference.id],
             invitees = invitees,
-            noninvitees = [conference.get_authors_id(number = note.number)],
+            noninvitees = [conference.get_authors_id(number = note.number), conference.id + '/Paper'+ str(note.number) + '/Secondary_Area_Chair'],
```
mohituniyal: Just checking, should this be:  `Secondary_Area_Chairs` ?

openreview/conference/builder.py
melisabok: it is already a comment stage, why do you need to add the work comment?

openreview/conference/invitation.py
melisabok: it is only one seconday area chair, why this should be plural?

zbialecki: Well, it is kind of a convention that group ids are plural even if there's only one member

melisabok: This group is similar to `AnonReviewer1` or `Area_Chair1`. I don't see a reason to use plural name. 

BTW this group is member of `Area_Chairs` group. 

I think it should be called `Area_Chair2` following the convention of the reviewers names.

openreview/conference/builder.py
mohituniyal: Official_name sounds like the name of some official involved with the paper. Maybe invitation_name would be better?

melisabok: I changed the value to `official_comment_name`

PR #585, opened 2020-03-04
custom load edge head set to group id for reviewer/ac

mohituniyal:


openreview/conference/matching.py
```diff
@@ -110,6 +110,16 @@ def _create_edge_invitation(self, edge_id):
...
         }
 
+        edge_head_type = 'Note'
+        edge_head_query = {
+            'invitation' : self.conference.get_blind_submission_id()
+        }
+        if 'Custom_Load' in edge_id:
+            edge_head_type = 'Group'
+            edge_head_query = {
+                'group' : edge_id.split('/-/')[0]
```
zbialecki: The query object should be `{'id': <reviewer's group id>}`

mohituniyal: Thanks @zbialecki . Updated.

PR #586, opened 2020-03-05
Submission email - update action in email body

mohituniyal:
I have updated the action in the email's body. 
The subject however seems invalid for a delete scenario. 

@melisabok @zbialecki any suggestions on language for the email subject?


melisabok:
Please, update the process function through the invitation editor and NOT the builder.

mohituniyal:
> Please, update the process function through the invitation editor and NOT the builder.

Yep, the live site has been updated just now.

PR #587, opened 2020-03-05
Fix/home page load

melisabok:
When a conference is not public we shouldn't show any tab with paper information.


melisabok:
Good catch. I should change this message. What should I say? the Author Console should be available soon?

zbialecki:
I would say, "The author console page for managing your submissions will be available soon."

PR #588, opened 2020-03-08
Fix/pc console sort bids

zbialecki:
Improve PC console: 
- Add option to sort by completed bids to AC tab
- Add links to edge browser to config tab

Also add `repliedEdges` param to all webfields that use the taskList component. Related to https://github.com/openreview/openreview/pull/1953


zbialecki:
This PR was going to include messaging reviewers and ACs from those tabs in the PC console, but it looks like that is going to be much more difficult that I thought. We need to discuss that feature more...

PR #590, opened 2020-03-09
Fix invitations for paper withdraw and paper desk rejection

mohituniyal:
Use separate var for mail recipients and submission readers.
Add conference id as readers of the submission.

### DIFF
--- a/openreview/conference/templates/desk_reject_process.py
+++ b/openreview/conference/templates/desk_reject_process.py
@@ -16,7 +16,6 @@ def process(client, note, invitation):
     if PAPER_AREA_CHAIRS_ID:
         committee.append(PAPER_AREA_CHAIRS_ID)
     committee.append(PROGRAM_CHAIRS_ID)
-    committee.append(CONFERENCE_ID)
 
     mail_recipients = [PAPER_AUTHORS_ID]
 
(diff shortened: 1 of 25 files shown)


openreview/conference/templates/desk_reject_process.py
```diff
@@ -16,6 +16,9 @@ def process(client, note, invitation):
     if PAPER_AREA_CHAIRS_ID:
         committee.append(PAPER_AREA_CHAIRS_ID)
     committee.append(PROGRAM_CHAIRS_ID)
+    committee.append(CONFERENCE_ID)
```
melisabok: if program chairs are already in the committee you don't need to add the conference id. 

I think the ECCV invitation doesn't have this version.

openreview/conference/templates/withdraw_process.py
```diff
@@ -16,6 +16,9 @@ def process(client, note, invitation):
     if PAPER_AREA_CHAIRS_ID:
         committee.append(PAPER_AREA_CHAIRS_ID)
     committee.append(PROGRAM_CHAIRS_ID)
+    committee.append(CONFERENCE_ID)
```
melisabok: same here.

ECCV version only has authors in the committee

openreview/conference/templates/desk_reject_process.py
mohituniyal: Right, the ECCV version is missing the reviewers and ACs. When this PR is merged, I can update the Desk_Reject and Withdraw invitations with this version.

melisabok: I don't think we need to merge this PR, we need to use the version of master where AC/Reviewers are readers of the withdrawn/desk rejected notes and they also get email notifications. 

We already hace assigned ACs and we are going to have assigned Reviewers. They should get email notifications.

mohituniyal: Agreed. I will close this PR since this change is not needed anymore.

PR #597, opened 2020-03-11
Fix/filter non-author tasks

### LINKED ISSUE #591: Author console: show tasks for authors only
It seems we are showing all the invitations when a user is an author and Ac/Reviewer:

This is Rene Vidal author console:

Screen Shot 2020-03-09 at 3 51 46 PM

zbialecki:
Standardize author, reviewer, and AC consoles to filter invitations in the same way.

Fixes #591


zbialecki:
Yes, so was the area chair console, but it wasn't filtering tag invitation for some reason.

melisabok:
I know this request is not  related to this issue but could you remove the 'Schedule' tab from the AC/Reviewers/Author console?

We are not really using them and it can be confusing for the users. 

May be in the future we can automatically create the schedule from the existent invitations.

zbialecki:
Do you want to remove it entirely or just comment it out for now?

melisabok:
comment it out for now.

PR #606, opened 2020-03-17
Fix/set assignments

melisabok:


openreview/conference/matching.py
```diff
@@ -562,18 +562,17 @@ def deploy(self, assignment_title):
...
+        assignment_edges = list(openreview.tools.iterget_edges(
             client,
             invitation=self.conference.get_paper_assignment_id(self.match_group.id),
-            label=assignment_title)
+            label=assignment_title))
 
         paper_by_forum = {n.forum: n for n in submissions}
 
-        for edge in assignment_edges:
+        for edge in tqdm(assignment_edges, total=len(assignment_edges)):
```
carlosmondra: Why are you iterating over the `assignment_edges` twice?

melisabok: what do you mean? converting the iterator to a list and then iterate the list?

carlosmondra: Yeah, as far as I know converting it to a list iterates over the iterator, so why do it?

melisabok: I think I'm doing it for tqdm, I need to pass the total number the edges to show the progress bar, do you know another way of doing it?

carlosmondra: I was reading that doing `list(<iterator>)` is quite fast as it uses C++ under the hood, so maybe it's not impacting performance too much. Another way to do it would be to use `get_edges_count` and use the result for tqdm, if you really want the progress bar.

melisabok: Ok, I changed it getting the count first.

PR #611, opened 2020-03-23
Fix/avoid multiple bids for the same paper

xkopenreview:
copy of https://github.com/openreview/openreview-py/pull/605

moving page refresh logic to bidwebfield.js

This PR needs to be merged together with https://github.com/openreview/openreview/pull/1971


melisabok:
I'm seeing the error is being displayed and then the page is loaded, I think this is fine.

PR #616, opened 2020-03-31
Fix/review readers

melisabok:
- Use Enum class to set the different reader options for the review stage.
- Default readers of a review are: Program Chairs and signature of the review. 'Reviewers submitted' is not the default option anymore.
- Do not show the review content in the email notification because they may no have `title` or `review` anymore, like ECCV. This is also good, because we can support Markdown and no need to worry about the email content rendering.
- Request form support more options in the Review Stage invitation: https://github.com/openreview/openreview-scripts/pull/1175
- Fix tests.


melisabok:
@mohituniyal should be ready to review.

openreview/conference/builder.py
```diff
@@ -997,10 +1016,14 @@ def get_readers(self, conference, number):
         if conference.use_area_chairs:
             readers.append(conference.get_area_chairs_id(number = number))
 
-        if self.release_to_reviewers:
+        if self.release_to_reviewers is ReviewStage.Readers.REVIEWERS:
+            readers.append(conference.get_reviewers_id())
+        if self.release_to_reviewers is ReviewStage.Readers.REVIEWERS_ASSIGNED:
```
mohituniyal, line 1021: Can we use elif instead of if for the 3 subsequent ifs ? Since, self.release_to_reviewers can only be 1 value out of these 4, it makes sense to not check all conditions as soon as 1 is met.

openreview/conference/invitation.py
```diff
@@ -730,30 +731,39 @@ def __init__(self, conference, note):
...
+                'values-regex': signature_regex,
+                'description': 'How your identity will be displayed.'
+            }
+        }
+
+        has_copies = [r for r in readers if r.startswith('{') and r.endswith('}')]
+        if has_copies:
+            reply['readers'] = {
+                'description': 'Select all user groups that should be able to read this comment.',
+                'values-copied': readers
```
mohituniyal, line 758: Should this be `'values-copied': has_copies` instead ? 

Also, I would like to rename has_copies to something else, maybe copy_values. `has_copies` seems to imply a boolean.

openreview/conference/templates/programchairWebfield.js
```diff
@@ -617,7 +617,7 @@ var calcReviewersComplete = function(reviewerMap, officialReviews) {
 
 var calcPaperReviewsComplete = function(noteMap, officialReviewMap) {
   return _.reduce(noteMap, function(numComplete, reviewerMap, n) {
-    var allSubmitted = officialReviewMap[n] && Object.values(reviewerMap).length === Object.values(officialReviewMap[n]).length;
+    var allSubmitted = officialReviewMap[n] && Object.values(reviewerMap).length >0 && Object.values(reviewerMap).length === Object.values(officialReviewMap[n]).length;
```
zbialecki, line 620: Could we move `Object.values(reviewerMap).length` into a separate var so it's not calling Object.values twice?

openreview/conference/invitation.py
melisabok: `has_copies` is a boolean that checks if there is at least one element in the array between curly braces.

openreview/conference/builder.py
melisabok: I refactored to avoid the if else.

openreview/conference/templates/programchairWebfield.js
melisabok: Done.

Issue #615, opened 2020-03-31
Assignment Configuration: fix field descriptions


melisabok:
<img width="926" alt="Screen Shot 2020-03-31 at 7 54 37 AM" src="https://user-images.githubusercontent.com/545506/78023612-f387c680-7324-11ea-9cf6-4caaae355024.png">

PR #618, opened 2020-04-02
config for custom papers, users

mohituniyal:
Contains changes to assignment configuration invitation.

Fixes https://github.com/openreview/openreview-py/issues/615


openreview/conference/matching.py
```diff
@@ -430,6 +430,20 @@ def _build_config_invitation(self, scores_specification):
...
                         'order': 12
                     },
+                    'custom_user_demand_invitation': {
+                        'description': 'Invitation to store custom max number of users required by papers',
+                        'default': self.match_group.id + '/-/Custom_Max_Users',
+                        'order': 13,
+                        'value-regex': self.match_group.id + '/-/.*',
+                        'required': False
+                    },
+                    'custom_max_papers_invitation': {
```
melisabok, line 440: Doesn't this invitation replace the current 'custom_load_invitation'?

mohituniyal: Yes, both serve the same purpose. Which do you prefer to keep? I would keep the new one i.e. `custom_max_papers_invitation` just because the name and description are clearer.

openreview/conference/matching.py
```diff
@@ -361,23 +361,23 @@ def _build_config_invitation(self, scores_specification):
                         'description': 'Title of the configuration.',
                         'order': 1
                     },
-                    'max_users': {
+                    'user_demand': {
                         'value-regex': '[0-9]+',
                         'required': True,
-                        'description': 'Max number of reviewers that can review a paper',
+                        'description': 'Max number of users that can review a paper',
```
melisabok, line 367: Can we remove `max` from the description?

openreview/conference/matching.py
```diff
@@ -415,28 +415,38 @@ def _build_config_invitation(self, scores_specification):
...
-                    },
                     'assignment_invitation': {
                         'value': self.conference.get_paper_assignment_id(self.match_group.id),
                         'required': True,
                         'description': 'Invitation to store paper user assignments',
-                        'order': 12
+                        'order': 11
+                    },
+                    'custom_user_demand_invitation': {
+                        'description': 'Invitation to store custom max number of users required by papers',
```
melisabok, line 428: Can we remove max from this description too?

openreview/conference/matching.py
melisabok: Yes, I agree

openreview/conference/matching.py
mohituniyal: Done

openreview/conference/matching.py
mohituniyal: Done.

openreview/conference/matching.py
```diff
@@ -586,6 +598,6 @@ def deploy(self, assignment_title, is_area_chair, overwrite):
             if edge.head in paper_by_forum:
                 paper_number = paper_by_forum.get(edge.head).number
                 user = edge.tail
-                new_assigned_group = self.conference.set_assignment(user, paper_number, self.is_area_chair)
+                _ = self.conference.set_assignment(user, paper_number, self.is_area_chair)
```
melisabok, line 601: if `new_assigned_groups` is not being used then remove the assignment.

PR #619, opened 2020-04-03
Fix/reviewer console refactor

### LINKED ISSUE #625: PC Console: don't show AC stats when the venue doesn't have ACs
Same for bids and recommendations. If the values are empty then we shouldn't show any data.

### LINKED ISSUE #626: PC Console: reviewer progress should be counted based on real assignments and not all the reviewers
There are only 3 reviewers with assignments and not 20. 

Screen Shot 2020-04-12 at 5 29 44 PM

zbialecki:
Re-write review console webfield to bring it up to our current webfield standards.

Fixes #625 
Fixes #626


melisabok:
I'm sorry Zach, there are a few conflicts here because of my last PR merged :(

zbialecki:
No worries! I'll fix them

openreview/conference/templates/authorWebfield.js
```diff
@@ -1,14 +1,13 @@
...
-//
-// This webfield displays the conference header (#header), the submit button (#invitation),
-// and a tabbed interface for viewing various types of notes.
 // ------------------------------------
 
 // Constants
 var CONFERENCE_ID = '';
 var SUBMISSION_ID = '';
 var BLIND_SUBMISSION_ID = '';
+var OFFICIAL_REVIEW_NAME = '';
```
melisabok, line 9: can you pass these two variables in webfield.py?

openreview/conference/templates/authorWebfield.js
```diff
@@ -301,10 +285,30 @@ var buildTableRow = function(note, reviewerIds, completedReviews, metaReview) {
...
 
-  // Temporally hide review status
-  //return [cell0, cell1, cell2, cell3];
-  return [cell0, cell1];
-};
+  return [cell0, cell1, cell2, cell3];
+}
+
+function renderReviewSummary(data) {
+  if (!OFFICIAL_REVIEW_NAME) {
```
melisabok, line 292: Does it mean we need to set this variable when the assignments are deployed?

I would show this message if there are no reviews to show.

openreview/conference/templates/authorWebfield.js
```diff
@@ -151,113 +162,91 @@ function renderContent(authorNotes, invitations, edgeInvitations) {
...
+      reviewFormatted.rating = rating;
     }
-  }
+    var confidenceMatch = review.content.confidence && review.content.confidence.match(ratingExp);
+    var confidence = confidenceMatch ? parseInt(confidenceMatch[1], 10) : null;
+    if (confidence) {
+      confidences.push(confidence);
+      reviewFormatted.confidence = confidence;
+    }
+    reviewsFormatted.push(reviewFormatted);
```
melisabok, line 248: Can we pass the `referrer` parameter to the 'Read Review' link?

openreview/conference/templates/authorWebfield.js
```diff
@@ -151,113 +162,91 @@ function renderContent(authorNotes, invitations, edgeInvitations) {
...
+  $('#your-submissions .console-table th').eq(0).css('width', '4%');
+  $('#your-submissions .console-table th').eq(1).css('width', '36%');
+  $('#your-submissions .console-table th').eq(2).css('width', '30%');
+  $('#your-submissions .console-table th').eq(3).css('width', '30%');
+}
 
+function buildTableRow(note, completedReviews, metaReview) {
   // Paper number cell
-  var cell0 = { number: note.number};
+  var cell0 = { number: note.number };
```
melisabok, line 218: Can we pass the `referrer` parameter here too?

openreview/conference/templates/authorWebfield.js
melisabok: I think the if is not necessary here, showing 0 reviews is good:

Screen Shot 2020-04-08 at 4 23 46 PM

openreview/conference/templates/authorWebfield.js
```diff
@@ -19,128 +18,140 @@ var paperDisplayOptions = {
...
     });
 
-    invitationsP = Webfield.get('/invitations', {
-      regex: CONFERENCE_ID + '.*',
-      invitee: true,
-      duedate: true,
-      replyto: true,
-      type: 'notes',
-      details:'replytoNote,repliedNotes'
+    if (blindNoteIds.length) {
```
melisabok, line 56: I realized this code is necessary because the only way to know the submissions the user is an author is looking at the original notes but we should render the blinded note.

zbialecki: Yes, I was thinking there must be a better way, but the only other way I could think of was getting the note numbers from the Authors groups and using those numbers to query get the blinded notes (with original in details).

openreview/conference/templates/authorWebfield.js
zbialecki: OK, I'll remove this `if`

melisabok:
Reviewer Console looks good! I made a few comments with Author Console but it is working fine. Thanks for the refactor!

zbialecki:
@melisabok I think all the issues you identified are fixed. It should be ready to review again. To test paper ranking on the reviewer console you can use the dev server:

https://dev.openreview.net/group?id=thecvf.com/ECCV/2020/Conference/Reviewers
Username: reviewer@eccv2020.eu

Due to ECCV's unique review notes, I had to change 2 lines in the webfield code using the group editor to get the reviews to display properly. Line 251 and 252 were changed to:

```js
    reviewStatus.paperRating = officialReview.content.preliminary_rating;
    reviewStatus.review = officialReview.content.summary_of_contributions;
```

melisabok:
Thanks! I will take a look.

melisabok:
Looks good!, below some feedback:

- Author Console: can we hide the meta review column if the venue doesn't have area chairs? I think if META_REVIEW_NAME is empty then we shouldn't show it.
- PC Console:

When there are no ACs, the recruitment section is still visible:
Screen Shot 2020-04-14 at 8 56 00 PM

When Bids and Recommendations are empty, there are two lines rendered:
Screen Shot 2020-04-14 at 8 56 05 PM

- Reviewer Console:

Paper Ranking tags should be shown even when the invitation is expired, in read-only mode.

melisabok:
I've removed the webfield for the paper ranking invitation. Given the invitation has a duedate is being shown in the Tasks view, we have two options here:

1. Don't render invitations as Tasks when they don't have webfields. 
2. Invitations with no webfields will be rendered as Tasks but the link should go to the prefix group webfield. For example: thecvf.com/ECCV/2020/Conference/Area_Chairs/-/Paper_Ranking -> thecvf.com/ECCV/2020/Conference/Area_Chairs

zbialecki:
> 2\. Invitations with no webfields will be rendered as Tasks but the link should go to the prefix group webfield.

I kind of like this option actually. Are there any cases currently where this isn't true? Any invitations that don't have webfields and shouldn't be displayed?

melisabok:
> Are there any cases currently where this isn't true? Any invitations that don't have webfields and shouldn't be displayed?

I think in all the cases the invitation prefix is either the conference group or the reviewer/ac group.

The only issue here is the Tasks tab of the consoles will go the console!

zbialecki:
OK, I think all the issues are fixed and it should be ready for review again. The Reveiwer console and AC console on dev are set up to show read only tags.

openreview/conference/templates/reviewerWebfield.js
```diff
@@ -32,10 +32,20 @@ var main = function() {
 
       displayStatusTable(blindedNotes, officialReviews);
 
+      // Add paper ranking tag widgets to the table of submissions
+      // If tag invitation does not exist, get existing tags and display read-only
       var paperRankingInvitation = _.find(tagInvitations, ['id', PAPER_RANKING_ID]);
       if (paperRankingInvitation) {
         var paperRankingTags = paperRankingInvitation.details.repliedTags || [];
```
melisabok, line 39: I forgot about the details parameter to get the tags.

openreview/conference/templates/areachairWebfield.js
```diff
@@ -660,8 +662,8 @@ var postRenderTable = function(rows) {
       paperRanking ? [paperRanking] : [],
       {
         forum: noteId,
-        placeholder: (paperRankingInvitation && paperRankingInvitation.reply.content.tag.description) || (paperRankingInvitation && prettyId(paperRankingInvitation.id)),
-        label: paperRankingInvitation && view.prettyInvitationId(paperRankingInvitation.id),
+        placeholder: (paperRankingInvitation && paperRankingInvitation.reply.content.tag.description) || view.prettyInvitationId(invitationId),
```
melisabok, line 665: thanks for this change, tags look much better when the invitation is expired.

PR #621, opened 2020-04-06
Renaming options

melisabok:


melisabok:
This commit fixes the MIDL issue: 54675a8. We are not looking the reviewer by all emails and all names. I also found a bug in the `findProfile` function, @zbialecki  please review.

melisabok:
@xkopenreview @zbialecki I need your help making the dropdown wider:

Screen Shot 2020-04-06 at 5 28 24 PM

Screen Shot 2020-04-06 at 5 29 00 PM

This last change fixes three issues:

- Reviewers added with an email to the Reviewers group are shown with the Profile if exists.
- The list of options show the name + all the emails to disambiguate reviewers.
- User can search by name or any content of the email.

openreview/conference/templates/programchairWebfield.js
```diff
@@ -455,14 +455,16 @@ var findProfile = function(profiles, id) {
     return profiles[id];
   }
   var profile = _.find(profiles, function(p) {
-    return _.includes(p.allNames, id) || _.includes(p.allEmails, id);
+    return _.includes(_.map(p.allNames, 'username'), id) || _.includes(p.allEmails, id);
```
zbialecki, line 466: Rather than doing a map, then includes I think it would be better to just do a `find`
```suggestion
    return _.find(p.allNames, ['username', id]) || _.includes(p.allEmails, id);
```

openreview/conference/templates/programchairWebfield.js
```diff
@@ -1593,7 +1595,8 @@ var displayReviewerStatusTable = function() {
       var reviewerNum = 0;
       var reviewers = reviewerByNote[number];
       for (var revNumber in reviewers) {
-        if (reviewer == reviewers[revNumber].id) {
+        var profile = reviewers[revNumber];
+        if (_.includes(_.map(profile.allNames, 'username'), reviewer) || _.includes(profile.allEmails, reviewer)) {
```
zbialecki, line 1607: Same here

openreview/conference/templates/programchairWebfield.js
```diff
@@ -157,6 +157,14 @@ var main = function() {
       }
     }
 
+    conferenceStatusData.reviewerOptions = _.map(conferenceStatusData.reviewers, function(r) {
+      var profile = findProfile(conferenceStatusData.profiles, r);
+      return {
+        id: r,
+        description: view.prettyId(profile.name) + ' - (' + profile.allEmails + ')'
```
zbialecki, line 164: Won't this list all the emails a user has on their profile? I think we should just show `profile.email` which will be equal to the preferred email, or the first email in the email list.

I also don't think the dash in between is necessary - the () separate it enough and it's on a new line anyway

openreview/conference/templates/programchairWebfield.js
melisabok: Why don’t we save the username only in allNames, is there a reason to have the whole name object?

openreview/conference/templates/programchairWebfield.js
melisabok: I prefer to show all the emails(confirmed) so users can search reviewers by any email. The issue that Mohit this morning the reviewer was assigned using an alternate email instead of preferred.

zbialecki:
The width of the assign reviewers dropdown is defined here: https://github.com/openreview/openreview/blob/5d70b662dc239539ed1244fd6ba174a6f48572d5/public/less/legacy.less#L902

I would suggest replacing 180px with 240px on line 904 and 909.

openreview/conference/templates/programchairWebfield.js
zbialecki: You're right, that would be the better solution. I looked and that field is only ever used for the usernames.

openreview/conference/templates/programchairWebfield.js
zbialecki: OK, well in that case I think it would be better to do `profile.allEmails.join(', ')` so that the text will wrap.

melisabok:
This is way better, what do you think?

Screen Shot 2020-04-07 at 1 14 55 PM

zbialecki:
I like it!

melisabok:
This is ready to merge, it depends on https://github.com/openreview/openreview/pull/2033

PR #623, opened 2020-04-09
Remove [] from values-regex

carlosmondra:
This PR fixes one values-regex in an invitation so that openreview-py can be compatible with openreview/openreview/pull/2027

### DIFF
--- a/openreview/conference/invitation.py
+++ b/openreview/conference/invitation.py
@@ -901,7 +901,7 @@ class DecisionInvitation(openreview.Invitation):
             multiReply = False,
             reply = {
                 'writers': {
-                    'values-regex': [conference.get_program_chairs_id()],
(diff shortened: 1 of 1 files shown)


openreview/conference/invitation.py
```diff
@@ -901,7 +901,7 @@ def __init__(self, conference):
             multiReply = False,
             reply = {
                 'writers': {
-                    'values-regex': [conference.get_program_chairs_id()],
+                    'values-regex': conference.get_program_chairs_id(),
```
melisabok, line 904: good catch!

Given the string is not regex and it is valid group id, let's change the field type to  `values`, like signatures.

carlosmondra: Done!

Issue #612, opened 2020-03-24
Remove fuzzywuzzy dependency


melisabok:
We are not using it anymore.

zbialecki:
Also apparently it is GPLv2 licensed which would force us re-license the whole project.

PR #627, opened 2020-04-14
Fix/remove fuzzy

mohituniyal:
fixes https://github.com/openreview/openreview-py/issues/612


openreview/tools.py
```diff
@@ -166,68 +161,6 @@ def create_profile(client, email, first, last, middle = None, allow_duplicates =
...
-            return [clean_name(n) for n in [first, last, middle]]
-        else:
-            return []
-
-    if 'authorids' in note.content and 'authors' in note.content:
-        author_names = [a.replace('*', '') for a in note.content['authors']]
-        author_emails = [e for e in note.content['authorids']]
-        if len(author_names) == len(author_emails):
-
-            email_by_authorname = match_authors_to_emails(author_names, author_emails)
```
melisabok, line 201: can you remove this call and keep this function?

I think we should keep it. This function create profiles for current submissions.

tests/test_tools.py
```diff
@@ -158,56 +158,6 @@ def test_parallel_exec(self, client):
...
-        shuffled_emails = random.sample(
-            [p[1] for p in correct_pairs], len(correct_pairs))
-
-        result = openreview.tools.match_authors_to_emails(
-            shuffled_names, shuffled_emails)
-
-        for name, email in correct_pairs:
-            assert result[name] == email
-
-    def test_create_authorid_profiles(self, client):
```
melisabok, line 181: keep this test please

tests/test_tools.py
```diff
@@ -34,29 +34,19 @@ def test_add_members_to_group(self, client):
         posted_group = client.add_members_to_group(posted_group, ['test_subject_y1@mail.com', 'test_subject_y2@mail.com'])
         assert posted_group
         assert len(posted_group.members) == 3
-        assert 'test_subject_x@mail.com' in posted_group.members
-        assert 'test_subject_y1@mail.com' in posted_group.members
-        assert 'test_subject_y2@mail.com' in posted_group.members
+        assert all(list(map(lambda email: email in posted_group.members, ['test_subject_x@mail.com', 'test_subject_y1@mail.com', 'test_subject_y2@mail.com'])))
```
melisabok, line 37: is there a change in this PR making these tests fail? why are you changing them?

tests/test_tools.py
```diff
@@ -228,10 +196,9 @@ def test_replace_members_with_ids(self, client, test_client):
             signatories=['~Super_User1'],
             members=['test@mail.com', '~Test_User1', '~Another_Name1']
         ))
-
         assert posted_group
 
-        client.post_profile(openreview.Profile(
+        _ = client.post_profile(openreview.Profile(
```
melisabok, line 201: what is this doing?

tests/test_tools.py
melisabok: I prefer to keep the tests as simple as possible. Asserting each member individually is not a problem, it is clear for the users what the test is doing. 

If the test fails it is easier to know the reason. 

Also this is PR doesn't change any of this behavior, this change shouldn't be here.

tests/test_tools.py
mohituniyal: Test has been restored already.

openreview/tools.py
mohituniyal: `get_names` has been restored.

tests/test_tools.py
mohituniyal: Reverted to original asserts.

tests/test_tools.py
mohituniyal: I was trying to follow the changes I was making in other places in this file. Anyway, I have reverted this too now.

PR #632, opened 2020-04-16
Fix/create groups submission

melisabok:
Create Paper, Authors and Reviewers group inside of the process function. 

This is needed for ACL workshop, they want to assign reviewers right after the paper was submitted.


melisabok:
I'm sorry, this PR is still a DRAFT.

melisabok:
Now it should be ready to be reviewed.

openreview/conference/invitation.py
```diff
@@ -41,6 +41,21 @@ def __init__(self, conference):
...
             file_content = f.read()
             file_content = file_content.replace("var SHORT_PHRASE = '';", "var SHORT_PHRASE = '" + conference.get_short_name() + "';")
+            file_content = file_content.replace("var CONFERENCE_ID = '';", "var CONFERENCE_ID = '" + conference.get_id() + "';")
+            if submission_stage.email_pcs:
+                file_content = file_content.replace("var PROGRAM_CHAIRS_ID = '';", "var PROGRAM_CHAIRS_ID = '" + conference.get_program_chairs_id() + "';")
+            if submission_stage.create_groups:
+                file_content = file_content.replace("var CREATE_GROUPS = false;", "var CREATE_GROUPS = true;")
+                # Only supported for public reviews
+                if submission_stage.create_review_invitation:
+                    print('REPLACE OFFICIAL REVIEW')
```
mohituniyal, line 51: What is the purpose of these print statements?

tests/test_tools.py
```diff
@@ -10,7 +10,7 @@ class TestTools():
 
     def test_get_submission_invitations(self, client):
         invitations = openreview.tools.get_submission_invitations(client)
-        assert len(invitations) == 21, "Invitations could not be retrieved"
+        assert invitations, "Invitations could not be retrieved"
```
mohituniyal, line 13: Superb, thanks for this change @melisabok

openreview/conference/helpers.py
```diff
@@ -246,11 +249,13 @@ def get_decision_stage(client, request_forum):
             due_date = decision_due_date,
             public = request_forum.content.get('make_decisions_public', '').startswith('Yes'),
             release_to_authors = request_forum.content.get('release_decisions_to_authors', '').startswith('Yes'),
-            release_to_reviewers = request_forum.content.get('release_decisions_to_reviewers', '').startswith('Yes'))
+            release_to_reviewers = request_forum.content.get('release_decisions_to_reviewers', '').startswith('Yes'),
+            email_authors = True)
```
celestemartinez, line 253: One question: not sure where we use get_decision_stage, but why do we set email_authors=True here?

melisabok: get_decision_stage is being executed in the process function of 'Decision Stage' of the request form. 

I put True because I think in most of the cases where the venue is managed by the PCs they want to notify the authors. But we could add a checkbox in the Decision Stage form, what do you think?

celestemartinez: Oh I see. I think that makes sense, you are right most PCs do assume that when decisions are posted authors are emailed about it. Adding the checkbox wouldn't hurt.

melisabok: Ok, I implemented the option in the ui of the decision stage, please review this PR too: https://github.com/openreview/openreview-scripts/pull/1181

celestemartinez: Looks good!

PR #634, opened 2020-04-17
Make openreview-py compatible with Refresh and Access Tokens

carlosmondra:
Merge after openreview/openreview/pull/1835 is merged.

It looks like A LOT of changes, but it's simply adding an error handler when the token expires. This handler needs to be included in the `__handle_response` method. A new parameter was included to the method to allow a retry when a request is made and the token has expired.


carlosmondra:
Use this #660 instead

Issue #546, opened 2020-01-21
Documentation: write examples how to get all the data to run a matching 


melisabok:
ICML will need to use this documentation to run the matching by themselves. 

I think openreview-py it is the right place to do it.

mohituniyal:
This was resolved with #636

PR #636, opened 2020-04-22
Documentation for matching input data download

mohituniyal:


melisabok:
I'm not sure why we need this documentation. If the scores are being stored in edges why don't we run the matching using a config note and using the matching server?

if the reason is that Program Chairs want to run the matching in their local environments(?), we could support a main.py script that receives a config a id and run the matching from command line. 

I do think we should add documentation about what an Edge is and what represents in the paper matching scope.

mohituniyal:
> I'm not sure why we need this documentation. If the scores are being stored in edges why don't we run the matching using a config note and using the matching server?
> 
> if the reason is that Program Chairs want to run the matching in their local environments(?), we could support a main.py script that receives a config a id and run the matching from command line.
> 
> I do think we should add documentation about what an Edge is and what represents in the paper matching scope.

@melisabok Sure, I can do that. I actually created this documentation for the issue: https://github.com/openreview/openreview-py/issues/546 thinking that some other conference may need it in the future. Do you think the issue description needs to be modified?

melisabok:
Thanks for pointing out the issue, I don't remember why I created it. You are right, it is not clear. 

The thing this is issue is showing a command line to the matcher where openreview-py doesn't know everything about it. 

I think what is helpful is to write in openreview-py documentation how to get and download edges. 
And write documentation in the matching how to write the matcher from the command line. The latter is already done, right?

mohituniyal:
> Thanks for pointing out the issue, I don't remember why I created it. You are right, it is not clear.
> 
> The thing this is issue is showing a command line to the matcher where openreview-py doesn't know everything about it.
> 
> I think what is helpful is to write in openreview-py documentation how to get and download edges.
> And write documentation in the matching how to write the matcher from the command line. The latter is already done, right?

Right, matcher already has documentation on this. I will update this documentation to explain edges and their role in matching.

docs/source/getting_data.rst
```diff
@@ -240,3 +240,69 @@ You can also call this script with the `openreview` package::
...
+
+
+Since most of the common tasks performed using `Edges` require `Edges` to be grouped, it's also possible to query for already grouped `Edges`. Consider the following example that gets all reviewers grouped by papers they have conflicts with for the ICLR 2020 Conference ::
+
+
+	>>> grouped_conflict_edges = client.get_grouped_edges(
+		invitation='ICLR.cc/2020/Conference/Reviewers/-/Conflict',
+		groupby='head',
+		select='tail,weight,label'
+		)
```
zbialecki, line 297: I'm not sure what the convention is for python, either the `)` on the line with `select` or a new line, but I think the examples should be consistent. Personally I prefer it on the select line.

mohituniyal: It seems PEP8 is goes along with what you suggested.
Check the "Similar questions" section at the bottom of this link: https://blog.finxter.com/pep-8-hanging-indentation-and-closing-brackets-in-python/

I can make the other example consistent with this.

melisabok:
What should we do with this PR? Is it ready to merge?

PR #637, opened 2020-04-23
Fix/paper ranking readers

melisabok:
- Post ranking tags as a string with the format: '2 of 2' or '1 of 3', depending of the number of assigned papers for each AC/Reviewer.
- Sign the tags anonymously, using the Area_Chair1 or AnonReviewer group. 
- Allow ACs to read reviewer paper rankings.
- Show reviewer rankings in the AC Console. (Should we do the same in the PC Console?)
- Parametrize the review rating field name.

Depends on https://github.com/openreview/openreview/pull/2057

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -147,7 +147,14 @@ class Conference(object):
     def __create_review_stage(self):
 
         notes = list(self.get_submissions())
-        return self.invitation_builder.set_review_invitation(self, notes)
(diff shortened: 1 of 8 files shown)


melisabok:
It seems parseInt still works. It parses the first number of the string, for example:

```
parseInt('2 of 2') 
```

returns 2

zbialecki:
Cool, I didn't know that.

PR #638, opened 2020-04-24
Release author names of accepted papers only.

melisabok:


tests/test_double_blind_conference.py
```diff
@@ -1673,3 +1673,37 @@ def test_release_decisions(self, client, helpers):
...
+        assert len(invitation_panel.find_elements_by_tag_name('div')) == 0
+        notes_panel = selenium.find_element_by_id('notes')
+        assert notes_panel
+        tabs = notes_panel.find_element_by_class_name('tabs-container')
+        assert tabs
+        with pytest.raises(NoSuchElementException):
+            notes_panel.find_element_by_class_name('spinner-container')
+        assert tabs.find_element_by_id('accept-oral')
+        assert tabs.find_element_by_id('accept-poster')
+        assert tabs.find_element_by_id('reject')
```
mohituniyal, line 1709: Can we add an assert on the bibtex of any of the accepted submissions? I ask because that is one of the major changes we are doing.

melisabok:
The backend will ignore the `details` field but you are right we should avoid this field being sent in the POST call. I think the solution here is to change `to_json` implementation and don't serialize the details field. Let me try that.

melisabok:
Ok, I changed the serialization to not send the details property in the POST function.

PR #639, opened 2020-04-24
Fix/replace member invalid profiles

mohituniyal:
Currently tools.replace_members_with_ids would fail if there is a non-existent profile in the group that is being passed in the arguments. 

[Edit] Instead of calling `client.get_profile` I am now calling `tools.get_profile` which instead of throwing an exception, returns None in case the id does not have a corresponding profile.

NOTE: This will result in elimination of profile id group members when a profile corresponding to the id does not actually exist anymore.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -478,6 +478,7 @@ def replace_members_with_ids(client, group):
     """
     ids = []
     emails = []
+    invalid_ids = []
     for member in group.members:
         if '~' not in member:
             try:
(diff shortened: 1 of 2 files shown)


openreview/tools.py
```diff
@@ -489,9 +490,16 @@ def replace_members_with_ids(client, group):
                 else:
                     raise e
         else:
-            profile = client.get_profile(member)
-            ids.append(profile.id)
+            try:
```
melisabok, line 493: there is a function in tools that does exactly the same. It returns `None` if profile can not be found.

mohituniyal: Agreed @melisabok , this change has been made.

mohituniyal:
This should be ready now.

PR #642, opened 2020-04-28
Replace Object.values with lodash equivalent

zbialecki:
Apparently Object.values was not included in the current polyfill.


xkopenreview:
tried adding bundled corejs to vendor (didn't remove existing polyfill) and seems that the mathjax error is gone.

zbialecki:
If this PR is merged https://github.com/openreview/openreview/pull/2062 then these changes won't really be necessary.

melisabok:
I think we can close this issue, right?

zbialecki:
Yes, agreed.

PR #645, opened 2020-05-01
Fix/custom max users

mohituniyal:
With this change the builder will now create Custom_Max_Papers invitation instead of Custom_Load invitation.

### DIFF
--- a/openreview/conference/matching.py
+++ b/openreview/conference/matching.py
@@ -511,7 +511,7 @@ class Matching(object):
 
         self._create_edge_invitation(self.conference.get_paper_assignment_id(self.match_group.id))
         self._create_edge_invitation(self._get_edge_invitation_id('Aggregate_Score'))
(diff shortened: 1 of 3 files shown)


openreview/conference/matching.py
```diff
@@ -511,7 +511,7 @@ def setup(self, affinity_score_file=None, tpms_score_file=None, elmo_score_file=
 
         self._create_edge_invitation(self.conference.get_paper_assignment_id(self.match_group.id))
         self._create_edge_invitation(self._get_edge_invitation_id('Aggregate_Score'))
-        self._create_edge_invitation(self._get_edge_invitation_id('Custom_Load'))
+        self._create_edge_invitation(self._get_edge_invitation_id('Custom_Max_Papers'))
```
melisabok, line 514: could you create the invitation for custom paper demands too?

PR #646, opened 2020-05-01
Render the reviewer option with name and all the emails for disambigu…

melisabok:
…ation

### DIFF
--- a/openreview/conference/templates/areachairWebfield.js
+++ b/openreview/conference/templates/areachairWebfield.js
@@ -23,7 +23,7 @@ var PAPER_RANKING_ID = CONFERENCE_ID + '/' + AREA_CHAIR_NAME + '/-/Paper_Ranking
 var REVIEWER_PAPER_RANKING_ID = REVIEWER_GROUP + '/-/Paper_Ranking';
 
 var reviewerSummaryMap = {};
-var allReviewers = [];
+var reviewerOptions = [];
 var paperAndReviewersWithConflict = {};
 var paperRankingInvitation = null;
 var showRankings = false;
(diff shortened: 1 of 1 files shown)


openreview/conference/templates/areachairWebfield.js
```diff
@@ -258,12 +259,21 @@ var getReviewerGroups = function(noteNumbers) {
 };
 
 var formatData = function(blindedNotes, officialReviews, metaReviews, noteToReviewerIds, invitations, edgeInvitations, tagInvitations, allReviewers, acRankingByPaper, reviewerRankingByPaper) {
-  var uniqueIds = _.uniq(_.reduce(noteToReviewerIds, function(result, idsObj, noteNum) {
+  var uniqueIds = _.uniq(_.concat(_.reduce(noteToReviewerIds, function(result, idsObj, noteNum) {
```
zbialecki, line 262: I don't understand why you can't just do `var uniqueIds = allReviewers` Doesn't allReviewers contain all the members of the reviewers group?

openreview/conference/templates/areachairWebfield.js
```diff
@@ -258,12 +259,21 @@ var getReviewerGroups = function(noteNumbers) {
...
     return result.concat(_.values(idsObj));
-  }, []));
+  }, []), allReviewers));
 
   return getUserProfiles(uniqueIds)
   .then(function(profiles) {
+
+    reviewerOptions = _.map(allReviewers, function(r) {
```
zbialecki, line 269: Also you only iterate over allReviewers...

openreview/conference/templates/areachairWebfield.js
melisabok: yes, you are right. I thought about that, I didn't want to break anything.

melisabok: I prefer to keep it as it is because it may happen there is a review done by a reviewer that could have been removed from the group. 

The dropdown should only show the members of the reviewer group.

PR #653, opened 2020-05-04
Fix/PC console bid completion count

### LINKED ISSUE #652: PC Console: read taskCompletionCount from the invitation
Read the value to compute the bid progress from the bid invitation. Now, it is hardcoded to 40 and 60 for ACs.

zbialecki:
Load Bid and Recommendation invitations and to get correct task completion count.

Fixes #652


zbialecki:
Thanks, I missed that we were already getting invitations. I realized there was one other thing I needed to confirm before this was merged: currently the PC console is treating `taskCompletionCount` for AC Recommendations as the number required **per paper**. Is this correct?

melisabok:
yes, that's correct.

zbialecki:
OK, I think this is ready to merge then.

PR #654, opened 2020-05-07
Editable fields in assignment invitation

mohituniyal:
[Updated]
1. Added a test to check if default values for fields `custom_max_papers_invitation` and `custom_user_demand_invitation` are visible in the form.
2. Made the fields `aggregate_score_invitation` and `conflicts_invitation` editable.
3. Minor update to tools.py to ensure that we see a print for invalid ids only when they exist in the group passed to `replace_members_with_ids`

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -714,13 +714,13 @@ class Conference(object):
         authors_group = self.__create_group(self.get_authors_id(), self.id, public=True)
         return self.webfield_builder.set_author_page(self, authors_group)
 
-    def setup_matching(self, is_area_chair=False, affinity_score_file=None, tpms_score_file=None, elmo_score_file=None, build_conflicts=False):
+    def setup_matching(self, is_area_chair=False, affinity_score_file=None, tpms_score_file=None, elmo_score_file=None, build_conflicts=False, delete_existing_conflicts=True):
         if is_area_chair:
             match_group = self.client.get_group(self.get_area_chairs_id())
         else:
             match_group = self.client.get_group(self.get_reviewers_id())
         conference_matching = matching.Matching(self, match_group)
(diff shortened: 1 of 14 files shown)


openreview/conference/builder.py
```diff
@@ -714,13 +714,13 @@ def set_authors(self):
         authors_group = self.__create_group(self.get_authors_id(), self.id, public=True)
         return self.webfield_builder.set_author_page(self, authors_group)
 
-    def setup_matching(self, is_area_chair=False, affinity_score_file=None, tpms_score_file=None, elmo_score_file=None, build_conflicts=False):
+    def setup_matching(self, is_area_chair=False, affinity_score_file=None, tpms_score_file=None, elmo_score_file=None, build_conflicts=False, delete_existing_conflicts=True):
```
melisabok, line 717: why do you want to delete the conflicts and don't create them again?

build_conflicts = True does that. If you pass build_conflicts = False the conflicts shouldn't be deleted/created.

openreview/conference/matching.py
```diff
@@ -407,13 +408,15 @@ def _build_config_invitation(self, scores_specification):
                         'default': scores_specification
                     },
                     'aggregate_score_invitation': {
-                        'value': self._get_edge_invitation_id('Aggregate_Score'),
+                        'value-regex': '.*',
```
melisabok, line 411: shouldn't the regex starts al least with the conference id?

openreview/conference/builder.py
mohituniyal: Agreed, change has been reverted.

openreview/conference/matching.py
mohituniyal: Agreed, code updated accordingly.

PR #658, opened 2020-05-07
Only create review invitation when the review stage is set.

melisabok:


mohituniyal:
@celestemartinez You can merge this one whenever you are done reviewing.

PR #660, opened 2020-05-07
Add option to set expiration of token

carlosmondra:
This PR is similar to #634 in that it changes the cookie name. Refreshing the token is not possible but setting the time of the expiration of the token is now possible. This PR will fail until openreview/openreview/pull/1835 is merged


zbialecki:
I think this could be merged now, right?

carlosmondra:
Yes, this can be merged now

tests/test_client.py
```diff
@@ -77,7 +84,21 @@ def test_login_user(self):
...
+        client = openreview.Client(username = "openreview.net", password = "1234", tokenExpiresIn=3)
+        group = client.get_group("openreview.net")
+        assert group
+        assert group.members == ['~Super_User1']
+        time.sleep(4)
+        try:
+            group = client.get_group("openreview.net")
+        except openreview.OpenReviewException as e:
+            error = e.args[0]
+            assert e.args[0]['name'] == 'TokenExpiredError'
```
zbialecki, line 101: Not sure why, but this test is no longer passing. Seems like the API is returning a different error: `{'type': 'forbidden', 'path': 'id', 'value': 'openreview.net', 'user': 'guest_1620851435577'}`

carlosmondra: This is happening because I commented this code in the backend, because `TokenExpiredError` was not handled by UI to retrieve a new token. I'll comment this test too

zbialecki:
Can we merge this PR?

carlosmondra:
Yes, I think it's fine now, we just need to uncomment the test, once the UI has the refresh token implemented :D

melisabok:
Should we implement this in the new API 2 client? https://github.com/openreview/openreview-py/blob/master/openreview/api/client.py#L113

carlosmondra:
Yes, I think we should. I can make that change, it does not seem difficult to do

melisabok (approved):

PR #662, opened 2020-05-10
Check review completed by signature

melisabok:
Fix Reviewer Progress calculation, we should check if the user signed the official review with their anonReviewer group number. 

I'm still see a difference between this number and the number of reviewer to send a reminder for ECCV. 

Currently I see the progress: 698 / 2808 and the reminder tries to remind 2066 reviewers, that means the other 742 completed the reviews, the difference is 44 where 14 are reviewers with no assignments what is correct to not send the reminder, what about the other 30 reviewers?


melisabok:
The issue was the reviewerMap was built using all the current assignment without checking if the note was still active. Now the numbers are the same.

PR #669, opened 2020-05-14
Compute demand based on how many reviews per paper

melisabok:


openreview/conference/templates/programchairWebfield.js
```diff
@@ -139,7 +139,7 @@ var main = function() {
         invitationMap[REVIEWERS_ID + '/-/' + BID_NAME]
       ),
       reviewsCount: officialReviews.length,
-      assignedReviewsCount: calcAssignedReviewsCount(reviewerGroupMaps.byReviewers),
+      assignedReviewsCount: PAPER_REVIEWS_COMPLETE_THRESHOLD ? (blindedNotes.length * PAPER_REVIEWS_COMPLETE_THRESHOLD) : calcAssignedReviewsCount(reviewerGroupMaps.byReviewers),
```
zbialecki, line 142: I would prefer to put this calculation in the `calcAssignedReviews` function, there maybe some other variables that change the calculation in the future and I'd prefer to have all the logic in 1 place.

melisabok:
You're right, I'm going to close this PR and take more time to think this.

PR #672, opened 2020-05-20
disable dropdown when posting the tag, use only valid tags

melisabok:


openreview/conference/templates/areachairWebfield.js
```diff
@@ -705,6 +705,7 @@ var postRenderTable = function(rows) {
             ddate: deleted ? Date.now() : null
           };
           body = view.getCopiedValues(body, paperRankingInvitation.reply);
+          $('.tag-widget').find('button').attr('disabled', true);
```
zbialecki, line 708: It's more efficient to avoid using `find` when possible, so I would change this to `$('.tag-widget button')...`

openreview/conference/templates/reviewerWebfield.js
```diff
@@ -322,14 +327,15 @@ var displayPaperRanking = function(notes, paperRankingInvitation, paperRankingTa
             invitation: invitationId,
             ddate: deleted ? Date.now() : null
           };
+          $('.tag-widget').find('button').attr('disabled', true);
```
zbialecki, line 330: same here

openreview/conference/templates/reviewerWebfield.js
```diff
@@ -283,10 +283,15 @@ var displayPaperRanking = function(notes, paperRankingInvitation, paperRankingTa
   var invitation = paperRankingInvitation ? _.cloneDeep(paperRankingInvitation) : null;
   if (invitation) {
     var availableOptions = ['No Ranking'];
+    var validTags = [];
     notes.forEach(function(note, index) {
       availableOptions.push((index + 1) + ' of ' + notes.length );
+      noteTags = paperRankingTags.filter(function(tag) { return tag.forum == note.forum; })
+      if (noteTags.length) {
+        validTags.push(noteTags[0]);
+      }
```
zbialecki, line 292:
```suggestion
      var noteTag = paperRankingTags.find(function(tag) { return tag.forum === note.forum; })
      if (noteTag) {
        validTags.push(noteTag);
      }
```

melisabok: I'm keeping the filter here because I would need to check which is the most recent tag to use and ignore the others, if there are duplicate tags for the same paper.

zbialecki: I see, makes sense. Changing the == to === is the only change then.

openreview/conference/templates/areachairWebfield.js
melisabok: done

openreview/conference/templates/reviewerWebfield.js
melisabok: done

openreview/conference/templates/reviewerWebfield.js
melisabok: done

PR #674, opened 2020-05-21
read tmdate/tcdate

melisabok:


openreview/openreview.py
```diff
@@ -1389,6 +1391,8 @@ def from_json(Group,g):
         group = Group(g['id'],
             cdate = g.get('cdate'),
             ddate = g.get('ddate'),
+            tcdate = g.get('tcdate'),
```
mohituniyal, line 1394: I can see that you have added these Group attributes to the `init` and `from_json` methods, but for some reason when I do a `client.get_group` I am still not seeing these attributes in the returned Group object. 
I am sure that the endpoint returns these attributes because for the same group id I can see them on the browser by querying `/groups`
Are you observing the same @melisabok ?

openreview/openreview.py
```diff
@@ -1325,11 +1325,13 @@ class Group(object):
     :param details:
     :type details: optional
     """
-    def __init__(self, id, readers, writers, signatories, signatures, cdate = None, ddate = None, members = None, nonreaders = None, web = None, details = None):
+    def __init__(self, id, readers, writers, signatories, signatures, cdate = None, ddate = None, tcdate=None, tmdate=None, members = None, nonreaders = None, web = None, details = None):
```
mohituniyal, line 1328: @melisabok Can you also add these Group attributes to the docstring.

openreview/openreview.py
melisabok: tmdate date is retrieved correctly, I'm creating an issue on the openreview repo about the tcdate is None. 

Screen Shot 2020-06-09 at 9 58 44 PM

openreview/openreview.py
melisabok: done

Issue #680, opened 2020-06-05
Readthedocs is not displaying the documentation


carlosmondra:
For some time now the documentation in readthedocs is not showing up:
https://openreview-py.readthedocs.io/en/latest/api.html

mohituniyal:
Closed via #683

PR #683, opened 2020-06-09
Fix - docs

mohituniyal:
This fixes the issue with docs: https://github.com/openreview/openreview-py/issues/680.

Here's a RTD site connected with this branch: https://testorpybranch.readthedocs.io/en/latest/


openreview/agora/agora.py
```diff
@@ -1,5 +1,4 @@
-from .. import openreview
-
+import openreview
```
melisabok, line 1: does it mean that openreview has to be installed with pip and this import is linked to the local file in the repo folder?

mohituniyal: This means that instead of importing the local module openreview (coming from openreview.py), I am now importing the root level package openreview. 
This package would be actively linked to the local files if you installed the package in develop mode.

mohituniyal: I have reverted this change @melisabok

PR #698, opened 2020-06-19
ICLR 2021 proposed change to recruitment landing page

mohituniyal:
Updated based on Melisa's comments:
1. Showing user's email if available.
2. Showing `if any` for the instruction asking user to complete their tasks.

Screen Shot 2020-06-19 at 6 06 31 PM

### DIFF
--- a/openreview/conference/templates/recruitResponseWebfield.js
+++ b/openreview/conference/templates/recruitResponseWebfield.js
@@ -33,7 +33,7 @@ function render() {
           '<div class="row">',
             '<h4>Please complete the following steps now:</h4>',
               '<ol>',
-                '<li><p>Log in to your OpenReview account (if you do not already have an OpenReview account, you can sign up <a style="font-weight:bold;" href="/signup">here</a>).</p></li>',
+                '<li><p>Log in to your OpenReview account. If you do not already have an account, you can sign up <a style="font-weight:bold;" href="/signup">here</a>).</p></li>',
                 '<li><p>Ensure that the email address ' + email + ' that received this invitation is linked to your <a style="font-weight:bold;" href="/profile?mode=edit">profile page</a> and has been confirmed.</p></li>',
                 '<li><p>Complete your pending <a style="font-weight:bold;" href="/tasks">tasks</a> (if any) for ' + HEADER.subtitle + '.</p></li>',
               '</ol>',
--- a/openreview/conference/templates/recruitResponseWebfield.js
+++ b/openreview/conference/templates/recruitResponseWebfield.js
@@ -26,6 +26,7 @@ function render() {
       // Display response text
       var message = 'Thank you for accepting this invitation from ' + HEADER.title;
       $response.append('<div class="panel"><div class="row"><h3 style="line-height:normal;">' + message + '</h3></div></div>');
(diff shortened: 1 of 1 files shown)


melisabok:
This is nice, I like. Can we display the email address that we used to invite the user, I noticed that it is not always the same one when the user gets the email for some kind of redirection.

mohituniyal:
> This is nice, I like. Can we display the email address that we used to invite the user, I noticed that it is not always the same one when the user gets the email for some kind of redirection.

@melisabok Good idea. Sometimes we send out recruitment links with user set to tilde ids. Showing that as the email would be confusing. What do you think?

melisabok:
In that case, we don't show it.

mohituniyal:
> In that case, we don't show it.

Got it. I have made the change. Please take a look.

openreview/conference/templates/recruitResponseWebfield.js
```diff
@@ -26,12 +26,17 @@ function render() {
...
 
       $response.append([
         '<div class="panel">',
           '<div class="row">',
-            '<p>If you do not already have an OpenReview account, please sign up <a style="font-weight:bold;" href="/signup">here</a>.</p>',
-            '<p>If you have an existing OpenReview account, please ensure that the email address that received this invitation is linked to your <a style="font-weight:bold;" href="/profile?mode=edit">profile page</a> and has been confirmed.</p>',
+            '<h4>Please complete the following steps now:</h4>',
+              '<ol>',
+                '<li><p>Log in to your OpenReview account (if you do not already have an OpenReview account, you can sign up <a style="font-weight:bold;" href="/signup">here</a>).</p></li>',
+                '<li><p>Ensure that the email address ' + email + 'that received this invitation is linked to your <a style="font-weight:bold;" href="/profile?mode=edit">profile page</a> and has been confirmed.</p></li>',
```
zbialecki, line 37: I think that you're missing a space here after email

openreview/conference/templates/recruitResponseWebfield.js
```diff
@@ -26,12 +26,17 @@ function render() {
...
+      var email = args.user.indexOf('@') > -1 ? '(<strong>' + args.user + '</strong>) ' : '';
 
       $response.append([
         '<div class="panel">',
           '<div class="row">',
-            '<p>If you do not already have an OpenReview account, please sign up <a style="font-weight:bold;" href="/signup">here</a>.</p>',
-            '<p>If you have an existing OpenReview account, please ensure that the email address that received this invitation is linked to your <a style="font-weight:bold;" href="/profile?mode=edit">profile page</a> and has been confirmed.</p>',
+            '<h4>Please complete the following steps now:</h4>',
+              '<ol>',
+                '<li><p>Log in to your OpenReview account (if you do not already have an OpenReview account, you can sign up <a style="font-weight:bold;" href="/signup">here</a>).</p></li>',
```
zbialecki, line 36: I would make this 2 sentences. "Log in to your OpenReview account. If you do not already have an account, sign up for OpenReview here."

mohituniyal: Done. I have split it to "Log in to your OpenReview account. If you do not already have an account, you can sign up here."

Issue #696, opened 2020-06-17
Request form: reviews visibility


celestemartinez:
If PCs choose reviews to be private in the request form initially, the review stage form should not have the option to make reviews public.

PR #700, opened 2020-06-22
Venue request form refactor

mohituniyal:
(Updated)
- Ported venue request form logic to `openreview-py` package.
- Added submission revision stage as `Submission Revision Stage`.

Fixes https://github.com/openreview/openreview-py/issues/696
Fixes https://github.com/openreview/openreview-py/issues/705


openreview/venue_request/process/deployProcess.py
```diff
@@ -47,8 +47,8 @@ def process(client, note, invitation):
     readers = [conference.get_program_chairs_id(), SUPPORT_GROUP]
 
     client.post_invitation(openreview.Invitation(
-        id = SUPPORT_GROUP + '/-/Request' + str(forum.number) + '/Revision',
-        super = SUPPORT_GROUP + '/-/Revision',
+        id = SUPPORT_GROUP + '/-/Request' + str(forum.number) + '/Venue_Update',
```
melisabok, line 50: Why do you change the name?

mohituniyal: That is to avoid confusions. We now have another Revision in `Revision Stage` also. I mentioned this in the PR description as well. 
Let me know if you disagree with the change.

melisabok: What if we call the other invitation `Submission Revision Stage`? It more clear for the user that is a stage to enable revision on submissions. @celestemartinez what do you think?

mohituniyal: I actually named it exactly that earlier. But then I felt that it seems too long in a button. Anyway, if everyone agrees on using this, I can make the changes for that.
Also, the changes would mean that there will be a button called `Revision` and then another called `Submission Revision Stage`. While the latter is really clear in its objective, the former is still confusing as to exactly what it affects. And for that reason, I would like to retain the change of `Revision` to `Venue Update`. Your thoughts @melisabok and @celestemartinez ?

melisabok: I think you should use Submissions Revision Stage because there will be Review Revision Stage, etc. Unless we want all the configuration in a single invitation. Another option is to call it Submission Stage. 

I think keeping the current invitation as Revision is fine.

celestemartinez: I agree that keeping the venue update invitation as Revision should be fine, but maybe we can mention this button somewhere in the comment to PCs after the venue is deployed:
> If you need to make a change to the information provided in your request form, please feel free to revise it directly 
using the 'Revision' button.

or something like that

mohituniyal: Thanks, I have made these changes.

melisabok:
the changes look good to me but I'm not fully convinced that the support group should be listed everywhere, what if we run the different stages using the PC group?

mohituniyal:
> the changes look good to me but I'm not fully convinced that the support group should be listed everywhere, what if we run the different stages using the PC group?

As discussed, this would require updating the request forum to make it accessible to the PC group.

Issue #695, opened 2020-06-17
Builder: Author group empty


celestemartinez:
For a single-blind conference, paper author groups were created correctly, and main author group was created as well, but it was empty.

mohituniyal:
@celestemartinez Did you see this happen when you run `conference.create_paper_groups` ? I tried doing that but could not reproduce the issue.

mohituniyal:
I think I see what happened. The `create_paper_groups` method is only called in method `create_blind_submissions`. So, there might be an issue with Single blind conferences. Looking into how to fix this.

Issue #659, opened 2020-05-07
Builder: add venue to active_venues after submission deadline


melisabok:
We are currently doing this manually, we should add the venue to the group when we create the blind submissions or paper groups.

Issue #650, opened 2020-05-03
Builder: create withdraw and desk reject invitations for all the venues


melisabok:
I think the place to put these calls is when we call `create_blind_submissions`, thoughts?

mohituniyal:
Yes, right after the line https://github.com/openreview/openreview-scripts/blob/ed13c606f5c626f1fa1505a2c2ded5e9e244286a/venues/OpenReview.net/Support/revisionProcess.py#L19 would be a good place. This would mean that the change goes in openreview-scripts repo.
@melisabok is this what you have in mind or do you want the `create_blind_submissions` method to call `create_withdraw_invitations` and `create_desk_reject_invitations`?

melisabok:
I think we should do it inside of `create_blind_submissions`, enable this feature by default for all the venues. 

I'm not sure if the withdraw and desk reject invitation support venues with different configurations, like `single blind` , `public` or `private`.

mohituniyal:
@melisabok The withdraw and desk reject invitations do support double, single blind and also support both public and private submission type venues. Methods to enable each of these invitations take these 2 inputs `reveal_authors`, `reveal_submission` and that should allow it to work for all types of venue configurations.

Re: calling these 2 invitation setup functions inside create_blind_submissions, that would only enable these for double blind venues, right? If we want these 2 enabled for all venues, we may have to call them somewhere else.

melisabok:
You are right, but we have a dependency of  groups. They need to be created before we open these invitation. When the venue is single blind we should call to create paper groups instead of create_blind_submissions. 

May be we should have a everything in a single function, close_submissions? that calls either create blind submissions or create paper groups and then create the desk reject and withdraw invitations.

Close submissions should be called from all the process function.

mohituniyal:
> You are right, but we have a dependency of groups. They need to be created before we open these invitation. When the venue is single blind we should call to create paper groups instead of create_blind_submissions.
> 
> May be we should have a everything in a single function, close_submissions? that calls either create blind submissions or create paper groups and then create the desk reject and withdraw invitations.
> 
> Close submissions should be called from all the process function.

@melisabok `close_submissions` bears the primary responsibility of closing the submission invitation. 
Now, if I place the logic for creating blind submissions and paper groups inside `close_submissions`, I still can not call `close_submissions` method from the `stage` button processes on the request forum. This is because if a PC submits one of `Stage` forms, that would then close the submissions regardless of whether it should be closed at that point, and this can be problematic.

[...]

PR #701, opened 2020-06-23
Create paper groups + Add conference to active_venues

mohituniyal:
[Updated]
Note that now if you call `set_submission_stage` you need to make sure to pass the correct values for the params for withdraw and desk-reject to ensure that they are not set to their corresponding defaults.
This is because the desk-reject and withdraw params have been made a part of submission stage now.

Fixes https://github.com/openreview/openreview-py/issues/695, https://github.com/openreview/openreview-py/issues/659, https://github.com/openreview/openreview-py/issues/650

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -642,13 +642,13 @@ class Conference(object):
         return blinded_notes
 
     def setup_venue_papers(self):
-        if self.submission_stage.due_date < datetime.datetime.now():
+        if not self.submission_stage.due_date or self.submission_stage.due_date < datetime.datetime.now():
             # Due date is in the past
             if self.submission_stage.double_blind:
                 # Double Blind venue
                 self.create_blind_submissions()
             else:
-                # Single Blind OR Open venue
+                # Single Blind or Open venue
                 self.create_paper_groups(authors=True, reviewers=True, area_chairs=True)
             
             reveal_authors = False
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -641,7 +641,7 @@ class Conference(object):
 
         return blinded_notes
 
-    def setup_venue_papers(self):
+    def setup_post_submission_stage(self):
         if not self.submission_stage.due_date or self.submission_stage.due_date < datetime.datetime.now():
             # Due date is in the past
             if self.submission_stage.double_blind:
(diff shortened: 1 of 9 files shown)


mohituniyal:
I am going to make some more changes based on the comments on https://github.com/openreview/openreview-py/issues/650. Converting it to draft till then.

openreview/conference/invitation.py
```diff
@@ -1501,7 +1501,7 @@ def set_paper_ranking_invitation(self, conference, group_id, start_date, due_dat
                     "tag": {
                         "description": "Select value",
                         "order": 1,
-                        "value-regex": 'No Ranking|[0-9]+\sof\s[0-9]+',
+                        "value-regex": r'No Ranking|[0-9]+\sof\s[0-9]+',
```
melisabok, line 1504: Why are you changing this? does it return an string?

openreview/venue_request/process/revisionProcess.py
```diff
@@ -14,8 +14,7 @@ def process(client, note, invitation):
 
     invitation_type = invitation.id.split('/')[-1]
     if invitation_type in ['Bid_Stage', 'Review_Stage', 'Meta_Review_Stage', 'Decision_Stage']:
-        if conference.submission_stage.double_blind and conference.submission_stage.due_date < datetime.datetime.now():
-            conference.create_blind_submissions()
+        conference.setup_venue_papers()
         conference.set_authors()
```
melisabok, line 18: is it possible to move these three lines to `setup_venue_papers`?

```
        conference.set_authors()
        conference.set_reviewers()
            conference.set_area_chairs()
```

It seems to be the same. We could rename the name of the function like `post_submission_stage`?

openreview/conference/builder.py
```diff
@@ -636,6 +641,22 @@ def create_blind_submissions(self, force=False, hide_fields=[]):
...
+                # Double Blind venue
+                self.create_blind_submissions()
+            else:
+                # Single Blind or Open venue
+                self.create_paper_groups(authors=True, reviewers=True, area_chairs=True)
+            
+            reveal_authors = False
+            if not self.submission_stage.double_blind:
+                reveal_authors = True
+            self.create_withdraw_invitations(reveal_authors=reveal_authors, reveal_submission=False)
```
melisabok, line 657: I think there are a few tests that are calling this directly, it would be nice to remove the calls and see if the tests are still passing.

openreview/conference/invitation.py
mohituniyal: That is just a way to indicate to pylint that there are backslash chars in the string.

openreview/venue_request/process/revisionProcess.py
mohituniyal: Done. Method renamed to `setup_post_submission_stage`.

melisabok:
The changes look good to me, thanks @mohituniyal . It would be nice to have a way to call `setup_post_submission_stage` only once and not after each stage is called.

PR #711, opened 2020-07-08
Fix/messages

melisabok:


openreview/openreview.py
```diff
@@ -1133,7 +1133,7 @@ def delete_group(self, group_id):
         return response.json()
 
 
-    def post_message(self, subject, recipients, message, ignoreRecipients=None, sender=None):
+    def post_message(self, subject, recipients, message, ignoreRecipients=None, sender=None, replyTo=None):
```
mohituniyal, line 1136: Can we add this field `replyTo` in the docstring.

openreview/openreview.py
```diff
@@ -1152,7 +1152,8 @@ def post_message(self, subject, recipients, message, ignoreRecipients=None, send
             'subject': subject ,
             'message': message,
             'ignoreGroups': ignoreRecipients,
-            'from': sender
+            'from': sender,
+            'replyTo': replyTo
```
mohituniyal, line 1156: Doc-string needed here as well.

openreview/openreview.py
melisabok: done

openreview/openreview.py
melisabok: done

PR #712, opened 2020-07-08
Remove legacy classes row and panel that conflict with bootstrap CSS

zbialecki:
This is important for https://github.com/openreview/openreview-web/pull/63, however it could be deployed before the migration is complete as it does not contain any breaking changes.

### DIFF
--- a/openreview/conference/templates/programchairWebfield.js
+++ b/openreview/conference/templates/programchairWebfield.js
@@ -786,7 +786,7 @@ var displayStatsAndConfiguration = function(conferenceStats, conferenceConfig) {
   };
 
   // Conference statistics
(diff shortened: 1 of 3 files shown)


melisabok:
Should we migrate current webfields when we switch to openreview-web?

zbialecki:
Well, the existing recruitment webfields will not be viewed again, so we don't need to migrate those, but it would be nice to migrate some of the active PC console webfields. The only downside to not migrating them is that the table won't quite fill the whole width of the page, it will have some extra padding on the sides. So I'd say it's not totally necessary.

melisabok:
It looks good to me, I prefer to merge this before sending invitations to ICLR reviewers. waiting fro @mohituniyal review.

PR #722, opened 2020-07-28
Fix the pdf size for submission invitation

mohituniyal:
This fixes the pdf file size in submission invitation reply template. The default size right now is incorrect.


melisabok:
@carlosmondra do you want to take a look?

PR #723, opened 2020-07-30
Check is user is in invited group

### LINKED ISSUE #628: Reviewer recruitment: check the user is in the invited group
In the process function of the recruitment invitation we check the key matches with user invited, we should also check the user is in the invited group.

mohituniyal:
Fixes #628

### DIFF
--- a/openreview/conference/templates/recruitReviewersProcess.js
+++ b/openreview/conference/templates/recruitReviewersProcess.js
@@ -63,13 +63,10 @@ function() {
       } else {
         done('Invalid response', note.content.response);
       }
-
       return true;
     } else {
       done('Invalid key or user not in invited group', note.content.key, user);
       return false;
     }
-  })
-  .then(result => done())
-  .catch(error => done(error));
+  });
 }
--- a/openreview/conference/templates/recruitReviewersProcess.js
+++ b/openreview/conference/templates/recruitReviewersProcess.js
@@ -4,6 +4,7 @@ function() {
   var SHORT_PHRASE = '';
   var CONFERENCE_NAME = '';
   var REVIEWER_NAME = '';
+  var REVIEWERS_INVITED_ID = '';
   var REVIEWERS_ACCEPTED_ID = '';
   var REVIEWERS_DECLINED_ID = '';
   var HASH_SEED = '';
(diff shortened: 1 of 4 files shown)


openreview/conference/templates/recruitReviewersProcess.js
```diff
@@ -12,56 +13,61 @@ function() {
...
 
-  if (hashKey === note.content.key) {
-    if (note.content.response === 'Yes') {
-      console.log('Invitation replied Yes');
-      //if a user is in the declined group, remove them from that group and add them to the reviewers group
-      var text = 'Thank you for accepting the invitation to be a ' + REVIEWER_NAME + ' for ' + SHORT_PHRASE + '.\n';
-      text += 'The ' + SHORT_PHRASE + ' program chairs will be contacting you with more information regarding next steps soon. ';
-      text += 'In the meantime, please add noreply@openreview.net to your email contacts to ensure that you receive all communications.\n\n';
-      text += 'If you would like to change your decision, please click the Decline link in the previous invitation email.';
+  var invitedGroupP = or3client.or3request(or3client.grpUrl + '?id=' + REVIEWERS_INVITED_ID, {}, 'GET', token);
```
melisabok, line 16: can you add the parameter `member` to this call and check the result is empty or not?

melisabok:
it looks good to me. the only issue that I see here is that the user will think that the invitation link worked because the invitation UI doesn't check this and there is no way to do it in the webfield because of reader permissions. Guest users can not access to the invited group.

openreview/conference/templates/recruitReviewersProcess.js
mohituniyal: As discussed, we are not making this suggested change right now because of the issue https://github.com/openreview/openreview/issues/2156. I can make the change once this gets resolved.

mohituniyal:
> it looks good to me. the only issue that I see here is that the user will think that the invitation link worked because the invitation UI doesn't check this and there is no way to do it in the webfield because of reader permissions. Guest users can not access to the invited group.

Agreed that the user will not realize whether s/he was able to successfully accept or reject the invitation.

In case the recipient of the recruitment mail switches the email address/tilde from `x` to `y` in the recruitment link OR the email/tilde in the link is not a member of the invited group, then maybe in the process function we can send an email to the address`y` (since that is what we receive in the process function) the correct status based on these checks.

@melisabok What do you think?

melisabok:
I think we should show in the UI the error but currently there is no way to do it. I'm going to merge this. it is good to have the checking in the process function.

PR #727, opened 2020-07-31
Default header update for expertise invitation

mohituniyal:
I wanted to add a link directly to the edit profile page but the end point for that is different in openreview and openreview-web.
Currently, I can use `/profile?mode=edit` but on openreview-web it is `/profile/edit`. Not wanting to add inconsistent links I skipped on adding this link to the instructions on expertise page.

### DIFF
--- a/openreview/conference/webfield.py
+++ b/openreview/conference/webfield.py
@@ -113,7 +113,7 @@ class WebfieldBuilder(object):
                         <b>By default, we consider all of these papers to formulate your expertise. Please click on \"Exclude\" for papers that you do  NOT want to be used to represent your expertise.</b>
                         <br>
                         <br>
-                        Your previously authored papers from selected conferences were imported automatically from <a href=https://dblp.org>DBLP.org</a>. The keywords in these papers will be used to rank submissions for you during the bidding process, and to assign submissions to you during the review process. If there are missing DBLP papers, you can add them by editing your <a href=/profile?mode=edit>OpenReview profile</a> and then clicking on 'Add DBLP Papers to Profile'.
+                        Your previously authored papers from selected conferences were imported automatically from <a href=https://dblp.org>DBLP.org</a>. The keywords in these papers will be used to rank submissions for you during the bidding process, and to assign submissions to you during the review process. If there are DBLP papers missing, you can add them by editing your <a href=/profile?mode=edit>OpenReview profile</a> and then clicking on 'Add DBLP Papers to Profile'.
                 </p>
                 <br>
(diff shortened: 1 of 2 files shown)


openreview/conference/webfield.py
```diff
@@ -113,7 +113,7 @@ def set_expertise_selection_page(self, conference, invitation):
                         <b>By default, we consider all of these papers to formulate your expertise. Please click on \"Exclude\" for papers that you do  NOT want to be used to represent your expertise.</b>
                         <br>
                         <br>
-                        Your previously authored papers from selected conferences were imported automatically from <a href=https://dblp.org>DBLP.org</a>. The keywords in these papers will be used to rank submissions for you during the bidding process, and to assign submissions to you during the review process.
+                        Your previously authored papers from selected conferences were imported automatically from <a href=https://dblp.org>DBLP.org</a>. The keywords in these papers will be used to rank submissions for you during the bidding process, and to assign submissions to you during the review process. If you need to upload more <a href=https://dblp.org>DBLP.org</a> papers, you can do so by going to your OpenReview profile edit page and then clicking on 'Add DBLP Papers to Profile'.
```
melisabok, line 116: I would say:

"If you need to upload more " -> "If there are DBLP papers missing, you can go to your profiles"

@zbialecki @celestemartinez please advise.

melisabok:
I think we should keep having this URL('/profile?mode=edit') available because there are other places where we use it. @zbialecki can we keep backward compatibility?

openreview/conference/webfield.py
zbialecki: I would change the last part to "you can do so by editing your OpenReview profile then clicking on 'Add DBLP Papers to Profile'."

zbialecki:
I would prefer to just change those links to match the new URL, but I suppose we could also add logic to the profile page to detect that param and redirect to the correct url.

melisabok:
> we could also add logic to the profile page to detect that param and redirect to the correct url

Yes, please. We will correct the places where we have the old url after the migration.

mohituniyal:
> > we could also add logic to the profile page to detect that param and redirect to the correct url
> 
> Yes, please. We will correct the places where we have the old url after the migration.

@melisabok Sorry, I am not quite clear as to what we are going to do here. 
If we add the redirection logic on the profile page, then why do we need to update the old URLs?

melisabok:
> If we add the redirection logic on the profile page, then why do we need to update the old URLs?

to keep the code always up to date and use always the same URL. 

you should use `/profile?mode=edit` that is the one that works in the live site now.

openreview/conference/webfield.py
mohituniyal: I have included both suggestions and it now says: 

> If there are DBLP papers missing, you can add them by editing your <a href=/profile?mode=edit>OpenReview profile</a> and then clicking on 'Add DBLP Papers to Profile'.

PR #728, opened 2020-08-04
don't reset super user password

melisabok:
It depends on https://github.com/openreview/openreview/pull/2164


mohituniyal:
The change in this PR look good to me. Waiting for openreview/openreview#2164 to be merged.

melisabok:
this should be ready now

PR #730, opened 2020-08-06
Feature/moderate comments

mohituniyal:
Open Question: 
1. Do we want the comment author to be able to see the revisions posted to their comment?


melisabok:
> Do we want the comment author to be able to see the revisions posted to their comment?

Yes, they should the edit made by the PC. the comment should display a 'Revisions' link visible to the comment author and PCs.

melisabok:
is this PR ready for review?

mohituniyal:
> is this PR ready for review?

There are still some issues with this. 

1. During moderation we introduce two new fields (via a reference) namely `moderation_reason` and `justification`. During the undo action, I remove these 2 fields using the Undo reference invitation. There are 2 ways to do this undo action:
a. Show the PC the `moderation_reason` check boxes selected and `justification` text entered during moderation. The PC then has to de-select the checkboxes and clear the text before submitting the undo reference. **Issue** with this method is that it REQUIRES the PC to ensure that the check-boxes and text are cleared before submitting. This is already working in this branch.
b. Other way is to have the undo moderation invitation defined with already empty `moderation_reason` and `justification` fields. I tested this and it works fine. **Issue** with this method is that the note editor for responding to undo invitation looks like this:
Screen Shot 2020-08-18 at 11 41 21 AM

2. Another issue is that during undo the `moderation_reason` (uses values-checkbox) field returns an empty array (i.e. []) when all fields are cleared, but the `justification` (uses value-regex) field returns nothing when the text is emptied, and so the reference posted by the UI has only the `moderation_reason` field. This requires me to post another reference in the undo process function to empty the `justification` text.

3. Once a comment has been moderated, the `Public Comment` button should not be available to reply to it because it will cause an error. This is because of the readership of the moderated comment which is now limited to the PC and the comment author but the public comment has _everyone_ in its readers list.

4. Before merging this PR we now have to ensure that openreview-web has the changes for referentInvitation.

melisabok:
1.b: what is the undo process function removes the fields from the note?
2. This will be solved doing everything in the process function. 
3. This is not a problem, the moderated comment will visible to the author comment and PC and the error correctly shown in the UI.
4. I can make the change in openreview-web.

melisabok:
@mohituniyal could you check this is working? https://github.com/openreview/openreview-web/pull/180

mohituniyal:
> 1.b: what is the undo process function removes the fields from the note?
> 2. This will be solved doing everything in the process function.
> 3. This is not a problem, the moderated comment will visible to the author comment and PC and the error correctly shown in the UI.
> 4. I can make the change in openreview-web.

@melisabok I am not sure if I understood your suggestion completely. Do you mean the undo invitation would not have the 2 fields `moderation_reason` and `justification`? If so, then what field would it have, because invitation is not allowed to have empty content. Also, if you introduce a new field in this undo invitation, we will then need to worry about how to remove that new field.
The undo invitation not having these 2 fields is also problematic because then NEITHER can you post a reference to empty these fields using the undo invitation NOR post an updated comment note.
Let me know if I am missing something.

openreview/conference/invitation.py
```diff
@@ -637,12 +636,12 @@ def __init__(self, conference, note):
                 },
                 'writers': {
                     'values-copied': [
-                        conference.get_id(),
                         '{signatures}'
                     ]
                 },
                 'signatures': {
-                    'values-regex': '~.*|\\(anonymous\\)' if comment_stage.anonymous else '~.*',
+                    # 'values-regex': '~.*|\\(anonymous\\)' if comment_stage.anonymous else '~.*',
```
melisabok, line 643: can we remove this line?

PR #732, opened 2020-08-11
Fix withdraw and desk-reject options

mohituniyal:


melisabok:
it looks good to me but how are you going to migrate the data?

mohituniyal:
> it looks good to me but how are you going to migrate the data?

I found 11 request forms which have the old withdraw and desk-reject options selected. Once this PR is merged, I will update the request form invitation on the live site. With that I can simply manually update these 11 request forms by posting a `Revision` to each.

I could do this successfully locally.

PR #733, opened 2020-08-11
Fix/deploy writers

melisabok:
- fix the writers of the deploy revision.
- create a new invitation for the Support group to call the post submission stage (create blind submissions, groups, etc)

### DIFF
--- a/openreview/venue_request/venue_request.py
+++ b/openreview/venue_request/venue_request.py
@@ -417,6 +417,7 @@ class VenueRequest():
         self.setup_request_form()
         self.setup_venue_comments()
         self.setup_venue_deployment()
+        self.setup_venue_post_submission()
         self.setup_venue_recruitment()
 
         # Setup for venue stages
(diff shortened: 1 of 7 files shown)


openreview/venue_request/venue_request.py
```diff
@@ -710,6 +711,46 @@ def setup_venue_deployment(self):
...
+    def setup_venue_post_submission(self):
+
+        post_submission_content = {
+            'force': {
+                'value-radio': ['Yes', 'No'],
+                'required': True,
+                'description': 'Force to create the blind submission if conference if double-blind'
+            },
+            'hide_fields': {
+                'values-regex': '.*',
```
mohituniyal, line 723: This needs to be `value-regex`, otherwise the value for `hide_fields` is saved as a list instead of a string. This causes failure down the line when the process function of `post_submission` invitation calls `split` over the value of this field (and `split` can not be called on a `list`).

openreview/venue_request/venue_request.py
```diff
@@ -710,6 +711,46 @@ def setup_venue_deployment(self):
                 }
             ))
 
+    def setup_venue_post_submission(self):
+
+        post_submission_content = {
+            'force': {
+                'value-radio': ['Yes', 'No'],
+                'required': True,
+                'description': 'Force to create the blind submission if conference if double-blind'
```
mohituniyal, line 720: What do you think about: `Force creating blind submissions if conference is double blind`

openreview/venue_request/venue_request.py
melisabok: Thanks for caching this. I removed the split call.

openreview/venue_request/venue_request.py
melisabok: Fixed.

melisabok:
> If we create the invitation in deploy process, I think that will solve this issue.

Good point, I made the change

PR #734, opened 2020-08-17
Fix/request form matching options

mohituniyal:


openreview/venue_request/webfield/supportRequestsWeb.js
```diff
@@ -40,12 +40,13 @@ var instructionDetails = [
   as well as a web interface for browsing, viewing statistics, and modifying the resulting assignments. \
   The OpenReview Matching System supports multiple optimization objectives and arbitrary numerical inputs as features for the match. \
   If you would like to use the OpenReview Paper Matching System for your venue, please indicate which features you would like to use in the form below. \
+  If you do not wish to use the OpenReview Paper Matching System, you can indicate that by deselecting all the options for the field "Paper Matching" in the request form. \
```
melisabok, line 43: I think we should say and 'want to make the manual assignments..'

melisabok:
@celestemartinez could you take a look to this?

openreview/venue_request/process/revisionProcess.py
```diff
@@ -17,9 +17,9 @@ def process(client, note, invitation):
         conference.setup_post_submission_stage()
 
     if invitation_type == 'Bid_Stage':
-        conference.setup_matching(build_conflicts=True)
+        conference.setup_matching(build_conflicts=('Conflict Detection' in forum_note.content.get('Paper Matching', [])))
```
melisabok, line 20: what happens with the current venues that don't have `Conflict Detection` option?

Also I think the manual assignment in the PC console is checking the conflicts, right? so may be the conflict computation should be always true.

mohituniyal: Ok, I have removed the option for Conflict Detection from the request form and we are now always building conflicts.

openreview/venue_request/webfield/supportRequestsWeb.js
```diff
@@ -40,12 +40,12 @@ var instructionDetails = [
...
   Our most commonly used features are described in the following list: \
   <ul>\
-  <li><strong>OpenReview Affinity</strong>: our in-house paper-reviewer affinity model based on TF-IDF vector similarity. </li>\
-  <li><strong>TPMS</strong>: <a href="http://torontopapermatching.org/webapp/profileBrowser/about_us/">The Toronto Paper Matching System</a>, a third-party affinity model (if selected, venues are responsible for coordinating with TPMS)</li>\
-  <li><strong>Reviewer Bid Scores</strong>: Reviewers indicate their preference to review papers, which are then converted to scores.</li>\
+  <li><strong>Reviewer Bid Scores</strong>: Reviewers indicate their preference to review papers via bids, which are then converted to scores.</li>\
   <li><strong>Reviewer Recommendation Scores</strong>: Reviewers, Area Chairs, and/or Program Chairs recommend specific reviewers to review papers. These recommendations are then converted to scores according to your specifications.</li>\
+  <li><strong>OpenReview Affinity</strong>: Our in-house paper-reviewer affinity model based on TF-IDF vector similarity. </li>\
```
melisabok, line 48: This is not accurate "TF-IDF vector similarity", can we remove it?

openreview/venue_request/venue_request.py
```diff
@@ -498,13 +498,11 @@ def setup_request_form(self):
...
                 'values-checkbox': [
-                    'Organizers will assign papers manually',
                     'Reviewer Bid Scores',
                     'Reviewer Recommendation Scores',
-                    'OpenReview Affinity',
-                    'TPMS'
+                    'OpenReview Affinity'
                 ],
                 'order': 17,
                 'required': True,
```
melisabok, line 508: If manual assignments implies no selection, then required must be False. @celestemartinez can you check this?

openreview/venue_request/webfield/supportRequestsWeb.js
melisabok: @celestemartinez can you fix this?

openreview/venue_request/venue_request.py
celestemartinez: You are right, I also added instruction to deselect all options if manually assigning reviewers.

PR #741, opened 2020-08-25
added allow_zero_score_assignments to match config invitation

mohituniyal:
Change needed for https://github.com/openreview/openreview-matcher/pull/183


openreview/conference/matching.py
```diff
@@ -466,6 +466,12 @@ def _build_config_invitation(self, scores_specification):
                         'value-regex': '.*',
                         'required': False,
                         'order': 17
+                    },
+                    'allow_all_affinities': {
```
melisabok, line 470: I would like something more description, like 'allow_zero_score_assignments` and a description of this parameter.

PR #748, opened 2020-09-04
Fix/iclr submission stage

melisabok:


melisabok:
@celestemartinez could you take a look at these changes and we can discuss them on a video call.

celestemartinez:
Just one question, are we adding a field for 'second_due_date' to the request form, or how are we setting this date?

melisabok:
> Just one question, are we adding a field for 'second_due_date' to the request form, or how are we setting this date?

I think we discussed that we shouldn't add it because most of the venues don't have a second deadline. I could add this field to the ICLR request form only, or we can set the SubmissionStage from python.

celestemartinez (approved):
Looks good!

PR #754, opened 2020-09-17
fix remove review options

celestemartinez:
When remove_review_form_options was just one word (i.e. 'title'), it was not removed from the review form.
Now this field supports one word, lists, words separated by spaces and words separated by new lines.


openreview/conference/helpers.py
```diff
@@ -170,11 +170,7 @@ def get_review_stage(client, request_forum):
 
     review_form_additional_options = request_forum.content.get('additional_review_form_options', {})
 
-    review_form_remove_options = request_forum.content.get('remove_review_form_options', '')
-    if ',' in review_form_remove_options:
-        review_form_remove_options = [field.strip() for field in review_form_remove_options.strip().split(',') if field.strip()]
-    if not isinstance(review_form_remove_options, list):
-        review_form_remove_options = []
+    review_form_remove_options = request_forum.content.get('remove_review_form_options', '').replace(',', ' ').split()
```
melisabok, line 173: why don't split by ',' and remove the replace call?

melisabok: can you add a remove option here: https://github.com/openreview/openreview-py/blob/master/tests/test_venue_request.py#L373

and verify the field was removed from the review invitation?

celestemartinez: I know it says to add a comma-separated list, but I wanted to make sure it could also handle spaces and new lines just in case, and the split() handles that nicely. However, I am ok with using just split(',') and strip() since we do specify that it must be a comma-separated list

melisabok (approved):

PR #755, opened 2020-09-23
Fix/notifications

melisabok:
Send `parentGroup` parameter when sending the reminders to the PCs can see all theses messages in the messages UI.

### DIFF
--- a/openreview/conference/templates/reviewerWebfield.js
+++ b/openreview/conference/templates/reviewerWebfield.js
@@ -378,21 +378,3 @@ $('#group-container').on('click', 'a.note-contents-toggle', function(e) {
   var updated = $(this).text() === hiddenText ? visibleText : hiddenText;
   $(this).text(updated);
 });
-
-$('#group-container').on('click', 'a.send-reminder-link', function(e) {
-  var userId = $(this).data('userId');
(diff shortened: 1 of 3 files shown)


openreview/conference/templates/reviewerWebfield.js
```diff
@@ -378,21 +378,3 @@ $('#group-container').on('click', 'a.note-contents-toggle', function(e) {
   var updated = $(this).text() === hiddenText ? visibleText : hiddenText;
   $(this).text(updated);
 });
-
-$('#group-container').on('click', 'a.send-reminder-link', function(e) {
```
zbialecki, line 382: Why was this handler removed? Is it no longer being used?

zbialecki (commented):
Looks good, just one question

openreview/conference/templates/reviewerWebfield.js
melisabok: do we send reminders from the reviewers console?

I think this was a copy/paste from the AC console, correct me if I'm wrong.

zbialecki: I took at look and I think you are right.

zbialecki (approved):

celestemartinez (approved):

PR #756, opened 2020-09-28
Fix/iclr setup

melisabok:
- use tqdm for all the iterations.
- fix authorids field description.
- make the pdf field mandatory for the revision invitation

### DIFF
--- a/openreview/conference/templates/bidWebfield.js
+++ b/openreview/conference/templates/bidWebfield.js
@@ -8,10 +8,11 @@ var SHORT_PHRASE = '';
 var BLIND_SUBMISSION_ID = '';
 var BID_ID = '';
 var SUBJECT_AREAS = '';
-var AFFINITY_SCORE_ID = '';
 var CONFLICT_SCORE_ID = '';
+var SCORE_IDS = [];
 
 // Bid status data
+var selectedScore = SCORE_IDS.length && SCORE_IDS[0];
 var activeTab = 0
 var noteCount = 0;
 var conflictIds = [];
@@ -24,11 +25,6 @@ var bidsById = {
   'Low': []
 };
 var sections = [];
-var scoresByName = {
-  'TPMS': 'thecvf.com/ECCV/2020/Conference/Reviewers/-/TPMS_Score',
-  'Affinity': 'thecvf.com/ECCV/2020/Conference/Reviewers/-/Affinity_Score'
-}
-var selectedScore = 'Affinity';
 
 var paperDisplayOptions = {
   pdfLink: true,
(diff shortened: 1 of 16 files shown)


openreview/conference/templates/bidWebfield.js
```diff
@@ -345,5 +321,58 @@ function addEdgesToNotes(validNotes, edgesMap) {
...
 // Go!
 main();
+
+// Event handlers
+$('#invitation-container').on('change', 'form.notes-search-form .score-dropdown', function(e) {
+  console.log('change!', $(this).val());
+  selectedScore = $(this).val();
+        return getPapersSortedByAffinity(0, selectedScore)
+        .then(function(notes) {
+          return renderNotes(prepareNotes(notes, conflictIds, bidsByNote));
```
melisabok, line 376: replacing renderNotes with:

```
          $('.submissions-list', '#allPapers').replaceWith(
            Handlebars.templates['partials/noteList']({
              notes: prepareNotes(notes, conflictIds, bidsByNote),
              options: paperDisplayOptions
            })
          );
```

seems to fix the issues with the event but it doesn't reload the pagination bar, so it keeps in the same page. I'm not sure if we should reset the offset value to 0.

openreview/conference/templates/bidWebfield.js
```diff
@@ -345,5 +250,127 @@ function addEdgesToNotes(validNotes, edgesMap) {
   return validNotes;
 }
 
+function updateNotes(notes) {
```
melisabok, line 253: I fixed the issue with the tag widget events, I'm calling to re render the whole tab panel when the sorting criteria changes, it should be working now.

zbialecki:
Does dev have the current webfield code?

melisabok:
Thanks @zbialecki , I just deployed the changes in the dev site, could you take another look? I want to make sure the approach of reloading the whole tab panel is correct.

openreview/conference/templates/bidWebfield.js
```diff
@@ -10,7 +10,35 @@ var BID_ID = '';
...
+  'High': [],
+  'Neutral': [],
+  'Very Low': [],
+  'Low': []
+};
+var sections = [];
+var scoresByName = {
+  'TPMS': 'thecvf.com/ECCV/2020/Conference/Reviewers/-/TPMS_Score',
+  'Affinity': 'thecvf.com/ECCV/2020/Conference/Reviewers/-/Affinity_Score'
+}
```
zbialecki, line 30: I think this is fine for now, but I think we should come up with a more general method before this is merged into master.

zbialecki:
I think it's working correctly, the only thing I noticed is that it's missing the referrer banner

zbialecki (approved):

celestemartinez (approved):

openreview/conference/templates/bidWebfield.js
```diff
@@ -207,124 +211,19 @@ function renderContent(notes, conflictIds, bidEdges) {
...
-    sections[activeTab].active = true;
-
-    $('#notes .tabs-container').remove();
-
-    Webfield.ui.tabPanel(sections, {
-      container: '#notes',
-      hidden: true
+  $('#invitation-container').on('change', 'form.notes-search-form .score-dropdown', function(e) {
+    selectedScore = $(this).val();
+    return getPapersSortedByAffinity(0, selectedScore)
```
xkopenreview, line 216: passing selectedScore to getPapersSortedByAffinity is not necessary as the function takes only 1 argument which is the offset

PR #763, opened 2020-10-09
Feature/get pdfs

melisabok:


openreview/conference/builder.py
```diff
@@ -1111,6 +1113,38 @@ def set_homepage_decisions(self, invitation_name = 'Decision', decision_heading_
...
+        submissions = list(self.get_submissions(accepted))
+        pbar = tqdm(total=len(submissions), desc='Downloading files...')
+
+        if not os.path.exists(folder_path):
+            os.makedirs(folder_path)
+
+        def get_attachment_file(submission):
+            pbar.update(1)
+            if field_name in submission.content:
+                attachment_url = '{0}{1}'.format(self.client.baseurl, submission.content[field_name])
```
carlosmondra, line 1127: I think it's not important, but this attachment url will only work with `/pdf/<something-here>`. It does not work if it starts with `/attachment`.

carlosmondra (approved):
Oh nice! This should make much faster!

openreview/conference/builder.py
melisabok: I think the `attachment_url` is not being used, I can remove this.

celestemartinez (approved):

PR #766, opened 2020-10-09
Comparisons to singletons - PEP8 style guide

EwertonBello:
* Code based on [PEP 8 Style Guide](https://www.python.org/dev/peps/pep-0008/)

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.


melisabok (approved):
👍 Thanks!

PR #779, opened 2020-10-28
Fix/ Replace old frontend URLs in tests

zbialecki:
Fix frontend URLs in cases that were still referencing the old UI.

Required for https://github.com/openreview/openreview/pull/2218


melisabok (approved):
Oops, thanks for fixing this, I thought I had changed all the urls.

PR #780, opened 2020-10-29
Make sure PCs appear in invitees, writers and signatures

celestemartinez:
Added PCs to invitees, writers and signatures of Comments, Reviews, ReviewRebuttal, ReviewRevision and ReviewRating.


celestemartinez:
One comment:
- since the creation of ReviewRebuttal, ReviewRevision and ReviewRating invitations depend on the review's signature, then we end up with invitations like this (because when the PC edits the review, the signature changes): 

Screen Shot 2020-10-29 at 4 42 12 PM

Additionally, this is a problem if multiple reviews are edited by the PCs before these additional invitations are created.

Not sure if I'm missing something here @melisabok

melisabok:
I think we can do two things:

- Allow the PCs to post the review on behalf of the reviewer, that means they could sign the note using the AnonReviewer group.
- When creating the invitation that depend on the signature of the review, check if the signature is a unique group id, (eg: is AnonReviewer) and create the invitation otherwise, throw an error or ignore that review.

I think the PCs will use the PCs group to sign the review they want to delete.

openreview/conference/invitation.py
```diff
@@ -882,19 +887,19 @@ def __init__(self, conference, review):
...
+                'values-copied': [conference.get_id(), '{signatures}']
             }
         }
 
         super(PaperReviewRebuttalInvitation, self).__init__(id = signature + '/-/' + review_rebuttal_stage.name,
             super = conference.get_invitation_id(review_rebuttal_stage.name),
             writers = [conference.id],
             signatures = [conference.id],
-            invitees = [paper_group + '/Authors', conference.support_user],
+            invitees = [paper_group + '/Authors', conference.get_program_chairs_id(), conference.support_user],
```
melisabok, line 902: I was wrong before when I asked to add the PCs to be invitees, we don't need that otherwise the PC user will have a lot of pending tasks. 

Given that PCs are writers of the invitation I think they should be able to see the buttons in the UI, if this is not happening, may be we need to change the API.

melisabok:
These two PRs should allow PCs to edit submissions after the deadline:

https://github.com/openreview/openreview/pull/2247
https://github.com/openreview/openreview-web/pull/327

melisabok (approved):

PR #784, opened 2020-10-30
Feature/journal with edits

melisabok:
To setup the journal run:

```
pytest tests/test_journal.py
```


openreview/journal/webfield/actionEditorWebfield.js
```diff
@@ -0,0 +1,206 @@
...
+  ];
+
+  Webfield.ui.tabPanel(tabsList);
+};
+
+var renderTableAndTasks = function(venueStatusData) {
+
+  Webfield2.ui.renderTasks('#action-editor-tasks', venueStatusData.invitations, { referrer: encodeURIComponent('[Action Editor Console](/group?id=' + ACTION_EDITOR_ID + '#action-editor-tasks)')});
+
+  Webfield2.ui.renderTable('#assigned-papers', ['<input type="checkbox" id="select-all-papers">', '#', 'Paper Summary',
```
zbialecki, line 153: Nice! This looks a lot better than what we are currently using. 

My only suggestion would be to move the headings array and the row render functions array into the options object. So the function signature would be: `renderTable(container, data, options = {})`. I think it will make it even more clear.

melisabok: does it mean the headings and templates are optionals? what should I use if they are not present?

melisabok: I like the idea of having a default template for any object where we can show the key-values in a div.

zbialecki: Yeah I was thinking the headings could be optional (not shown if not provided), and the render functions could be like you said, just printing the values.

xkopenreview:
i think the author console is missing a line to remove empty message after rendering the table
image

openreview/journal/webfield/editorsInChiefWebfield.js
```diff
@@ -0,0 +1,360 @@
...
+            id: submission.id,
+            noteId: submission.id,
+            invitationId: VENUE_ID + '/Paper' + number + '/-/Review'
+          })
+        }
+        var reviewerStatus = reviewerStatusById[reviewer.id];
+        if (reviewerStatus) {
+          reviewerStatus.reviewerProgressData.numPapers += 1;
+          reviewerStatus.reviewerStatusData.numPapers += 1;
+          reviewerStatus.reviewerProgressData.papers.push({ note: formattedSubmission, review: { forum: completedReview.forum, status: status }});
```
xkopenreview, line 152: looks like completedReview might be undefined here

melisabok: I see, I can fix this.

xkopenreview:
the console with multiple tabs(checked eic console) seems to have the same id for sort dropdown and sort button
there's warning in browser Console
image

openreview/journal/webfield/editorsInChiefWebfield.js
```diff
@@ -0,0 +1,364 @@
...
+            id: submission.id,
+            noteId: submission.id,
+            invitationId: VENUE_ID + '/Paper' + number + '/-/Review'
+          })
+        }
+        var reviewerStatus = reviewerStatusById[reviewer.id];
+        if (reviewerStatus) {
+          reviewerStatus.reviewerProgressData.numPapers += 1;
+          reviewerStatus.reviewerStatusData.numPapers += 1;
+          reviewerStatus.reviewerProgressData.papers.push({ note: formattedSubmission, review: completedReview ? { forum: completedReview.forum, status: status } : null});
```
xkopenreview, line 152: opening http://localhost:3030/group?id=.TMLR/Editors_In_Chief as su has an error that completedReview is undefined
image

xkopenreview: open the url without logging in has an error in getGroup in webfield-v2 that group is undefined.

carlosmondra (approved):

PR #787, opened 2020-11-04
Allow PCs to release notes and reveal authors

celestemartinez:
This PR creates two more venue request buttons:
- Reveal Authors
- Release Notes

Venue is removed from active venues if any of these buttons is clicked.


melisabok:
This is great! The only suggestion that I would make is to create the 'Release Papers' and 'Release Authors' invitations in a later stage, for example during the Decision Stage process function. If we create them during the deploy stage we can have:

1. too many buttons.
2. PCs can try to release papers/authors before the decisions were created. 

What do you think?

celestemartinez:
Two comments:
-I named the invitation Post Decision Stage. It is not a stage so I am open to other names (perhaps just Post Decision)
-I believe combining the reveal_authors and release_notes functions would not make the code any shorter. A PC might want to release all papers to the public but only reveal author names of accepted papers, so they need to be handled separately. 

Let me know what you think @melisabok

melisabok (approved):

PR #796, opened 2020-11-20
parametrize home page tab name through the post decision stage

melisabok:


melisabok:
I'm going to merge this so I can enable the button for the NeurIPS workshopw.

PR #797, opened 2020-12-04
Fix/improve webfields

melisabok:
- Support referrer parameters in all the webfields.
- Show banner to the parent group for all the webfields, including landing pages.
- Make the decision home page optional when running post decision stage.
- Handle 'Reject' tab as optional.
- Compute 'venue' and 'venueid' for accepted papers. (In the future we can show the decision page using the venueid instead of querying the decision notes).
- Refactor home page webfield, some options were in the header json and we don't need them. 

UPDATE:
- Removed property `set_homepage_override()`
- Added a new line at the beginning of each webfield template to indicate the webfeld can be overwritten. 

Please review, feedback is welcome!

### DIFF
--- a/openreview/conference/webfield.py
+++ b/openreview/conference/webfield.py
@@ -21,6 +21,25 @@ class WebfieldBuilder(object):
 
         return merged_options
 
+    def __should_update(self, entity):
+        return entity.details.get('writable', False) and (not entity.web or entity.web.startswith('// webfield_template'))
+
+    def __update_invitation(self, invitation, content):
+        current_invitation=self.client.get_invitation(invitation.id)
+        if self.__should_update(current_invitation):
+            current_invitation.web = content
+            return self.client.post_invitation(current_invitation)
+        else:
+            return current_invitation
+
+    def __update_group(self, group, content):
+        current_group=self.client.get_group(group.id)
+        if self.__should_update(current_group):
+            current_group.web = content
+            return self.client.post_group(current_group)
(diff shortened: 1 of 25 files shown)


zbialecki (approved):
The webfield changes all look good to me!

melisabok:
Thanks @zbialecki , I have a question for you: for the roots groups I'm not using the banner, I'm calling to the `wecolme` function, do you think we should add a link to go back to the openreview home page or all venues?

zbialecki:
Honestly for that case I would probably not show the banner at all. If you don't call one of the banner functions it defaults to being hidden.

melisabok:
> Honestly for that case I would probably not show the banner at all. If you don't call one of the banner functions it defaults to being hidden.

Do you mean to not call `OpenBanner.welcome()`?

zbialecki:
Yes

openreview/conference/webfield.py
```diff
@@ -16,11 +16,33 @@ def __build_options(self, default, options):
...
+        if self.__should_update(current_invitation):
+            current_invitation.web = content
+            return self.client.post_invitation(current_invitation)
+        else:
+            return current_invitation
+
+    def __update_group(self, group, content, signature=None):
+        current_group=self.client.get_group(group.id)
+        if signature:
+            current_group.signatures=[signature]
```
celestemartinez, line 38: shouldn't we post the group here too if signature changes?

PR #798, opened 2020-12-08
Feature/eswc setup

melisabok:
- support two deadlines for single blind venues.
- support different readership for private venues.


openreview/conference/builder.py
```diff
@@ -1169,13 +1179,20 @@ def is_release_authors(is_note_accepted):
...
+        REVIEWERS = 3
+        REVIEWERS_ASSIGNED = 4
+
     def __init__(
             self,
             name='Submission',
             start_date=None,
             due_date=None,
             second_due_date=None,
-            public=False,
```
celestemartinez, line 1178: Shouldn't we still keep this public here? We keep using self.public in some places

melisabok: I'm defining the property below, I would like to remove it so we start using `readers`, I'm keeping the `public` property in the builder anyway.

openreview/conference/builder.py
```diff
@@ -1214,22 +1231,57 @@ def __init__(
         self.email_pcs_on_desk_reject = email_pcs_on_desk_reject
         self.author_names_revealed = author_names_revealed
         self.papers_released = papers_released
+        self.public = self.Readers.EVERYONE in self.readers
```
melisabok, line 1234: I'm defining the property here.

celestemartinez (approved):

PR #803, opened 2021-01-05
Make sure later stages do not overwrite readers and author reveal

### LINKED ISSUE #801: Request form: Submission Revision Stage overrides the settings of Post Decision Stage
When the PCs decide to release the papers to the public, then enable camera ready revisions, the papers go back to their original readers. 

We need to keep them with the same readers selected in the Post Decision Stage.

celestemartinez:
Fixes #801 

Any stage that calls setup_post_submission_stage() was overwriting changes made by post_decision_stage because we were not setting `submission_stage.author_names_revealed` and `submission_stage.papers_released` correctly.


melisabok (approved):
👍

PR #816, opened 2021-01-25
Add export button to AC console

celestemartinez:


melisabok (approved):

zbialecki (approved):
This is the same code that used in the PC console for CSV downloads right? If so then I think it's good to merge. If it's different maybe you could add a comment about how it's different.

PR #818, opened 2021-01-26
Feature/check queue

carlosmondra:
We have been using time.sleep(2) for process functions and messages that is not very reliable, so this PR checks the queue instead and waits until the elements in the queue are empty before moving on.


melisabok (approved):
Thanks, I thought we were already checking the status of the queue in some tests, I was wrong.

celestemartinez (approved):

Issue #811, opened 2021-01-14
Request form forum: allow markdown in the comment invitations


melisabok:
Allow markdown to the conversation between the support and the program chairs.

Issue #810, opened 2021-01-14
Recruit reviewers: put the committee role in the message subject


melisabok:
When we invite reviewers and acs, the emails subject is the same and the user can receive emails from both invitations clicking all of them thinking they were the same. 

I suggest to add the role name in the message subject.

PR #819, opened 2021-01-29
Add markdown and modify recruitment emails

celestemartinez:
Closes:
- #811 
- #810 

I changed recruit_reviewers email to be the same as the one used in the request form, which includes venue role in subject and message (previous message always said 'to serve as reviewer' even if ACs were being recruited which would cause confusion).

Added some formatting to deployment comment.

Added abstract to remove submission options.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -951,15 +951,15 @@ class Conference(object):
 
         invitation = self.invitation_builder.set_reviewer_recruiter_invitation(self, options)
         invitation = self.webfield_builder.set_recruit_page(self.id, invitation, self.get_homepage_options())
+
+        role = 'reviewer' if reviewers_name == 'Reviewers' else 'area chair'
         recruit_message = '''Dear {name},
 
-        You have been nominated by the program chair committee of ''' + self.short_name + ''' to serve as a reviewer.  As a respected researcher in the area, we hope you will accept and help us make the conference a success.
+        You have been nominated by the program chair committee of '''+ self.get_short_name() + ''' to serve as '''+ role +'''. As a respected researcher in the area, we hope you will accept and help us make ''' + self.short_name + ''' a success.
 
-        Reviewers are also welcome to submit papers, so please also consider submitting to the conference!
+        You are also welcome to submit papers, so please also consider submitting to '''+ self.get_short_name() + '''.
 
-        We will be using OpenReview.net and a reviewing process that we hope will be engaging and inclusive of the whole community.
-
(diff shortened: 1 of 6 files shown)


openreview/conference/builder.py
```diff
@@ -951,15 +951,15 @@ def recruit_reviewers(self, invitees = [], title = None, message = None, reviewe
 
         invitation = self.invitation_builder.set_reviewer_recruiter_invitation(self, options)
         invitation = self.webfield_builder.set_recruit_page(self.id, invitation, self.get_homepage_options())
-        recruit_message = '''Dear {name},
 
-        You have been nominated by the program chair committee of ''' + self.short_name + ''' to serve as a reviewer.  As a respected researcher in the area, we hope you will accept and help us make the conference a success.
+        role = 'reviewer' if reviewers_name == 'Reviewers' else 'area chair'
```
melisabok, line 955: How do we do this in the request form when the reviewers_name is 'Area_Chairs'? Do we need to replace underscores with spaces?

openreview/venue_request/venue_request.py
```diff
@@ -24,8 +24,8 @@ def setup_venue_revision(self):
         }
         revision_content['remove_submission_options'] = {
             'order': 19,
-            'values-dropdown':  ['keywords', 'pdf', 'TL;DR'],
-            'description': 'Fields to remove from the default form: keywords, pdf, TL;DR'
+            'values-dropdown':  ['abstract','keywords', 'pdf', 'TL;DR'],
```
melisabok, line 27: Thanks for this!

openreview/conference/builder.py
```diff
@@ -973,14 +973,15 @@ def recruit_reviewers(self, invitees = [], title = None, message = None, reviewe
...
+        If you have any questions, please contact info@openreview.net.
 
         Cheers!
 
         Program Chairs
 
         '''
-        recruit_message_subj = self.id + ': Invitation to Review'
+
+        recruit_message_subj = '[' + self.get_short_name() + ']: Invitation to serve as ' + role.capitalize()
```
melisabok, line 984: I like this new subject!

openreview/conference/builder.py
celestemartinez: if reviewer_name == 'Area_Chairs', then role would be 'area chair', is this what you mean?

melisabok: I see, I think this is ok for this PR but we should translate the role from the group id so we can support other names than reviewers and area chairs.

melisabok (approved):

PR #821, opened 2021-02-01
Fix/test permissions

melisabok:
- Remove deprecated code `open_submissions` and `close_submissions`, use the SubmissionStage class or `setup_post_submissions_stage`.

- Fix some tests where the replies can not read the forum.

### DIFF
--- a/openreview/conference/invitation.py
+++ b/openreview/conference/invitation.py
@@ -734,6 +734,10 @@ class OfficialCommentInvitation(openreview.Invitation):
                 'forum': note.id,
                 'replyto': None,
                 'readers': reply_readers,
+                'nonreaders': {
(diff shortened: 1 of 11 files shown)


openreview/conference/invitation.py
```diff
@@ -734,6 +734,10 @@ def __init__(self, conference, note):
                 'forum': note.id,
                 'replyto': None,
                 'readers': reply_readers,
+                'nonreaders': {
+                    'values-dropdown': readers,
+                    'required': False
+                },
```
carlosmondra, line 740: `required` is only available inside content fields :D
```suggestion
                'nonreaders': {
                    'values-dropdown': readers
                },
```

carlosmondra (approved):

PR #823, opened 2021-02-02
Recruitment Process Function

### LINKED ISSUE #808: Reviewer Recruitment process function: validate if the user is already an AC before adding them as a member
In order to avoid the issue of having users member of ACs and Reviewers group because they clicked on both invitations links. 

We can add a validation to check if the user is already an AC them don't the user to the reviewers group and send an email with the reason.

celestemartinez:
-Fixes #808 
- I am adding the reviewer to the declined group if they are already an AC (not sure about this)
- Open to suggestions about the wording of the email sent if they are already an AC/reviewer
- Will need to change the webfield if user responds 'Yes' but they are not added to reviewer group since they are already an AC

### DIFF
--- a/openreview/conference/invitation.py
+++ b/openreview/conference/invitation.py
@@ -1458,17 +1458,23 @@ class InvitationBuilder(object):
         }
         reply = self.__build_options(default_reply, options.get('reply', {}))
 
-        with open(os.path.join(os.path.dirname(__file__), 'templates/recruitReviewersProcess.js')) as f:
+        with open(os.path.join(os.path.dirname(__file__), 'templates/recruitReviewersProcess.py')) as f:
             content = f.read()
-            content = content.replace("var SHORT_PHRASE = '';", "var SHORT_PHRASE = '" + conference.get_short_name() + "';")
-            content = content.replace("var CONFERENCE_NAME = '';", "var CONFERENCE_NAME = '" + conference.get_id() + "';")
-            content = content.replace("var REVIEWER_NAME = '';", "var REVIEWER_NAME = '" + options.get('reviewers_name', 'Reviewers').replace('_', ' ')[:-1] + "';")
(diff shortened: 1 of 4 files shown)


openreview/conference/templates/recruitReviewersProcess.py
```diff
@@ -0,0 +1,56 @@
...
+    REDUCED_LOAD_INVITATION_NAME = ''
+
+    user = urllib.parse.unquote(note.content['user'])
+
+    hashkey = HMAC.new(HASH_SEED.encode(), digestmod=SHA256).update(user.encode()).hexdigest()
+    
+    invited_group = client.get_group(REVIEWERS_INVITED_ID)
+    if (hashkey == note.content['key'] and user in invited_group.members):
+        if (note.content['response'] == 'Yes'):
+            if (AREA_CHAIRS_ACCEPTED_ID and user in client.get_group(AREA_CHAIRS_ACCEPTED_ID).members):
```
melisabok, line 22: In order to check any other user email/tilde id, we can do the following:

```
client.get_groups(regex=AREA_CHAIRS_ACCEPTED_ID, member=user)
```

If the result is not empty then the user accepted to be an area chair.

openreview/conference/invitation.py
```diff
@@ -1458,17 +1458,23 @@ def set_reviewer_recruiter_invitation(self, conference, options = {}):
         }
         reply = self.__build_options(default_reply, options.get('reply', {}))
 
-        with open(os.path.join(os.path.dirname(__file__), 'templates/recruitReviewersProcess.js')) as f:
+        with open(os.path.join(os.path.dirname(__file__), 'templates/recruitReviewersProcess.py')) as f:
```
melisabok, line 1461: Thanks for doing this

celestemartinez:
> Cool! Should I merge it?

Sure! If you think everything looks good

melisabok (approved):

PR #828, opened 2021-02-06
Fix/recruitment process

celestemartinez:
Raises exception for:
- invalid key
- invalid response


celestemartinez:
This test fails if we don't decode the url: https://github.com/openreview/openreview-py/blob/deada312139e42d3bfed482ad89fa657f1604741/tests/test_double_blind_conference.py#L549 but not sure if it would actually happen in real life

melisabok:
I understand what is doing now, some email client will decode the '@' character in the url and we should decode it to compute the hash key.

melisabok (approved):

PR #829, opened 2021-02-09
Add second due date to request form

celestemartinez:
-Add second due date to request form
-Make pdf optional if there is a second due date

### DIFF
--- a/openreview/conference/helpers.py
+++ b/openreview/conference/helpers.py
@@ -94,17 +94,6 @@ def get_conference_builder(client, request_form_id, support_user='OpenReview.net
     if isinstance(submission_additional_options, str):
         submission_additional_options = json.loads(submission_additional_options.strip())
 
-    if submission_second_due_date:
-        submission_additional_options['pdf'] = {
-            'description': 'Upload a PDF file that ends with .pdf',
-                'order': 9,
-                'value-file': {
-                    'fileTypes': ['pdf'],
-                    'size': 50
-                },
-                'required':False
-        }
-
     submission_remove_options = note.content.get('remove_submission_options', [])
     withdrawn_submission_public = 'Yes' in note.content.get('withdrawn_submissions_visibility', '')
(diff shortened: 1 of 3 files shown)


openreview/conference/helpers.py
```diff
@@ -85,6 +94,17 @@ def get_conference_builder(client, request_form_id, support_user='OpenReview.net
     if isinstance(submission_additional_options, str):
         submission_additional_options = json.loads(submission_additional_options.strip())
 
+    if submission_second_due_date:
```
melisabok, line 97: can we do this inside of the builder?, I guess we already do this.

melisabok: here: https://github.com/openreview/openreview-py/blob/master/openreview/conference/builder.py#L1281

if second_due_date exists then content['pdf']['required'] = False

melisabok: make sure 'pdf' exists in content

openreview/conference/helpers.py
```diff
@@ -42,15 +42,24 @@ def get_conference_builder(client, request_form_id, support_user='OpenReview.net
         submission_start_date = None
 
     submission_due_date_str = 'TBD'
-    submission_due_date = note.content.get('Submission Deadline', '').strip()
-    if submission_due_date:
+    submission_second_due_date = note.content.get('Submission Deadline', '').strip()
```
melisabok, line 45: I think it is possible the user enter an abstract deadline but not a submission deadline, what should we do in that case?

celestemartinez:
I was looking at the issue of having `due_date == second_due_date`

I don' think it'll be a problem as https://github.com/openreview/openreview-py/blob/47b6271bc336ab02ba3e9ae5d318bd088e63f9f7/openreview/conference/builder.py#L708 would not be entered and only `setup_final_deadline_stage` would be executed

melisabok (approved):

PR #832, opened 2021-02-10
Fix/ Bid webfield hide note

zbialecki:
Remove notes from bid-specific tabs when bid is changed.


zbialecki:
You can test these changes out here: https://dev.openreview.net/invitation?id=NeurIPS.cc%2F2021%2FConference%2FReviewers%2F-%2FBid

melisabok (approved):
💯

PR #838, opened 2021-02-18
Fix/submission process function

celestemartinez:
One comment: I think we need to handle if there is no due_date separately. Now, if post_submission_stage() is called on a venue with no submission deadline, the submission process function will be removed. 

Maybe we can pass under_submission as an argument to setup_final_deadline_stage() accordingly.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -723,7 +723,7 @@ class Conference(object):
             if self.submission_stage.second_due_date < now:
                 self.setup_final_deadline_stage(force, hide_fields)
         else:
-            if force or self.submission_stage.due_date < datetime.datetime.now():
+            if force or (self.submission_stage.due_date and self.submission_stage.due_date < datetime.datetime.now()):
                 self.setup_final_deadline_stage(force, hide_fields)
 
     ## Deprecated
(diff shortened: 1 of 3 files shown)


melisabok:
> Maybe we can pass under_submission as an argument to setup_final_deadline_stage() accordingly.

Do we need to call `setup_final_deadline_stage` when there is no due_date? It seems that we shouldn't call it becase we will never reach any deadline.

openreview/conference/builder.py
```diff
@@ -723,7 +723,7 @@ def setup_post_submission_stage(self, force=False, hide_fields=[]):
             if self.submission_stage.second_due_date < now:
                 self.setup_final_deadline_stage(force, hide_fields)
         else:
-            if force or not self.submission_stage.due_date or self.submission_stage.due_date < datetime.datetime.now():
+            if force or self.submission_stage.due_date < datetime.datetime.now():
```
melisabok, line 726: What happens with due_date is None? 

I'm getting this error in a python console:

```

TypeError: '<' not supported between instances of 'NoneType' and 'datetime.datetime'

```

celestemartinez: I am now checking that due date is not None, if it is it is just ignored

melisabok (approved):

PR #841, opened 2021-02-23
set the reduced load from the request form

melisabok:
PCs can select a reduce load option if the user declines the invitation, please check the wording 🙏 

`'Please enter a comma separated values of the reduced load options if the invitee rejected the invitation. (Available for reviewer role only)'`

### DIFF
--- a/openreview/venue_request/process/deployProcess.py
+++ b/openreview/venue_request/process/deployProcess.py
@@ -178,7 +178,7 @@ Program Chairs'''.replace('{Abbreviated_Venue_Name}', conference.get_short_name(
                     'order': 2
                 },
                 'invitee_reduced_load': {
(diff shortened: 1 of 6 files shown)


openreview/venue_request/process/deployProcess.py
```diff
@@ -118,23 +118,30 @@ def process(client, note, invitation):
                     'required': True,
                     'order': 2
                 },
+                'invitee_reduced_load': {
+                    'description': 'Please enter a comma separated value of the reduced load options if the invitee rejected the invitation. (Only for reviewer role)',
```
celestemartinez, line 122: Maybe something like:
"Please enter a comma separated list of reduced load options. If an invitee rejects the reviewing invitation, they will be able to choose a reduced load from this list. (For reviewer role only)"

Might seem wordy but I think it's a bit more understandable.

celestemartinez (approved):

PR #844, opened 2021-03-01
Fix/post submission button

### LINKED ISSUE #826: Stages references: should not be editable by the Program Chairs
They can not edit the references because it doesn't rollback the changes.

### LINKED ISSUE #842: Post Submission button: make it visible to the Program Chairs
I think we should make it visible to the PCs, sometimes they want to create the blind submissions before the deadline so we should allow them to do that. 

Also update the FAQ with instructions to do this.

celestemartinez:
Fixes #842 
Fixes #826

### DIFF
--- a/openreview/venue_request/process/revisionProcess.py
+++ b/openreview/venue_request/process/revisionProcess.py
@@ -6,9 +6,6 @@ def process(client, note, invitation):
     conference = openreview.helpers.get_conference(client, note.forum, SUPPORT_GROUP)
     forum_note = client.get_note(note.forum)
 
-    note.writers = []
-    note = client.post_note(note)
-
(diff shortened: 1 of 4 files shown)


openreview/venue_request/process/revisionProcess.py
```diff
@@ -6,6 +6,9 @@ def process(client, note, invitation):
     conference = openreview.helpers.get_conference(client, note.forum, SUPPORT_GROUP)
     forum_note = client.get_note(note.forum)
 
+    note.writers = []
```
melisabok, line 9: instead of doing this, why don't we set the writers empty in each Stage invitation?

melisabok (approved):
💯

PR #857, opened 2021-03-15
Check if under submission when setting submission stage

celestemartinez:
If submission deadline has passed, pass `under_submission=False` to `set_submission_invitation()`

Notice this same error was also happening when using `post_decision_stage()` even from the venue request.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -171,9 +171,8 @@ class Conference(object):
         return len(invitations)
 
     def __create_submission_stage(self):
-        if self.submission_stage.due_date and self.submission_stage.due_date < datetime.datetime.utcnow():
-            return self.invitation_builder.set_submission_invitation(self, under_submission=False)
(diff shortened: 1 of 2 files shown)


openreview/conference/builder.py
```diff
@@ -171,6 +171,8 @@ def __expire_invitations(self, name):
         return len(invitations)
 
     def __create_submission_stage(self):
+        if self.submission_stage.due_date and self.submission_stage.due_date < datetime.datetime.utcnow():
```
melisabok, line 174: Could you avoid the if statement calling `set_submission_invitation(self, not(self.submission_stage.due_date and self.submission_stage.due_date < datetime.datetime.utcnow()))`

or 

```
under_submission=not self.submission_stage.due_date or self.submission_stage.due_date > datetime.datetime.utcnow()
set_submission_invitation(self, under_submission=under_submission)
```

melisabok (approved):

PR #862, opened 2021-03-24
Fix/Capital letter words in bibtex can be corrupted

xkopenreview:
use the conversion rule of pylatexenc instead of capitalize_title()

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -7,7 +7,7 @@ import openreview
 import re
 import datetime
 import time
-from pylatexenc.latexencode import utf8tolatex
+from pylatexenc.latexencode import utf8tolatex, unicode_to_latex, UnicodeToLatexConversionRule, UnicodeToLatexEncoder, RULE_REGEX
 from Crypto.Hash import HMAC, SHA256
 from multiprocessing import Pool
 from tqdm import tqdm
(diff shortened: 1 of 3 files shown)


xkopenreview:
it looks like the result of utf8tolatex and unicode_to_latex is not exactly the same
for the string we used in test 'Paper title has an Ô' (the result is the same string passing it through capitalize_title func)
utf8tolatex converted it to 
```python
'Paper title has an {\^O}'
```
and 
unicode_to_latex converted it to
```python
'Paper title has an \^O'
```

however the results when used as reference are the same 

image

melisabok:
Thanks for fixing this! could you add the test case that you are fixing in test_bibtex.py?

xkopenreview:
> Thanks for fixing this! could you add the test case that you are fixing in test_bibtex.py?

i modified the note title of existing test to include a capital letter word. do you think that's enough?

melisabok:
There are tests computing the bibtex and it seems one test is failing:

test_double_blind_conference.py

xkopenreview:
it looks like the extra escape of curly braces in title was the expected behavior.

the expected bibtex title in test_double_blind_conference is
```bibtex
title={Paper title {\{}REVISED{\}} {\{}AGAIN{\}}},
```

maybe the latex rule has changed.

melisabok:
Can you verify that bibtex is correctly rendered in a pdf?

xkopenreview:
> Can you verify the bibtex is correctly rendered in a pdf?

yes tested on overleaf.com

image

openreview/tools.py
```diff
@@ -366,12 +355,22 @@ def capitalize_title(title):
...
+        conversion_rules=[
+            UnicodeToLatexConversionRule(
+                rule_type=RULE_REGEX,
+                rule=[
+                    (re.compile(r'[A-Z]{2,}'), r'{\g<0>}')
+                ]),
+            'defaults'
+        ]
+    )
+    bibtex_title = u.unicode_to_latex(note.content['title'])
```
melisabok, line 368: why do you apply this to the title only and the other fields still use `utf8tolatex`, it seems that `utf8tolatex` is deprecated: https://pylatexenc.readthedocs.io/en/latest/latexencode/

xkopenreview: because title is transformed with capitalize_title fn so that it will be transformed wrongly, the other fields won't have this problem.

utf8tolatex will work fine and is clearer otherwise we will have to construct two separate encoders for title and other fields

melisabok (approved):

PR #867, opened 2021-03-29
Recruitment

celestemartinez:
- Check if user has a profile and remove both email and profile.id from accepted/rejected groups.


melisabok (approved):

melisabok:
the change looks good to me but could you add a test in test_neurips_conference.py that call replace_members and the response process function does what is expected?

- recruitment ac1
- ac1 accepts the invitation
- replace members with id in the AC group
- ac1 rejects the invitation
- ac1 is not member of AC group, ac1 is members of AC/Declined

PR #868, opened 2021-04-01
Support ac console when a paper has multiple ACs

celestemartinez:


openreview/conference/templates/areachairWebfield.js
```diff
@@ -71,11 +71,11 @@ var getRootGroups = function() {
     var paperNumbers = [];
     _.forEach(paperGroups, function(g) {
       var num = getNumberfromGroup(g.id, 'Paper');
-      g.members.forEach(function(member, index) {
-          var anonGroup = anonGroups.find(function(g) { return g.id.startsWith(CONFERENCE_ID + '/Paper' + num) && g.members[0] == member; });
-          anonids[num] = anonGroup.id;
-      })
-      paperNumbers.push(parseInt(num));
+      var anonGroup = anonGroups.find(function(g) { return g.id.startsWith(CONFERENCE_ID + '/Paper' + num) && g.members[0] == user.id; });
```
melisabok, line 74: This should work:
```suggestion
      var anonGroup = anonGroups.find(function(g) { return g.id.startsWith(CONFERENCE_ID + '/Paper' + num) && (user.emails.indexOf(g.members[0]) >= 0 || user.usernames.indexOf(g.members[0]) >= 0); });
```

melisabok: Now I'm realizing the query is already asking the group where the user is a member so we shouldn't do this check. 

Only checking the anonGroup startWith(CONFERENCE_ID + '/Paper' + num) is enough.

melisabok (approved):

PR #886, opened 2021-04-26
use directReplies

melisabok:
ECCV using this version in live site: https://openreview.net/group?id=thecvf.com/ECCV/2020/Conference/Program_Chairs

ICLR using this version in live site: https://openreview.net/group?id=ICLR.cc/2021/Conference/Program_Chairs

I think the improvement is significant. I would like to remove the call to get recommendation edges but it does too many calls or try to the get the edges grouped by signature.


zbialecki (commented):
The code looks good to me, but while testing the ICLR 2021 homepage on live I got the follow JS error when I tried to do an export:

```
webfieldCode.js:2919 Uncaught TypeError: Cannot read property '1954' of undefined
```

In the `buildCSV` function it seems like the var `completedReviews` is undefined

zbialecki:
Also, in the sorting seems to be broken on the Reviewer Status tab. 
```
webfieldCode.js:2001 Uncaught TypeError: Cannot read property 'numPapers' of undefined
    at Papers_with_Reviews_Missing (webfieldCode.js:2001)
```

openreview/conference/templates/programchairWebfield.js
```diff
@@ -1703,20 +1806,40 @@ var displayAreaChairsStatusTable = function() {
...
-      return [number, summaryHtml, progressHtml, statusHtml];
+      if (SENIOR_AREA_CHAIRS_ID) {
+        return [number, summaryHtml, seniorSummaryHtml, progressHtml, statusHtml];;
+      } else {
+        return [number, summaryHtml, progressHtml, statusHtml];
+      }
     });
 
+    var headings;
+    if (SENIOR_AREA_CHAIRS_ID) {
```
melisabok, line 1820: here is the change @zbialecki

zbialecki (approved):
I think this can be merged too.

celestemartinez (approved):

PR #887, opened 2021-04-26
Fix/author ranking and recruit other committee groups

melisabok:
1 - Enable Author Paper Ranking
2 - Allow the recruitment of other committee groups that can overlap with SAC/AC/Reviewers invited groups.


zbialecki (commented):
The webfield changes look good to me

PR #894, opened 2021-04-30
Add recruitment pre-process

### LINKED ISSUE #881: Recruitment invitation: add a pre-process function to check valid key and user is invitee
Before posting the response note, the pre-process function can check if the key is valid and the user is invited or not.

celestemartinez:
Users can see one if three errors:
- `Group Not Found` if group doesn't exist
- `User not in invited group, please accept the invitation using the email address you were invited with` if they change their email to an an email whose group exists in OR
- `Wrong key, please refer back to the recruitment email` when the key is different

Would maybe like to make errors more clear or write instructions somewhere on what to do to accept/decline invitation

Fixes #881


openreview/conference/invitation.py
```diff
@@ -1609,36 +1609,41 @@ def set_reviewer_recruiter_invitation(self, conference, options = {}):
...
-                readers = ['everyone'],
-                nonreaders = [],
-                invitees = ['everyone'],
-                noninvitees = [],
-                writers = [conference.get_id()],
-                signatures = [conference.get_id()],
-                reply = reply,
-                process_string = content,
-                web_string = current_invitation.web if current_invitation else None)
+        with open(os.path.join(os.path.dirname(__file__), 'templates/recruitReviewersPreProcess.py')) as pre:
```
melisabok, line 1612: could you rename the files with underscores like `recruit_reviewers_pre_process.py`? I think this is the Python standard.

tests/test_double_blind_conference.py
```diff
@@ -662,16 +662,16 @@ def test_recruit_reviewers(self, client, selenium, request_page, helpers):
...
         request_page(selenium, accept_url, alert=True)
 
         helpers.await_queue()
 
         logs = client.get_process_logs()
         assert logs
-        assert logs[0]['status'] == 'error'
-        assert logs[0]['error'] == {'message': 'Invalid key or user not in invited group', 'name': 'Error', 'status': 400}
+
+        assert logs[0]['status'] == 'ok'
```
melisabok, line 674: there shouldn't be any logs so you can remove this.

melisabok: `helpers.await_queue()` too

melisabok (approved):

PR #895, opened 2021-04-30
Senior Area Chair console

melisabok:
- Implement Paper Status and AC Status tab in the SAC console. 
- Fix other permissions related to NeurIPS 2021.


zbialecki (approved):
I think it looks good. I couldn't figure out why calling displayAreaChairsStatusTable on load was necessary, but I think it's fine for now.

celestemartinez (approved):

Issue #893, opened 2021-04-29
PC console: unexpected behavior pressing enter key in search input


xkopenreview:
it looks like when pressing enter key in search input has the following behavior:
when export button is present, it's having focus and pressing enter key triggers csv download.

https://user-images.githubusercontent.com/60613434/116608421-28d45e80-a901-11eb-8d70-322d4cfc0adb.mp4

when there's no export button, it seems like the switch sort option button is having focus and pressing enter key in search input will reverse sort direction
**this is local env**

xkopenreview:
it looks like the behavior is not consistent 
in prod the sort by option is cleared when enter key is pressed

https://user-images.githubusercontent.com/60613434/116617387-c765bd00-a90b-11eb-8c6e-f1fbeeb978e8.mp4

melisabok:
I fixed this issue in the NeurIPS PC console but I didn't push the change. Zach helped me to find the bug: all the buttons must have the property `type="button"`

xkopenreview:
pressing enter key is clearing sort by dropdown is caused by 
```javaScript
$(container + ' .form-sort').val('Reviewer');
```
in searchResults function
do you know what this does?

my guess is that this is for resetting sort by option, then Reviewer should be changed to 'Reviewer_Name'

melisabok:
@xkopenreview I think this is fixed, can I close this issue?

xkopenreview:
> @xkopenreview I think this is fixed, can I close this issue?

yes

PR #897, opened 2021-05-01
Fix/allow query search in pc console

xkopenreview:
allow query search in pc console;
possible fix of https://github.com/openreview/openreview-py/issues/893

type + in search input to enter query mode
press enter key to do query search
e.g. 
+id=123
+id="12 3"
+number<5 OR number>100
+number>2 AND number<5

webfield related change is in https://github.com/openreview/openreview-web/pull/472


openreview/conference/templates/programchairWebfield.js
```diff
@@ -37,6 +37,28 @@ var SENIOR_AREA_CHAIRS_INVITED_ID = SENIOR_AREA_CHAIRS_ID ? SENIOR_AREA_CHAIRS_I
 var ENABLE_REVIEWER_REASSIGNMENT = false;
 var PAPER_REVIEWS_COMPLETE_THRESHOLD = 3;
 var PAGE_SIZE = 25;
+var filterOperators = ['!=','>=','<=','>','<','=']; // sequence matters
+var propertiesAllowed ={
+    number:['note.number'],
+    id:['note.id','summary.id'],// multi tab same prop
+    title:['note.content.title'],
+    author:['note.content.authors','note.content.authorids'], // multi props
+    subject:['note.content.subject'],
```
melisabok, line 46: do you mean `abstract` here? There are no venues using the property `subject`.

melisabok: or keywords?

zbialecki (commented):
I think the changes to the webfields here look pretty good, but we should have a discussion about the changes to the query syntax that you made Xukun. I think it's cumbersome and more confusing for users to have to type AND or OR between every filter. It's also unclear to me how combining ands and ors will work. Will it be evaluated `(filter1 && filter2) || filter3`? Or `filter1 && (filter2 || filter3)`?

melisabok:
I added a description of the current implementation in the NeurIPS document: https://docs.google.com/document/d/1JKNmfEBC2yVXCU-3YLn4MQd4KKWrP-hU9TmAWaBzZCs/edit?ts=608b1beb#

The only restriction that I see with the current implementation is that () are not supported so users can not group the conditions.

openreview/conference/templates/programchairWebfield.js
```diff
@@ -1226,8 +1274,14 @@ var displayPaperStatusTable = function() {
 
     var filteredRows;
     if (searchText) {
-      filteredRows = _.orderBy(_.filter(rowData, filterFunc), sortOptions['Paper_Number'], 'asc');
-      matchingNoteIds = filteredRows.map(function(row) { return row.note.id; });
+      if(isQueryMode){
+        ({ filteredRows, queryIsInvalid } = Webfield.filterCollections(rowData, searchText.slice(1), filterOperators, propertiesAllowed, 'note.id'))
```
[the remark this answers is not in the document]
xkopenreview: thanks. 
besides the newly added code, there's also a saveAs call at the end which seems to have an extra comma and would fail with syntax error in ie.

so it looks like for the past 10 months no pc had used internet explorer to access the pc console.
maybe it's time to revisit support of internet explorer

openreview/conference/templates/programchairWebfield.js
```diff
@@ -1077,10 +1096,45 @@ var displaySortPanel = function(container, sortOptions, sortResults, searchResul
...
+        }).hover(function (e) {
+          $(e.target).tooltip({
+            title: "query format is {filter}=value<br/>\nand can be combined with AND/OR<br/>\neg. +number<50 AND reviewers=\"first last\"<br/>\navailable filters:".concat(Object.keys(propertiesAllowed).join('<br/>')),
+            html: true,
+            placement: 'bottom'
+          });
+        }));
+      }
+  
+      if (e.key === 'Enter') {
```
zbialecki, line 1123: This should also check for tab

zbialecki (commented):
I think that the only thing missing now is to add a modal with instructions on how to query and a list of all available filters. This modal could open when you click the info icon. I would still keep the tooltip on it, with the text "Show available query options"

melisabok:
1. Search by paper number must be an exact match:

```
+number=555
```

should return the paper 555 or an empty result. Now it is returning paper 5.

2. When I'm typing a search term, the button is changing between 'export' and 'export filtered' and it seems weird. I would keep 'export' in both cases. 

Screen Shot 2021-05-10 at 12 10 43 PM

3. I would change the placeholder text from 'Search all papers...' to 'Enter search term or type + to start a query and press enter...'.

4. I think we should give more space for the query. What about the search textbox uses all the space and move the 'Message Reviewers' and 'Export' to a new row?

Screen Shot 2021-05-10 at 12 15 13 PM

5. When I sort the papers, the search is reset, is it possible to keep the filtered rows?

6. Is the title filter doing an exact match? For example:

```
+title="Machine"
```

Return an empty result. Can we use a regex to match titles. I don't think the PCs will enter the full title to search a paper.

7. Do the match case insensitive:

Return results:
```
+keywords="machine"
```

Return no results:
```
+keywords="Machine"
```

8. When an operator is not equals, what is the behavior for strings?

This filter return results:

```
+keywords>"Machine"
```

9. Num of reviewers done is returning an empty result:

Screen Shot 2021-05-10 at 12 30 44 PM

10. It seems I can not use single quotes to enter a string term. Is it possible to validate this?

11. Rename 'metaReviewer' to 'areaChair', we will need to support 'seniorAreaChair'. 

12. I would like to be able to have spaces between the operators, field names and values.

openreview/conference/templates/programchairWebfield.js
xkopenreview: looks like we can't catch tab key up
not sure will it worth to add a keydown handler for this

xkopenreview:
how about changing message and export to icon buttons, like the sort switch button on the right
> 2. I think we should give more space for the query. What about the search textbox uses all the space and move the 'Message Reviewers' and 'Export' to a new row?
> 
> Screen Shot 2021-05-10 at 12 15 13 PM

melisabok:
> how about changing message and export to icon buttons, like the sort switch button on the right @melisabok

I like it, be aware the message icon should display a menu. Or we should show the options in the modal window.

zbialecki (commented):
- I'm not a fan of completely removing the text from the message button. I would prefer to keep the work "Message" in there to make it clear, especially since the dropdown options don't make sense without any text.
- Changing the Export button to Export Filtered takes up a lot of space that could be used for the Message button, what about opening a window.confirm dialog telling the user that the export will only contain the matching rows and asking if they want to continue?
- The i icon should be inside the `strong` tag before the ":": 
  Screen Shot 2021-05-12 at 5 23 48 PM
- Now that there is grouping of filters I think it's even more important to have a modal with better instructions on how to do use the query language. The tooltip is too hard to read and can't be formatted.

xkopenreview:
> * Changing the Export button to Export Filtered takes up a lot of space that could be used for the Message button, what about opening a window.confirm dialog telling the user that the export will only contain the matching rows and asking if they want to continue?

if the rows are filtered, i added a (Filtered) in the title of the exported file.

xkopenreview:
@melisabok @zbialecki 
change sorting dropdown and direction will not clear search in 8318d4a
but added for the paper status tab only.

i'm not sure if we should do the same for the other 2 tabs.
do you know what's the rational of current logic?(clearing search when sort option is changed)

zbialecki:
I don't think there was a good reason... it was probably just done to make the logic simpler.

openreview/conference/templates/programchairWebfield.js
```diff
@@ -1224,14 +1277,30 @@ var displayPaperStatusTable = function() {
...
+      matchingNoteIds = filteredRows.map(function (row) {
+        return row.note.id;
+      });
+      order = 'asc'
     } else {
       filteredRows = rowData;
       matchingNoteIds = [];
     }
+    if (rowData.length !== filteredRows.length) {
+      conferenceStatusData.filteredNotes = filteredRows.map(function (p) {
```
melisabok, line 1300: I would like to keep the status of `filteredRows` instead of `filteredNotes`, then the export button doesn't need to compute the rows again to export the stats. I know this is not related to this PR but I would reduce the duplicated code.

xkopenreview:
attached updated ac console code for midl short conference
[midl_short_new.js.zip](https://github.com/openreview/openreview-py/files/6534606/midl_short_new.js.zip)

zbialecki:
I tested this out with ICLR 2021 and ran into an issue with the `buildAreaChairGroupMaps` function. It seems this code will only work with groups that use the new anon AC groups. I added a basic check so it didn't crash the page, but I don't think it should be used on venues that use the numbered AC groups like "ICLR.cc/2021/Conference/Paper100/Area_Chair1".

Other than that it seems like the querying is working fine. I had to add 2 constants `SENIOR_AREA_CHAIRS_ID` and `SENIOR_AREA_CHAIRS_INVITED_ID` to the old webfield when testing it.

zbialecki (approved):

melisabok:
@xkopenreview can you add the search to the SAC console too?

xkopenreview:
function handling click of export button in sac console is missing, so remove the button for now

zbialecki:
I think this can be merge now, right?

melisabok (approved):

PR #900, opened 2021-05-03
check profile ids

melisabok:


carlosmondra:
@melisabok I think this is no longer needed after fixing it [here](https://github.com/openreview/openreview/pull/2390), right?

PR #908, opened 2021-05-16
Feature/ Lite webfield templates

zbialecki:
Create new stripped-down versions of the conference homepage and author console.

- Author console no longer loads any invitations and does not show an Author Tasks tab. Also, it does not attempt to load any reviews or tag widgets
- Conference homepage no longer includes an Activity tab, and just displays a Your Consoles tab and a Your Submissions tab. The call to get user groups does use a wildcard, but it should be OK since it's just a post-fix.

### DIFF
--- a/openreview/conference/templates/liteConferenceWebfield.js
+++ b/openreview/conference/templates/liteConferenceWebfield.js
@@ -130,11 +130,6 @@ function renderConferenceTabs() {
       heading: 'All Submissions',
       id: 'all-submissions',
     });
-  } else {
-    sections.push({
-      heading: 'Your Submissions',
-      id: 'user-submissions',
-    });
   }
 
   Webfield.ui.tabPanel(sections, {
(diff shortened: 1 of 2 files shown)


openreview/conference/templates/liteConferenceWebfield.js
```diff
@@ -0,0 +1,249 @@
...
+    id: 'your-consoles',
+  }];
+
+  if (PUBLIC) {
+    sections.push({
+      heading: 'All Submissions',
+      id: 'all-submissions',
+    });
+  } else {
+    sections.push({
```
melisabok, line 134: I don't think we should show 'Your submissions' tab, the author has the Author console link to see them.

zbialecki: Really? If we hide that the only thing they will see is a link to the Author Console, and some people won't even know what that is. I think it's kind of nice to be able to see the titles of the papers right on the conference homepage.

melisabok: this is no happening now in the live site. I'm worried they may think the submission is public because is listed in the home page.

zbialecki: OK, I see. Well I can either remove it, or I can put a message at the top of that tab saying that these submissions aren't public.

melisabok: I wouldn't show it, same as the live site now: it currently shows: 'Console' and 'Recent Activity' tabs. Removing the 'Recent Activity' is enough.

melisabok (approved):

PR #924, opened 2021-06-04
Exclude invited/declined groups from readers of own group

### LINKED ISSUE #903: Recruitment: members of declined/invited groups shouldn't be readers of the group
I noticed the group/invited is readers of their own group, meaning all the members of the invited group can see the group(all other members), I think we should remove them from the readers collection. It shouldn't affect any functionality.

celestemartinez:
Fixes #903

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -92,10 +92,10 @@ class Conference(object):
         self.area_chair_identity_readers = []
         self.senior_area_chair_identity_readers = []
 
-    def __create_group(self, group_id, group_owner_id, members = [], is_signatory = True, additional_readers = [], exclude_self=False):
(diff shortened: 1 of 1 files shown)


openreview/conference/builder.py
```diff
@@ -92,12 +92,13 @@ def __init__(self, client):
         self.area_chair_identity_readers = []
         self.senior_area_chair_identity_readers = []
 
-    def __create_group(self, group_id, group_owner_id, members = [], is_signatory = True, additional_readers = []):
+    def __create_group(self, group_id, group_owner_id, members = [], is_signatory = True, additional_readers = [], exclude_self=False):
```
melisabok, line 95: Can you call it `exclude_self_reader`? so it is clear we are not making the group.id reader of it self?

melisabok (approved):

PR #930, opened 2021-06-10
add submission fields to blind inv

### LINKED ISSUE #929: Blind submission: put all the submission content field in the invitation to the respect the order when the note is rendered
If the fields are not present in the Blind Submission invitation then the UI doesn't know which order to use to render the fields. 

Build the invitation with all the fields and use 'value-regex' for most of it.

celestemartinez:
Fixes #929

### DIFF
--- a/openreview/conference/invitation.py
+++ b/openreview/conference/invitation.py
@@ -76,18 +76,28 @@ class BlindSubmissionsInvitation(openreview.Invitation):
 
     def __init__(self, conference, hide_fields):
 
-        content = {
-            'authors': {
-                'values': ['Anonymous']
-            },
-            'authorids': {
-                'values-regex': '.*'
-            }
-        }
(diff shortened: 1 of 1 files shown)


openreview/conference/invitation.py
```diff
@@ -76,18 +76,28 @@ class BlindSubmissionsInvitation(openreview.Invitation):
...
+                content[key] = {
+                    'value-regex': '.*',
+                    'order': original_content[key]['order']
+                }
+
         for field in hide_fields:
-            content[field] = {
-                'value-regex': '.*'
-            }
+            if field not in content:
```
melisabok, line 97: do we keep hide_fields because they could have fields that are not part of the original content?

celestemartinez: yes, I believe the ECCV test was failing because of this.

melisabok (approved):

Issue #482, opened 2019-10-25
Get conflicts: automatically add a conflict edge where the reviewer is author of the paper. 


melisabok:
If the user doesn't have a profile and the only email address is 'gmail' then the conflict is not detected.

PR #933, opened 2021-06-16
Feature/reviewer specific conflicts

### LINKED ISSUE #595: Matching: add an option to compute conflicts for a specific reviewer
For ECCV 2020 there are reviewers that were added after we computed the conflicts. 

We need to make sure **ALL** the reviewers have conflict edges before running the paper matching.

haroldrubio:
Closes #595

I've added the function `set_matching_conflicts()` in the `Conference` object - which creates a `Matching` object, and calls the new public helper function `append_note_conflicts()`.
The `append_note_conflicts()` function is very similar to `_build_note_conflicts()`:

1. Sets up required objects that would have been created in a previous call to `setup()` - but the use case is that this is to be called after `setup()` was already performed
2. Adapts the user and invitation objects so that they are compatible with the conflict building loop from `_build_note_conflicts()`
3. Builds the conflicts for a single user
4. Removes existing + adds new conflict edges for the user and performs a sanity check

### DIFF
--- a/openreview/conference/matching.py
+++ b/openreview/conference/matching.py
@@ -251,6 +251,12 @@ class Matching(object):
             return self._build_profile_conflicts(ac_profiles, user_profiles)
         return self._build_note_conflicts(submissions, user_profiles, get_profile_info)
 
+    def _append_note_conflicts(self, submissions, user_profile, get_profile_info):
+        '''
+        Create conflict edges between the given Notes and a single profile
+        '''
+        pass
+
     def _build_note_conflicts(self, submissions, user_profiles, get_profile_info):
         '''
         Create conflict edges between the given Notes and Profiles
--- a/openreview/conference/matching.py
+++ b/openreview/conference/matching.py
@@ -251,11 +251,91 @@ class Matching(object):
             return self._build_profile_conflicts(ac_profiles, user_profiles)
         return self._build_note_conflicts(submissions, user_profiles, get_profile_info)
 
-    def _append_note_conflicts(self, submissions, user_profile, get_profile_info):
+    def append_note_conflicts(self, user_profile, build_conflicts=None):
         '''
         Create conflict edges between the given Notes and a single profile
         '''
-        pass
+
+        # Re-setup information that would have been initialized in setup()
+        submissions = list(openreview.tools.iterget_notes(
+            self.conference.client,
+            invitation=self.conference.get_blind_submission_id(),
+            details='original'))
+        get_profile_info = openreview.tools.get_neurips_profile_info if build_conflicts == 'neurips' else openreview.tools.get_profile_info
+        self.match_group = openreview.tools.replace_members_with_ids(self.client, self.match_group)
+        if not all(['~' in member for member in self.match_group.members]):
+            print(
+                'WARNING: not all reviewers have been converted to profile IDs.',
+                'Members without profiles will not have metadata created.')
+        
+        # Adapt single profile to multi-profile code
+        user_profiles = [user_profile]
+        user_profiles = _get_profiles(self.client, user_profiles, with_publications=build_conflicts)
+        user_profiles_info = [get_profile_info(p) for p in user_profiles]
+
+        # Fetch conflict invitation
+        try:
+            invitation = self.client.get_invitation(self.conference.get_conflict_score_id(self.match_group.id))
+        except:
+            print('Conflict invitation not found')
+
+        edges = []
+        
+        # Redo submission-author-user loop from _build_note_conflicts
+        for submission in tqdm(submissions, total=len(submissions), desc='_build_conflicts'):
+            # Get author profiles
+            authorids = submission.content['authorids']
+            if submission.details and submission.details.get('original'):
+                authorids = submission.details['original']['content']['authorids']
+
+            # Extract domains from each profile
+            author_profiles = _get_profiles(self.client, authorids, with_publications=True)
+            author_domains = set()
+            author_emails = set()
+            author_relations = set()
+            author_publications = set()
+
+            for author_profile in author_profiles:
+                author_info = get_profile_info(author_profile)
+                author_domains.update(author_info['domains'])
+                author_emails.update(author_info['emails'])
+                author_relations.update(author_info['relations'])
+                author_publications.update(author_info['publications'])
+
+            # Compute conflicts for the user and all the paper authors
+            for user_info in user_profiles_info:
+                conflicts = set()
+                conflicts.update(author_domains.intersection(user_info['domains']))
+                conflicts.update(author_relations.intersection(user_info['emails']))
+                conflicts.update(author_emails.intersection(user_info['relations']))
+                conflicts.update(author_emails.intersection(user_info['emails']))
+                conflicts.update(author_publications.intersection(user_info['publications']))
+
+                if conflicts:
+                    edges.append(openreview.Edge(
+                        invitation=invitation.id,
+                        head=submission.id,
+                        tail=user_info['id'],
+                        weight=-1,
+                        label='Conflict',
+                        readers=self._get_edge_readers(tail=user_info['id']),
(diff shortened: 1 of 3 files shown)


haroldrubio:
One thing I'm not sure about is - when a user is made into a reviewer after a large group of reviewers has already been set, would that new reviewer's affinity scores also be appended to this `.csv` file that's created:
https://github.com/openreview/openreview-py/blob/a38aa47608fa2599932a30108028f9f0570bea8b/tests/test_eccv_conference.py#L656-L662

Or would that be independent of creating conflict edges?

melisabok:
> One thing I'm not sure about is - when a user is made into a reviewer after a large group of reviewers has already been set, would that new reviewer's affinity scores also be appended to this .csv file that's created in parallel

To compute the scores for a single user, we need to compute the scores for all of them. For now, we don't have a way to incrementally recompute scores for new users. I think this is a good question for Haw-Shiuan.

openreview/conference/matching.py
```diff
@@ -251,6 +251,92 @@ def _build_conflicts(self, submissions, user_profiles, get_profile_info):
...
+        Create conflict edges between the given Notes and a single profile
+        '''
+
+        # Re-setup information that would have been initialized in setup()
+        submissions = list(openreview.tools.iterget_notes(
+            self.conference.client,
+            invitation=self.conference.get_blind_submission_id(),
+            details='original'))
+        get_profile_info = openreview.tools.get_neurips_profile_info if build_conflicts == 'neurips' else openreview.tools.get_profile_info
+        self.match_group = openreview.tools.replace_members_with_ids(self.client, self.match_group)
```
melisabok, line 265: You don't need to run this. It can be very expensive. I would make sure the profile_id is a valid profile.

openreview/conference/matching.py
```diff
@@ -251,6 +251,92 @@ def _build_conflicts(self, submissions, user_profiles, get_profile_info):
             return self._build_profile_conflicts(ac_profiles, user_profiles)
         return self._build_note_conflicts(submissions, user_profiles, get_profile_info)
 
+    def append_note_conflicts(self, user_profile, build_conflicts=None):
```
melisabok, line 254: Can you call it `profile_id` same as `set_matching_conflicts`?

```suggestion
    def append_note_conflicts(self, profile_id, build_conflicts=None):
```

openreview/conference/matching.py
```diff
@@ -251,6 +251,92 @@ def _build_conflicts(self, submissions, user_profiles, get_profile_info):
...
+        # Adapt single profile to multi-profile code
+        user_profiles = [user_profile]
+        user_profiles = _get_profiles(self.client, user_profiles, with_publications=build_conflicts)
+        user_profiles_info = [get_profile_info(p) for p in user_profiles]
+
+        # Fetch conflict invitation
+        try:
+            invitation = self.client.get_invitation(self.conference.get_conflict_score_id(self.match_group.id))
+        except:
+            print('Conflict invitation not found')
```
melisabok, line 280: can you raise an OpenReview Exception here?

openreview/conference/matching.py
```diff
@@ -251,6 +251,92 @@ def _build_conflicts(self, submissions, user_profiles, get_profile_info):
...
+        get_profile_info = openreview.tools.get_neurips_profile_info if build_conflicts == 'neurips' else openreview.tools.get_profile_info
+        self.match_group = openreview.tools.replace_members_with_ids(self.client, self.match_group)
+        if not all(['~' in member for member in self.match_group.members]):
+            print(
+                'WARNING: not all reviewers have been converted to profile IDs.',
+                'Members without profiles will not have metadata created.')
+        
+        # Adapt single profile to multi-profile code
+        user_profiles = [user_profile]
+        user_profiles = _get_profiles(self.client, user_profiles, with_publications=build_conflicts)
```
melisabok, line 273: `_get_profiles` returns a dummy profile when the profile_id doesn't exist. I think we should throw an error if there is no profile found. 

We could compute the conflicts associated with emails instead of profile id but it can be very problematic for later. I would prefer to always compute the conflicts against existing OpenReview profiles.

melisabok (changes requested):
This PR looks really good! Thanks, this will be very helpful to compute conflicts for reviewers that join the committee late. I request a few small changes.

openreview/conference/matching.py
haroldrubio: I was also thinking of incorporating a solution to this issue in this pull request since they seem related:
https://github.com/openreview/openreview-py/issues/482. If I were to detect:
1. No OpenReview profile detected
2. Only using a Gmail account

It would be at this step, right? If those are the two things that the issue is pointing to. Would throwing an exception if the user does not have an OpenReview profile be a reasonable solution to https://github.com/openreview/openreview-py/issues/482? I would also have to add that same check into the `_build_X_conflicts()` functions as well.

openreview/conference/matching.py
```diff
@@ -262,7 +262,6 @@ def append_note_conflicts(self, profile_id, build_conflicts=None):
             invitation=self.conference.get_blind_submission_id(),
             details='original'))
         get_profile_info = openreview.tools.get_neurips_profile_info if build_conflicts == 'neurips' else openreview.tools.get_profile_info
-        self.match_group = openreview.tools.replace_members_with_ids(self.client, self.match_group)
         if not all(['~' in member for member in self.match_group.members]):
```
melisabok, line 265: you can remove this if statement too

openreview/conference/matching.py
```diff
@@ -251,7 +251,7 @@ def _build_conflicts(self, submissions, user_profiles, get_profile_info):
             return self._build_profile_conflicts(ac_profiles, user_profiles)
         return self._build_note_conflicts(submissions, user_profiles, get_profile_info)
 
-    def append_note_conflicts(self, user_profile, build_conflicts=None):
+    def append_note_conflicts(self, profile_id, build_conflicts=None):
```
melisabok, line 254: it seems that you need to replace user_profile with profile_id when you _get_profiles.

openreview/conference/matching.py
```diff
@@ -251,6 +251,91 @@ def _build_conflicts(self, submissions, user_profiles, get_profile_info):
...
+            self.conference.client,
+            invitation=self.conference.get_blind_submission_id(),
+            details='original'))
+        get_profile_info = openreview.tools.get_neurips_profile_info if build_conflicts == 'neurips' else openreview.tools.get_profile_info
+        
+        # Adapt single profile to multi-profile code
+        user_profiles = [profile_id]
+        user_profiles = _get_profiles(self.client, user_profiles, with_publications=build_conflicts)
+        # Check for existing OpenReview profile - perform dummy check
+        if user_profiles[0].active == None:
```
melisabok, line 270: can you add this validation at the beginning of the function? so we don't nee to load all the submissions if the profile doesn't exists.

tests/test_eccv_conference.py
```diff
@@ -676,6 +676,12 @@ def test_recommend_reviewers(self, conference, test_client, helpers, selenium, r
             tpms_score_file=os.path.join(os.path.dirname(__file__), 'data/temp.csv')
         )
 
+        # Test adding reviewer after conflicts are built
+        r5_client = helpers.create_user('reviewer5@fb.com', 'Reviewer', 'ECCV Five')
+        conference.set_reviewers(['~Reviewer_ECCV_One1', '~Reviewer_ECCV_Two1', '~Reviewer_ECCV_Three1', '~Reviewer_ECCV_Four1', '~Reviewer_ECCV_Five1'])
+        assert r5_client.get_edges_count() == 0
+        conference.set_matching_conflicts(r5_client.profile.id)
```
melisabok, line 683: can you add a call with a profile that doesn't exist or it is not active and test that you expect the exception?

melisabok (approved):

melisabok:
@celestemartinez please take a look.

PR #934, opened 2021-06-16
Enable forum tabs in NeurIPS tests

zbialecki:
Add config to Blind_Submission invitation to enable forum tabs.

### DIFF
--- a/tests/test_neurips_conference.py
+++ b/tests/test_neurips_conference.py
@@ -1486,13 +1486,33 @@ OpenReview Team'''
 
     def test_review_stage(self, conference, helpers, test_client, client):
 
-
         ## make submissions visible to assigned commmittee
         invitation = client.get_invitation('NeurIPS.cc/2021/Conference/-/Submission')
         invitation.reply['readers'] = {
             'values-regex': '.*'
         }
         client.post_invitation(invitation)
+
+        invitation2 = client.get_invitation('NeurIPS.cc/2021/Conference/-/Blind_Submission')
+        invitation2.replyForumViews = '''[
+            {
+                "id": "authors",
+                "label": "Author Discussion",
+                "filter": "readers:NeurIPS.cc/2021/Conference/Paper${note.number}/Authors"
+            },
+            {
+                "id": "committee",
+                "label": "Committee Discussion",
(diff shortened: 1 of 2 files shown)


zbialecki:
I there a way we could add a review rebuttal stage to these tests? I would like to check that the forum works with authors commenting.

melisabok:
Yes, I can add this to the PR.

tests/test_neurips_conference.py
```diff
@@ -1515,13 +1515,33 @@ def test_deployment_stage(self, conference, client, helpers):
...
-
         ## make submissions visible to assigned commmittee
         invitation = client.get_invitation('NeurIPS.cc/2021/Conference/-/Submission')
         invitation.reply['readers'] = {
             'values-regex': '.*'
         }
         client.post_invitation(invitation)
+
+        invitation2 = client.get_invitation('NeurIPS.cc/2021/Conference/-/Blind_Submission')
+        invitation2.replyForumViews = '''[
```
melisabok, line 1526: how did you make this work? do you see the tabs in the forum page?

zbialecki: Thanks for fixing this. I added this configuration using the invitation editor as well so I didn't notice this wasn't working.

melisabok:
I logged in as an author (test@mail) and I can see the 'committee discussion' that I think the author shouldn't see. When I click on the tab I see a Review note where authors is reader. 

Screen Shot 2021-06-18 at 4 32 56 PM

melisabok:
It seems that after a note is posted, no tab is selected, is this the expected behavior? 

I think if the posted note readers match with the current tab then we should keep that tab selected otherwise switch to the 'all' tab.

zbialecki:
It's difficult to know what tab should be selected in certain cases. For example, if the user is on Author Discussion and changes the reader filter, should we stay on the same tab, or should no tab be selected now? I think that the tab should only be selected if the filters specified by the tab match exactly with what's selected on the page.

I agree with you about posting a note - if the new note matches the selected filters they should remain in place, but if it would be hidden given the selected filters then the should reset so the posted note is visible.

zbialecki:
These comments will be addressed by the openreview-web PR though, so I think it's OK to merge.

melisabok (approved):

PR #936, opened 2021-06-17
Fix/neurips sac fixes

melisabok:


melisabok:
These are some fixes for NeurIPS so I will merge these after the build is finished.

PR #954, opened 2021-07-15
pc console show count of query results

xkopenreview:
add total records count
should fix https://github.com/openreview/openreview-web/issues/582


zbialecki:
I like the idea of showing the count in the Export button, but could you attach a screenshot of how it looks?

xkopenreview:
> I like the idea of showing the count in the Export button, but could you attach a screenshot of how it looks?

the change has been made to the pc console of neurips, so you can use neurips pc console to see it

melisabok (approved):

PR #956, opened 2021-07-21
Fixed invitation pattern for reviews

joaquinvanschoren:
Reviews are apparently retrieved using the pattern:
ICLR.cc/2019/Conference/Paper.*/-/Official_Review

The documentation used to say
ICLR.cc/2019/Conference/-/Paper.*/Official_Review

which doesn't return anything.

There may be a similar issue for meta-reviews and comments.


joaquinvanschoren:
Closing because this template seems to be only loosely defined and depends on the conference.

PR #957, opened 2021-07-21
Add matching stage to request form

celestemartinez:
- Add Matching Stage to request form
- Return status of reviewers with no profiles to PCs
- Allow for csv upload of scores


openreview/venue_request/process/deployProcess.py
```diff
@@ -320,3 +323,63 @@ def process(client, note, invitation):
         },
         signatures=['~Super_User1']
     ))
+
+    if (forum.content.get('Submission Deadline', None) and forum.content.get('Paper Matching', None)):
```
melisabok, line 327: `None` is already the default value for get, no need to define it as a parameter.

openreview/venue_request/process/deployProcess.py
```diff
@@ -320,3 +323,63 @@ def process(client, note, invitation):
...
+                'readers' : {
+                    'description': 'The users who will be allowed to read the above content.',
+                    'values' : readers
+                },
+                'writers': {
+                    'values':[],
+                },
+                'content': {
+                    'title': {
+                        'value': 'Matcher',
```
melisabok, line 350: What about `Paper Matching initialization`? or `Paper Matching setup`?

openreview/venue_request/process/deployProcess.py
```diff
@@ -320,3 +323,63 @@ def process(client, note, invitation):
...
+                        'required': True,
+                        'order': 1
+                    },
+                    'matching_group': {
+                        'description': 'Please select the group you want to set up matching for.',
+                        'value-radio': ['reviewer'],
```
melisabok, line 356: Can you define the roles directly here?

```
'value-radio': ['reviewer', 'area chair'] if forum.content.get('Area Chairs (Metareviewers)') == "Yes, our venue has Area Chairs" else ['reviewer']
```

What about Senior Area Chairs? I think the setup matching works for that? I'm not sure if conflicts and scores are supported.

Also I would like to start using the name of the group so we don't need to do any translation:

VENUE_ID/Reviewers
VENUE_ID/Area_Chairs

if you use value-dropdown the options are formatted with prettyId so it would say 'NeurIPS 2021 Conference Reviewers'.

If you think that is too large the option then we can remove the venue id and add it later.

openreview/venue_request/process/matchingProcess.py
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    matching_group = conference.get_id() + '/' + roles[note.content['matching_group'].strip()]
+    scores = note.content.get('affinity_scores', None)
+    file_name=None
+
+    if scores:
+        scores_attachment = client.get_attachment(id=note.id, field_name='affinity_scores')
+        with open('affinity_scores.csv','wb') as file:
+            file.write(scores_attachment)
```
melisabok, line 22: I would prefer to avoid writing the result into a file. I think the setup matching can read the content as a stream instead of a physical file.

openreview/conference/matching.py
```diff
@@ -929,7 +933,50 @@ def setup(self, affinity_score_file=None, tpms_score_file=None, elmo_score_file=
         if build_conflicts:
             self._build_conflicts(submissions, user_profiles, openreview.tools.get_neurips_profile_info if build_conflicts == 'neurips' else openreview.tools.get_profile_info)
 
+        if scores_stream:
```
celestemartinez, line 936: this code is very similar to the code inside `_build_note_scores()`

melisabok: I would re-define two functions:

_build_note_scores_from_file
_build_note_scroes_from_stream

and define a function _build_note_scores(self, score_invitation_id, scores, submissions)

where scores is an array of scores than can be built from a file or from a stream:

```
scores = [input_line for input_line in scores_stream.decode().split('\n')]
```

or 

```
with open(score_file) as file_handle:
      scores = [row for row in csv.reader(file_handle)]
```

openreview/conference/matching.py
```diff
@@ -605,6 +613,26 @@ def _build_subject_area_scores(self, submissions):
             raise openreview.OpenReviewException('Failed during bulk post of edges! Scores found: {0}, Edges posted: {1}'.format(len(edges), edges_posted))
         return invitation
 
+    def _compute_scores(self, score_invitation_id, submissions, matching_status):
+        invitation = self._create_edge_invitation(score_invitation_id)
+
+        job_id = self.client.request_expertise(name=self.conference.get_short_name(), group_id=self.match_group.id, paper_invitation=self.conference.get_blind_submission_id(), model='specter+mfr')
```
melisabok, line 619: I think we need to add the expertise selection invitation to send to the API to ignore the excluded papers.

openreview/conference/matching.py
```diff
@@ -605,6 +613,26 @@ def _build_subject_area_scores(self, submissions):
...
+            if call_count == 30:
+                break
+            time.sleep(30)
+        if status:
+            result = self.client.get_expertise_results(job_id['job_id'])
+            scores = [[entry['submission'],entry['user'],entry['score']] for entry in result['results']]
+            matching_status['no_profiles'] = result['metadata']['no_profile']
+            matching_status['no_publications'] = result['metadata']['no_publications']
+            return self._build_note_scores(invitation, scores, submissions), matching_status
+        return None, matching_status
```
melisabok, line 634: can you add a try/except in case there is an error with the API and we should post a comment with the error.

openreview/venue_request/process/matchingProcess.py
```diff
@@ -0,0 +1,42 @@
...
+
+    build_conflicts=note.content.get('build_conflicts')
+
+    matching_group = note.content['matching_group']
+    compute_scores = note.content.get('compute_affinity_scores') == 'Yes'
+    scores = note.content.get('upload_affinity_scores')
+    file_name=None
+
+    if scores:
+        scores_stream = client.get_attachment(id=note.id, field_name='upload_affinity_scores')
```
melisabok, line 16: what happens if they says 'Yes' and also pass a file? which process wins? I think the API, right?

openreview/conference/builder.py
```diff
@@ -1137,14 +1137,14 @@ def set_impersonators(self, emails = []):
 
         return self.webfield_builder.set_impersonate_page(self, impersonate_group)
 
-    def setup_matching(self, committee_id=None, affinity_score_file=None, tpms_score_file=None, elmo_score_file=None, build_conflicts=None, alternate_matching_group=None):
+    def setup_matching(self, committee_id=None, affinity_score_file=None, tpms_score_file=None, elmo_score_file=None, build_conflicts=None, alternate_matching_group=None, scores_stream=None, compute_scores=False):
```
melisabok, line 1140: I would like to merge affinity_score_file, score_stream and compute_scores in a single parameter, could we use the type of the value to know that?

let's call it `compute_affinity_scores`:

- if `compute_affinity_scores` is String then we assume it is a file name and we should read the scores from the file.
- if `compute_affinity_scores` is a Stream then we read the scores from the stream.
- if `compute_affinity_scores` is  boolean and is True then we read the scores from the API.
- if `compute_affinity_scores` is boolean and is False then don't compute/read any scores. 


Does it work?

I know that removing the `affinity_score_file` parameter from the function can break a lot of things. I would suggest to mark this function as deprecated and create a new one:

`def setup_committee_matching(self, committee_id, compute_scores=False, compute_conflicts=False, alternate_matching_group)`

I'm open to other function names.

openreview/venue_request/process/matchingProcess.py
celestemartinez: maybe we can add this check to the preprocess function?

melisabok: Good idea!

openreview/conference/builder.py
```diff
@@ -1137,14 +1137,24 @@ def set_impersonators(self, emails = []):
...
         if committee_id is None:
             committee_id=self.get_reviewers_id()
         if self.use_senior_area_chairs and committee_id == self.get_senior_area_chairs_id() and not alternate_matching_group:
             alternate_matching_group = self.get_area_chairs_id()
         conference_matching = matching.Matching(self, self.client.get_group(committee_id), alternate_matching_group)
 
-        return conference_matching.setup(affinity_score_file, tpms_score_file, elmo_score_file, build_conflicts, scores_stream, compute_scores)
+        return conference_matching.setup(affinity_score_file, tpms_score_file, elmo_score_file, build_conflicts)
+
+    def setup_committee_matching(self, committee_id=None, compute_affinity_scores=False, build_conflicts=False, alternate_matching_group=None):
```
melisabok, line 1150:
```suggestion
    def setup_committee_matching(self, committee_id=None, compute_affinity_scores=False, compute_conflicts=False, alternate_matching_group=None):
```

openreview/venue_request/process/deployProcess.py
```diff
@@ -327,3 +329,68 @@ def process(client, note, invitation):
...
+        activation_date = forum.content.get('Submission Deadline').strip()
+        try:
+            activation_date = datetime.strptime(activation_date, '%Y/%m/%d %H:%M')
+        except ValueError:
+            activation_date = datetime.strptime(activation_date, '%Y/%m/%d')
+        matching_invitation = openreview.Invitation(
+            id = SUPPORT_GROUP + '/-/Request' + str(forum.number) + '/Matching_Stage',
+            super = SUPPORT_GROUP + '/-/Matching_Stage',
+            invitees = readers,
+            cdate = openreview.tools.datetime_millis(activation_date),
```
melisabok, line 343: do you update the cdate when the `Submission Deadline` is updated?

openreview/venue_request/process/deployProcess.py
```diff
@@ -327,3 +329,68 @@ def process(client, note, invitation):
...
+                        'required': True,
+                        'order': 1
+                    },
+                    'matching_group': {
+                        'description': 'Please select the group you want to set up matching for.',
+                        'value-dropdown' : [conference.get_id() + '/Reviewers', conference.get_id() + '/Area_Chairs'] if forum.content.get('Area Chairs (Metareviewers)') == "Yes, our venue has Area Chairs" else [conference.get_id() + '/Reviewers'],
+                        'required': True,
+                        'order': 2
+                    },
+                    'build_conflicts': {
```
melisabok, line 366:
```suggestion
                    'compute_conflicts': {
```

I would like to call it compute instead of build, it is more self descriptive

openreview/venue_request/process/deployProcess.py
```diff
@@ -327,3 +329,68 @@ def process(client, note, invitation):
...
+                        'description': 'Please select whether you want to compute conflicts of interest between the matching group and submissions. By default, conflicts will be computed.',
+                        'value-radio': ['Yes', 'No'],
+                        'default': 'Yes',
+                        'required': True,
+                        'order': 3
+                    },
+                    'compute_affinity_scores': {
+                        'description': 'Please select whether you would like affinity scores to be computed and uploaded automatically.',
```
melisabok, line 374:
```suggestion
                        'description': 'Please select whether you would like affinity scores to be computed by our expertise API and uploaded automatically.',
```

openreview/venue_request/process/deployProcess.py
```diff
@@ -327,3 +329,68 @@ def process(client, note, invitation):
...
+                },
+                'content': {
+                    'title': {
+                        'value': 'Paper Matching Setup',
+                        'required': True,
+                        'order': 1
+                    },
+                    'matching_group': {
+                        'description': 'Please select the group you want to set up matching for.',
+                        'value-dropdown' : [conference.get_id() + '/Reviewers', conference.get_id() + '/Area_Chairs'] if forum.content.get('Area Chairs (Metareviewers)') == "Yes, our venue has Area Chairs" else [conference.get_id() + '/Reviewers'],
```
melisabok, line 362: can we use the same role names that we use in the recruitment form so it is consistent?

celestemartinez: Ok, so we don't want to use the group names as suggested here: https://github.com/openreview/openreview-py/pull/957#discussion_r682810374?

melisabok: Haha I was about to remove this comment, you caught me! ignore this one. I will delete it.

celestemartinez:
The status comment looks like this now:

if all users have profiles:
> Affinity scores and/or conflicts were successfully computed. To run the matcher, click on the 'Reviewers Paper Assignment' link in the PC console: https://openreview.net/group/edit?id=TEST.cc/2030/Conference/Program_Chairs
> 
>  Please refer to the FAQ for pointers on how to run the matcher: https://openreview.net/faq#question-edge-browswer

if there are users with no profiles:
> 1 Reviewers without a profile: ['some_user@mail.com']
> Affinity scores and/or conflicts could not be computed for these users. Please ask these users to sign up in OpenReview and upload their papers. 
> 
> Alternatively, you can remove these users from the Reviewers group.
> 
> Please check the Reviewers group to see more details: https://openreview.net/group?id=TEST.cc/2030/Conference/Reviewers

Note that either option can contain a status on profiles with no publications.

openreview/venue_request/process/matchingProcess.py
```diff
@@ -0,0 +1,61 @@
...
+        no_profiles_status = matching_status.get('no_profiles')
+        if no_profiles_status:
+            profiles_status=f'''
+{len(no_profiles_status)} {role_name} without a profile: {no_profiles_status}
+
+Affinity scores and/or conflicts could not be computed for these users. Please ask these users to sign up in OpenReview and upload their papers. Alternatively, you can remove these users from the {role_name} group. 
+
+Please check the {role_name} group to see more details: https://openreview.net/group?id={matching_group}'''
+        else:
+            profiles_status=f'''Affinity scores and/or conflicts were successfully computed. To run the matcher, click on the '{role_name} Paper Assignment' link in the PC console: https://openreview.net/group/edit?id={conference.get_program_chairs_id()}
```
melisabok, line 49:
```suggestion
            profiles_status=f'''Affinity scores and/or conflicts were successfully computed. To run the matcher, click on the '{role_name} Paper Assignment' link in the PC console: https://openreview.net/group?id={conference.get_program_chairs_id()}
```

openreview/conference/matching.py
```diff
@@ -890,26 +932,52 @@ def setup(self, affinity_score_file=None, tpms_score_file=None, elmo_score_file=
             invitation=self.conference.get_blind_submission_id(),
             details='original'))
 
+        if not self.match_group.members:
+            raise openreview.OpenReviewException(f'The match group is empty: {self.match_group.id}')
+        if self.alternate_matching_group:
+            other_matching_group = self.client.get_group(self.alternate_matching_group)
+            if not other_matching_group.members:
+                raise openreview.OpenReviewException(f'The alternate match group is empty: {self.alternate_matching_group}')
+        elif not submissions:
```
celestemartinez, line 941: I don't think we can remove this elif, because we only want to check if there are submissions if there is no `alternate_matching_group`.

melisabok: You are right! ignore my comment.

melisabok (approved):
💯

PR #959, opened 2021-08-02
Fix/ Update username 'Test'

xkopenreview:
'test' or some name that includes 'test' is an invalid username
so all "test" name in the tests need to be updated to something else


melisabok (approved):
Thanks!

PR #960, opened 2021-08-04
Keep blind papers blind when adding revision after decisions

celestemartinez:
When the submission revision is enabled for accepted papers only after decisions are posted (but before author names are revealed) and we add 'venue' and `venueid` to the content of the blind submission, we stop hiding author names. 

Now, I check if submission is still blind and then add `Anonymous` and the author group as `authors` and `authorids` to the blind copy's content.

### DIFF
--- a/openreview/conference/templates/submission_revision_process.py
+++ b/openreview/conference/templates/submission_revision_process.py
@@ -38,17 +38,22 @@ To view your submission, click here: https://openreview.net/forum?id={}'''.forma
         notes=client.get_notes(original=forum.id)
         if notes:
             bibtex_note=notes[0]
+            released_note = bibtex_note.content.get('authors') != ['Anonymous']
             bibtex_note.content = {
                 'venue': bibtex_note.content.get('venue'),
                 'venueid': bibtex_note.content.get('venueid')
             }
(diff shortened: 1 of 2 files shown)


openreview/conference/templates/submission_revision_process.py
```diff
@@ -38,17 +38,22 @@ def process_update(client, note, invitation, existing_note):
...
         if notes:
             bibtex_note=notes[0]
+            released_note = bibtex_note.content.get('authors') != ['Anonymous']
             bibtex_note.content = {
                 'venue': bibtex_note.content.get('venue'),
                 'venueid': bibtex_note.content.get('venueid')
             }
+            if not released_note:
+                bibtex_note.content['authors'] = ['Anonymous']
+                bibtex_note.content['authorids'] = CONFERENCE_ID + '/Paper' + str(bibtex_note.number) + '/' + AUTHORS_NAME
```
celestemartinez, line 48: I was hesitant to use `author_group` here because of the statement 
`if note.content.get('authorids'):`

Is there any circumstance where `authorids` might not be in the content of a note?

melisabok (approved):

PR #968, opened 2021-08-18
Fix homepage tab names

celestemartinez:
When PCs use their own decision options, the dictionary of homepage tab names is not being built correctly.

### DIFF
--- a/openreview/venue_request/process/revisionProcess.py
+++ b/openreview/venue_request/process/revisionProcess.py
@@ -53,11 +53,15 @@ def process(client, note, invitation):
                     'No, I don\'t want to reveal any author identities.'],
                 'required': True
             }
-
+        decision_options = forum_note.content.get('decision_options')
+        if decision_options:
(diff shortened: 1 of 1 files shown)


openreview/venue_request/process/revisionProcess.py
```diff
@@ -53,11 +53,15 @@ def process(client, note, invitation):
                     'No, I don\'t want to reveal any author identities.'],
                 'required': True
             }
-
+        decision_options = forum_note.content.get('decision_options')
+        if decision_options:
+            decision_options = [s.translate(str.maketrans('', '', '"\'')).strip() for s in decision_options.split(',')]
+        else:
+            decision_options = ['Accept (Oral)', 'Accept (Poster)', 'Reject']
```
melisabok, line 60: Can you add a test here: https://github.com/openreview/openreview-py/blob/fix/homepage_tabs/tests/test_venue_request.py#L861

and verify the Post_Decision_Stage invitation default value is correctly computed?

PR #974, opened 2021-08-26
Fix/ Update email tests

zbialecki:
Update all email tests to check for existence of certain phrases, rather than an exact string match.

To find tests that check the message content search for `['content']['text']` in the repo.


zbialecki:
I think some of the tests are still failing because of that indentation issue. That would be my guess as least.

melisabok:
They are failing because this branch is not in master yet: https://github.com/openreview/openreview/pull/2457

I run them in my localhost and they all pass.

melisabok:
> I tried fixing the tests for test_arr_venue.py, but I'm not sure how to assert the converted messages to HTML. Maybe one of you knows how to assert it

I copy what the /messages endpoint shows and pasted it in the assert :) I can fix this.

PR #976, opened 2021-09-08
Fix/replace month name

xkopenreview:
calendar.month_abbr has Sep for september but in ui it shows Sept so ln 1683 should have a key error

### DIFF
--- a/tests/test_eccv_conference.py
+++ b/tests/test_eccv_conference.py
@@ -1680,7 +1680,7 @@ thecvf.com/ECCV/2020/Conference/Reviewers/-/Bid'
             due_date_list[2] = due_date_list[2][:-1]
             due_date_list.pop()
             date_obj = datetime.datetime(year=int(due_date_list[2]),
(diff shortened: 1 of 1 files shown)


tests/test_eccv_conference.py
```diff
@@ -1680,7 +1680,7 @@ def convert_text_to_datetime(text):
             due_date_list[2] = due_date_list[2][:-1]
             due_date_list.pop()
             date_obj = datetime.datetime(year=int(due_date_list[2]),
-                                         month=mnt_to_idx[due_date_list[1]],
+                                         month=mnt_to_idx[due_date_list[1].replace('sept','sep')],
```
melisabok, line 1683: I don't know why this test is not failing in CircleCI

xkopenreview:
eccv pc console is doing 
date.toLocaleDateString('en-GB',{...}...

which should output Sept
image

zbialecki:
This test is failing on my machine also.

melisabok (approved):

PR #977, opened 2021-09-09
invite assignments instead of deploying them

melisabok:
These changes are intended to be used by ARR but it may apply to other venues:

New method `set_invite_assignments` that instead of moving the proposed assignments to group members and assignments edges:

1. Creates Invite_Assignments edges. 
2. Show the edge browser url in the AC console.
3. Mark the configuration note as Deployed so the PCs can monitor the assingments.

### DIFF
--- a/openreview/conference/templates/invite_assignment_post_process.py
+++ b/openreview/conference/templates/invite_assignment_post_process.py
@@ -10,6 +10,8 @@ def process(client, edge, invitation):
     INVITED_LABEL = ''
     INVITE_LABEL = ''
     EMAIL_TEMPLATE = ''
+    IS_REVIEWER = 'Reviewers' in ASSIGNMENT_INVITATION_ID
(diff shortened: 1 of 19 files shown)


openreview/conference/templates/invite_assignment_post_process.py
```diff
@@ -50,7 +52,7 @@ def process(client, edge, invitation):
         decline_url = url + "No"
 
         # format the message defined above
-        subject=f'[{SHORT_PHRASE}] Invitation to review paper titled {submission.content["title"]}'
+        subject=f'[{SHORT_PHRASE}] Invitation {ACTION_STRING} paper titled {submission.content["title"]}'
         if EMAIL_TEMPLATE:
             message=EMAIL_TEMPLATE.format(
                 title=submission.content['title'],
```
celestemartinez, line 58:
```suggestion
                title=submission.content['title'],
                number=submission.number,
```

celestemartinez (approved):

PR #978, opened 2021-09-09
Change review rating readers to include signature

celestemartinez:
Sometimes papers may have multiple assigned ACs, and we don't want an AC of a paper to see the reviewer feedback posted by another AC.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -1837,7 +1837,7 @@ class ReviewRatingStage(object):
         readers = [ conference.get_program_chairs_id()]
 
         if conference.use_area_chairs:
-            readers.append(conference.get_area_chairs_id(number = number))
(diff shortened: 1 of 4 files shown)


openreview/conference/builder.py
```diff
@@ -1837,7 +1837,7 @@ def get_readers(self, conference, number, review_signature):
         readers = [ conference.get_program_chairs_id()]
 
         if conference.use_area_chairs:
```
celestemartinez, line 1839: Not sure why we check if conference uses ACs, since ACs are always the invitees of these invitations

melisabok: Maybe ReviewRating can be used for authors too in venues where there are no ACs. I would leave this for now.

melisabok (approved):

PR #982, opened 2021-09-20
setup matching for groups vs groups

melisabok:
Setup matching when we want to match people vs people, for example: SAC vs AC or Mentors vs Mentees. 

@carlosmondra you can use this to setup the matching for Mentors and Mentees:

```
conference.setup_matching(committee_id=conference.id + '/Reviewers_Mentors',
affinity_score_file=os.path.join(os.path.dirname(__file__), 'data/mentors_affinity_scores.csv'),
alternate_matching_group=conference.id + '/Reviewers_Mentees')
```


carlosmondra (approved):
It works!

PR #986, opened 2021-09-21
Feature/parametrize meta review stage name

carlosmondra:
I need this for ICLR. It worked in my python notebook but I'm not sure if this is OK or if there are other places where this needs to be changed.

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -1925,10 +1925,10 @@ class CommentStage(object):
 
 class MetaReviewStage(object):
 
-    def __init__(self, start_date = None, due_date = None, public = False, additional_fields = {}, remove_fields=[], process = None):
+    def __init__(self, name='Meta_Review', start_date = None, due_date = None, public = False, additional_fields = {}, remove_fields=[], process = None):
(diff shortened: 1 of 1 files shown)


openreview/conference/builder.py
```diff
@@ -1925,10 +1925,10 @@ def get_invitees(self, conference, number):
 
 class MetaReviewStage(object):
 
-    def __init__(self, start_date = None, due_date = None, public = False, additional_fields = {}, remove_fields=[], process = None):
+    def __init__(self, start_date = None, due_date = None, public = False, additional_fields = {}, remove_fields=[], process = None, name='Meta_Review'):
```
melisabok, line 1928:
```suggestion
    def __init__(self, name='Meta_Review', start_date = None, due_date = None, public = False, additional_fields = {}, remove_fields=[], process = None):
```

carlosmondra: I tried it like this, but some tests don't pass with this change. I'm not sure why.

celestemartinez (approved):

melisabok (approved):

PR #987, opened 2021-09-22
Add parametrization of readers for metareviews

celestemartinez:
Many PCs ask for metareviews to be released to authors and/or reviewers, so I am adding this to the metareview stage as well as to the request form.


celestemartinez:
- I wonder if we should have an option for which reviewers can see the metareview (or only have one option to release to all assigned reviewers)
- Should we add a process function to metareviews? I don't think there is one at the moment

melisabok:
> I wonder if we should have an option for which reviewers can see the metareview (or only have one option to release to all assigned reviewers)

Assigned reviewers for now it is ok.

> Should we add a process function to metareviews? I don't think there is one at the moment

I'm not sure. NeurIPS explicitly asked to add the reviewers as readers but don't message them. I wouldn't add a process function for now.

melisabok (approved):
💯

carlosmondra (approved):

PR #992, opened 2021-09-28
Fix/ add decision as query parameter in pc console

xkopenreview:
'No Decision' is a fallback value rendered when the actual value is undefined(not found)
set it as the string value "No Decision" so that query filter can compare it the user input

### DIFF
--- a/openreview/conference/templates/programchairWebfield.js
+++ b/openreview/conference/templates/programchairWebfield.js
@@ -57,7 +57,7 @@ var propertiesAllowed ={
     confidenceMax:['reviewProgressData.maxConfidence'],
     confidenceMin:['reviewProgressData.minConfidence'],
     replyCount:['reviewProgressData.forumReplyCount'],
-    decision: ['decision'],
+    decision: ['decision.content.decision'],
 }
 
 // Page State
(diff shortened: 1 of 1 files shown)


openreview/conference/templates/programchairWebfield.js
```diff
@@ -847,7 +848,7 @@ var calcMetaReviewCount = function(blindedNotes) {
 var calcDecisions = function(blindedNotes) {
   var finalDecisions = [];
   blindedNotes.forEach(function(note) {
-    if (note.details.decision) {
+    if (note.details.decision && note.details.decision !== 'No Decision') {
```
melisabok, line 851: I'm not sure when `note.details.decision` is a string

xkopenreview: when there's no decision i set it to the string 'No Decision'

melisabok: Could you instead create a decision object with 'No Decision'?

and there is no need to change openreview-web:

`var paperDecisions = _.find(submission.details.directReplies, ['invitation', getInvitationId(DECISION_NAME, submission.number)]) ?? [{ content: 'No Decision' }]`

melisabok: I would prefer to avoid that otherwise you need to check that everywhere.

xkopenreview: i think there was some difficulty with sorting when defaulting to object here but i can try again

melisabok: I don't think you need the second condition anymore, right?

melisabok, line 851:
```suggestion
    if (note.details.decision) {
```

melisabok:
It looks good! I just made a suggested change. Please take a look.

openreview/conference/templates/programchairWebfield.js
xkopenreview: thank it's reading a wrong property here but we can't remove this
removing this condition will cause the decision progress data in overview tab to be wrong where "No Decision" is considered as a final decision and the decision progress data will always be 100%

melisabok (approved):

PR #994, opened 2021-09-30
Fix/Allow export of PC console AC status tab

xkopenreview:
add export button to AC status tab of PC console

added a variable to store data that is displayed in table to avoid rebuilding the data during export


xkopenreview:
it looks like that the export paper and export reviews csv button in all tabs are rebuilding the table data.
how about changing that to read from tableDataInDisplay? @melisabok @celestemartinez

melisabok:
Yeah, I'm aware of that, I'm making it better for the new API. Feel free to change it if you want but you don't need to.

melisabok (approved):

PR #988, opened 2021-09-23
added venue name to subject of new venue email

nlbahy:
Fixes #971

PR #999, opened 2021-10-06
added venue name to subj line of email

nlbahy:
#988 did not actually add the venue to the subject line, it added it to the body of the email.


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "terminal.integrated.env.osx": { "PATH": "" },
+    "terminal.integrated.defaultProfile.osx": "zsh"
+  }
```
carlosmondra, line 4: I think this does not need to be in the PR :)

PR #1007, opened 2021-10-14
Add max note count as option for the bid stage

carlosmondra:


melisabok:
You could the max_count in the venue request form so the PCs can decide if they want to limit it or not:

https://github.com/openreview/openreview-py/blob/master/openreview/venue_request/venue_request.py#L63

carlosmondra:
I realized that this does not work and I have not had the time to really fix it.

melisabok:
Ok, I can assign it to somewhere else. 

Sometimes venues ask for this.

melisabok:
@carlosmondra , can @xkopenreview add this feature to the new react bidding console? should we wait until we have a first version and add this later?

carlosmondra:
> @carlosmondra , can @xkopenreview add this feature to the new react bidding console? should we wait until we have a first version and add this later?

I think it's fine if he can add it. I realized that my approach in this PR was wrong, he would need to fix that first. If it's too troublesome, maybe we can wait until later to implement it.

xkopenreview:
what's the rational of limiting the number of papers retrieved?

melisabok:
>what's the rationale of limiting the number of papers retrieved?

Some PCs don't want reviewers to see more than the top N papers. They haven't request this lately so I suggest to close this PR, agree?

carlosmondra:
Yes, we need to rethink the approach anyway.

PR #1012, opened 2021-10-19
Add helper function that uses concurrency for requests

carlosmondra:
I thought that this could be helpful. Let me know what you think.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -16,12 +16,25 @@ import urllib.parse as urlparse
 from tqdm import tqdm
 from concurrent.futures import ThreadPoolExecutor
 
-def concurrent_requests(request_func, params):
+def concurrent_requests(request_func, params, max_workers=6):
+    """
+    Returns a list of results given for each request_func param execution. It shows a progress bar to know the progress of the task.
+
+    :param request_func: a function to execute for each value of the list.
+    :type request_func: function
+    :param params: a list of values to be executed by request_func.
+    :type params: list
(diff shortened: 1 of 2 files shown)


melisabok:
can you give me an example of how to use it? for example, can I pass a whole note object to the function? can I share data with the function without passing it as a parameter?

carlosmondra:
> can you give me an example of how to use it? for example, can I pass a whole note object to the function? can I share data with the function without passing it as a parameter?

There is an example in the PR itself, but here's another one

```py
mentee_assignments = list(openreview.tools.iterget_edges(client, invitation='ICLR.cc/2022/Conference/Reviewer_Mentees/-/Assignment'))
reviewers_group_id = 'ICLR.cc/2022/Conference/Reviewers'

def assign_reviewer(assignment):
    note_id = assignment.head
    profile_id = assignment.tail
    note_number = client.get_note(note_id).number
    paper_group = f'ICLR.cc/2022/Conference/Paper{note_number}/Reviewers'
    client.add_members_to_group(paper_group, [profile_id])
    client.add_members_to_group(reviewers_group_id, [profile_id])


a = openreview.tools.concurrent_requests(assign_reviewer, mentee_assignments)   
```

you can share data with the function even if it's not passed to the function like it's shown here I'm not passing `reviewers_group_id`, but I can still read it. It's the same with `client`. In this example the function takes an element of the list at a time. Each element is an edge object. But the array can have anything as long as the function you defined handles it. It's very similar to `.map` in JS. Let me know if I didn't answer your questions

melisabok (approved):

PR #1022, opened 2021-11-10
Fix/update accepted authors group

nlbahy:


openreview/conference/templates/decisionProcess.js
```diff
@@ -1,4 +1,4 @@
-function (){
+function processUpdate(){
```
melisabok, line 1: This was an easy fix!

melisabok: I would change the verb in the email based on the action: posted when the decision is created and updated when the decision was updated.

melisabok: I'm realizing this change affect the author notification. If the PCs decides to release the decisions after they are posted then an email will be sent to each author if author notification is available? @celestemartinez do we want this? or should we send email only when the decision is created and no email when it is updated?

celestemartinez: I think we should always notify when then decision has been changed (and authors are readers) but not notify if all that is changing are the readers. Is this possible?

melisabok: I think we can do a diff between the current decision and new one and check if the change is only the readers of the note.

melisabok: You are right, we should only notify the authors when the decision value has changed.

nlbahy: What if only the comment is changed?

melisabok: yes, I think we should check if at least one field of the content has changed to send an email to the authors. That means if the only change is in the readers field then no email is sent.

melisabok: I would suggest to translate this process function to python, it will be easier to make the changes.

nlbahy: Okay I will get started on that now!

openreview/conference/templates/decision_process.py
```diff
@@ -0,0 +1,44 @@
...
+    ACCEPTED_AUTHORS_NAME = 'Authors/Accepted'
+    EMAIL_AUTHORS = False
+    baseUrl='https://openreview.net'
+
+    forum_note = client.get_note(note.forum)
+   
+    AUTHORS_ID = CONFERENCE_ID + '/Paper' + str(forum_note.number) + '/' + AUTHORS_NAME
+    
+    # Check if the content of the existing note is the same as the updated note
+    if (existing_note):
```
melisabok, line 15: no need to use parenthesis here.

openreview/conference/templates/decision_process.py
```diff
@@ -0,0 +1,44 @@
...
+      equal = True
+      for key in existing_note.content:
+        if existing_note.content[key] != note.content[key]:
```
melisabok, line 18: it seems the indentation is off here, are the process function working?

openreview/conference/templates/decision_process.py
```diff
@@ -0,0 +1,44 @@
...
+          equal = False
+      # If the content is not equal, want to send emails 
+      if(not equal):
+        if (EMAIL_AUTHORS and 'everyone' in note.readers) or (AUTHORS_ID in note.readers):
```
melisabok, line 22:
```suggestion
        if EMAIL_AUTHORS and ('everyone' in note.readers or AUTHORS_ID in note.readers):
```

melisabok: we should send the email only when EMAILS_AUTHORS is true.

openreview/conference/templates/decision_process.py
```diff
@@ -0,0 +1,44 @@
+def process_update(client, note, invitation, existing_note):
+
+    CONFERENCE_ID = ''
+    SHORT_PHRASE = ''
+    AUTHORS_NAME = ''
+    ACCEPTED_AUTHORS_NAME = 'Authors/Accepted'
```
melisabok, line 6: this is failing because the function replace matches AUTHOR_NAME here too. I would suggest to use another variable name `ACCEPTED_GROUP_NAME`.

openreview/conference/templates/decision_process.py
nlbahy: Yes, I believe it is working now even though the indentation does look shallow

openreview/conference/templates/decision_process.py
nlbahy: I resolved this in my most recent commit

openreview/conference/templates/decision_process.py
```diff
@@ -0,0 +1,66 @@
...
+    if existing_note:
+      equal = True
+      for key in existing_note.content:
+        if note.content[key] and (existing_note.content[key] != note.content[key]):
+          equal = False
+      # If the content is not equal, want to send emails 
+      if(not equal):
+        if (EMAIL_AUTHORS and ('everyone' in note.readers or AUTHORS_ID in note.readers)):
+          groups = [AUTHORS_ID]
+          subject = '{SHORT_PHRASE} Decision posted to your submission - Paper number: {paper_number}, Paper title: {paper_title}'.format(
```
celestemartinez, line 24: You should add square brackets around the venue short name to match how emails are formatted. 

Also, do we want to say that the decision has been updated when there is a change on the content, instead of posted?

melisabok (approved):

celestemartinez (approved):

PR #1026, opened 2021-11-12
Fix/sync ac vs sac

melisabok:
- Create Assignment edges when SAC are deployed so the PCs can make reassignments using the edge browser.
- Sync SAC group when a AC assignment changes.
- Sync SAC groups when a SAC assignment changes.

### DIFF
--- a/openreview/conference/matching.py
+++ b/openreview/conference/matching.py
@@ -1236,16 +1236,18 @@ class Matching(object):
                         sac_group.members.append(sac)
                         self.client.post_group(sac_group)
 
-                        assignment_edges.append(openreview.Edge(
-                            invitation=assignment_invitation_id,
-                            head=ac,
-                            tail=sac_assignment['tail'],
-                            readers=sac_assignment['readers'],
-                            nonreaders=sac_assignment['nonreaders'],
-                            writers=sac_assignment['writers'],
-                            signatures=sac_assignment['signatures'],
-                            weight=sac_assignment.get('weight')
-                        ))
+        for head, sac_assignments in proposed_assignment_edges.items():
+            for sac_assignment in sac_assignments:
+                assignment_edges.append(openreview.Edge(
+                    invitation=assignment_invitation_id,
+                    head=head,
+                    tail=sac_assignment['tail'],
+                    readers=sac_assignment['readers'],
+                    nonreaders=sac_assignment['nonreaders'],
+                    writers=sac_assignment['writers'],
+                    signatures=sac_assignment['signatures'],
+                    weight=sac_assignment.get('weight')
+                ))
 
         print('POsting assignments edges', len(assignment_edges))
         openreview.tools.post_bulk_edges(client=self.client, edges=assignment_edges)
--- a/openreview/conference/matching.py
+++ b/openreview/conference/matching.py
@@ -1330,7 +1330,7 @@ class Matching(object):
                     weight=sac_assignment.get('weight')
                 ))
 
-        print('POsting assignments edges', len(assignment_edges))
+        print('Posting assignments edges', len(assignment_edges))
         openreview.tools.post_bulk_edges(client=self.client, edges=assignment_edges)
 
 
--- a/openreview/conference/matching.py
+++ b/openreview/conference/matching.py
@@ -1204,8 +1204,10 @@ class Matching(object):
 
         papers = list(openreview.tools.iterget_notes(self.client, invitation=self.conference.get_blind_submission_id()))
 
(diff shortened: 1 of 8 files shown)


openreview/conference/matching.py
```diff
@@ -1315,8 +1317,21 @@ def deploy_sac_assignments(self, assignment_title, overwrite):
...
+                    head=head,
+                    tail=sac_assignment['tail'],
+                    readers=sac_assignment['readers'],
+                    nonreaders=sac_assignment['nonreaders'],
+                    writers=sac_assignment['writers'],
+                    signatures=sac_assignment['signatures'],
+                    weight=sac_assignment.get('weight')
+                ))
+
+        print('POsting assignments edges', len(assignment_edges))
```
carlosmondra, line 1333:
```suggestion
        print('Posting assignments edges', len(assignment_edges))
```

tests/test_neurips_conference.py
```diff
@@ -1020,6 +1024,74 @@ def test_setup_matching(self, conference, client, helpers):
...
+        ac_assignment.ddate = openreview.tools.datetime_millis(datetime.datetime.utcnow())
+        pc_client.post_edge(ac_assignment)
+
+        helpers.await_queue()
+
+        assert len(pc_client.get_edges(invitation='NeurIPS.cc/2021/Conference/Senior_Area_Chairs/-/Assignment')) == 3
+        assert len(pc_client.get_edges(invitation='NeurIPS.cc/2021/Conference/Area_Chairs/-/Assignment')) == 4
+
+        assert [] == pc_client.get_group('NeurIPS.cc/2021/Conference/Paper5/Area_Chairs').members
+        assert [] == pc_client.get_group('NeurIPS.cc/2021/Conference/Paper5/Senior_Area_Chairs').members
```
carlosmondra, line 1047: Why is the group `'NeurIPS.cc/2021/Conference/Paper5/Senior_Area_Chairs'` empty? Was it empty before the AC was unassigned or did it become empty after the AC was removed?

tests/test_neurips_conference.py
```diff
@@ -1020,6 +1024,74 @@ def test_setup_matching(self, conference, client, helpers):
...
+
+        ## Add AC assignment
+        ac_assignment.ddate = None
+        pc_client.post_edge(ac_assignment)
```
carlosmondra, line 1051: So if an AC is added/removed, its corresponding SAC is also added/removed?

tests/test_neurips_conference.py
```diff
@@ -1020,6 +1024,74 @@ def test_setup_matching(self, conference, client, helpers):
...
+        assert sac_assignment.tail == '~SeniorArea_GoogleChair1'
+
+        ## Add SAC assignment
+        sac_assignment.ddate = openreview.tools.datetime_millis(datetime.datetime.utcnow())
+        pc_client.post_edge(sac_assignment)
+
+        helpers.await_queue()
+
+        assert len(pc_client.get_edges(invitation='NeurIPS.cc/2021/Conference/Senior_Area_Chairs/-/Assignment')) == 2
+        assert len(pc_client.get_edges(invitation='NeurIPS.cc/2021/Conference/Area_Chairs/-/Assignment')) == 5
```
carlosmondra, line 1071: If you remove an SAC, the AC assignment does not change?

tests/test_neurips_conference.py
melisabok: it is empty because the AC assignment was remove, check the previous lines where the AC assignment edge is removed.

tests/test_neurips_conference.py
melisabok: yes, they need to be synced. Otherwise who is the SAC for that submission?

tests/test_neurips_conference.py
melisabok: the AC shouldn't never change by an SAC change. This test removes the SAC assignment, meaning the AC doesn't have an SAC anymore, that's why the SAC paper groups are empty. The PCs must find another SAC for that AC and then the group will have the new SAC as member.

tests/test_neurips_conference.py
carlosmondra: Oh yes, I still had the idea that SAC were directly linked to the submission

tests/test_neurips_conference.py
carlosmondra: Yes, OK the same explanation as before the SAC is linked to the submission through the AC. Thanks!

carlosmondra (approved):

celestemartinez (approved):

PR #1027, opened 2021-11-15
Fix/decline after assignment

### LINKED ISSUE #1024: Expire recruitment invitation
Users are declining invitation after being assigned to papers, so we should expire recruitment invitations. 

Alternatively, we should check if they have been assigned to papers already and not let them decline invitation, ask them to contact PCs/ACs.

celestemartinez:
Do no let a user decline a recruitment invitation if they have already been assigned to a paper. 

Fixes #1024

### DIFF
--- a/openreview/conference/templates/recruit_reviewers_pre_process.py
+++ b/openreview/conference/templates/recruit_reviewers_pre_process.py
@@ -17,6 +17,6 @@ def process(client, note, invitation):
     if note.content['response'] == 'No':
         memberships = client.get_groups(regex=VENUE_ID+'/Paper.*/Reviewers$', member=user)
         if memberships:
-            raise openreview.OpenReviewException('You have already been assigned to a paper. Please contact the paper\'s area chair to be unassigned.')
+            raise openreview.OpenReviewException('You have already been assigned to a paper. Please contact the paper area chair or program chairs to be unassigned.')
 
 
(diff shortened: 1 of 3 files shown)


openreview/conference/templates/recruit_reviewers_pre_process.py
```diff
@@ -13,4 +14,9 @@ def process(client, note, invitation):
     if not client.get_groups(id=REVIEWERS_INVITED_ID, member=user):
         raise openreview.OpenReviewException('User not in invited group, please accept the invitation using the email address you were invited with')
 
+    if note.content['response'] == 'No':
+        memberships = client.get_groups(regex=VENUE_ID+'/Paper.*/Reviewers$', member=user)
```
melisabok, line 18: a small suggestion to avoid having the string `Reviewers` here, you can use:

`conference.get_reviewers_id(number='.*')` as a constant in the template and then use that constant in the regex.

celestemartinez: Done! I also realized that this preprocess function is also used for example when we use Assignment_Recruitment. In this case I believe we want users to be able to decline (as they are removed from paper anon group) so I added a check. Let me know if I am missing anything!

melisabok (approved):
🥇

PR #1028, opened 2021-11-16
Add option to enable SUPER algorithm for sorting in the bidding console

carlosmondra:
This PR:
- [x] Implements Tanner's algorithm in the Bid Console. Paper of algorithm is [here](https://arxiv.org/abs/2007.07079). The algorithm can be enabled/disabled from the params at the top of the file. The algorithm is configurable and can be used for any of the computed scores as long as their values are between 0 and 1 like the OpenReview affinity scores.
- [x] Adds option to request the use of the algorithm from the request form
- [x] Allows to select bid options in the request form
- [x] Allows to select positive bid options in the request form. This is necessary for the algorithm. A positive bid can be `Very High` and/or `High`, for example, if the bid options are the following: `[ 'Very High', 'High', 'Neutral', 'Low', 'Very Low' ]`
- [x] Removes the conflict parameter as it can be specified in the bid options
- [x] Builds translate_map based on bid options specified in the invitation


melisabok:
Is it possible to make all these changes in a new template file and keep the current bid webfield unchanged? we could use either one or the other file when setting up the invitation.

melisabok (changes requested):
Please submit these changes in a different template file,  the builder can use one file or the other if the SUPER algorithm is on/off

melisabok:
@carlosmondra could you add a new parameter in the BidStage to decide which Bidding template to use? 

https://github.com/openreview/openreview-py/blob/master/openreview/conference/builder.py#L1724

The default value should always be the current implementation.

Then if we want the PCs to decide this from the UI, you would need to add a radio button here:

https://github.com/openreview/openreview-py/blob/master/openreview/venue_request/venue_request.py#L63

carlosmondra:
One option that is necessary is a list of what is considered to be a positive bid. For example, `very high` and `high` can be considered positive bids, but I'm not sure where the user should be asked about this. The request form does not seem to be the right place. @melisabok

melisabok (approved):
Cool!

tests/test_matching.py
```diff
@@ -80,6 +80,18 @@ def conference(self, client, helpers):
...
+        assert translate_map == {'Very High': 1.0, 'High': 0.5, 'Neutral': 0.0, 'Low': -0.5, 'Very Low': -1.0}
+        translate_map = openreview.matching._build_translate_map(['Very High', 'High', 'Neutral', 'Low', 'Very Low', 'Conflict'])
+        assert translate_map == {'Very High': 1.0, 'High': 0.6, 'Neutral': 0.2, 'Low': -0.2, 'Very Low': -0.6, 'Conflict': -1.0}
+        translate_map = openreview.matching._build_translate_map(['High', 'Neutral', 'Low'])
+        assert translate_map == {'High': 1.0, 'Neutral': 0.0, 'Low': -1.0}
+        translate_map = openreview.matching._build_translate_map(['High', 'Low'])
+        assert translate_map == {'High': 1.0, 'Low': -1.0}
+        translate_map = openreview.matching._build_translate_map(['Very High', 'High', 'Neutral', 'Low'])
+        assert translate_map == {'Very High': 1.0, 'High': 0.333, 'Neutral': -0.333, 'Low': -1.0}
+
```
carlosmondra, line 94: @melisabok I created a function that builds default translate maps from the bid options. Let me know what you think.

melisabok: That's great, thank you!

melisabok (approved):
the changes look good to me!

openreview/venue_request/venue_request.py
```diff
@@ -73,6 +73,18 @@ def setup_bidding_stage(self):
             'bid_count': {
                 'description': 'Minimum bids one should make to mark bidding task completed for them. Default is 50.',
                 'value-regex': '[0-9]*'
+            },
+            'bid_options': {
+                'description': 'Comma separated values e.g. Very High, High, Neutral, Low, Very Low. Default values are Very High, High, Neutral, Low, Very Low',
+                'value-regex': '[\w\s,]+'
```
melisabok, line 79: one small details, can you add a default field here with the values `['Very High', 'High', 'Neutral', 'Low', 'Very Low']`

most of the PCs don't know which options to use, having a default value is helpful

carlosmondra: Sounds good! It's done!

melisabok:
The UI looks good, I would suggest two things:

1. Set "Affinity Score" as the default selected option in the "Sort By" dropdown. When there is not bid data, the sorting doesn't say anything so I would prefer to show the papers sorted by affinity score.
2. Show a number that indicates the sorting criteria. Is it possible to show this number in the UI? 
```-1 * (partialScore * affinityScore + TRADE_OFF * (Math.pow(2, affinityScore) - 1))```

I also noticing the papers are not being re-sorted after each new bid posted. I'm not sure this is the intended but we could use a button to refresh the search or automatically re-sort the papers after each bid posted(this can be very expensive). Most of the users are bidding their papers at the same time so it would be nice to have the papers sorted in real time.

carlosmondra:
> 1. Set "Affinity Score" as the default selected option in the "Sort By" dropdown. When there is not bid data, the sorting doesn't say anything so I would prefer to show the papers sorted by affinity score.

How about we leave the option to be SUPER from the dropdown and if there are no bids the computed score is the affinity scores?

> 2. Show a number that indicates the sorting criteria. Is it possible to show this number in the UI?
>    `-1 * (partialScore * affinityScore + TRADE_OFF * (Math.pow(2, affinityScore) - 1))` 

Let me check this, maybe I have a very outdated bid console web field in this PR :/

> I also noticing the papers are not being re-sorted after each new bid posted. I'm not sure this is the intended but we could use a button to refresh the search or automatically re-sort the papers after each bid posted(this can be very expensive). Most of the users are bidding their papers at the same time so it would be nice to have the papers sorted in real time.

The only way to have the papers sorted again is by refreshing the page and the reason I didn't do it was because it was too expensive, also, now with rate limiting, I don't think it's a good idea.

melisabok:
Should we close this PR?

carlosmondra:
> Should we close this PR?

Sure, it seems like no one is interested anymore.

PR #1044, opened 2021-12-09
Set Author Group ID as authorids during post decision stage

purujitgoyal:


melisabok:
There are a few tests that are calling to post_submission_stage, could you call it using reveal_authors=False and add an assertion?

celestemartinez (approved):

melisabok (approved):
thanks!

PR #1045, opened 2021-12-09
Adds publications info to get_profile_info

purujitgoyal:


openreview/tools.py
```diff
@@ -1577,6 +1577,7 @@ def get_profile_info(profile):
     emails = set()
     relations = set()
     publications = set()
+    publication_years = 3
```
melisabok, line 1580: do you think the default should be last 3 years? I was thinking to use all the existing publications as the default implementation.

purujitgoyal: Depends on how much extra time would it take.

melisabok: It shouldn't take too much time to compare the author relationship. Then we can parametrize the last year for publications,  history of domains and relations.

purujitgoyal: I have made the default to consider all publications.

melisabok (approved):

celestemartinez (approved):

PR #1047, opened 2021-12-10
Fix/journal conflicts

melisabok:
- create new journal modules: groups, recruitment and assignment
- deprecate `_getprofiles` in the matching module of the conference, move the implementation to tools.
- compute conflicts for action editors and reviewers.

### DIFF
--- a/openreview/journal/assignment.py
+++ b/openreview/journal/assignment.py
@@ -1,0 +1,91 @@
+from .. import openreview
+from .. import tools
+
+from openreview.api import Edge
+
+import random
+
+class Assignment(object):
+
+    def __init__(self, journal):
+        self.client = journal.client
+        self.journal = journal
+
+
+    def setup_ae_assignment(self, note):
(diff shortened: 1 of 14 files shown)


openreview/journal/assignment.py
```diff
@@ -0,0 +1,117 @@
...
+        ## Create conflict and affinity score edges
+        for ae in action_editors:
+
+            edge = Edge(invitation = self.journal.get_ae_affinity_score_id(),
+                readers = [venue_id, authors_id, ae],
+                writers = [venue_id],
+                signatures = [venue_id],
+                head = note.id,
+                tail = ae,
+                weight=round(random.random(), 2)
```
celestemartinez, line 36: why do we assign a random affinity score here, is this a placeholder for a future affinity score computation?

melisabok: yes, I'm waiting for the expertise API to be ready to do the integration.

celestemartinez (approved):

PR #1050, opened 2021-12-13
Add HIDE_FIELDS to withdraw and desk-reject process functions

purujitgoyal:


tests/test_eccv_conference.py
```diff
@@ -1016,7 +1016,7 @@ def test_desk_reject_submission(self, conference, client, test_client, selenium,
 
     def test_withdraw_submission(self, conference, client, test_client, selenium, request_page, helpers):
 
-        conference.setup_post_submission_stage(force=True)
+        conference.setup_post_submission_stage(force=True, hide_fields=['_bibtex'])
```
melisabok, line 1019: I believe the submission doesn't have a _bibtex value, can you try to hide another field? 

Didn't NeurIPS test work?

purujitgoyal: Withdraw submission does have the `_bibtex` field, however, desk-reject doesn't.

No, NeurIPS didn't work. It had a completely different error.

melisabok: I added back the NeurIPS test and fixed the error. The Withdrawn and Desk Rejected invitations must have the hidden fields declared in their reply.

openreview/conference/invitation.py
```diff
@@ -302,6 +302,9 @@ def __init__(self, conference, reveal_authors, reveal_submission, hide_fields):
                     'values-regex': '[^;,\\n]+(,[^,\\n]+)*',
                     'required': False,
                     'order': 2
+                },
+                'keywords': {
```
melisabok, line 306: no need to make this change, I already added it in hide_fields.

openreview/conference/invitation.py
```diff
@@ -310,11 +310,12 @@ def __init__(self, conference, reveal_authors, reveal_submission, hide_fields):
             if not reveal_authors:
                 content['authors'] = {'values': ['Anonymous']}
                 content['authorids'] = {'values-regex': '.*'}
-            for field in hide_fields:
-                content[field] = {
-                    'values-regex': '.*',
-                    'required': False
-                }
+            if hide_fields:
```
melisabok, line 313: can we default `hide_fields` to be an empty array so we avoid this if?

openreview/conference/invitation.py
melisabok: @purujitgoyal can you remove this change? 

the variable `hide_fields` should have the key `keywords`. Not all the invitations have this field.

purujitgoyal: Done.

openreview/conference/invitation.py
purujitgoyal: Done.

melisabok (approved):

PR #1057, opened 2021-12-20
Adds rename_edges to openreview.py

nlbahy:


carlosmondra:
@nlbahy could you add a simple test that renames edges?

tests/test_edges.py
```diff
@@ -71,7 +71,24 @@ def test_save_bulk (self, client, test_client):
         openreview.tools.post_bulk_edges(client, edges)
         posted_edges = list(openreview.tools.iterget_edges(client, invitation=inv1.id))
         assert len(edges) == len(posted_edges)
-
+    
+    def test_rename_edges(self, client):
+        guest = openreview.Client()
+        to_profile = guest.register_user(email = 'lbahy@mail.com', first = 'Nadia', last = 'LBahy', password = '1234')
```
carlosmondra, line 77: This line is failing when all tests run with:
```js
{'name': 'Error', 'message': 'User not confirmed. Please click on "Didn\'t receive email confirmation?" to complete the registration.', 'status': 400}
```

This happened because test_client.py already registers this email. One solution is to change the email and the name to something else. Only changing the email would not be sufficient because then, the new profile would be assigned ~Nadia_LBahy2 if all tests are run and ~Nadia_LBahy1 if only test_edges.py is run.

nlbahy: Thank you for pointing this out, I'll fix that now!

tests/test_edges.py
```diff
@@ -71,7 +71,24 @@ def test_save_bulk (self, client, test_client):
...
+        client.rename_edges('~Super_User1', '~Nadia_LBahy2')
+        nadias_edges = list(openreview.tools.iterget_edges(client, tail="~Nadia_LBahy2"))
+        super_edges_ids = [edge.id for edge in super_user_edges]
+        nadia_edges_ids = [edge.id for edge in nadias_edges]
+        for edge in super_edges_ids:
+            assert edge in nadia_edges_ids
+        assert len(super_user_edges)==len(nadias_edges)
+        super_user_edges = list(openreview.tools.iterget_edges(client, tail="~Super_User1"))
+        assert len(super_user_edges) == 0
+        client.rename_edges('~Nadia_LBahy2','~Super_User1')
```
carlosmondra, line 90: Does this test pass if you only run the following?
```py
pytest tests/test_edges.py
```

carlosmondra (approved):

PR #1067, opened 2022-01-05
Feature/speed up setup matching

carlosmondra:
This PR adds the methods `concurrent_get` that work very similarly to `iterget`. The difference is that it performs concurrent requests, so it's faster than `iterget` and it does not return an iterator but the list of all the documents. I have replaced `iterget` with `concurrent_get` when possible. Other requests can also be optimized using `concurrent_get` or `concurrent_requests`. This should speed up the setup matching as well, but I haven't made any tests with large conferences.


openreview/tools.py
```diff
@@ -133,11 +132,18 @@ def get_profiles(client, ids_or_emails, with_publications=False):
...
 
         client_v1 = openreview.Client(baseurl=baseurl_v1, token=client.token)
-        #client_v2 = openreview.api.OpenReviewClient(baseurl=baseurl_v2, token=client.token)
-        for profile in profiles:
-            notes_v1 = list(iterget_notes(client_v1, content={'authorids': profile.id}))
-            #notes_v2 = list(iterget_notes(client_v2, content={'authorids': profile.id}))
-            profile.content['publications'] = notes_v1 #+ notes_v2
+        client_v2 = openreview.api.OpenReviewClient(baseurl=baseurl_v2, token=client.token)
+
+        notes_v1 = concurrent_requests(lambda profile : list(iterget_notes(client_v1, content={'authorids': profile.id})), profiles)
```
melisabok, line 137: what about using `get_all_notes` here?

carlosmondra: Done!

melisabok (approved):

celestemartinez (approved):

PR #1068, opened 2022-01-05
Fixes process function to use content.get(key)

nlbahy:


openreview/conference/templates/decision_process.py
```diff
@@ -13,13 +13,13 @@ def process_update(client, note, invitation, existing_note):
     
     # Check if the content of the existing note is the same as the updated note
     if existing_note:
-      equal = True
-      for key in existing_note.content:
-        if note.content[key] and (existing_note.content[key] != note.content[key]):
-          equal = False
+      if (EMAIL_AUTHORS and ('everyone' in note.readers or AUTHORS_ID in note.readers)):
+        equal = True
```
melisabok, line 17: some variable naming recommendation: I would always start with a boolean variable with the value False. In this case you can call "has_changed = False" and set True when there is at least a field that is not equal. Then you would send the email only `if has_changed` that is easier to read.

celestemartinez (approved):

melisabok (approved):

PR #1072, opened 2022-01-06
Fix/meta super invitations

melisabok:
- Support domain field
- Create all the invitations using another invitation
- Create meta invitations for the journal
- Add dateprocesses for Official Recommendation
- Add Review reminder as a proof of concept
- Refactor invitation module to have the journal object as a property instead of a parameter


xkopenreview:
i think client.py will need to be updated to work with domain property in meta invitation

melisabok:
Let me check this and get back to you.

carlosmondra:
There is just one empty file that I'm not sure is needed: openreview/journal/templates.py

carlosmondra (approved):

melisabok:
Thanks for the catch, I deleted the file.

celestemartinez (approved):

PR #1077, opened 2022-01-07
Fix/journal feedback

melisabok:
- Fix anonymized bibtex entry when the submission is under review
- Remove EIC as noninvitees, they can perform all the actions
- Show the submission full content in the consoles.
- Remove existing edges if they exist before posting the new ones.
- get profiles with publications from both APIs.


celestemartinez (approved):
Looks good to me!

openreview/journal/assignment.py
```diff
@@ -13,6 +13,12 @@ def __init__(self, journal):
         self.client = journal.client
         self.journal = journal
 
+    def post_submission_edges(self, edges):
+        if edges:
+            ## Remove current edges if they exists
+            self.client.delete_edges(invitation=edges[0].invitation, head=edges[0].head, wait_to_finish=True)
+            return tools.post_bulk_edges(self.client, edges)
```
carlosmondra, line 20: Shouldn't this be:
```suggestion
        if edges:
            ## Remove current edges if they exists
            self.client.delete_edges(invitation=edges[0].invitation, head=edges[0].head, wait_to_finish=True)
        return tools.post_bulk_edges(self.client, edges)
```

tests/test_journal.py
```diff
@@ -431,6 +439,15 @@ def test_submission(self, journal, openreview_client, test_client, helpers):
...
         assert note.content['venueid']['value'] == '.TMLR/Withdrawn_Submission'
+#         assert note.content['_bibtex']['value'] == '''@article{
+# userpaper,
+# title={Paper title {UPDATED}},
+# author={Test User and Andrew McCallum},
+# journal={Submitted to Transactions of Machine Learning Research},
+# year={2022},
+# url={https://openreview.net/forum?id=''' + note_id_1 + '''},
+# note={Under review}
+# }'''
```
carlosmondra, line 450: Should this be removed?

openreview/tools.py
```diff
@@ -107,8 +121,23 @@ def get_profiles(client, ids_or_emails, with_publications=False):
...
+            baseurl_v1 = 'https://api.openreview.net'
+            baseurl_v2 = 'https://api2.openreview.net'
+
+        client_v1 = openreview.Client(baseurl=baseurl_v1, token=client.token)
+        client_v2 = openreview.api.OpenReviewClient(baseurl=baseurl_v2, token=client.token)
         for profile in profiles:
-            profile.content['publications'] = list(iterget_notes(client, content={'authorids': profile.id}))
+            notes_v1 = list(iterget_notes(client_v1, content={'authorids': profile.id}))
+            notes_v2 = list(iterget_notes(client_v2, content={'authorids': profile.id}))
+            profile.content['publications'] = notes_v1 + notes_v2
```
carlosmondra, line 140: Not necessary, but you could speed this up like this:
```py
notes_v1 = concurrent_requests(lambda profile : list(iterget_notes(client_v1, content={'authorids': profile.id})), profiles)
for idx, publication in enumerate(notes_v1):
    profiles[idx].content['publications'] = publication

notes_v2 = concurrent_requests(lambda profile : list(iterget_notes(client_v2, content={'authorids': profile.id})), profiles)
for idx, publication in enumerate(notes_v2):
    if profiles[idx].content['publications']:
        profiles[idx].content['publications'] = profiles[idx].content['publications'] +  publication
    else:
        profiles[idx].content['publications'] = publication

```

melisabok: I don't think a user would have more 1k publications, let's keep it as it is now. Thanks for the shared code.

tests/test_journal.py
melisabok: I would like to keep it because there is bibtex entry for withdrawn invitations and I think we should, I need to confirm with Hugo which note  text to put.

openreview/tools.py
carlosmondra: Yes, I think it's OK. To explain the code a bit: This gets all the publications of the profiles in parallel, it does not go one by one for each profile. In other words, the `for profile in profiles` is removed. That's why it's faster :)

carlosmondra (approved):

openreview/tools.py
melisabok: ohhhh I see now, let me do the change now

PR #1079, opened 2022-01-10
Fix adding impersonators to conference group

carlosmondra:
Fixes setting impersonation for a conference. This PR only adds the emails or tilde ids to the conference group to the field `impersonators`. I removed the impersonate webfield since we are using the same URL the super user uses to impersonate.


melisabok (approved):
it looks good to me, adding @nlbahy  as a reviewer so she can verify impersonators work for ARR

nlbahy (approved):

PR #1081, opened 2022-01-10
update reviewer assigned email per Hugo's request

fabianp:
@melisabok : I tried to use a branch instead of a fork as per your advice but I don't have write access to this repo


melisabok:
I added you writer permission, if you don't mind can you create the branch?

PR #1085, opened 2022-01-11
update email to reviewers for TMLR

fabianp:


melisabok:
There is a test that check the test of the email, could you update it? https://github.com/openreview/openreview-py/blob/fabianp-patch-1/tests/test_journal.py#L466

fabianp:
@melisabok : the last test failure seems unrelated to this PR, could you kindly check that?

melisabok:
I fixed the tests in master but it seems there is another TMLR test that needs to be updated.

melisabok (approved):

PR #1086, opened 2022-01-12
Speed up decision stage

carlosmondra:
Uses concurrent requests to create decision invitations. Locally, processing time is reduced by half


openreview/conference/invitation.py
```diff
@@ -1545,13 +1545,13 @@ def set_meta_review_invitation(self, conference, notes):
...
-        invitations = []
         self.client.post_invitation(DecisionInvitation(conference))
-        for note in tqdm(notes, total=len(notes), desc='set_decision_invitation'):
+        def post_invitation(note):
             invitation = self.client.post_invitation(PaperDecisionInvitation(conference, note))
             self.__update_readers(note, invitation)
-            invitations.append(invitation)
+            return invitation
+
+        invitations = tools.concurrent_requests(post_invitation, notes)
```
melisabok, line 1554: Thanks for doing this! Is there a way to show a progress bar when doing concurrent requests?

carlosmondra: A progress bar is already implemented within concurrent_requests ([here](https://github.com/openreview/openreview-py/blob/9c775a0e6978aff1966f68f32601c3be3952a232/openreview/tools.py#L34)). So there is no need to add it here.

melisabok: Awesome!

melisabok (approved):

openreview/conference/invitation.py
melisabok: You are invited to do the same with the other stages 🙈

Issue #1053, opened 2021-12-16
PC console: add a new tab that shows the Desk Rejected and Withdrawn submission


melisabok:
So PCs can look at them and export them if they want.

PR #1091, opened 2022-01-13
fix/PC console: export all the fields of the content

xkopenreview:
current pc console will only export title, abstract and authors

this pr should export all content fields

### DIFF
--- a/openreview/conference/templates/programchairWebfield.js
+++ b/openreview/conference/templates/programchairWebfield.js
@@ -2967,10 +2967,14 @@ var buildCSV = function(){
   var areachairIds = conferenceStatusData.areaChairGroups.byNotes;
   var isFiltered = conferenceStatusData.filteredRows ? true : false
   var notes = isFiltered ? conferenceStatusData.filteredRows : conferenceStatusData.blindedNotes
+  var originalNotes = notes.map(p => isFiltered
+    ? p.note.details.original || p.note
+    : p.details.original || p
+  )
   
   var noteContentFields = [
     ...new Set(
-      notes
+      originalNotes
         .map((p) => {
           var content = isFiltered ? p.note.content : p.content;
           if (!content) return []
@@ -3031,7 +3035,7 @@ var buildCSV = function(){
     var originalNote = paperTableRow.note.details.original || paperTableRow.note;
 
     var contents = noteContentFields.map(field=>{
-      const contentValue = paperTableRow.note.content[field]
+      const contentValue = originalNote.content[field]
       if (!contentValue) return ''
       if (field === 'authors') return originalNote.content.authors ? originalNote.content.authors.join('|') : ''
       if(Array.isArray(contentValue)) return contentValue.join('|')
--- a/openreview/conference/templates/programchairWebfield.js
+++ b/openreview/conference/templates/programchairWebfield.js
@@ -1460,7 +1460,9 @@ var displayPaperStatusTable = function() {
       conferenceStatusData.filteredRows = filteredRows
     }
     renderTable(container, filteredRows);
-    $(container + ' .btn-export-data').text(`Export ${filteredRows.length} records`)
+    $(container + " .btn-export-data").text(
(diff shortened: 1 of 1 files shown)


nlbahy (approved):

xkopenreview:
@celestemartinez please try 071a0ac

openreview/conference/templates/programchairWebfield.js
```diff
@@ -3020,9 +3030,14 @@ var buildCSV = function(){
 
     var originalNote = paperTableRow.note.details.original || paperTableRow.note;
 
-    var title = paperTableRow.note.content.title.replace(/"/g, '""');
-    var abstract = paperTableRow.note.content.abstract.replace(/"/g, '""');
-    var authors = originalNote.content.authors ? originalNote.content.authors : [];
+    var contents = noteContentFields.map(field=>{
```
melisabok, line 3033: can we use arrow functions here?

melisabok:
I think the SAC and AC console also have export functions, I was wondering if we can move this code to Webfield.js.

celestemartinez:
I get no error now! But I noticed that we are only getting the values from the original note for the field 'authors'. This means that PCs are not getting author emails ('authorids') or any value for fields that have been hidden in the blind submission.

xkopenreview:
discussed with @celestemartinez that it should use the original note to extract the info
this also applies to https://github.com/openreview/openreview-py/issues/1053

xkopenreview:
> I think the SAC and AC console also have export functions, I was wondering if we can move this code to Webfield.js.

i think sac webfield does not have export function
and the function in pc console ac console don't really converge

xkopenreview:
the new tab currently shows 3 columns: paper number; paper summary and reason (desk rejected or withdrawn)

openreview/conference/templates/programchairWebfield.js
```diff
@@ -2967,13 +2967,27 @@ var buildCSV = function(){
   var areachairIds = conferenceStatusData.areaChairGroups.byNotes;
   var isFiltered = conferenceStatusData.filteredRows ? true : false
   var notes = isFiltered ? conferenceStatusData.filteredRows : conferenceStatusData.blindedNotes
-
+  var originalNotes = notes.map(p => isFiltered
+    ? p.note.details.original || p.note
+    : p.details.original || p
+  )
+  
+  var noteContentFields = [
```
melisabok, line 2975: @xkopenreview can you create a utils function in webfield.js to get all the content fields or to exports the notes into an array of string? having the code in webfield.js you can use ES6 features

xkopenreview: ok. for your info the changes are generated by babel with target "since 2017"

xkopenreview: discussed with Melisa about this and decided not to have a util function in webfield.js as where the util function can be reused is not clear

zbialecki:
It looks like there's still a lot of new JS syntax here. Another reason to stick with ES5 is that it's too hard to keep track of what's supported by which browsers, `=>` functions might be fine for our users but the `??` operator isn't... I noticed you use the `.flat()` function which wasn't supported by Edge until after 2020.

openreview/conference/templates/programchairWebfield.js
```diff
@@ -1460,7 +1460,9 @@ var displayPaperStatusTable = function() {
       conferenceStatusData.filteredRows = filteredRows
     }
     renderTable(container, filteredRows);
-    $(container + ' .btn-export-data').text(`Export ${filteredRows.length} records`)
+    $(container + " .btn-export-data").text(
+      "Export ".concat(filteredRows.length, " records")
```
zbialecki, line 1464: Generally if you aren't using template strings you would use `+` for string concatenation

zbialecki (approved):
Looks good.

melisabok (approved):
it looks good!!!!!!

celestemartinez (approved):

PR #1092, opened 2022-01-13
Catches exception and Post comment for venue request process functions

purujitgoyal:


openreview/venue_request/process/postSubmissionProcess.py
```diff
@@ -1,6 +1,31 @@
 def process(client, note, invitation):
-
+    GROUP_PREFIX = ''
```
melisabok, line 2: are you setting this variable?

purujitgoyal: It's the same as in deployProcess and revisionProcess.

melisabok: right, you need to set the value in the python code. Like we are doing here:

https://github.com/openreview/openreview-py/blob/65bfa35d2a242674abd4b7cf3a8d9f3762b12a11/openreview/venue_request/venue_request.py#L10

You added this variable to the postSubmissionProcess.py but you are not setting any value:

https://github.com/openreview/openreview-py/blob/65bfa35d2a242674abd4b7cf3a8d9f3762b12a11/openreview/venue_request/venue_request.py#L984

purujitgoyal: Thanks, this worked.

purujitgoyal:
I've added full error stack trace here but it looks weird on UI, 
Screen Shot 2022-01-18 at 4 48 09 PM

is this okay? should I remove the stack trace?

carlosmondra:
> I've added full error stack trace here but it looks weird on UI, Screen Shot 2022-01-18 at 4 48 09 PM
> 
> is this okay? should I remove the stack trace?

I believe this is happening because the stack string contains `\n`. I'm not sure how those can be handled though.

purujitgoyal:
There are `\n` in the first 3 lines as well. Why are those 3 in a separate box?

carlosmondra:
> There are `\n` in the first 3 lines as well. Why are those 3 in a separate box?

I'm not sure how you are parsing the error, but  it's a JSON that looks like this:
```js
"error": {
  "name": "Error",
  "message": "ValueError: unconverted data remains:  11:59",
  "status": 500,
  "stack": "Error: ValueError: unconverted data remains:  11:59\n    at convertToOpenReviewError (/home/openreview/deploy/openreview/helpers/errors.js:20:14)\n    at formatProcessError (/home/openreview/deploy/openreview/helpers/errors.js:647:12)\n    at Process.executeProcessFunction (/home/openreview/deploy/openreview/helpers/entityProcess.js:176:22)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)"
}
```

So I am not sure how you are getting the first 3 `\n`. The `\n` should only appear in the stack property

purujitgoyal:
I'm adding those myself, this is the `content.comment` string that I pass to the note:

`f'''
                {invitation_type} Process failed due to the following error: \n 
                {repr(e)} \n 
                {traceback.format_exc()}
                '''
`

carlosmondra:
Ah, I see, my bad. I thought you were reading the logs.  Have you tried formatting it manually instead of using repr(e)? Perhaps that way works.

melisabok:
I think we need to save the stack trace somewhere, if we catch the error then the process logs don't save it so I think we should show it in the comment. @carlosmondra ideas?

carlosmondra:
> I think we need to save the stack trace somewhere, if we catch the error then the process logs don't save it so I think we should show it in the comment. @carlosmondra ideas?

The process logs do save the stack of the error, but it's only visible by us.
I think it's fine that a comment is posted with the error and the stacktrace is there, we just need to figure out how to post it correctly right? Or am I misunderstanding the question?

melisabok:
Purujit is catching the exception so the process function won't fail. The API won't know there was an exception. There are two options:

1. Post the stack trace in the comment note so it can help us to know why it failed: PCs entered the wrong configuration or there was an error in our code. 
2. Don't post the stack trace in the comment note, catch the exception and raise it again. So the process function will fail and we can use the process logs to see what happening. In this case we need to check all the emails from errors where in most of the cases the PCs entered the wrong configuration.

purujitgoyal:
Is there a way to get the process logs id? we can add it to the comment and raise the exception after posting the comment. This way we won't have to go through all the emails.

carlosmondra:
OK, so if we want to reduce the number of emails with failed process functions then we need to catch and NOT raise an exception. So in that case we should POST the stack trace in the comment. The process function will not appear as failed in the backend, like you said, but we could still add a print statement that will log that there was an error and was POSTed as a comment.

The problem with using the log id is that it requires special reader permissions. I would need to change the backend to add readers to those logs so that PCs can see it.

carlosmondra:
In that case, since we are not going to receive an email, perhaps we should still add the jobId to the comment for us (the Super User) to check. What do you think?

melisabok:
The comment is visible to the PCs too. What if we add more context in the comment?

I see that you are putting as title the invitation name. We should add the link to the reference that triggered the process function. This way we can know the reference id and look into the process logs.

purujitgoyal:
Do you mean something like this?
Screen Shot 2022-01-19 at 3 41 14 PM
I've added note ID in the title, to add a link to the note, I believe we will have to make changes in the UI?

melisabok:
Yes, I like that, in the comment text you can add the link to the reference, it should be `https://api.openreview/references?id=3-wG5H7QY`

openreview/venue_request/process/recruitmentProcess.py
```diff
@@ -83,6 +83,12 @@ def process(client, note, invitation):
         error_status=f'''No recruitment invitation was sent to the following users due to the error(s) in the recruitment process: \n
         {recruitment_status.get('errors') }'''
 
-        comment_note.content['comment'] += f'''\n Error: {error_status}'''
+        comment_note.content['comment'] += f'''
```
melisabok, line 86: I don't think we need to change this file. The errors are correctly handled in the conference object.

openreview/venue_request/process/remindRecruitmentProcess.py
```diff
@@ -32,7 +32,7 @@ def process(client, note, invitation):
         writers = [],
         signatures = [SUPPORT_GROUP],
         content = {
-            'title': 'Remind Recruitment Status',
+            'title': f'Remind Recruitment Status',
```
melisabok, line 35: Same here, you can need to change this file.

tests/test_venue_request.py
```diff
@@ -467,6 +467,37 @@ def test_venue_remind_recruitment(self, client, test_client, selenium, request_p
...
+        helpers.await_queue()
+        process_logs = client.get_process_logs(id=bid_stage_note.id)
+        assert len(process_logs) == 1
+        assert process_logs[0]['invitation'] == '{}/-/Request{}/Bid_Stage'.format(venue['support_group_id'], venue['request_form_note'].number)
+        assert process_logs[0]['status'] == 'ok'
+
+        comment_invitation = '{}/-/Request{}/Comment'.format(venue['support_group_id'],
+                                                             venue['request_form_note'].number)
+        last_comment = client.get_notes(invitation=comment_invitation)[0]
+        error_string = 'Bid Stage Process failed due to the following error:'
```
melisabok, line 498: Is it possible to show in the assertion which is the error? is not clear to me why the Bid Stage is failing

openreview/conference/builder.py
```diff
@@ -1332,6 +1333,7 @@ def recruit_reviewers(self,
                         if repr(e) not in recruitment_status['errors']:
                             recruitment_status['errors'][repr(e)] = []
                         recruitment_status['errors'][repr(e)].append(reviewer_id)
+                        print(traceback.format_exc())
```
melisabok, line 1336: we don't need to print the stack trace here

openreview/conference/builder.py
```diff
@@ -1358,6 +1360,7 @@ def recruit_reviewers(self,
...
```
melisabok, line 1363: we don't need to print the stack trace here

openreview/conference/builder.py
```diff
@@ -1397,6 +1400,7 @@ def recruit_reviewers(self,
                     if repr(e) not in recruitment_status['errors']:
                         recruitment_status['errors'][repr(e)] = []
                     recruitment_status['errors'][repr(e)].append(email)
+                    print(traceback.format_exc())
```
melisabok, line 1403: we don't need to print the stack trace here

celestemartinez:
Seems like this doesn't catch exceptions for the /Revision invitation (besides trying to update the activation date of the paper matching invitation). 

I did a test where I put in a wrong JSON for `Additional Submission Options` but there was an error that was not caught. This is because this is happening not in the try-catch but before in 
`conference = openreview.helpers.get_conference(client, note.forum, SUPPORT_GROUP)`

This is one of the most common errors I think from the request form, so it would be nice to catch as well?

melisabok:
Yeah, @purujitgoyal can you add a test that tries to add additional submissions options with typos?

melisabok (approved):

celestemartinez (approved):

PR #1096, opened 2022-01-18
add conflict last n years

melisabok:


openreview/tools.py
```diff
@@ -1628,7 +1628,7 @@ def get_conflicts(author_profiles, user_profile, policy='default'):
         author_relations.update(author_info['relations'])
```
nlbahy, line 1628: Does it not need to pass n_years when getting the author info?

nlbahy: I meant to link the comment to this line: author_info = info_function(profile)

openreview/tools.py
```diff
@@ -1628,7 +1628,7 @@ def get_conflicts(author_profiles, user_profile, policy='default'):
...
         author_publications.update(author_info['publications'])
 
-    user_info = info_function(user_profile)
+    user_info = info_function(user_profile, n_years)
```
melisabok, line 1631: i'm passing the parameter here

openreview/tools.py
melisabok: oops you are right!
I missed this part.

nlbahy (approved):

PR #1103, opened 2022-01-19
Fix/eic readers notes

melisabok:
- Add EIC as readers of all the replies
- Force to be only profiles for authorids
- Add retraction invitation to retract accepted papers
- Allow author release for withdrawn papers.


celestemartinez (approved):
Looks good!

melisabok:
Thanks! I'm going to merge the changes now so the beta testers that tests these changes tomorrow when the decision phase starts.

PR #1104, opened 2022-01-19
marking group members without ~ or @ as invalid

purujitgoyal:


tests/test_tools.py
```diff
@@ -224,7 +224,7 @@ def test_replace_members_with_ids(self, client, test_client):
             writers=['~Super_User1'],
             signatures=['~Super_User1'],
             signatories=['~Super_User1'],
-            members=['test@mail.com', '~SomeFirstName_User1', '~Another_Name1']
+            members=['test@mail.com', '~SomeFirstName_User1', '~Another_Name1', 'random_member']
```
melisabok, line 227: Can you test this using a group id with capital letters like 'ICLR.cc/2021/Conference/Reviewers'? this group id shouldn't change after the replacement.

purujitgoyal: Do you want to keep it as a member of the group?

purujitgoyal: I updated it to check for the group if it's not an email or a tilde ID. But `get_group()` returns a group for an invalid ID. Is that expected?

melisabok: > Do you want to keep it as a member of the group?
Yes, please

> get_group() returns a group for an invalid ID. Is that expected?
if the group doesn't exist then it will throw an error but all the members of a group should exist. If you get an exception then you can print it out but don't remove it so the PCs can fix this.

openreview/tools.py
```diff
@@ -596,21 +596,27 @@ def replace_members_with_ids(client, group):
...
-        else:
+        elif '~' in member:
             profile = get_profile(client, member)
             if profile:
-                return ('ids', profile.id)
+                return 'ids', profile.id
+            else:
+                return 'invalid_ids', member
+        else:
+            _group = get_group(client, member)
```
melisabok, line 615: no need to get the group here, all the members of a group are also groups.

melisabok: ohhh I see you are doing this to check if the group is valid or not, ok

tests/test_tools.py
```diff
@@ -218,13 +218,22 @@ def test_subdomains(self):
         assert openreview.tools.subdomains('   ') == []
 
     def test_replace_members_with_ids(self, client, test_client):
+        new_group = client.post_group(
```
melisabok, line 221: no need to create the group, once a member is added to a group the backend created the group automatically.

tests/test_tools.py
```diff
@@ -245,7 +254,8 @@ def test_replace_members_with_ids(self, client, test_client):
 
         replaced_group = openreview.tools.replace_members_with_ids(client, posted_group)
         assert replaced_group
-        assert replaced_group.members == ['~SomeFirstName_User1']
+        assert replaced_group.members == ['~SomeFirstName_User1', 'NewGroup']
+        assert 'random_member' not in replaced_group.members
```
melisabok, line 258: I don't think you need this assertion

tests/test_tools.py
```diff
@@ -218,13 +218,12 @@ def test_subdomains(self):
...
 
     def test_replace_members_with_ids(self, client, test_client):
-
         posted_group = client.post_group(openreview.Group(id='test.org',
             readers=['everyone'],
             writers=['~Super_User1'],
             signatures=['~Super_User1'],
             signatories=['~Super_User1'],
-            members=['test@mail.com', '~SomeFirstName_User1', '~Another_Name1']
+            members=['test@mail.com', '~SomeFirstName_User1', '~Another_Name1', 'NewGroup']
```
melisabok, line 226: if the group has a member with a tilde id where the number is missing, the member is removed from the group and prints this:

`Invalid profile/group id in group aclweb.org/ACL/ARR/2021/December/Paper1/Reviewers : ~Melisa_Bok`

I would like to keep ~Melisa_Bok in the group, the replace function should only replace, don't remove members.

purujitgoyal: wouldn't this cause problems later on?

purujitgoyal:
@melisabok @celestemartinez do we want to raise an exception when there's no profile found for a member with a `~` ID?

melisabok (approved):

PR #1105, opened 2022-01-19
check if year is in content first

melisabok:


openreview/tools.py
```diff
@@ -1725,7 +1725,9 @@ def get_neurips_profile_info(profile, n_years=3):
 
     ## Publications section: get publications within last n years
     for pub in profile.content.get('publications', []):
-        if pub.cdate:
+        if 'year' in pub.content and isinstance(pub.content['year'], str) and len(pub.content['year']) == 4:
```
purujitgoyal, line 1728: is the length check required?

melisabok: I'm not sure, should I try to convert to int and catch the error if the string doesn't represent a number?

purujitgoyal: Yes, that'd be better but ideally, we should not need to do that. I'd rather have checks in place when the `year` value is set in `pub.content`.

purujitgoyal (approved):

PR #1134, opened 2022-02-14
fix/PC console desk rejected and withdrawn submission tab - notes with no details

xkopenreview:
the query to get desk rejected/withdrawn submissions use invitation and details=original

when a note has no original, the details field will not be returned so when a rejected/withdrawn note has no original the page breaks trying to read undefined

in single blind venues such as MIDL, the original note is modified to have desk rejected/withdrawn invitation so the concept of original does not apply


openreview/conference/templates/programchairWebfield.js
```diff
@@ -1748,7 +1748,7 @@ var displayRejectedWithdrawnPaperStatusTable = function () {
       paperNumbers.push(note.number);
 
       var numberHtml = '<strong class="note-number">' + note.number + '</strong>';
-      var summaryHtml = Handlebars.templates.noteSummary(note.details.original || note);
+      var summaryHtml = Handlebars.templates.noteSummary(note.details ? note.details.original || note : note);
```
melisabok, line 1751: don't you need to check if `original` exists too? We can use the details for single blind submissions...

celestemartinez (approved):
Tested with MIDL 2022 and looks good!

nlbahy:
Tested this locally with a new single blind venue and it worked as well!

openreview/conference/templates/programchairWebfield.js
melisabok: I see the || now!, thanks

melisabok (approved):

PR #1135, opened 2022-02-14
Obfuscated Emails Exception

purujitgoyal:


openreview/tools.py
```diff
@@ -1718,7 +1718,7 @@ def get_conflicts(author_profiles, user_profile, policy='default', n_years=5):
     conflicts.update(author_domains.intersection(user_info['domains']))
     conflicts.update(author_relations.intersection(user_info['emails']))
     conflicts.update(author_emails.intersection(user_info['relations']))
-    conflicts.update(author_emails.intersection(user_info['emails']))
+    # conflicts.update(author_emails.intersection(user_info['emails']))
```
melisabok, line 1721: I believe we need to keep doing this for profiles that contain enterprise email addresses like gmail.com, outlook.com where the domains are ignored.

openreview/tools.py
```diff
@@ -1798,11 +1800,14 @@ def get_neurips_profile_info(profile, n_years=3):
...
+        if email.startswith("****@"):
+            raise openreview.OpenReviewException("You do not have the required permissions as some emails are obfuscated. Please login with the correct account or contact support.")
         emails.add(email)
-    ## if institution section is empty, add email domains
-    if not domains:
-        for email in profile.content['emails']:
-            domains.update(openreview.tools.subdomains(email))
+        domains.update(openreview.tools.subdomains(email))
+    # ## if institution section is empty, add email domains
+    # if not domains:
```
melisabok, line 1808: Why do you remove this?

melisabok:
You can reproduce this issue trying to get two profiles as a guest user and call to compute the conflicts, you should expect an exception.

openreview/tools.py
purujitgoyal: Not completely removed, just removed the `if not domains:` condition. It'll update the domains irrespective of whether there was institution section or not.

openreview/tools.py
purujitgoyal: Okay, do we need to calculate all of them, if we have already found a conflict?

openreview/tools.py
```diff
@@ -1716,10 +1716,14 @@ def get_conflicts(author_profiles, user_profile, policy='default', n_years=5):
 
     conflicts = set()
     conflicts.update(author_domains.intersection(user_info['domains']))
-    conflicts.update(author_relations.intersection(user_info['emails']))
-    conflicts.update(author_emails.intersection(user_info['relations']))
-    conflicts.update(author_emails.intersection(user_info['emails']))
-    conflicts.update(author_publications.intersection(user_info['publications']))
+    if not conflicts:
```
melisabok, line 1719: why do you add an if here? what does it change?

openreview/tools.py
```diff
@@ -1798,11 +1804,10 @@ def get_neurips_profile_info(profile, n_years=3):
 
     ## Emails section
     for email in profile.content['emails']:
+        if email.startswith("****@"):
+            raise openreview.OpenReviewException("You do not have the required permissions as some emails are obfuscated. Please login with the correct account or contact support.")
         emails.add(email)
-    ## if institution section is empty, add email domains
-    if not domains:
```
melisabok, line 1803: this implementation was added by NeurIPS Organizers, I'm not sure why you want to remove the if here?

The comment says if there is not institution listed then use the email as a domain conflict.

purujitgoyal: I combined both of them as I was trying to remove email-email conflict computation. I don't think it will change things either way, but I can revert it.

openreview/tools.py
melisabok: sometimes we want to have the full list of conflicts for debugging purposes.

openreview/tools.py
melisabok: where do you combine them?

Could you write a test for this scenario? using a profile with an email address but no history

purujitgoyal: it's updated [here](https://github.com/openreview/openreview-py/blob/cd624bbb5795a779fba067d1a8b4408df3f4b3dd/openreview/tools.py#L1806). It's same as the regular `get_profiles_info` function.

openreview/tools.py
purujitgoyal: Reverted.

melisabok:
Can you run `test_journal.py` using this branch? I believe the Journal is having this exception in one of the process functions

purujitgoyal:
`test_journal.py` fails [here](https://github.com/openreview/openreview-py/blob/1fad0cf45775bb3efa7ebe5ab76121f45f44bef9/openreview/journal/process/under_review_submission_process.py#L7). What's the client that gets used by journal here?

melisabok:
Can you show me the error? /logs/process should show it.

melisabok:
this is happening because the signature of the Under_Review invitation runs the process function and doesn't have permission to load the profiles with full email addresses. Can you try to change the signature value and use the `journal.get_editors_in_chief_id()` instead?

https://github.com/openreview/openreview-py/blob/master/openreview/journal/invitation.py#L1201

melisabok (approved):
Thanks for fixing the Journal code!

PR #1143, opened 2022-02-17
fix/PC console - some issues

xkopenreview:
should fix issue 2 and 3


xkopenreview:
messageCount retrieved from localStorage is of type string so it requires type conversion

melisabok:
@nlbahy and @celestemartinez could you check this one?

celestemartinez:
When I tried to send 0 messages, I was allowed to get to the confirmation step in message modal and was able to click on the 'Confirm and Send Messages' button. I then got an error saying 
`Error: Could not send emails at this time. Please refresh the page and try again.`

Looks like this is the expected behavior? I think it might be a bit confusing if a PC doesn't realize there are no recipients and only sees the error message (but not sure how common that might be)

celestemartinez (approved):

PR #1145, opened 2022-02-18
remove edge browser text

purujitgoyal:
Fix for openreview/openreview#12


purujitgoyal:
Tests are passing on my local.

celestemartinez:
Looks good, but I don't particularly like having just the link without any other text. I remember ARR complaining about this before and we changed the text to **Reviewer Browser**: https://openreview.net/group?id=aclweb.org/ACL/ARR/2021/October/Area_Chairs
If you think this is still not clear enough we can remove all text before the link

purujitgoyal:
I think it works, but yes, **Reviewer Browser** does make it more clear than **Edge Browser**. The only issue I can think of is if we do change it to **Reviewer Browser**, we should make sure we do it everywhere else someone will come up with another issue.

melisabok:
What about "Reviewer Assignment Browser"?

melisabok:
@purujitgoyal instead of removing "Edge Browser" could you replace it with "Reviewer Assignment Browser"?

celestemartinez (approved):

melisabok (approved):

PR #1157, opened 2022-03-01
Undo withdraw submission

purujitgoyal:


openreview/conference/templates/withdraw_process.py
```diff
@@ -1,4 +1,4 @@
-def process(client, note, invitation):
+def process(client, note, invitation, existing_note):
```
melisabok, line 1: don't you need to change the signature to `process_update` here? so the process function is being called when the note is edited or deleted.

purujitgoyal: I did that but it's failing with `InvitationExpiredError` now.

melisabok:
The error in the test is saying the Withdraw invitation has expired, that means you can not delete the note. 

This is invitation is expired when the paper is withdrawn, could you try to remove 'Withdraw' from the regex here: https://github.com/openreview/openreview-py/blob/master/openreview/conference/templates/withdraw_process.py#L71?

melisabok:
The changes look good! Could you verify in the test the message sent to the reviewers/ACs when the submission is restored? 

I'm not sure the text would like right.

melisabok (approved):

PR #1160, opened 2022-03-03
update comment invitation readers to PC group

purujitgoyal:


melisabok (approved):
Awesome! @celestemartinez do you want to take a look?

openreview/venue_request/process/deployProcess.py
```diff
@@ -390,4 +390,18 @@ def process(client, note, invitation):
             signatures = ['~Super_User1']
         )
 
-        client.post_invitation(matching_invitation)
\ No newline at end of file
+        client.post_invitation(matching_invitation)
```
celestemartinez, line 393: Looks like the comment invitation's reply.readers are updated only when we setup the paper matching, it should be outside of the if.

purujitgoyal: Fixed.

celestemartinez (approved):

celestemartinez:
Looks good!

openreview/venue_request/process/deployProcess.py
```diff
@@ -390,4 +390,18 @@ def process(client, note, invitation):
...
+
+    client.post_invitation(openreview.Invitation(
+        id=SUPPORT_GROUP + '/-/Request' + str(forum.number) + '/Comment',
+        super=SUPPORT_GROUP + '/-/Comment',
+        reply={
+            'forum': forum.id,
+            'referent': forum.id,
```
melisabok, line 400: here is the issue, @purujitgoyal this should be `replyto` instead. 

Could you fix it and add a test?

melisabok: `replyto` must be None in order the comment to be present in all the replies.

purujitgoyal: what about `forum`?

purujitgoyal: and test for adding a new PC and posting a comment?

melisabok: forum should stay, the invitation should only replies to `forum.id` only.

> and test for adding a new PC and posting a comment?

That would be great, I added a comment in the new PR, you can also tests the existing replies were correctly updated.

PR #1162, opened 2022-03-04
Adds Status Invitations

purujitgoyal:


openreview/venue_request/process/invitationStatusProcess.js
```diff
@@ -0,0 +1,25 @@
...
+  or3client.or3request(or3client.notesUrl + '?id=' + note.forum, {}, 'GET', token)
+  .then(function(result) {
+
+    var forumNote = result.notes[0];
+
+    var message = {
+      groups: note.readers,
+      ignoreGroups: [note.tauthor, SUPPORT_GROUP],
+      subject: 'Comment posted to your request for service: ' + forumNote.content.title,
+      message: 'A comment was posted to your service request. \n\nComment title: ' + note.content.title + '\n\nComment: ' + note.content.comment + '\n\nTo view the comment, click here: ' + baseUrl + '/forum?id=' + note.forum + '&noteId=' + note.id
```
melisabok, line 16: The note is going to be an instance of Paper_Matching_Setup_Status, Recruitment_Status and Remind_Recruitment_Status. Could you format the email based on the schema of these notes? 

The subject should say "Paper Matching Setup Status for service...." or "Recruitment Status for service..."

openreview/venue_request/process/matchingProcess.py
```diff
@@ -59,26 +59,25 @@ def process(client, note, invitation):
     else:
         no_profiles_status = matching_status.get('no_profiles')
         if no_profiles_status:
-            profiles_status=f'''
+            without_profiles_status = f'''
```
melisabok, line 62: I would prefer that without_profiles_status and no publications_status are arrays of emails or profile ids and move the text to the comment section.

openreview/venue_request/venue_request.py
```diff
@@ -1145,6 +1146,74 @@ def setup_venue_recruitment(self):
...
+            },
+            'invited': {
+                'value-regex': '.*',
+                'description': 'No. of users invited',
+                'required': True,
+                'markdown': True,
+                'order': 2
+            },
+            'already_invited': {
+                'value-regex': '[\\S\\s]{0,100000}',
```
melisabok, line 1163: if this is a list then it should be `values-regex: '.*'` so the value is an array of strings instead of an string

openreview/venue_request/process/matchingProcess.py
melisabok: @celestemartinez  please let us know what you think.

celestemartinez: right now, the already_invited field looks like this:

`No recruitment invitation was sent to the following users because they have already been invited:
{'testvenue.org/Test_Venue/2022/Conference/Reviewers/Invited': ['celeste@mail.com']}`

Do you mean @melisabok you only want to have `['celeste@mail.com']` here? and the text in the comment?

melisabok: Yes. What do you think?

We can keep a string but having an array could be easier to parse the data.

openreview/venue_request/process/invitationStatusProcess.js
purujitgoyal: Done.

celestemartinez:
Now that we have more invitations, the Super User sees a lot more buttons on the request form (which is not a problem) However, the PCs can see the 'Error Status' button and they shouldn't

openreview/venue_request/process/matchingProcess.py
celestemartinez: this change hasn't been made yet

openreview/venue_request/process/deployProcess.py
```diff
@@ -404,3 +404,59 @@ def process(client, note, invitation):
     for reply in replies:
         reply.readers = readers
         client.post_note(reply)
+
+    client.post_invitation(openreview.Invitation(
+        id=SUPPORT_GROUP + '/-/Request' + str(forum.number) + '/Recruitment_Status',
+        super=SUPPORT_GROUP + '/-/Recruitment_Status',
+        reply={
+            'forum': forum.id,
+            'referent': forum.id,
```
melisabok, line 413: referent shouldn't be here. The referent value is used when we want to revise the request form note. In this case, we want to make replies.

openreview/venue_request/process/deployProcess.py
```diff
@@ -404,3 +404,59 @@ def process(client, note, invitation):
...
+        },
+        signatures=['~Super_User1']
+    ))
+
+    client.post_invitation(openreview.Invitation(
+        id=SUPPORT_GROUP + '/-/Request' + str(forum.number) + '/Paper_Matching_Setup_Status',
+        super=SUPPORT_GROUP + '/-/Paper_Matching_Setup_Status',
+        reply={
+            'forum': forum.id,
+            'referent': forum.id,
```
melisabok, line 441: same here

openreview/venue_request/process/matchingProcess.py
```diff
@@ -36,21 +36,22 @@ def process(client, note, invitation):
...
-        invitation = note.invitation.replace('Paper_Matching_Setup', 'Comment'),
+        invitation = note.invitation.replace('Paper_Matching_Setup', 'Paper_Matching_Setup_Status'),
         forum = note.forum,
         replyto = note.id,
         readers = [conference.get_program_chairs_id()] + [SUPPORT_GROUP],
         writers = [],
         signatures = [SUPPORT_GROUP],
         content = {
-            'title': 'Matching Status [{note_id}]'.format(note_id=note.id),
+            'title': 'Paper Matching Setup Status',
```
melisabok, line 46: I think we can remove the `title` field from the invitation. The UI automatically shows the invitation id as status.

openreview/venue_request/process/matchingProcess.py
```diff
@@ -36,21 +36,22 @@ def process(client, note, invitation):
...
             'comment': ''
         }
     )
 
     if matching_status.get('error'):
-        error_status = f'''{len(matching_status.get('error'))} error(s): {matching_status.get('error')}'''
-        comment_note.content['comment'] += error_status
+        error_status = f'''{len(matching_status.get('error'))} error(s):
```
melisabok, line 52: if there could be more than one error,`errors_status` could be an array of strings.

melisabok: The paper matching status shouldn't have this text `To check references for the note: https://api.openreview.net/references?id={note.id}`

openreview/venue_request/process/recruitmentProcess.py
```diff
@@ -69,31 +70,29 @@ def process(client, note, invitation):
...
-        invitation = note.invitation.replace('Recruitment', 'Comment'),
+        invitation = note.invitation.replace('Recruitment', 'Recruitment_Status'),
         forum = note.forum,
         replyto = note.id,
         readers = [conference.get_program_chairs_id(), SUPPORT_GROUP],
         writers = [],
         signatures = [SUPPORT_GROUP],
         content = {
-            'title': f'Recruitment Status [{note.id}]',
+            'title': f'Recruitment Status',
```
melisabok, line 80: same here, no need to have a title

openreview/venue_request/process/recruitmentProcess.py
```diff
@@ -69,31 +70,29 @@ def process(client, note, invitation):
...
-
-        comment_note.content['comment'] += f'''
-Error: {error_status}
+        error_status = f'''No recruitment invitation was sent to the following users due to the error(s) in the recruitment process:
+```python
+{json.dumps(recruitment_status.get('errors'), indent=2)}
+```
 '''
 
+        comment_note.content['error'] = error_status
```
melisabok, line 96:
```suggestion
        recruitment_status_note.content['error'] = recruitment_status['errors']
```

openreview/venue_request/process/revisionProcess.py
```diff
@@ -125,25 +127,28 @@ def process(client, note, invitation):
...
+            invitation=SUPPORT_GROUP + '/-/Request' + str(forum_note.number) + '/Error_Status',
             forum=forum_note.id,
             replyto=forum_note.id,
             readers=comment_readers,
             writers=[SUPPORT_GROUP],
             signatures=[SUPPORT_GROUP],
             content={
-                'title': '{invitation} Status [{note_id}]'.format(invitation=invitation_type.replace("_", " "), note_id=note.id),
-                'comment': error_status
+                'title': '{invitation} Process Failed'.format(invitation=invitation_type.replace("_", " ")),
```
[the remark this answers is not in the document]
melisabok: we could add a new field called `reference_url` to store the reference url and don't show it in the comment.

openreview/venue_request/process/revisionProcess.py
```diff
@@ -125,25 +127,28 @@ def process(client, note, invitation):
...
-        print("Following error in the process function was posted as a comment:")
-        print(traceback.format_exc())
-
         forum_note = client.get_note(note.forum)
 
+        from openreview import OpenReviewException
+        error_status = json.dumps(e.args[0], indent=2) if isinstance(e, OpenReviewException) else repr(e)
         comment_note = openreview.Note(
-            invitation=SUPPORT_GROUP + '/-/Request' + str(forum_note.number) + '/Comment',
+            invitation=SUPPORT_GROUP + '/-/Request' + str(forum_note.number) + '/Error_Status',
```
melisabok, line 135:
```suggestion
            invitation=SUPPORT_GROUP + '/-/Request' + str(forum_note.number) + '/Stage_Error_Status',
```

openreview/venue_request/venue_request.py
```diff
@@ -596,6 +596,7 @@ def __init__(self, client, support_group_id, super_user):
         self.support_process = os.path.join(os.path.dirname(__file__), 'process/supportProcess.js')
         self.support_pre_process = os.path.join(os.path.dirname(__file__), 'process/request_form_pre_process.py')
         self.comment_process = os.path.join(os.path.dirname(__file__), 'process/commentProcess.js')
+        self.invitation_status_process = os.path.join(os.path.dirname(__file__), 'process/invitationStatusProcess.py')
```
melisabok, line 599:
```suggestion
        self.invitation_status_process = os.path.join(os.path.dirname(__file__), 'process/status_process.py')
```

openreview/venue_request/process/deployProcess.py
purujitgoyal: Done.

openreview/venue_request/process/deployProcess.py
```diff
@@ -404,3 +404,59 @@ def process(client, note, invitation):
...
+        },
+        signatures=['~Super_User1']
+    ))
+
+    client.post_invitation(openreview.Invitation(
+        id=SUPPORT_GROUP + '/-/Request' + str(forum.number) + '/Error_Status',
+        super=SUPPORT_GROUP + '/-/Error_Status',
+        reply={
+            'forum': forum.id,
+            'referent': forum.id,
```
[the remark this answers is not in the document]
purujitgoyal: Done.

openreview/venue_request/process/matchingProcess.py
purujitgoyal: We need `title` when sending email to the PCs.

melisabok: Could you have a different process function per invitation? I do think we need to customize the email message for each status case.

purujitgoyal: okay

openreview/venue_request/process/matchingProcess.py
purujitgoyal: Done.

openreview/venue_request/process/revisionProcess.py
purujitgoyal: Done.

openreview/venue_request/process/revisionProcess.py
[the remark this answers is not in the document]
purujitgoyal: Added `reference_url`

openreview/venue_request/venue_request.py
purujitgoyal: Done.

openreview/venue_request/venue_request.py
purujitgoyal: It's a dictionary actually, do you want me to combine all the lists?

openreview/venue_request/process/error_status_process.py
```diff
@@ -0,0 +1,13 @@
+def process(client, note, invitation):
+    GROUP_PREFIX = ''
+    SUPPORT_GROUP = GROUP_PREFIX + '/Support'
+
+    forum_note = client.get_note(note.forum)
+    subject = f'''{note.content['title']} for service: {forum_note.content['title']}'''
+    message = f'''A comment was posted to your service request.
```
melisabok, line 7: Given you are going to have a process function per invitation you can customize the email text like: 'An error occurred while running the stage... for your service request.'

openreview/venue_request/venue_request.py
purujitgoyal: It's a dictionary ([here](https://github.com/openreview/openreview-py/blob/717c9e7927d5d019a0161fcd197cc47f210ee230/openreview/conference/builder.py#L1251)). It maps `group_id` to the list of profiles. ([here](https://github.com/openreview/openreview-py/blob/717c9e7927d5d019a0161fcd197cc47f210ee230/openreview/conference/builder.py#L1376))

melisabok: I see, because the user could come from different groups. 

How does the UI render this value? the field type should `value-json` instead of `value-regex`.

openreview/venue_request/process/matchingProcess.py
```diff
@@ -23,62 +23,62 @@ def process(client, note, invitation):
...
         else:
-            matching_status['error'] = [str(e)]
+            matching_status['error'] = str(e)
 
     print("Following error in the process function was posted as a comment:")
     print(traceback.format_exc())
 
     comment_note = openreview.Note(
-        invitation = note.invitation.replace('Paper_Matching_Setup', 'Comment'),
+        invitation = note.invitation.replace('Paper_Matching_Setup', 'Paper_Matching_Setup_Status'),
```
melisabok, line 39: All the existing venues must have this invitation created so we need to keep backward compatibility if we don't want to create these invitations for previous venues. 

I suggest to check if the Paper_Matching_Setup_Status invitation exists then use it otherwise keep using the comment invitation.

openreview/venue_request/process/remindRecruitmentProcess.py
```diff
@@ -34,28 +35,28 @@ def process(client, note, invitation):
     )
 
     comment_note = openreview.Note(
-        invitation = note.invitation.replace('Remind_Recruitment', 'Comment'),
+        invitation = note.invitation.replace('Remind_Recruitment', 'Remind_Recruitment_Status'),
```
melisabok, line 38: also here, keep backward compatibility.

openreview/venue_request/process/matchingProcess.py
purujitgoyal: It's best to make invitations for the existing venues. If we switch to `Comment` invitation, we will have to make a different new note which matches the `Comment` invitation structure.

melisabok: Ok, please remember to prepare the script to create the invitations for all the venues before we deploy this change in the live site.

openreview/venue_request/venue_request.py
```diff
@@ -1292,3 +1418,132 @@ def setup_matching(self):
...
+                            'markdown': True
+                        },
+                        'reference_url': {
+                            'order': 4,
+                            'value-regex': '.{1,500}',
+                            'description': 'URL to check references',
+                            'required': False,
+                            'hidden': True
+                        },
+                        'stage_name': {
```
melisabok, line 1540: If you are going to put the stage name in the title, I don't think this field is needed.

purujitgoyal: I added this so that we can have flexibility with the Title. I can change it back.

melisabok: ok, let's keep it then.

melisabok:
The changes looks good to me. I notice the issue about showing too many buttons for the super user, I'm not sure how to solve it. 

Maybe we should stop using the super user and use our own account. 

@celestemartinez could you do a final review?

openreview/venue_request/process/matchingProcess.py
```diff
@@ -23,62 +23,62 @@ def process(client, note, invitation):
...
     )
 
     if matching_status.get('error'):
-        error_status = f'''{len(matching_status.get('error'))} error(s): {matching_status.get('error')}'''
-        comment_note.content['comment'] += error_status
-        comment_note.content['comment'] += f'''
-
-To check references for the note: https://api.openreview.net/references?id={note.id}
-'''
+        error_status = f'''{len(matching_status.get('error'))} error(s): \n
```
celestemartinez, line 52: Looks like  `{len(matching_status.get('error'))}` here is just getting the lenght of the error string? It shouldn't say there are 136 error(s) (and the error string has 136 characters)

Screen Shot 2022-04-01 at 2 09 52 PM

purujitgoyal: Removed the count of errors as it was always 1.

purujitgoyal:
> PCs can still see the 'Stage Error Status' button, is this intended? I assume we don't want PCs to see this button, as it is only to post error status?

It's not intended. How to fix this?

celestemartinez:
> > PCs can still see the 'Stage Error Status' button, is this intended? I assume we don't want PCs to see this button, as it is only to post error status?
> 
> It's not intended. How to fix this?

I think changing the invitees to only support would fix this

PR #1165, opened 2022-03-07
request form paper matching fixes

celestemartinez:
- Always creates Paper_Matching_Setup invitation on deployment
- Include instructions on what to do if there are users with no profiles
- Fix link to reviewers group
- Set allow_zero_score_assignements == Yes as default


melisabok:
Thanks!

nlbahy (approved):

PR #1167, opened 2022-03-08
Fix for Comment Invitation

purujitgoyal:


melisabok:
thanks for the quick fix, before I merge this. Could you add this code after the comment invitation is updated?

```
replies = client.get_notes(forum=forum.id, invitation=SUPPORT_GROUP + '/-/Request' + str(forum.number) + '/Comment')
for reply in replies:
     reply.readers = readers
     client.post_note(reply)
```

Please check in one of the tests the existing comments are updated with the new list of readers.

purujitgoyal:
> thanks for the quick fix, before I merge this. Could you add this code after the comment invitation is updated?
> 
> ```
> replies = client.get_notes(forum=forum.id, invitation=SUPPORT_GROUP + '/-/Request' + str(forum.number) + '/Comment')
> for reply in replies:
>      reply.readers = readers
>      client.post_note(reply)
> ```
> 
> Please check in one of the tests the existing comments are updated with the new list of readers.

Isn't `deployProcess.py` run only once at the start? or did you mean to add this somewhere else?

melisabok:
> Isn't deployProcess.py run only once at the start? or did you mean to add this somewhere else?

Deploy process runs when we deploy it but the comment invitation exists since the request is posted. Some PCs may post a comment asking questions before the deployment.

purujitgoyal:
I tried posting a comment before deployment but it's giving me this error: `"ForbiddenError: User Support is not writer of the group tom@mail.com`.

@melisabok is this correct?
```
        client.post_note(openreview.Note(
            content={
                'title': 'Urgent',
                'comment': 'Please deploy ASAP.'
            },
            forum=request_form_note.forum,
            invitation='{}/-/Request{}/Comment'.format(venue.support_group_id, request_form_note.number),
            readers=[
                support_group_id,
                'new_test_user@mail.com',
                'tom@mail.com'
            ],
            replyto=None,
            signatures=['~NewFirstName_User1'],
            writers=['~NewFirstName_User1']
        ))
        helpers.await_queue()
```

melisabok:
> I tried posting a comment before deployment but it's giving me this error: "ForbiddenError: User Support is not writer of the group tom@mail.com.

I think this error is happening when it tries to send a mail to all the readers of the comment. The process function runs with the permissions of the signature of the invitation, in this case is openreview.net/Support and it seems that doesn't have permissions to send the message. 

This is currently working in the live site. Is tom@mail.com listed as a PC in the request form?

Can you try to post a comment using the UI and check if the error happens too?

purujitgoyal:
> This is currently working in the live site. Is [tom@mail.com](mailto:tom@mail.com) listed as a PC in the request form?
Yes, tom@mail.com is listed as a PC.

> Can you try to post a comment using the UI and check if the error happens too?
There was no error.

melisabok:
> There was no error.

The process function didn't fail?

melisabok:
can you compare how the browser is doing the POST /notes? then your test should be doing the same.

melisabok:
Please push your changes and I can take a look.

purujitgoyal:
> can you compare how the browser is doing the POST /notes? then your test should be doing the same.

`process function not executed` I get this in the logs when I post a comment from UI.

purujitgoyal:
Not sure why but all tests are passing on circleci, even with the new change of posting a comment before deployment.

melisabok:
I found an [issue](https://github.com/openreview/openreview-api-v1/issues/2576) in the API with the recipient validation. 

I realized this test should always fail because the support group doesn't have any permission on the PC emails to send them a message. The rules to be able to message them is:

- Being writer of at least one group where the PC email is a member.
- Being writer of at least one note where the PC emails is an authorid.

I think the fix here is to use "openreview.net" or "~Super_User1" as signature of the Comment invitation so the process function can send emails to any email address.

cc @carlosmondra

melisabok:
There is another issue here: https://github.com/openreview/openreview-py/pull/1167/files#diff-e3ae2f10352929825bba9040318c444e8bf4b97395b036eec532aacae837bf72R144

Could you see the writers as "Support, PC emails and contact Emails"?

That's the reason the test is failing when you run it individually.

purujitgoyal:
I tried the suggested changes @melisabok, but it's still failing.

melisabok:
Can you push your changes?

openreview/venue_request/process/supportProcess.js
```diff
@@ -32,8 +33,8 @@ function(){
             values: note.content['program_chair_emails'].concat([SUPPORT_GROUP])
         }
       },
-      writers: [SUPPORT_GROUP],
-      signatures: [SUPPORT_GROUP]
+      writers: [SUPER_USER],
```
melisabok, line 36: no need to change the writers, only the signatures.

melisabok (approved):

PR #1170, opened 2022-03-11
Fix/tmlr adjustments

melisabok:
- Show Reviewer Responsibility Acknowledgemente in the PC console.
- Fix text in Responsibility task.
- Fix typo Responsibility task name.
- Enable Revisions for authors.

### DIFF
--- a/openreview/journal/invitation.py
+++ b/openreview/journal/invitation.py
@@ -50,7 +50,7 @@ class InvitationBuilder(object):
         self.set_meta_invitation()
         self.set_ae_recruitment_invitation()
         self.set_reviewer_recruitment_invitation()
-        self.set_reviewer_responsability_invitation()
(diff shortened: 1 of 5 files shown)


celestemartinez (approved):

openreview/journal/invitation.py
```diff
@@ -2581,33 +2578,6 @@ def set_revision_submission(self, note):
                                 'markdown': True
```
zbialecki, line 2578: I think we should disable markdown here. There are actually 2 other locations that define the abstract field that should be changed as well.

zbialecki: I fixed this in [84282c8](https://github.com/openreview/openreview-py/pull/1170/commits/84282c808a1d64ae705b00f455e1b491b2248991)

PR #1173, opened 2022-03-15
fix/ Add delay to reviewer assignment process

xkopenreview:
add delay 5 seconds to reviewer assignment process so that reviewer won't be notified if the edge is removed within 5 seconds

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -1687,8 +1687,8 @@ class OpenReviewClient(object):
         response = self.__handle_response(response)
         return response.json()
 
-    def request_expertise(self, name, group_id, paper_invitation, exclusion_inv=None, model=None, baseurl=None):
-
+    def request_expertise(self, name, group_id, paper_invitation, alternate_match_group = None, exclusion_inv=None, model=None, baseurl=None):
+        
         # Build entityA from group_id
         entityA = {
             'type': 'Group',
@@ -1697,12 +1697,18 @@ class OpenReviewClient(object):
         if exclusion_inv:
             expertise = {'exclusion': { 'invitation': exclusion_inv }}
             entityA['expertise'] = expertise
-
-        # Build entityB from paper_invitation
-        entityB = {
-            'type': 'Note',
-            'invitation': paper_invitation
-        }
+        
+        # Build entityB from alternate_match_group or paper_invitation
+        if alternate_match_group:
+            entityB = {
+                'type': 'Group',
+                'memberOf': alternate_match_group
+            }
+        else:
+            entityB = {
+                'type': 'Note',
+                'invitation': paper_invitation
+            }
 
         expertise_request = {
             'name': name,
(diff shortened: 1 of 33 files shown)


xkopenreview:
this pr should work with openreview-v1 branch

this pr should also fix the issue that assignment acknowledge invitation stay expired when a reviewer is assigned,unassigned and assigned again

openreview/journal/invitation.py
```diff
@@ -492,6 +520,7 @@ def set_reviewer_assignment_acknowledgement_invitation(self):
                     'signatures': { 'const': [editors_in_chief_id] },
                     'maxReplies': { 'const': 1 },
                     'duedate': { 'const': '${params.duedate}' },
+                    # 'expdate': { 'const': None },
```
xkopenreview, line 523: currently this line will cause invitation validation to fail
discussed with @carlosmondra and he will make the changes to allow const null

when the changes are added, unexpire invitation can be removed.

openreview/api/client.py
```diff
@@ -271,7 +271,7 @@ def get_edge(self, id):
         return: Edge object with its information
         :rtype: Edge
         """
-        response = requests.get(self.tags_url, params = {'id': id}, headers = self.headers)
+        response = requests.get(self.edges_url, params = {'id': id, 'trash': 'true' if trash == True else 'false'}, headers=self.headers)
```
melisabok, line 274: I don't think you need to translate the value.

xkopenreview: do you mean params = {'id': id, 'trash': trash} ?
this will fail api validation that trash must be a boolean value

melisabok: No need of translation, the API should receive a boolean, take a look at this one:

https://github.com/openreview/openreview-py/blob/master/openreview/openreview.py#L1012

melisabok (approved):

tests/test_journal.py
```diff
@@ -497,8 +514,28 @@ def test_submission(self, journal, openreview_client, test_client, helpers):
...
+            nonreaders=[f"{venue_id}/Paper1/Authors"],
+            writers=[venue_id, f"{venue_id}/Paper1/Action_Editors"],
+            signatures=[f"{venue_id}/Paper1/Action_Editors"],
+            head=note_id_1,
+            tail='~David_Belanger1',
+            weight=1
+        ))
 
+         # wait for process function delay (5 seconds) and check email has been sent
+        time.sleep(6)
```
xkopenreview, line 538: @melisabok in e0db1c32ff703053bd40401c452125e60ada11ac delay is change from 5 to 30
so for the test to pass, Ln 537 and 554 will need to change it to wait at least 30s
maybe it can be shorter than 30 seconds?

melisabok: Let me fix the test, sorry about this

openreview/api/client.py
```diff
@@ -2077,23 +2074,10 @@ def __init__(self,
         self.details = details
         self.reply_forum_views = reply_forum_views
         self.web = None
```
xkopenreview, line 2076: i think this line should be removed

xkopenreview:
@melisabok please take a look at [1efc0de](https://github.com/openreview/openreview-py/pull/1173/commits/1efc0de5374a4b1fec3440f5abab7db82fa137d7)
if the change is correct i think this pr can be merged.

xkopenreview:
the failure seems to be very random

PR #1177, opened 2022-03-21
set paper revs/acs/sacs as default readers unless venue does bidding

### LINKED ISSUE #1021: Submission readers should be assigned revs, ACs, SACs
If bidding is not enabled, default readers of private submissions should be assigned reviews, assigned Acs, assigned SACs.

celestemartinez:
Sets paperX revs, ACs and SACs as default readers unless PCs selected 'Reviewer Bid Scores' or 'Reviewer Recommendation Scores'.

I am open to other names for the parameter `bidding_enabled`, since there might be some venues who would like all reviewers,ACs and SACs to be readers of all papers (independently of whether they have bidding or not)

Fixes #1021 

Edit: I removed the `bidding_enabled` param and have added `submission_readers` to the request form.


tests/test_eccv_conference.py
```diff
@@ -248,7 +248,8 @@ def conference(self, client):
             withdrawn_submission_reveal_authors=False,
             email_pcs_on_withdraw=True,
             desk_rejected_submission_public=False,
-            desk_rejected_submission_reveal_authors=False)
+            desk_rejected_submission_reveal_authors=False,
+            bidding_enabled=True)
```
melisabok, line 252: is there a test to verify when bidding_enabled is False the readers are PaperX groups?

openreview/conference/builder.py
```diff
@@ -2257,12 +2257,15 @@ def set_submission_stage(
             email_pcs_on_desk_reject=True,
             author_names_revealed=False,
             papers_released=False,
-            readers=None
+            readers=None,
+            bidding_enabled=False
```
melisabok, line 2261: We are having too many parameters to decide the readers of the notes: public, papers_release, readers and bidding _enabled. I wish we can merge this into one.

I would prefer to use readers param and don't add a new param. Also don't you need to check if there are SAC or AC in the conference? (I guess this was working like this before but we should fix it)

celestemartinez: Sounds good. 

submission_stage.get_readers() checks if the conference has SACs and ACs

openreview/conference/helpers.py
```diff
@@ -125,6 +125,11 @@ def get_conference_builder(client, request_form_id, support_user='OpenReview.net
 
     name = note.content.get('submission_name', 'Submission').strip()
 
+    readers = None
+
+    if 'Reviewer Bid Scores' in note.content.get('Paper Matching', '') or 'Reviewer Recommendation Scores' in note.content.get('Paper Matching', ''):
+        readers = [openreview.SubmissionStage.Readers.SENIOR_AREA_CHAIRS, openreview.SubmissionStage.Readers.AREA_CHAIRS, openreview.SubmissionStage.Readers.REVIEWERS]
```
melisabok, line 131: I wonder if we should allow the PCs to define the readers in the "Post Submission Stage" instead of us deciding them. 

I'm saying this because in the future I would like to remoce the field "Paper Matching" from the request form and create BidStage always. The PCs can decide if they want to use it or not.

celestemartinez: The only problem I see with this is that not all PCs use the 'Post Submission' button, I think most of the times `setup_post_submission_stage()` is being called from the process fucntion of one of the stages, should we remove it and force PCs to use the button?

melisabok (approved):

PR #1178, opened 2022-03-23
increase selenium timeout

melisabok:


zbialecki:
You could try adding changing the command to build openreview-web to `NODE_ENV=production SUPER_USER=openreview.net npm run start` (the command to run the server should be modified in the same way). This may enable some React and Next.js optimizations.

.circleci/config.yml
```diff
@@ -59,13 +59,13 @@ jobs:
             cd ~/openreview-web
             npm install
             cp .env.example .env
```
zbialecki, line 61:
```suggestion
            cp .env.example .env.local
```

PR #1181, opened 2022-03-24
Fix bug in recruit invitation webfield

zbialecki:
Fix bug in webfield template causing JS errors


zbialecki:
Closing this in favor of https://github.com/openreview/openreview-web/pull/911

PR #1187, opened 2022-03-25
Fix/params serialization

carlosmondra:
This PR formats the boolean params from python to json. For example: True -> 'true' and False -> 'false'

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -8,6 +8,7 @@ if sys.version_info[0] < 3:
 else:
     string_types = [str]
 
+from .. import tools
 import requests
 import pprint
 import os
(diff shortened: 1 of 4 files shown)


melisabok:
I like the changes, thank you so much for doing this!

A little suggestion: format_params doesn't need to be a tool function, it could be a function of the Client class, nobody will use it expect the client.

carlosmondra:
> I like the changes, thank you so much for doing this!
> 
> A little suggestion: format_params doesn't need to be a tool function, it could be a function of the Client class, nobody will use it expect the client.

I see, I had done it this way to avoid repeating code in the client of the old and new API. Is this OK?

melisabok (approved):

PR #1189, opened 2022-03-31
Fix/wait for element

carlosmondra:
This PR waits for selenium elements to be loaded before asserting their presence.


melisabok (approved):
Thanks for doing this!!! 🙏

PR #1193, opened 2022-04-06
Fix/short name double quotes

melisabok:


openreview/conference/invitation.py
```diff
@@ -1691,8 +1691,8 @@ def set_reviewer_recruiter_invitation(self, conference, options = {}):
                 pre_content = pre.read()
                 post_content = post.read()
                 pre_content = pre_content.replace("REVIEWERS_REGEX = ''", "REVIEWERS_REGEX = '" + conference.get_committee_id(name=options.get('reviewers_name', 'Reviewers'), number='.*') + "'")
-                pre_content = pre_content.replace("var CHECK_DECLINE = False;", "var CHECK_DECLINE = True;")
-                post_content = post_content.replace("SHORT_PHRASE = ''", "SHORT_PHRASE = '" + conference.get_short_name() + "'")
+                pre_content = pre_content.replace("CHECK_DECLINE = False;", "CHECK_DECLINE = True;")
```
celestemartinez, line 1694: Just noticed that this is never being replaced because this is a python pre-process, so there is no `;` at the end of `CHECK_DECLINE = False`

Also noticed in some places you are adding a semi-colon at the end of `SHORT_PHRASE` for python process functions

nlbahy (approved):

openreview/conference/invitation.py
```diff
@@ -1691,8 +1691,8 @@ def set_reviewer_recruiter_invitation(self, conference, options = {}):
...
+                post_content = post_content.replace("SHORT_PHRASE = ''", f'SHORT_PHRASE = "{conference.get_short_name()}";')
```
celestemartinez, line 1695: Extra `;` here for example

melisabok: thanks for the catch, copy and paste :(

openreview/conference/invitation.py
melisabok: yeah, I saw the value `var` in the python code so I removed it. I will remove the `;`. Do you remember what is this for `CHECK_DECLINE`?

celestemartinez: it's so a user can't decline the invitation after being assigned to a paper

melisabok: we don't have a test for this? the test should have failed. 

I could add one here.

celestemartinez (approved):

openreview/conference/invitation.py
melisabok: my next question is: what do we have a boolean that we are always setting to true. Is there a case when we don't have to check this?

celestemartinez: there is the test in [test_eccv_conference.py](https://github.com/openreview/openreview-py/pull/1027/files#diff-6c3e5b84f16f0115a679610fa37e2f87d687bd1780efc911b3e6bc59fe51eaa0) that should have failed?

melisabok: it is probably showing another error message. We should assert the error message to be "You have already been assigned to a paper. Please contact the paper area chair or program chairs to be unassigned."

I will take a look.

melisabok: The email to find the invitation message was wrong, I fixed it and added an assertion of the error message:

[ce52209](https://github.com/openreview/openreview-py/pull/1193/commits/ce52209bf897078c20b64933f48dd02f18edc36b)

PR #1197, opened 2022-04-08
add option to hide rejected papers

celestemartinez:
Add the option to hide rejected papers to the request form and builder

- Post decision stage expires post_submission stage and uses `submission_readers` to set readers of papers


celestemartinez:
@melisabok, I noticed now that many tests use `post_decision_stage()` directly (this method is now `post_decision_stage_old()` for now until I figure out what to do with it). The problem about the new `post_submission_stage()` is that it uses `submission_readers` to set the readers of the submissions, so to use it directly from the builder we would need to first set the submission stage readers and then call `post_submission_stage()`. 

Let me know what you think.

melisabok:
I would prefer to set the submission stage with the new submission_readers before running post_decision_stage
 
or

Support submission_readers as an optional parameter in post_decision_stage so the implementation can override the submission_readers from submission_stage before calling to get_readers

I think I prefer the latter.

openreview/conference/builder.py
```diff
@@ -1477,62 +1477,123 @@ def get_attachment_file(submission):
...
+                submission.content['_bibtex'] = tools.get_bibtex(
+                    submission,
+                    venue_fullname=self.name,
+                    year=str(self.year),
+                    url_forum=submission.forum,
+                    accepted=note_accepted,
+                    anonymous=False
+                )
+            #add venue_id if note accepted
+            if note_accepted:
```
melisabok, line 1525: I think we should add a `venue_id` for rejected papers like `Submitted to {venue_id}`. What do you think?

celestemartinez: Done

openreview/conference/helpers.py
```diff
@@ -138,7 +139,7 @@ def get_conference_builder(client, request_form_id, support_user='OpenReview.net
     create_review_invitation = create_groups and note.content.get('make_reviews_public', None)
 
     author_names_revealed = 'Reveal author identities of all submissions to the public' in note.content.get('reveal_authors', '') or 'Reveal author identities of only accepted submissions to the public' in note.content.get('reveal_authors', '')
-    papers_released = 'Release all submissions to the public'in note.content.get('release_submissions', '') or 'Release only accepted submission to the public' in note.content.get('release_submissions', '')
+    papers_released = 'Release all submissions to the public'in note.content.get('release_submissions', '') or 'Release only accepted submission to the public' in note.content.get('release_submissions', '') or 'Everyone for accepted submissions and only assigned program committee for rejected submissions' in note.content.get('submission_readers', '')
```
celestemartinez, line 142: @melisabok how should we define `papers_released` now? I think the only place we need this for is to not override `venueid` and `venue` for blind submissions, so perhaps we can just check in `create_blind_submissions()` if these fields exist already and add them to content? Since we are using `submission_readers` everywhere now we can't really distinguish if `post_decision` has been called or not

used here: https://github.com/openreview/openreview-py/blob/c8854ef6cf6a8892322b5370d20d690d8bd5fe6a/openreview/conference/builder.py#L887

melisabok: That's a good idea. Let's remove `papers_released` and keep the values `venue`  and `venueid` if they already exist, I think `_bibtex` too?

openreview/conference/builder.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+from math import e
```
melisabok, line 2: why do we need this import?

openreview/conference/builder.py
```diff
@@ -827,7 +835,7 @@ def create_blind_submissions(self, hide_fields=[]):
 
                 blind_note = self.client.post_note(blind_note)
 
-                if self.submission_stage.public:
+                if self.submission_stage.public and blind_content['_bibtex'] is None:
```
melisabok, line 838: I think what you need to check here is that `venue` doesn't exist. When `venue` exists, it means the post decision stage was ran and all the submissions were marked with a venue value and a bibtex entry including the rejected submissions. 

When the bibtex was set during the post decision stage we shouldn't override it if create_blind_submissions is called after that. Does it make sense?

openreview/conference/builder.py
celestemartinez: we don't

openreview/conference/builder.py
celestemartinez: So what I did was always set the bibtex to the bibtex set in post_decision if it exists, therefore if there is already a bibtex in `blind_content` it will not be overwritten (only calculate bibtex if it is None), is this wrong?

celestemartinez: I do this earlier so that if nothing is being changed in the blind submission, then we won't even get this far because of this line:

`if not existing_blind_note or existing_blind_note.content != blind_content or existing_blind_note.readers != blind_readers:`

melisabok: If we ran create_blind_submissions twice and there was a change in the authors list then the bibtex is not being updated. I'm confused about this, let's have a call.

PR #1198, opened 2022-04-08
EIC console: add under dicussion tab

melisabok:
and show counters for under discussion and under decision notes.


zbialecki (approved):
Looks good!

PR #1208, opened 2022-04-14
Fix tests: add sort option to all the get_notes calls

melisabok:
- Add the sort parameters to all the get_notes calls from the tests.
- Add the sort parameters to all the get_notes calls from the webfields.

### DIFF
--- a/openreview/conference/templates/tabsConferenceWebfield.js
+++ b/openreview/conference/templates/tabsConferenceWebfield.js
@@ -99,7 +99,8 @@ function load() {
   if (user && !_.startsWith(user.id, 'guest_')) {
     activityNotesP = Webfield.api.getSubmissions(WILDCARD_INVITATION, {
       pageSize: PAGE_SIZE,
-      details: 'forumContent,invitation,writable'
+      details: 'forumContent,invitation,writable',
+      sort: 'tmdate:desc'
     });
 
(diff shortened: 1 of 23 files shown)


carlosmondra (approved):

openreview/conference/templates/tabsConferenceWebfield.js
```diff
@@ -99,7 +99,8 @@ function load() {
   if (user && !_.startsWith(user.id, 'guest_')) {
     activityNotesP = Webfield.api.getSubmissions(WILDCARD_INVITATION, {
       pageSize: PAGE_SIZE,
-      details: 'forumContent,invitation,writable'
+      details: 'forumContent,invitation,writable',
+      sort: 'tmdate:desc'
```
zbialecki, line 103: I thought you agreed that it didn't really matter what order the submissions were in on the venue homepage. I'm ok with having it be sorted in this way, but it would reduce the load on mongo to not sort.

melisabok: this query is to get the latest activity so I think we need to sort them here.

zbialecki: Oh I see. In that case it looks good!

zbialecki (approved):

PR #1209, opened 2022-04-15
Bulk Decision Upload

purujitgoyal:


openreview/conference/builder.py
```diff
@@ -1533,6 +1534,28 @@ def is_release_authors(is_note_accepted):
...
 
+    def post_decisions(self, decisions_file):
+        with open(decisions_file) as file_handle:
+            decisions = [row for row in csv.reader(file_handle)]
+        for paper_id, decision in decisions:
+            paper_note = self.client.get_notes(id=paper_id)[0]
+            self.client.post_note(openreview.Note(
+                invitation='{}/Paper{}/-/Decision'.format(self.id, paper_note.number),
+                writers=[self.get_program_chairs_id()],
+                readers=[self.get_program_chairs_id(), self.get_senior_area_chairs_id(), self.get_area_chairs_id()],
```
melisabok, line 1545: usually decision reader are the PCs and authors. You need to set the readers based on what the decision invitation says and the decision invitation is created based on the settings `released_to_authors`, etc.

I would move this code to the DecisionStage class if possible.

openreview/conference/builder.py
```diff
@@ -1533,6 +1534,28 @@ def is_release_authors(is_note_accepted):
...
+                nonreaders=['{}/Paper{}/Authors'.format(self.id, paper_note.number)],
+                signatures=[self.get_program_chairs_id()],
+                content={
+                    'title': 'Paper Decision',
+                    'decision': decision,
+                    'comment': 'Good paper. I like!',
```
melisabok, line 1551: We could offer a third column in the CSV file to add a comment, this field is optional.

tests/test_venue_request.py
```diff
@@ -1234,25 +1242,67 @@ def test_venue_decision_stage(self, client, test_client, selenium, request_page,
...
+            )
+        )[0]
+
+        decision_note.content = {
+            'title': 'Paper Decision',
+            'decision': 'Accept',
+            'comment':  'Good paper. I like!',
+            'suggestions': 'Add more results for camera ready.'
+        }
+        decision_note = test_client.post_note(decision_note)
```
melisabok, line 1305: shouldn't be the decision note posted by the process function?

melisabok: here you need to check if decision notes are posted and they match with the content described in decisions.csv

openreview/conference/builder.py
```diff
@@ -1533,6 +1536,35 @@ def is_release_authors(is_note_accepted):
             self.set_homepage_decisions(decision_heading_map=decision_heading_map)
         self.client.remove_members_from_group('active_venues', self.id)
 
+    def post_decisions(self, decisions_file):
+        decisions_data = list(csv.reader(StringIO(decisions_file.decode()), delimiter=","))
+
+        def post_decision(paper_decision):
+            try:
+                paper_id, decision, comment = paper_decision
+            except IndexError:
```
melisabok, line 1545: what is an IndexError?

tests/test_venue_request.py
purujitgoyal: I can add that check before this. I kept this test to check if a single decision note can also be posted.

celestemartinez:
should we check if the decision has already posted (and update it if it does). Running the decision stage multiple times with a file will post multiple decision notes

openreview/conference/builder.py
purujitgoyal: It should be `ValueError`, but I did this to keep the `comment` field optional, so if the decisions csv only have paperIDs and decisions it will run without any error.

melisabok: When is it happening the `ValueError`?  AFAIK `paper_decision` is a tuple with three values and you are assign them to three different variable, right?

melisabok:
> should we check if the decision has already posted (and update it if it does)? Running the decision stage multiple times with a file will post multiple decision notes

Good point, decision notes must be updated with the new value comings from the CSV file. There should be one decision per submission.

openreview/conference/builder.py
celestemartinez: I believe since the comment if optional, then sometimes paper_decision will only have two values which we will try to assign to three variables

celestemartinez: I get this error when only passing paper_id and decision:

`ValueError('not enough values to unpack (expected 3, got 2)')`

purujitgoyal: Yes, I'll update it to `ValueError` in place of `IndexError`.

melisabok: ohhhh I see, it can be simpler to treat `paper_decision` as an array. We need to make sure that `paper_id` and `decision` are always present. So the array could be length 2 or 3.

celestemartinez: Also, might be a good idea to strip() the decisions? If an extra space is added before or after, it will not match the invitation definition

tests/test_venue_request.py
```diff
@@ -1258,10 +1264,61 @@ def test_venue_decision_stage(self, client, test_client, selenium, request_page,
...
+        process_logs = client.get_process_logs(id=decision_stage_note.id)
         assert len(process_logs) == 1
         assert process_logs[0]['status'] == 'ok'
 
+        for sub in submissions:
+            sub_decision_note = test_client.get_notes(
+                invitation='{venue_id}/Paper{number}/-/Decision'.format(
+                    venue_id=venue['venue_id'], number=sub.number
+                )
+            )[0]
```
melisabok, line 1317: check the length of decisions is always 1. 

After this assertion you can run the Decision stage again with different decision types and assert that the decisions have been updated.

openreview/conference/builder.py
```diff
@@ -1533,6 +1536,62 @@ def is_release_authors(is_note_accepted):
...
+
+        def post_decision(paper_decision):
+            if len(paper_decision) == 3:
+                paper_id, decision, comment = paper_decision
+            elif len(paper_decision) == 2:
+                paper_id, decision = paper_decision
+                comment = ''
+            else:
+                raise ValueError(
+                    "Too many values provided in the decision file. Expected values are: paper_id, decision, comment"
```
melisabok, line 1559: is it possible that `len(paper_decision) == 0`? in this case should we say "Wrong values in the decision file. Expected values are: paper_id, decision, comment". 

I would also throw an OpenReviewException instead.

openreview/conference/builder.py
```diff
@@ -1533,6 +1536,62 @@ def is_release_authors(is_note_accepted):
...
+            paper_decision_note = decision_notes.get(paper_id, None)
+            if paper_decision_note:
+                paper_decision_note.content = {
+                    'title': 'Paper Decision',
+                    'decision': decision.strip(),
+                    'comment': comment,
+                }
+            else:
+                paper_decision_note = openreview.Note(
+                    invitation='{}/Paper{}/-/Decision'.format(self.id, paper_note.number),
```
melisabok, line 1577: please get the "Decision" world from `decisiont_stage.name` and the value "Paper" is parametrized so you can use the following:

```
self.get_invitation_id(name=decision_stage.name, number=paper_note.number)
```

melisabok:
@nlbahy do you have a real CSV file that you can use to test this PR? Can you share one with Purujit in order to run this with a real venue?

nlbahy:
> @nlbahy do you have a real CSV file that you can use to test this PR? Can you share one with Purujit in order to run this with a real venue?

I can check, but any csvs I'd have would be for venues that already posted their decisions. Is that okay?

melisabok:
> I can check, but any csvs I'd have would be for venues that already posted their decisions. Is that okay?

Yes, the code should re-upload the decisions. Purujit can make a few changes in the file and then make sure the decision is updated.

openreview/venue_request/process/revisionProcess.py
```diff
@@ -42,6 +42,11 @@ def process(client, note, invitation):
 
         elif invitation_type == 'Decision_Stage':
             conference.set_decision_stage(openreview.helpers.get_decision_stage(client, forum_note))
+            decisions_file = forum_note.content.get('upload_decisions')
```
melisabok, line 45: move this behavior to the `DecisionStage` class so we don't need to update the revision process function and the functionality keeps encapsulated.

openreview/venue_request/venue_request.py
```diff
@@ -500,6 +500,15 @@ def setup_decision_stage(self):
                 'required': False,
                 'description': 'Configure additional options in the decision form. Use lowercase for the field names and underscores to represent spaces. The UI will auto-format the names, for example: supplementary_material -> Supplementary Material. Valid JSON expected.'
             },
+            'upload_decisions': {
```
melisabok, line 503: What about calling this `decisions_file` or `final_decisions_file`?

openreview/conference/builder.py
```diff
@@ -1540,6 +1547,63 @@ def is_release_authors(is_note_accepted):
...
+        decision_notes = {
+            n.forum: n for n in self.client.get_all_notes(
+                invitation=self.get_invitation_id(self.decision_stage.name, '.*')
+            )}
+        paper_notes = {n.forum: n for n in self.get_submissions()}
+
+        def post_decision(paper_decision):
+            if len(paper_decision) < 2:
+                raise OpenReviewException(
+                    "Not enough values provided in the decision file. Expected values are: paper_id, decision, comment")
```
melisabok, line 1561: Could you add a test for this scenario?

openreview/conference/builder.py
```diff
@@ -1540,6 +1547,63 @@ def is_release_authors(is_note_accepted):
...
+            if len(paper_decision) > 3:
+                raise OpenReviewException(
+                    "Too many values provided in the decision file. Expected values are: paper_id, decision, comment"
```
melisabok, line 1564: Could you add a test for this scenario too?

openreview/conference/builder.py
```diff
@@ -1540,6 +1547,63 @@ def is_release_authors(is_note_accepted):
...
+                paper_id, decision, comment = paper_decision
+            else:
+                paper_id, decision = paper_decision
+                comment = ''
+
+            print(f"Posting Decision {decision} for Paper {paper_id}")
+            paper_note = paper_notes.get(paper_id, None)
+            if not paper_note:
+                raise OpenReviewException(
+                    f"Paper with ID: {paper_id} not found. Please check the submitted paperIDs."
```
melisabok, line 1576: same here. In this case, we are going to throw an error but we may have submitted the previous decisions. I think we should return this in the status message.

openreview/conference/builder.py
```diff
@@ -1540,6 +1547,63 @@ def is_release_authors(is_note_accepted):
...
+                        'decision': decision.strip(),
+                        'comment': comment,
+                    },
+                    forum=paper_note.forum,
+                    replyto=paper_note.forum
+                )
+            self.client.post_note(paper_decision_note)
+            print(f"Decision posted for Paper {paper_id}")
+
+        tools.concurrent_requests(post_decision, decisions_data)
```
melisabok, line 1604: I think we need to return the status of the decision upload given that some of the note id or decisions must be wrong. I suggest to keep track of the decisions with errors and the decisions with no errors and put this in the decision status note.

openreview/conference/builder.py
```diff
@@ -1540,6 +1547,63 @@ def is_release_authors(is_note_accepted):
...
+            else:
+                paper_decision_note = openreview.Note(
+                    invitation=self.get_invitation_id(name=self.decision_stage.name, number=paper_note.number),
+                    writers=[self.get_program_chairs_id()],
+                    readers=self.decision_stage.get_readers(conference=self, number=paper_note.number),
+                    nonreaders=self.decision_stage.get_nonreaders(conference=self, number=paper_note.number),
+                    signatures=[self.get_program_chairs_id()],
+                    content={
+                        'title': 'Paper Decision',
+                        'decision': decision.strip(),
```
melisabok, line 1595: I would also check if the decision is part of the decision options of the decision stage and report this in the decision status note.

openreview/conference/builder.py
```diff
@@ -340,6 +343,10 @@ def set_submission_revision_stage(self, stage):
 
     def set_decision_stage(self, stage):
         self.decision_stage = stage
+
+        if self.decision_stage.decisions_file is not None:
```
celestemartinez, line 347: Looks like we are trying to post decisions before running `__create_decision_stage()`, this will give an error if the decision invitations haven't been created yet (or if they run the decision stage again to make a change to the decision options for example)

celestemartinez, line 347: This might be minor because I am not sure in what scenario this would happen but:
if the PCs upload a decisions file and for some reason it does not work and then try to just run the decision stage without a file (clicking the X next to upload_decisions), then `self.decision_stage_file` will be `''` instead of `None`, and there will be an error when we try to get the file.

openreview/conference/builder.py
purujitgoyal: Done.

openreview/conference/builder.py
purujitgoyal: Done.

openreview/venue_request/venue_request.py
```diff
@@ -504,8 +504,72 @@ def setup_decision_stage(self):
...
+                'required': False,
+                'markdown': True
+            }
+        }
+
+        with open(self.venue_request.decision_upload_status_process, 'r') as f:
+            file_content = f.read()
+            file_content = file_content.replace("GROUP_PREFIX = ''", "GROUP_PREFIX = '" + self.venue_request.super_user + "'")
+            self.venue_request.decision_upload_super_invitation = self.venue_request.client.post_invitation(
+                openreview.Invitation(
```
nlbahy, line 551: So does this mean that existing venues won't have this feature? I got a process function error when trying this for MIDL: "The Invitation OpenReview.net/Support/-/Request265/Decision_Upload_Status was not found". Is there a way to enable it for already deployed venues?

purujitgoyal: Yes, we will just have to create the invitations for `Decision_Upload_Status` for all venues once we deploy this.

PR #1210, opened 2022-04-15
TMLR recruitment feedback

celestemartinez:
Adds `reinvite` param to `invite_reviewers`, which can be used to decide if we should send recruitment invitation to reviewers that have already been invited.


melisabok:
Looks good to me. 

Do you want to send an email to the AE when the reviewer accepts/declines the invitations?

I think the best way to do this is to post a comment as a reply of the Recruitment note in the Journal Request forum. You can try to find all the recruitment notes that where made by AEs and try to match the email address/tilde id, if it matches then post a comment notifying the response. This way we can notify all the AEs that invite the same reviewer.

celestemartinez:
If a user that has accepted/decline was invited by an AE, a comment is posted to the request form and an email is sent to EICs and AEs (this is the only time an email is sent to AEs as to not spam them)

openreview/journal/journal_request/journal_request.py
```diff
@@ -168,7 +168,7 @@ def setup_comment_invitation(self, note_id, action_editors_id=None):
                 writers = [],
                 signatures = ['~Super_User1'],
                 edit = {
-                    'signatures': { 'regex': f'~.*|{self.support_group_id}', 'type': 'group[]' },
+                    'signatures': { 'regex': f'~.*|{venue_id}|{self.support_group_id}', 'type': 'group[]' },
```
celestemartinez, line 171: had to add this here to allow the `venue_id` to sign the comment (since we couldn't sign it using `support_group.id` from the process function)

openreview/journal/invitation.py
```diff
@@ -211,6 +211,10 @@ def set_reviewer_recruitment_invitation(self):
             process_content = process_content.replace("ACTION_EDITOR_ACCEPTED_ID = ''", f"ACTION_EDITOR_ACCEPTED_ID = '{reviewers_id}'")
             process_content = process_content.replace("ACTION_EDITOR_DECLINED_ID = ''", f"ACTION_EDITOR_DECLINED_ID = '{reviewers_declined_id}'")
             process_content = process_content.replace("HASH_SEED = ''", f"HASH_SEED = '{self.journal.secret_key}'")
+            if self.journal.get_request_id():
+                process_content = process_content.replace("JOURNAL_REQUEST_ID = ''", "JOURNAL_REQUEST_ID = '" + self.journal.get_request_id() + "'")
+            process_content = process_content.replace("SUPPORT_GROUP = ''", f"SUPPORT_GROUP = 'openreview.net/Support'")
```
celestemartinez, line 216: @melisabok what would be the best way to get the value for `SUPPORT_GROUP` here? We don't really need it to find the recruitment notes but we do for the readers of the comment

openreview/journal/journal_request/journal_request.py
melisabok: I think it is ok that the venue_id can sign that note.

openreview/journal/invitation.py
melisabok: What about getting the request comment invitation and read the readers from there?

openreview/journal/journal_request/process/ae_recruitment_process.py
```diff
@@ -30,14 +30,11 @@ def process(client, edit, invitation):
     if not inviter_email:
         inviter_email = inviter_profile.content.get('emails')[0]
 
-    status = journal.invite_reviewers(message, subject, [email], [name], replyTo=inviter_email)
+    status = journal.invite_reviewers(message, subject, [email], [name], replyTo=inviter_email, reinvite=True)
 
     non_invited_status = f'''No recruitment invitation was sent to the following user because they have already been invited as reviewer:
```
melisabok, line 35: what about this message? I think they can reinvite invited reviewers, right?

openreview/journal/process/recruit_process.py
```diff
@@ -46,6 +48,37 @@ def process(client, note, invitation):
 
 '''.format(ACTION_EDITOR_NAME=ACTION_EDITOR_NAME, SHORT_PHRASE=SHORT_PHRASE)
 
-            return client.post_message(subject, [user], message, parentGroup=ACTION_EDITOR_DECLINED_ID)
+            response =  client.post_message(subject, [user], message, parentGroup=ACTION_EDITOR_DECLINED_ID)
+
+        action = 'accepted' if response == 'Yes' else 'declined'
+
+        if JOURNAL_REQUEST_ID:
+            recruitment_notes = list(openreview.tools.iterget_notes(client, invitation=f'.*/Journal_Request.*/-/Reviewer_Recruitment_by_AE', replyto=JOURNAL_REQUEST_ID))
```
melisabok, line 56: using `.*` at the beginning of the regex is not a good idea. Is this because we don't know the openreview support group?

I'm not sure the difficulties of adding the support group as constance here.

openreview/journal/process/recruit_process.py
```diff
@@ -46,6 +48,37 @@ def process(client, note, invitation):
...
+            print(len(recruitment_notes), 'recruitment notes!')
+            for note in recruitment_notes:
+                invitee = note.content['invitee_email']['value'].strip()
+                id_or_email = user
+                if '~' in user:
+                    profile = openreview.tools.get_profile(user)
+                    id_or_email = profile.id
+                if invitee == id_or_email:
```
melisabok, line 64: don't you need to get the profile of the invitee too? 

AEs can invitee reviewers using email addresses, right?

openreview/journal/process/recruit_process.py
celestemartinez: would using `invitation=(openreview.net|OpenReview.net)/Support/-/Journal_Request.*/-/Reviewer_Recruitment` be better or should I just find a way of adding the support group as a constant here?

melisabok (approved):

PR #1221, opened 2022-04-27
Feature/remove count

carlosmondra:
This PR:
- Removes count when offset is present in the query
- Improves get_all methods and relies on getting count on the first query


openreview/openreview.py
```diff
@@ -625,7 +625,7 @@ def get_groups(self, id = None, regex = None, member = None, signatory = None, w
         response = self.__handle_response(response)
         groups = [Group.from_json(g) for g in response.json()['groups']]
 
-        if with_count:
+        if with_count and params.get('offset') is None:
```
melisabok, line 628: if `offset` is 0, I think we should get the count too, right?

melisabok (approved):

PR #1225, opened 2022-04-28
filter notes by invitation prefix instead of a full regex

melisabok:
Fix slow queries with this filter:

```
"invitation": {
"$regex": "^ICLR.cc/2022/Workshop/DL4C/Reviewers/-/.*Form"
}
```

### DIFF
--- a/openreview/conference/templates/programchairWebfield.js
+++ b/openreview/conference/templates/programchairWebfield.js
@@ -296,8 +296,12 @@ var getRegistrationForms = function() {
   var prefixes = [ REVIEWERS_ID, AREA_CHAIRS_ID ];
   var promises = _.map(prefixes, function(prefix) {
     return Webfield.getAll('/notes', {
-      invitation: prefix + '/-/.*Form'
-    });
+      invitation: prefix + '/-/.*',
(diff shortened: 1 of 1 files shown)


openreview/conference/templates/programchairWebfield.js
```diff
@@ -296,8 +296,12 @@ var getRegistrationForms = function() {
...
   var promises = _.map(prefixes, function(prefix) {
     return Webfield.getAll('/notes', {
-      invitation: prefix + '/-/.*Form'
-    });
+      invitation: prefix + '/-/.*',
+      signature: CONFERENCE_ID
+    })
+    .then(function(notes) {
+      return notes.filter(function(note) { return note.invitation.endsWith('Form'); })
+    })
```
carlosmondra, line 304: Are all the fields in the Form Notes necessary? We can use the `select` param to reduce the amount of data transferred to the client.

melisabok: good point, I will use `select=id`

melisabok: Done!

PR #1231, opened 2022-05-02
Fix: render submissions using venue instead of decision

celestemartinez:


openreview/conference/templates/tabsConferenceDecisionsWebfield.js
```diff
@@ -138,35 +149,52 @@ function createConsoleLinks(allGroups) {
...
-      var decisionKey = getElementId(tabName);
-      if (notesDict[d.forum] && papersByDecision[decisionKey]) {
-        papersByDecision[decisionKey].push(notesDict[d.forum]);
-      }
-    }
-
-  });
+  $(container).empty();
+  if (submissionCount) {
+    var SearchResultsListOptions = _.assign({}, paperDisplayOptions, {
```
melisabok, line 158: I think this should be `searchResultsListOptions`, right?

celestemartinez: Fixed.

tests/test_venue_request.py
```diff
@@ -2003,3 +2003,32 @@ def test_supplementary_material_revision(self, client, test_client, selenium, re
...
+        #Post a post decision note
+        now = datetime.datetime.utcnow()
+        start_date = now - datetime.timedelta(days=2)
+        due_date = now + datetime.timedelta(days=3)
+        post_decision_stage_note = test_client.post_note(openreview.Note(
+            content={
+                'reveal_authors': 'No, I don\'t want to reveal any author identities.',
+                'submission_readers': 'Everyone (submissions are public)',
+                'home_page_tab_names': {
+                    'Accept': 'Accept',
```
celestemartinez, line 2016: @melisabok this test should let you test this change

tests/test_double_blind_conference.py
```diff
@@ -2037,9 +2037,9 @@ def test_release_accepted_notes_without_revealing_authors(self, client, request_
         assert tabs
         with pytest.raises(NoSuchElementException):
             notes_panel.find_element_by_class_name('spinner-container')
-        assert tabs.find_element_by_id('accepted-poster-papers')
+        # assert tabs.find_element_by_id('accepted-poster-papers')
```
melisabok, line 2040: why did you comment out this? can you modify the test and assert the new element id?

celestemartinez: so these two tab names are correct but they don't appear because there are no papers with these decisions. I can remove the lines instead

melisabok: maybe we should update the webfield and always show all the tabs even if there is not data. What do you think?

openreview/conference/builder.py
```diff
@@ -1532,22 +1526,29 @@ def is_release_authors(is_note_accepted):
                     paper_status = 'accepted' if note_accepted else 'rejected',
                     anonymous=False
                 )
-            #add venue_id if note accepted
+            #add venue_id if note accepted and build venue_heading_map
             venue = self.short_name
+            decision_option = decision_note.content['decision'] if decision_note else None
```
melisabok, line 1531: instead of build the map here when you are iterating all the notes where some decisions are duplicated, why don't you build it before calling set_homepage_decisions:

```
for decision in decisionOptions:
     venue_heading_map[self.decisionToVenue(decision)] = decision_heading_map[decision]
```

We can define a new function `decisionToVenue` that convert the decision string to a venue value.

openreview/conference/builder.py
```diff
@@ -1541,13 +1544,19 @@ def is_release_authors(is_note_accepted):
                 if decision:
                     venue += ' ' + decision
                 submission.content['venueid'] = venueid
-                submission.content['venue'] = venue
+                submission.content['venue'] = venue.strip()
```
melisabok, line 1547: you can use the function `decision_to_venue` here and we avoid duplicate code.

tests/test_double_blind_conference.py
```diff
@@ -2096,9 +2096,9 @@ def test_release_accepted_notes(self, client, request_page, selenium):
...
```
melisabok, line 2099: what about this assertion? can you enable it again?

celestemartinez: yes! I thought I had, sorry

openreview/conference/builder.py
melisabok: you can pass `note_accepted` to `decision_to_venue` so the implementation doesn't need to check again `if 'Accept' in...`, then:

```
submission.content['venue'] = decision_to_venue(decision_note.content['decision'], note_accepted)
```

melisabok: I think this won't work because when you are translating the map you don't have `note_accepted`. I think you can have another function `is_accepted(decision)` that is being called from `decision_to_venue` and to set `note_accepted`.

openreview/conference/builder.py
```diff
@@ -1499,6 +1491,17 @@ def post_decision_stage(self, reveal_all_authors=False, reveal_authors_accepted=
         def is_release_authors(is_note_accepted):
             return reveal_all_authors or (reveal_authors_accepted and is_note_accepted)
 
+        def decision_to_venue(decision_option):
```
melisabok, line 1494: could you move this function to the tools module? I think we are going to need it to update the `content.venue` for previous conferences.

openreview/conference/builder.py
celestemartinez: done

melisabok (approved):

melisabok:
Looks good! thanks for all the changes, adding Zach so he can take a final look to the UI changes.

openreview/conference/templates/tabsConferenceDecisionsWebfield.js
```diff
@@ -100,17 +114,20 @@ function renderConferenceTabs() {
...
-  tabArray.forEach(function(tabName) {
+
+  var tabNames = []
+
+  Object.keys(DECISION_HEADING_MAP).forEach(function(key) {
+    venueIds.push(key)
+    tabNames.push(DECISION_HEADING_MAP[key])
+  })
+
+  tabNames.forEach(function(tabName) {
```
zbialecki, line 123: It would be good to combine this forEach with the one above. The tabNames array isn't needed outside of that forEach, so you could remove the array and just use a local var inside the function.

openreview/conference/templates/tabsConferenceDecisionsWebfield.js
```diff
@@ -173,30 +208,8 @@ function renderContent(notes, decisionNotes, withdrawnNotes, deskRejectedNotes,
...
-
-  $('#group-container').on('hidden.bs.tab', 'ul.nav-tabs li a', function(e) {
-    var containerSelector = $(e.target).attr('href');
-    var containerId = containerSelector.substring(1);
-    if (!papersByDecision.hasOwnProperty(containerId) || !$(containerSelector).length) {
-      return;
-    }
-
-    Webfield.ui.spinner(containerSelector, { inline: true });
+  _.forEach(notesArray, function(notes, index) {
```
zbialecki, line 209: We want to try to avoid using the lodash functions when native versions are supported, so it would be better to do notesArray.forEach here.

openreview/conference/templates/tabsConferenceDecisionsWebfield.js
```diff
@@ -135,32 +152,50 @@ function createConsoleLinks(allGroups) {
...
-  });
-  
-  if (DECISION_HEADING_MAP['Reject']) {
-    var decisionKey = getElementId(DECISION_HEADING_MAP['Reject']);
-    papersByDecision[decisionKey] = papersByDecision[decisionKey].concat(withdrawnNotes.concat(deskRejectedNotes));
+  var container = '#' + getElementId(DECISION_HEADING_MAP[venueId])
+
+  $(container).empty();
+  if (submissionCount) {
+    var searchResultsListOptions = _.assign({}, paperDisplayOptions, {
```
zbialecki, line 159: You could use Object.assign here

celestemartinez:
@melisabok I added some more code to prevent the homepage from being overwritten once they have been set, let me know if there are any issues

openreview/conference/helpers.py
```diff
@@ -201,6 +201,11 @@ def get_conference_builder(client, request_form_id, support_user='OpenReview.net
     builder.set_review_stage(get_review_stage(note))
     builder.set_ethics_review_stage(get_ethics_review_stage(note))
 
+    decision_heading_map = note.content.get('home_page_tab_names')
```
melisabok, line 204: are you sure this work? becauase `get_result` is going to call set_home_page: https://github.com/openreview/openreview-py/blob/master/openreview/conference/builder.py#L2782

Could you add a new property in the conference call that store the `decision_heading_map` and inside of the `set_home_page` implementation, you can check if that map is present then set the decisions layout otherwise set the tab layout.

tests/test_venue_request.py
```diff
@@ -2037,3 +2042,20 @@ def test_supplementary_material_revision(self, client, test_client, selenium, re
         assert all(x not in revision_invitations[0].reply['content'] for x in ['title','authors', 'authorids','abstract','keywords', 'TL;DR'])
         assert revision_invitations[0].duedate
         assert revision_invitations[0].expdate
+
+        #make sure homepage webfield was not overwritten after doing get_conference()
```
celestemartinez, line 2046: this is testing the decision tabs are still there even after using the request form

openreview/conference/helpers.py
melisabok: Sorry I just realized that, the changes are great!

tests/test_venue_request.py
melisabok: excelent!

zbialecki (approved):

PR #1232, opened 2022-05-03
Adds Expiration Date to Withdraw Invitation

purujitgoyal:


openreview/venue_request/venue_request.py
```diff
@@ -875,21 +875,27 @@ def setup_request_form(self):
                 'default': 'No, wait until the submission deadline has passed to make them public.',
                 'order': 24
             },
+            'withdraw_submission_deadline': {
```
melisabok, line 878: I would call this `withdraw_submission_expiration`

celestemartinez:
It looks like you set the `Withdrawn_Submission` invitation as the super invitation for the `PaperX/-/Withdraw` invitations. Would it be better to create another super invitation, because the `Withdrawn_Submission` has different reply and no process function?

@melisabok I am not sure if this may cause confusion or an error at some point

melisabok:
Yes, Purujit please create a new invitation as a super of PaperX/-/Withdraw. Withdrawn_Submission invitation is being used to mark the submission as withdrawn.

openreview/conference/invitation.py
```diff
@@ -335,6 +336,37 @@ def __init__(self, conference, reveal_authors, reveal_submission, hide_fields=[]
...
         )
 
+
+class PaperWithdrawSuperInvitation(openreview.Invitation):
+
+    def __init__(self, conference):
+        content = invitations.withdraw.copy()
+        exp_date = tools.datetime_millis(conference.submission_stage.withdraw_submission_exp_date) if conference.submission_stage.withdraw_submission_exp_date else None
+        super(PaperWithdrawSuperInvitation, self).__init__(
+            id=conference.get_invitation_id("Paper_Withdraw"),
```
melisabok, line 346: Could you call it `Withdraw` only so it matches the same name with the PaperX/-/Withdraw invitations?

openreview/conference/invitation.py
```diff
@@ -416,9 +448,8 @@ def __init__(self, conference, note, reveal_authors, reveal_submission, email_pc
 
             super(PaperWithdrawInvitation, self).__init__(
                 id=conference.get_invitation_id('Withdraw', note.number),
+                super=conference.get_invitation_id('Paper_Withdraw'),
```
melisabok, line 451: given the invitation is created using a super invitation, all the properties that are defined in the super invitation don't need to be defined in the child invitation, for example `content`, you can remove it.

openreview/venue_request/venue_request.py
```diff
@@ -998,21 +998,27 @@ def setup_request_form(self):
                 'default': 'No, wait until the submission deadline has passed to make them public.',
                 'order': 24
             },
+            'withdraw_submission_expiration': {
+                'value-regex': r'^[0-9]{4}\/([1-9]|0[1-9]|1[0-2])\/([1-9]|0[1-9]|[1-2][0-9]|3[0-1])(\s+)?((2[0-3]|[01][0-9]|[0-9]):[0-5][0-9])?(\s+)?$',
+                'description': 'By when authors can withdraw their submission? Please specify the due date in GMT using the following format: YYYY/MM/DD HH:MM(e.g. 2019/01/31 23:59)',
```
melisabok, line 1003:
```suggestion
                'description': 'By when authors can withdraw their submission? Please specify the expiration date in GMT using the following format: YYYY/MM/DD HH:MM(e.g. 2019/01/31 23:59)',
```

nlbahy: I might reword this to "When should authors lose the option to withdraw their submissions?"

nlbahy:
This looks good! Do PCs ever ask to set a custom activation date for the withdraw invitations?

melisabok:
>  Do PCs ever ask to set a custom activation date for the withdraw invitations?

That's a good question, withdraw invitations become active after the first or unique deadline.

nlbahy (approved):

openreview/conference/invitation.py
```diff
@@ -335,11 +336,40 @@ def __init__(self, conference, reveal_authors, reveal_submission, hide_fields=[]
...
+                'readers': {
+                        'values': ['everyone']
+                    },
+                'writers': {
+                    'values': [],
+                },
+                'signatures': {
+                    'values-regex': '~.*|{}'.format(conference.support_user)
+                },
+                'content': content
```
melisabok, line 365: could you set the process function here in the super invitation? then it will be easier to update from here rather than update every child invitation.

purujitgoyal: There are some paper-specific (author_id, reviewers_id, etc.) things that are set while creating child invitations.

melisabok: Ohhh I see, would you mind to change those constants with the following:

```
 PAPER_AUTHORS_ID = f'aclweb.org/ACL/ARR/2021/September/Paper{number}/Authors'
 PAPER_REVIEWERS_ID = f'aclweb.org/ACL/ARR/2021/September/Paper{number}/Reviewers'
 PAPER_AREA_CHAIRS_ID = f'aclweb.org/ACL/ARR/2021/September/Paper{number}/Area_Chairs'
```

then you can do:

```
PAPER_AUTHORS_ID.format(number=forum_note.number)
```

for build the value of the constant you can use the method in the conference class:

```
self.get_authors_id(number='{number}')
```

purujitgoyal: This works everywhere `forum_note.number` is required, however, `SUBMISSION_READERS = note.readers` would not work this way. Any suggestions for that?

melisabok: Could you try assigning this instead of `note.readers`?

```
SUBMISSION_READERS =  self.submission_stage.get_readers(self, '{number}')
```

If this doesn't work then we should keep the process function in the children.

purujitgoyal: This will require access to the conference which is not there, should I add the `request_form_id` to the process function?

melisabok: No, you can setup the variable SUBMISSION_READERS here: https://github.com/openreview/openreview-py/blob/master/openreview/conference/invitation.py#L391

You will pass an array of values that may contain a group id with `{number}` so you would need to iterate all the elements of the array and replace `{number}` with `forumNote.number`.

celestemartinez:
Changes look good to me!

One comment: I noticed that if you change the withdrawn/desk-rejected visibility or author anonymity using the `Revision` button, no change is made until we run one of the stages (since using the `Revision` button does not call `setup_post_submission_stage()`) This might be confusing for PCs that might expect an immediate change (but not sure if this is a big deal)

melisabok:
> One comment: I noticed that if you change the withdrawn/desk-rejected visibility or author anonymity using the Revision button, no change is made until we run one of the stages (since using the Revision button does not call setup_post_submission_stage()) This might be confusing for PCs that might expect an immediate change (but not sure if this is a big deal)

But this was happening before, right? it is not related to the changes of this PR. I suggest to create an issue and fix this in another PR. What do you think?

melisabok (approved):

openreview/conference/templates/withdraw_process.py
```diff
@@ -17,15 +17,22 @@ def process_update(client, note, invitation, existing_note):
...
     HIDE_FIELDS = []
 
+    forum_note = client.get_note(note.forum)
+    for i in range(len(SUBMISSION_READERS)):
+        SUBMISSION_READERS[i] = SUBMISSION_READERS[i].format(number=forum_note.number)
+
+    PAPER_AUTHORS_ID = PAPER_AUTHORS_ID.format(number=forum_note.number)
+    PAPER_REVIEWERS_ID = PAPER_REVIEWERS_ID.format(number=forum_note.number)
+    PAPER_AREA_CHAIRS_ID = PAPER_AREA_CHAIRS_ID.format(number=forum_note.number)
+    PAPER_SENIOR_AREA_CHAIRS_ID = PAPER_SENIOR_AREA_CHAIRS_ID.format(numbers=forum_note.number)
```
celestemartinez, line 27:
```suggestion
    PAPER_SENIOR_AREA_CHAIRS_ID = PAPER_SENIOR_AREA_CHAIRS_ID.format(number=forum_note.number)
```

melisabok: thanks for the catch!

PR #1236, opened 2022-05-03
Adds option to return profiles as dict

purujitgoyal:


openreview/tools.py
```diff
@@ -134,18 +137,29 @@ def get_profiles(client, ids_or_emails, with_publications=False):
...
+    profiles_as_dict = {}
 
     batch_size = 100
     for i in range(0, len(ids), batch_size):
         batch_ids = ids[i:i+batch_size]
         batch_profiles = client.search_profiles(ids=batch_ids)
         profiles.extend(batch_profiles)
 
+    if as_dict:
+        for i, profile in enumerate(profiles):
```
melisabok, line 149: is this assuming the profiles are returned in the same order? also what happens with the ids that don't return any profile (it should happen). 

I would build the dictionary using the `ids` arrays instead, same as you are doing with the emails. You would need to find the profile that contains that `id` into the `profile.content['names']` array. Maybe using a dictionary where the key are all the alternate name can help.

openreview/tools.py
```diff
@@ -146,8 +146,14 @@ def get_profiles(client, ids_or_emails, with_publications=False, as_dict=False):
         profiles.extend(batch_profiles)
 
     if as_dict:
-        for i, profile in enumerate(profiles):
-            profiles_as_dict[ids[i]] = profile
+        for profile in profiles:
+            for name in profile.content.get("names", {}):
```
melisabok, line 150: names is an array so the default value should be [].

openreview/tools.py
```diff
@@ -146,8 +146,14 @@ def get_profiles(client, ids_or_emails, with_publications=False, as_dict=False):
...
+                if name.get("username") in ids:
+                    profiles_as_dict[name.get("username")] = profile
+                    continue
```
melisabok, line 153: If we pass two different tilde ids that belongs to the same profile we are going to get a profile for one tilde id and None for the other. I think you can reproduce this adding more tests. I believe there are profiles with alternate names. 

You must check that all the `ids` have a profile. I would suggest to do the following:

```
profiles_by_name = {}
for profile in profiles:
      for name in profile.content.get('names', []):
            if name.get('username'):              
                  profiles_by_name[name.get('username')] = profile

for id in ids:
      profiles_as_dict[id] = profile_by_name.get(id)
```

purujitgoyal: Emails will have same issue

melisabok: could you provide a test? that is the best way to verify if it fails.

openreview/tools.py
```diff
@@ -3,7 +3,9 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import json
+import logging
 import os
+from collections import OrderedDict
```
melisabok, line 8: do you need this import?

openreview/tools.py
```diff
@@ -3,7 +3,9 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import json
+import logging
```
melisabok, line 6: do you need this import?

openreview/tools.py
```diff
@@ -152,18 +155,37 @@ def get_profiles(client, ids_or_emails, with_publications=False):
...
     for i in range(0, len(ids), batch_size):
         batch_ids = ids[i:i+batch_size]
         batch_profiles = client.search_profiles(ids=batch_ids)
         profiles.extend(batch_profiles)
 
+    if as_dict:
+        profiles_by_name = {}
+        for profile in profiles:
+            for name in profile.content.get("names", []):
+                profiles_by_name[name.get("username")] = profile
```
melisabok, line 170: name.get("username") can return None here, is it ok to have a None key?

purujitgoyal: Yeah, it won't cause any issues.

openreview/tools.py
purujitgoyal: Removed.

openreview/tools.py
purujitgoyal: Removed.

melisabok (approved):

PR #1242, opened 2022-05-05
Fix: assignment invitation preprocess and process functions

celestemartinez:
Always set assignment process function when creating invitation


melisabok (approved):
🏅

nlbahy (approved):

PR #1248, opened 2022-05-10
Expire Non Accepted Revision Invitations

purujitgoyal:


melisabok:
Could you specify which issue you are fixing here?

setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='openreview-py',
-    version='1.2.3',
+    version='1.2.4',
```
melisabok, line 5: I discussed this with Celeste and we decided to increase the version number only when we are going to make a release in the instances of the live site. Please remove this change for this PR.

openreview/conference/builder.py
```diff
@@ -297,10 +297,15 @@ def __create_decision_stage(self):
         return self.invitation_builder.set_decision_invitation(self, notes)
 
     def __create_submission_revision_stage(self):
-
         invitation = tools.get_invitation(self.client, self.get_submission_id())
         if invitation:
             notes = self.get_submissions(accepted=self.submission_revision_stage.only_accepted, details='original')
+            if self.submission_revision_stage.only_accepted:
```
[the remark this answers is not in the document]
purujitgoyal: Do we want to expire all the existing revision invitations in case the name changes or everytime and create new ones?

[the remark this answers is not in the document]
melisabok: I think it is in the case the name changes. 

It happened to a venue they had a typo in the revision name, they fixed it and then the forum was showing two revision buttons.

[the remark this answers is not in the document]
melisabok: One approach: in the revision process function, find the latest reference of the submission revision stage and check if submission revision name is different than the current request form submission revision name, then expire the previous submission revision name (set an expdate in the super invitation).

to get the latest reference:

```
GET /references?referent=[request_form_id]&invitation=OpenReview.net/Support/-/Request[request_form_number]/Submission_Revision_Stage&sort=tmdate:desc&limit=1
```

melisabok:
Could you also add a test that describes the scenario described here? https://github.com/openreview/openreview-py/pull/1248#discussion_r870605459

openreview/venue_request/process/revisionProcess.py
```diff
@@ -194,6 +194,20 @@ def process(client, note, invitation):
...
+                limit=2
+            )
+            if len(submission_revision_stage_notes) > 1:
+                last_submission_revision_stage_note = submission_revision_stage_notes[-1]
+                expire_revision_stage_name = last_submission_revision_stage_note.content.get('submission_revision_name',
+                                                                                             'Revision')
+            else:
+                expire_revision_stage_name = 'Revision'
+            if expire_revision_stage_name != forum_note.content.get('submission_revision_name', '').strip():
+                conference._Conference__expire_invitations(expire_revision_stage_name)
```
melisabok, line 210: where is this method defined?

purujitgoyal: It's the `__expire_invitations` method, this is how python access methods with double leading underscores. https://peps.python.org/pep-0008/#descriptive-naming-styles

melisabok: Wow, I didn't know that!

melisabok: that only issue with this is that the implementation is expiring all the children invitations. I would prefer to change the expiration date of the super invitation only. Could you use `__expire_invitation` instead?

melisabok (approved):

openreview/venue_request/process/revisionProcess.py
```diff
@@ -194,6 +194,20 @@ def process(client, note, invitation):
...
+                limit=2
+            )
+            if len(submission_revision_stage_notes) > 1:
+                last_submission_revision_stage_note = submission_revision_stage_notes[-1]
+                expire_revision_stage_name = last_submission_revision_stage_note.content.get('submission_revision_name',
+                                                                                             'Revision')
+            else:
+                expire_revision_stage_name = 'Revision'
+            if expire_revision_stage_name != forum_note.content.get('submission_revision_name', '').strip():
+                conference.expire_invitation(conference.get_invitation_id(expire_revision_stage_name))
```
celestemartinez, line 210: When the old revision name is something like `Camera Ready Revision`, you need to replace spaces with `_`, otherwise you won't find the invitation and they won't be expired:

https://github.com/openreview/openreview-py/blob/71f501c59103089a6e18fad27da5e8b1f2f2d497/openreview/conference/helpers.py#L419

celestemartinez (approved):

PR #1250, opened 2022-05-11
Fix/ error in put attachment

xkopenreview:
this pr should fix the error in put_attachment when running the test by
1. add a util function to get mime type from a list of mimetype/extension dict
2. pass the mimetype to put call

the mimetype/extension list is the same as the one used in api


carlosmondra:
Looks good so far, but there is a failing test :(

melisabok:
it needs https://github.com/openreview/openreview-py/pull/1249

melisabok (approved):

carlosmondra (approved):
Awesome!

PR #1252, opened 2022-05-12
Fix: post_decision_stage adds new revision to papers

### LINKED ISSUE #1064: Bibtex generation: use a revision invitation to add the bibtex entry
Trying to add the bibtex entry as an update of the note can lead to different errors. The camera ready invitation may have extra fields that are not present in the submission, the process function trying to update the note can have validation error because the some fields are not present in the submission invitation.

I suggest to use a new revision invitation to add any extra fields to the note: bibtex, venue, venueid, etc.

celestemartinez:
- Add venue, venueid and bibtex to all papers using revision invitation

Fixes #1064  

Needs https://github.com/openreview/openreview-api-v1/pull/2629


celestemartinez:
I need to create the `/Venue_Revision` invitation for the tests to pass, @melisabok what would be the best place to do this? In the request form or in one of the tests?

We won't need to be changing this invitation in the live site often.

melisabok:
This invitation is going to be used across all the venues, let's create it in the setup script of the API, you have an example here: https://github.com/openreview/openreview-api-v1/blob/master/scripts/mongo_database.js#L396

celestemartinez:
I am now always adding the bibtex as a reference (even when creating blind submissions). It has to be done like this now because even if we were changing the bibtex later on using a reference, it was not overwriting the bibtex added at the beginning to the blind submission. 
@melisabok let me know if something does not make sense

Also, seems like some tests that use the request form are failing because the invitation `openreview.net/Support/-/Venue_Revision` is not found (I see other tests are adding references using the invitation `OpenReview.net/Support/-/Venue_Revision)`

openreview/conference/builder.py
```diff
@@ -939,15 +931,23 @@ def create_blind_submissions(self, hide_fields=[], number=None):
...
+                bibtex = tools.generate_bibtex(
                     note=note,
                     venue_fullname=self.name,
                     url_forum=blind_note.id,
                     year=str(self.get_year()))
 
-                blind_note.content = blind_content
-
-                blind_note = self.client.post_note(blind_note)
+                revision_note = self.client.post_note(openreview.Note(
```
melisabok, line 940: are you adding the revision to the original note or blind note?

openreview/conference/templates/submission_revision_process.py
```diff
@@ -36,23 +37,30 @@ def process_update(client, note, invitation, existing_note):
...
             venue_fullname=CONFERENCE_NAME,
             year=CONFERENCE_YEAR,
             url_forum=bibtex_note.id,
             paper_status='accepted',
-            anonymous=bibtex_note.content.get('authors', []) == ['Anonymous']
+            anonymous=anonymous_note
         )
-        client.post_note(bibtex_note)
+
+        revision_note = client.post_note(openreview.Note(
```
melisabok, line 54: is this a revision to the original note or blind note?

openreview/conference/builder.py
celestemartinez: they should all be revisions of the original note but I can double check

openreview/conference/helpers.py
```diff
@@ -17,8 +17,11 @@ def get_conference_builder(client, request_form_id, support_user='OpenReview.net
     if not note.content.get('venue_id') and not note.content.get('conference_id'):
         raise openreview.OpenReviewException('venue_id is not set')
 
+    support_user = note.invitation.split('/-/')[0]
     builder = openreview.conference.ConferenceBuilder(client, support_user)
     builder.set_request_form_id(request_form_id)
+    # support_user = note.invitation.split('/-/')[0]
```
melisabok, line 23: could you remove the commented code?

openreview/conference/templates/submission_revision_process.py
```diff
@@ -36,23 +37,30 @@ def process_update(client, note, invitation, existing_note):
...
+            invitation = f'{SUPPORT_GROUP}/-/Venue_Revision',
```
melisabok, line 55: instead of passing the support group, could you pass the whole invitation id? so we don't have `Venue_Revision` hardcoded everywhere and we can easily change the name in the future if needed.

melisabok (approved):
💯

PR #1255, opened 2022-05-12
Optimizes Replace Members with IDs function

purujitgoyal:


openreview/tools.py
```diff
@@ -796,38 +797,23 @@ def replace_members_with_ids(client, group):
...
 
-    results = concurrent_requests(classify_members, group.members)
+    member_profiles = get_profiles(client, group.members, as_dict=True)
 
-    for key, member in results:
-        if key == 'ids':
-            ids.append(member)
-        elif key == 'emails':
-            emails.append(member)
+    for member, profile in member_profiles.items():
```
melisabok, line 804: could you start the iteration with `group.members` to make sure each member is replaced or not found error. 

```
for member in groups.members:
   profile = member_profiles.get(member)
   if profile:
          update_members.append(profile.id)
   else:
          if '@' in member:
                update_members.append(profile.id)
         else:
               raise openreview.OpenReviewException(f"Profile Not Found for {member}")
```

We also make sure they are replaced in the same order.

carlosmondra: Why do they need to be replaced in the same order?

melisabok: I think maintaining the order is better but I may be wrong.

carlosmondra (approved):

openreview/tools.py
carlosmondra: Well, they are now being replaced in order, so I think it's fine.

melisabok (approved):

PR #1240, opened 2022-05-05
Fix NeurIPS COI

sherry-zyx:

PR #1260, opened 2022-05-13
Fix/neur ips coi

carlosmondra:


carlosmondra:
Close in favor of https://github.com/openreview/openreview-py/pull/1240

PR #1262, opened 2022-05-17
Adds review rating and confidence field names and options to Review Stage

purujitgoyal:


celestemartinez:
Maybe we should not have a separate field for the rating/confidence options and have PCs still use the `additional_review_form_options` field for this, as this might be confusing. I am not sure if this is what @melisabok meant, but I think all we need to know is what the name of the field is. 

Also, we need to know the name of the field to show these values in the PC console here: https://github.com/openreview/openreview-py/blob/f3abe1ce81be25fa8723512f651bc456e7781185/openreview/conference/templates/programchairWebfield.js#L14 since we have `rating` and `confidence` hard-coded here

openreview/venue_request/venue_request.py
```diff
@@ -164,14 +164,55 @@ def setup_review_stage(self):
...
                 'order': 27
             },
+            'review_rating_field_name': {
+                'description': "Name of the rating field for the review invitation, default is \'Rating\'",
+                'value-regex': '.*',
+                'required': False,
+                'default': 'Rating',
+                'order': 28
+            },
+            'review_rating_options': {
```
melisabok, line 174: as @celestemartinez says, we don't need this. They can describe this in additional review options.

openreview/venue_request/venue_request.py
```diff
@@ -164,14 +164,55 @@ def setup_review_stage(self):
...
+                'order': 29
+            },
+            'review_confidence_field_name': {
+                'description': "Name of the confidence field for the review invitation, default is \'Confidence\'",
+                'value-regex': '.*',
+                'required': False,
+                'default': 'Confidence',
+                'order': 30
+            },
+            'review_confidence_options': {
```
melisabok, line 197: same here.

openreview/venue_request/venue_request.py
```diff
@@ -164,14 +164,55 @@ def setup_review_stage(self):
                 'default': 'No, do not email program chairs about received reviews',
                 'order': 27
             },
+            'review_rating_field_name': {
+                'description': "Name of the rating field for the review invitation, default is \'Rating\'",
```
melisabok, line 168: You can add the description here: "Please follow the format number: description if you want to customize options and add one option per line."

openreview/venue_request/venue_request.py
```diff
@@ -164,14 +164,55 @@ def setup_review_stage(self):
...
+                'value-regex': '.*',
+                'required': False,
+                'default': 'Rating',
```
melisabok, line 171: I believe the default field name is rating all lowercase.

openreview/conference/helpers.py
```diff
@@ -275,7 +275,9 @@ def get_review_stage(request_forum):
         release_to_reviewers = release_to_reviewers,
         email_pcs = (request_forum.content.get('email_program_chairs_about_reviews', '').startswith('Yes')),
         additional_fields = review_form_additional_options,
-        remove_fields = review_form_remove_options
+        remove_fields = review_form_remove_options,
+        rating_field_name=request_forum.content.get('review_rating_field_name', 'rating').strip().lower().replace(" ", "_"),
```
melisabok, line 279: I'm not sure about replacing the spaces with _ here. The review_rating_field_name should always match with the field name described in Additional Review field options. 

If they define "Final Rating" as a key of the json then the value in review_rating_field_name should keep as "Final Rating" otherwise the UI won't find this value in the content.

openreview/venue_request/venue_request.py
```diff
@@ -164,14 +164,28 @@ def setup_review_stage(self):
                 'default': 'No, do not email program chairs about received reviews',
                 'order': 27
             },
+            'review_rating_field_name': {
+                'description': "Name of the rating field for the review invitation, default is \'rating\'. Please follow the format number: description if you want to customize options and add one option per line \'Additional Review Form Options\'.",
```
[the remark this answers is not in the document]
celestemartinez: could we add a link to the default review form here? https://docs.openreview.net/reference/default-forms/default-review-form

melisabok (approved):

PR #1265, opened 2022-05-17
Fix/messages tests

melisabok:


melisabok:
closing this for now until we can start using invitations to send messages.

PR #1268, opened 2022-05-18
Makes formatting consistent in emails

purujitgoyal:


openreview/venue_request/process/revisionProcess.py
```diff
@@ -118,8 +118,7 @@ def process(client, note, invitation):
                     content[f'{decision.lower().replace(" ", "_")}_email_content'] = {
                         'value-regex': '[\\S\\s]{1,10000}',
                         'description': 'Please carefully review the template below before you click submit to send out the emails. Make sure not to remove the parenthesized tokens.',
-                        'default': f'''
-Dear {{{{{{{{fullname}}}}}}}},
+                        'default': f'''Dear {{fullname}},
 
 Thank you for submitting your paper, {{submission_title}}, to {short_name}. We are delighted to inform you that your submission has been accepted. Congratulations!
```
melisabok, line 123: it seems that short name is still using single curly braces.

openreview/conference/builder.py
```diff
@@ -1701,6 +1701,7 @@ def send_notification(note):
                 submission_title=paper_note.content['title']
             )
             final_message = f'''{message.format(
+                fullname='{{fullname}}',
                 submission_title=paper_note.content['title'],
```
melisabok, line 1705: if submission_title has double curly braces, does the format remove all of them?

melisabok: Could change the assertion here: https://github.com/openreview/openreview-py/pull/1268/files#diff-e3ae2f10352929825bba9040318c444e8bf4b97395b036eec532aacae837bf72R1845

and assert the whole message? then you can verify all the token where correctly replaced.

melisabok: This is what I suspected. I think we should use find and replace instead of format.

purujitgoyal: But we will always get single curly braces, right?

melisabok: The UI should show all the tokens with double curly braces so they match with the backend supported tokens ({{firstname}} and {{fullname}}). 

Now I'm seeing single curly braces. 

Screen Shot 2022-05-19 at 9 50 59 AM

celestemartinez (approved):

melisabok (approved):

PR #1270, opened 2022-05-18
Feature/speed up revision invitation

carlosmondra:
This PR reduced the amount of time from 10 min to 3 min in NeurIPS 2022 when posting revision invitations.

### DIFF
--- a/openreview/conference/invitation.py
+++ b/openreview/conference/invitation.py
@@ -1671,13 +1671,12 @@ class InvitationBuilder(object):
         return invitations
 
     def set_revise_submission_invitation(self, conference, notes, content):
-
-        invitations = []
         self.client.post_invitation(SubmissionRevisionInvitation(conference, content))
-        for note in tqdm(notes, total=len(notes), desc='set_revise_submission_invitation'):
-            invitations.append(self.client.post_invitation(PaperSubmissionRevisionInvitation(conference, note, content)))
-
-        return invitations
+        return tools.concurrent_requests(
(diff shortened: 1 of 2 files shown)


openreview/conference/invitation.py
```diff
@@ -1699,13 +1699,12 @@ def post_invitation(note):
...
 
     def set_revise_submission_invitation(self, conference, notes, content):
-
-        invitations = []
         self.client.post_invitation(SubmissionRevisionInvitation(conference, content))
-        for note in tqdm(notes, total=len(notes), desc='set_revise_submission_invitation'):
-            invitations.append(self.client.post_invitation(PaperSubmissionRevisionInvitation(conference, note, content)))
-
-        return invitations
+        return tools.concurrent_requests(
```
melisabok, line 1703: most of the methods here are looping the notes to create invitations, do you think we should use concurrent requests for all of them?

melisabok: I can merge this PR and we start changing the other methods little by little

melisabok: I feel a little bit scared about using concurrent requests in python, most of the time works well for me but sometimes one request it gets stuck ... but I will give it a try.

I will merge this and see how it goes.

melisabok (approved):

openreview/conference/invitation.py
carlosmondra: I didn't know this was happening. This may be related to a server issue. Perhaps related to what Zach has reported as well.

PR #1273, opened 2022-05-20
sort the submissions by late tasks (earliest late due date)

melisabok:


zbialecki (approved):

zbialecki:
I'm not sure how you are displaying this information, but it might be nice to generalize this by creating a function that takes the complete tasks array and returns the all the late task ordered by due date.

melisabok:
It is just for sorting submissions and I need to get the min duedate for all the un complete tasks. I could do this in a better way, you are right:

```
tasks.filter(function(t) { return !t.complete; }).map(function(t) { return t.duedate}).sort(function(a, b) { return a - b; })[0]
```

PR #1276, opened 2022-05-23
implement new recruitment feature for ARR and NeurIPS venues

melisabok:


openreview/conference/webfield.py
```diff
@@ -323,33 +323,68 @@ def set_reduced_load_page(self, conference_id, invitation, options = {}):
...
+        
+            new_webfield_content = '''// Webfield component
+return {
+    component: 'RecruitmentForm',
+    version: 1,
+    properties: {
+        header: ''' + json.dumps(conference.get_homepage_options(), indent=2) + ''',
+        acceptMessage: ''' + json.dumps(accept_message) + ''',
+        declineMessage: ''' + json.dumps(decline_message) + ''',
+        ''' + properties + '''
```
zbialecki, line 350: I think it's better to list all the possible properties and then set some to `null` if they are not needed.

nlbahy:
Is it intended that they are able to either submit a reduced load request or submit a comment? It seems confusing that if I write a comment and then click “Submit comment”, I decline the invitation without my reduced load being logged. And then if I click “Submit Reduced Load Request”, it is not clear whether or not my comment is going through.

melisabok:
@nlbahy  I think your comment must be posted in the openreview-web PR, @xkopenreview can you take a look? I think we should have a single button to submit the reduced quota and/or comment.

nlbahy:
How are the reduced load notes now being posted? Are they no longer using the invitation/Reviewers/-/Reduced_Load? When I tried accepting with a reduced load request, I don't see the message in the reviewers' console that says "You agreed to review x papers". I also don't see any Reduced_Load notes

melisabok:
> How are the reduced load notes now being posted? Are they no longer using the invitation/Reviewers/-/Reduced_Load? When I tried accepting with a reduced load request, I don't see the message in the reviewers' console that says "You agreed to review x papers". I also don't see any Reduced_Load notes

The reduce load is stored in the same note with the response. I think I need to fix the reviewers' console and get the quota from a different invitation, thanks for the catch!

melisabok:
@zbialecki the tests should pass now.

melisabok:
> I think Zach had a comment in the web pr that comment field should disable markdown

Fixed

zbialecki (approved):

nlbahy (approved):

PR #1279, opened 2022-05-24
Journal recruit response note

celestemartinez:
Process function checks if response note exists and doesn't repost if recruitment response did not change

### DIFF
--- a/openreview/journal/process/recruit_process.py
+++ b/openreview/journal/process/recruit_process.py
@@ -66,26 +66,36 @@ If you would like to change your decision, please click the Accept link in the p
                     profile = openreview.tools.get_profile(client, user)
                     id_or_email = profile.id
                 if id_or_email in invitee_ids:
-                    recruitment_response_notes = list(openreview.tools.iterget_notes(client, replyto=note.id, sort='number:desc'))
-                    if recruitment_response_notes:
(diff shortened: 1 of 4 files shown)


openreview/journal/process/recruit_process.py
```diff
@@ -66,6 +66,10 @@ def process(client, note, invitation):
                     profile = openreview.tools.get_profile(client, user)
                     id_or_email = profile.id
                 if id_or_email in invitee_ids:
+                    recruitment_response_notes = list(openreview.tools.iterget_notes(client, replyto=note.id, sort='number:desc'))
+                    if recruitment_response_notes:
+                        if action in recruitment_response_notes[0].content['comment']['value']:
+                            break
```
melisabok, line 72: instead of avoiding the post of a comment if the comment already exists why don't we post an edit as revision of the existing comment. Then we can have track of all the responses in the edit history. 

I would like to avoid the use of `break` or `continue` because it is difficult to follow the code and it could be an early sign of bad design. If you want to stop processing the response then you can use `return` to finish the process function.

celestemartinez: Sounds good, I will also have to send an email on update of a comment, perhaps only when the comment content has changed?

melisabok (approved):

PR #1283, opened 2022-05-25
transfer conflicts from SACs to their assigned ACs

melisabok:


openreview/conference/matching.py
```diff
@@ -913,6 +913,46 @@ def _build_config_invitation(self, scores_specification):
...
+                        edges.append(Edge(
+                        invitation=self.conference.get_conflict_score_id(self.alternate_matching_group),
+                        head=submission.id,
+                        tail=member,
+                        weight=-1,
+                        label='Alternate Conflict',
+                        readers=self._get_edge_readers(tail=member),
+                        writers=[self.conference.id],
+                        signatures=[self.conference.id]
+                    ))
```
carlosmondra, line 950: From what I understand `matching_group` would be the ACs and `alternate_matching_group` would be the SACs.

If that's the case, shouldn't `invitation=self.conference.get_conflict_score_id(self.matching_group)`? Because you are adding edges to the ACs' conflicts right?

melisabok: matching_group is SAC and altermate_matching_group is AC.

melisabok: Yes, I think the name "Alternate Conflict" is not right, which name do you suggest? or should I use "Conflict"?

carlosmondra: What if we parametrize it? The value for NeurIPS would be 'SAC conflict' or 'Conflict inherited from SAC'

melisabok: I added a parameter for the conflict label.

carlosmondra (approved):

PR #1287, opened 2022-05-26
Update revision invitation if submission deadline is updated

purujitgoyal:


openreview/venue_request/process/revisionProcess.py
```diff
@@ -28,6 +28,9 @@ def process(client, note, invitation):
                 if matching_invitation:
                     matching_invitation.cdate = openreview.tools.datetime_millis(submission_deadline)
                     client.post_invitation(matching_invitation)
+                revision_invitation = openreview.tools.get_invitation(client, conference.get_invitation_id('Revision'))
+                if revision_invitation:
+                    conference.setup_post_submission_stage(hide_fields=forum_note.content.get('hide_fields', []))
```
melisabok, line 33: this means we are going to call this every time there is a revision invitation and sometimes there is a revision invitation after the deadline. The function `setup_post_submission_stage` is expensive so I suggest the following:

- Check if now is between first and second deadline
- Check if the Revision invitation exists.
- Update the Revision invitation duedate and expdate with the new second deadline value.

openreview/venue_request/process/revisionProcess.py
```diff
@@ -28,6 +28,11 @@ def process(client, note, invitation):
                 if matching_invitation:
                     matching_invitation.cdate = openreview.tools.datetime_millis(submission_deadline)
                     client.post_invitation(matching_invitation)
+                revision_invitation = openreview.tools.get_invitation(client, conference.get_invitation_id('Revision'))
+                if revision_invitation and conference.submission_stage.second_due_date:
```
melisabok, line 32: you are missing if now is between the first and second deadline.

The Revision invitation can be used later for authors to submit a new revision of the paper during the rebuttal or the camera ready version.

You can reproduce this calling the Revision stage when there is a Revision invitation happening that is not related to the second deadline.

For example here: https://github.com/openreview/openreview-py/blob/master/tests/test_venue_request.py#L1727

purujitgoyal: Why do we need to compare with the first deadline? `now < second_deadline` should be sufficient, right?

purujitgoyal: There's another issue here, if the second deadline is updated to sometime before now, it won't update the revision invitation.

melisabok: You are right, comparing the date with now is not correct. 

Another alternative is to update the Revision with the second_due_date only if there is no value for submission_revision_deadline. Because if there is a value, that means the PCs decided to start a revision stage and that should win the second_due_date.

melisabok (approved):

celestemartinez:
Looks good! I don't think the second part of the issue was fixed here, will this be a separate PR? @purujitgoyal 

> The edit and trash button should not appear because none of these actions update the groups and invitations.

purujitgoyal:
> Looks good! I don't think the second part of the issue was fixed here, will this be a separate PR? @purujitgoyal
> 
> > The edit and trash button should not appear because none of these actions update the groups and invitations.
When I login as PC, I cannot see the trash and edit button on the request forum page, so I assumed that's fixed.

Issue #1010, opened 2021-10-16
setup_matching is very slow


neubig:
Hello,

We have some code like the following for ACL ARR:

    or_client = openreview.Client(baseurl='https://api.openreview.net', username=args.username, password=args.password)
    conference = openreview.helpers.get_conference(or_client, args.confid)
    or_client.impersonate(conference.id)
    conference.setup_matching(committee_id=f'aclweb.org/ACL/ARR/2021/{args.month}/{args.group}', build_conflicts=True, affinity_score_file=args.affinity_file)

This works fine, but it's extremely slow, taking about 30-40 minutes over 360 papers and 300 area chairs (not to mention when we have 2500 reviewers). It'd be nice if this could be made more efficient.

melisabok:
I think the build conflicts task is the most expensive one because it has to get all the profiles for all the area chairs and authors and their publications, we do an API call to get the publication of each profile. We can try to add parallelism to use multiple API calls, that should reduce the execution time. 

In the meantime, we are happy to compute the setup_mathing in one of our server to reduce the network latency.

PR #1288, opened 2022-05-27
Compute conflicts: load all the author profiles at once

melisabok:
Speed up the computation of conflicts loading all the author profiles at once instead of get the author profiles per submission.

Add support to get profiles as a dict with publications. 

Fixes https://github.com/openreview/openreview-py/issues/1010


purujitgoyal (approved):

carlosmondra:
Nice improvement! I just wanted to know if we can increase the `batch_size` to 1000. Other that, I think it's fine

melisabok:
> I just wanted to know if we can increase the batch_size to 1000
Done!

PR #1291, opened 2022-06-01
Feat/version increase

celestemartinez:


melisabok:
is there a reason to move from 1.2.5 to 1.3.0? why don't use 1.2.6?

celestemartinez:
> is there a reason to move from 1.2.5 to 1.3.0? why don't use 1.2.6?

Yes! Trying to follow this guide: https://semver.org/. Seems like increasing patch number is only for bug fixes and we seem to have at least one new feature introduced here. (but happy to use 1.2.6 if that's better!)

melisabok (approved):
looks good!

PR #1298, opened 2022-06-06
Makes expected_submissions required field of venue request form

nlbahy:


celestemartinez (approved):

celestemartinez:
Looks good! We would need to either deploy all pending venue requests before deploying this change or edit them to include this field otherwise the deploy will fail. 

https://openreview.net/forum?id=r6TDQ5Y2Pgg
https://openreview.net/forum?id=zkiinKLi3E0

PR #1304, opened 2022-06-09
Fix Comment Stage

nlbahy:
Selecting to notify PCs of every comment was not being converted to email_pcs in the comment stage. This fixes that so that email_pcs will be set to true.


tests/test_arr_venue.py
```diff
@@ -1092,3 +1093,53 @@ def test_ethics_review_stage(self, venue, client, helpers):
...
         assert client.get_invitation('aclweb.org/ACL/ARR/2021/September/Paper4/-/Ethics_Review')               
         assert client.get_invitation('aclweb.org/ACL/ARR/2021/September/Paper5/-/Ethics_Review')
+   
+    def test_comment_email_pcs(self, venue, client, helpers, test_client): 
+        pc_client=openreview.Client(username='pc@aclrollingreview.org', password='1234')
+        super_client=openreview.Client(username='openreview.net', password='1234')
+        request_form=client.get_notes(invitation='openreview.net/Support/-/Request_Form', sort='tmdate')[0]
+        conference = openreview.helpers.get_conference(client, request_form.id)
+
+        # Post a review stage note
```
celestemartinez, line 1103:
```suggestion
        # Post a comment stage note
```

tests/test_arr_venue.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
+from tokenize import String
```
melisabok, line 2: I don't think you need this import, right?

tests/test_arr_venue.py
```diff
@@ -1092,3 +1092,46 @@ def test_ethics_review_stage(self, venue, client, helpers):
         assert client.get_invitation('aclweb.org/ACL/ARR/2021/September/Paper3/-/Ethics_Review')
         assert client.get_invitation('aclweb.org/ACL/ARR/2021/September/Paper4/-/Ethics_Review')               
         assert client.get_invitation('aclweb.org/ACL/ARR/2021/September/Paper5/-/Ethics_Review')
+   
+    def test_comment_email_pcs(self, venue, client, helpers, test_client): 
+        pc_client=openreview.Client(username='pc@aclrollingreview.org', password='1234')
+        super_client=openreview.Client(username='openreview.net', password='1234')
```
melisabok, line 1098: why do you need to create super_client and not using client instead?

tests/test_arr_venue.py
```diff
@@ -1092,3 +1092,46 @@ def test_ethics_review_stage(self, venue, client, helpers):
...
+                "comment": "Let's see if the PCs are notified of this comment"
+            },
+            signatures = [f"aclweb.org/ACL/ARR/2021/September/Paper{paper.number}/Authors"],
+            readers = ["aclweb.org/ACL/ARR/2021/September/Program_Chairs"],
+            writers = ['aclweb.org/ACL/ARR/2021/September', f'aclweb.org/ACL/ARR/2021/September/Paper{paper.number}/Authors'],
+            forum = paper.forum
+        ))
+        helpers.await_queue()
+        messages = super_client.get_messages(to = 'pc@aclrollingreview.org', subject = '[ARR 2021 - September] An author commented on a paper. Paper Number: 5, Paper Title: "Paper title 5"')
+        assert messages
```
melisabok, line 1136: please assert that `len(messages) == 1` and `My test comment in messages[0]['content']['text']`

tests/test_arr_venue.py
nlbahy: When the tests were failing I wanted to make sure I was using the super user client. I just changed it back

melisabok (approved):
👍

PR #1309, opened 2022-06-14
TMLR feature requests

melisabok:
- Release the edit history to the public once the submission is marked as "Under Review"
- New Reviewer report form so the AE can report the reviewer's behavior
- Remind reviewers/AE every 4 weeks to switch from "Unavailable" to "Available"
- Allow author reorder during the Camera Ready Revision.
- Enable pagination in the EIC console
- Add AE as replyTo for all the reviewer reminders.
- Compute the reviewer affinity scores after the submission and instead of after under review status.
- Add author reminder for AE recommendation
- Allow same reviewer assignment when the paper is resubmitted


tests/test_journal.py
```diff
@@ -1453,6 +1500,8 @@ def test_submission(self, journal, openreview_client, test_client, helpers):
                 forum=note_id_1,
                 content={
                     'title': { 'value': 'Paper title VERSION 2' },
+                    'authors': { 'value': ['Melisa Bok', 'Test User'] },
```
xkopenreview, line 1503: i think this does not match with value defined in the invitation

xkopenreview (approved):

carlosmondra (approved):

PR #1310, opened 2022-06-15
Adds "Cancelled" as Matcher Status

purujitgoyal:


melisabok (approved):

melisabok:
@xkopenreview could be take into account this new state in the assignment page? I think the UI should enable the "Run Matcher"

PR #1315, opened 2022-06-16
Expire proposed assignment invitation

purujitgoyal:


purujitgoyal:
@melisabok when the deploy is called from the matcher, it always passes `enable_reviewer_reassignment` as True which won't expire the invitation.

melisabok:
You always need to expire the Proposed_Assignment invitation, it doesn't matter if `enable_reviewer_reassignment` is true or false. 

After the assignments are deployed, the Assignment invitation is being used for reassignments.

purujitgoyal:
okay, but after this change, most of the tests in `test_matching` and `test_matching_anonids.py` become invalid as they try to post new edges using the `Proposed Assignment` invitation.

PR #1318, opened 2022-06-21
Set multi-reply=true if single blind

purujitgoyal:


openreview/conference/builder.py
```diff
@@ -1012,7 +1012,7 @@ def setup_first_deadline_stage(self, force=False, hide_fields=[], allow_author_r
             additional_fields=self.submission_stage.additional_fields,
             remove_fields=self.submission_stage.remove_fields,
             only_accepted=False,
-            multiReply=False,
+            multiReply=False if self.submission_stage.double_blind else True,
```
melisabok, line 1015: the change looks good, you can add a test in a single blind submission with a second deadline. I think ESWC has that.

purujitgoyal: @melisabok what exactly needs to be tested here?

melisabok: the Revision invitation has the multireply property in True.

PR #1328, opened 2022-06-24
Fix/ Editor-in-Chief Console: link to new forum page

zbialecki:
Add option to EiC console to make all forum links go to new forum page.

Related to https://github.com/openreview/openreview-web/pull/899

### DIFF
--- a/openreview/journal/webfield/editorsInChiefWebfield.js
+++ b/openreview/journal/webfield/editorsInChiefWebfield.js
@@ -305,7 +305,8 @@ var formatData = function(
         content[currentValue] = submission.content[currentValue].value;
         return content;
       }, {}),
-      referrerUrl: referrerUrl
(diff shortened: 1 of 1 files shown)


melisabok:
Is this PR still valid?

melisabok:
does it mean the EICs will see the new forum page first? instead of having the option to switch using the banner?

zbialecki:
Yes, this option changes the forum link in the tables to /forum-new. If you'd rather they continue being directed to the old forum we don't have to add the option on live.

melisabok:
I would like to give them the option to choose if they want to go to the new forum. There are links to the forum everywhere and it can be confusing that one of the links go to the new one and the others don't.

zbialecki:
OK, then I guess I'll close this PR for now.

PR #1329, opened 2022-06-27
deletes assignment and proposed assignment edges for desk-rejected/withdrawn note

purujitgoyal:


openreview/conference/templates/desk_reject_process.py
```diff
@@ -98,6 +98,16 @@ def process_update(client, note, invitation, existing_note):
 
         forum_note = client.post_note(forum_note)
 
+        # Delete any assignment and proposed assignment edges
+        proposed_assignment_invitations = client.get_invitations(regex=CONFERENCE_ID + '/.*/-/Proposed_Assignment$')
```
melisabok, line 102: to avoid using a regex that can be very expensive, I suggest to make the query per committee group. You would need to pass SENIOR_AREA_CHAIRS_ID to check if there are senior area chairs:

```
if SENIOR_AREA_CHAIRS_ID:
      client.delete_edges(invitation=SENIOR_AREA_CHAIRS_ID/-/Assignment, head=forum_note.id, soft_delete=True)
      client.delete_edges(invitation=SENIOR_AREA_CHAIRS_ID/-/Proposed_Assignment, head=forum_note.id, soft_delete=True)

if AREA_CHAIRS_ID:
     ....
```

other ideas @celestemartinez ?

celestemartinez: I agree, pass `SENIOR_AREA_CHAIRS_ID` and `AREA_CHAIRS_ID` only when the conference uses them.

purujitgoyal: What if there are extra roles other than `SENIOR_AREA_CHAIRS` and `AREA_CHAIRS`?

melisabok (approved):

celestemartinez (approved):

PR #1330, opened 2022-06-27
don't update comment invitation when the request form is updated

melisabok:
I'm having a test failing because the comment invitation of the request from is being updated twice: when the deploy process updates it and when the process function of the request form runs when the writers are removed from the request note. 

To fix this, the update of the request form doesn't update the comment invitation if there is already a 'venue_id'.


melisabok:
@carlosmondra once you disable filter member in both APIs we can merge this PR and make a new release cc @celestemartinez

purujitgoyal (approved):

PR #1331, opened 2022-06-28
Uses participants to determine the invitees of the comment stage

purujitgoyal:
Also separates the usage of participants and readers


openreview/venue_request/venue_request.py
```diff
@@ -344,6 +345,21 @@ def setup_comment_stage(self):
                 'default': ['Program Chairs'],
                 'order': 29
             },
+            'additional_readers': {
+                'description': 'Select who should only be allowed to view the comments on submissions (other than the participants)',
+                'values-checkbox': [
+                    'Program Chairs',
+                    'Paper Senior Area Chairs'
```
celestemartinez, line 352:
```suggestion
                    'Paper Senior Area Chairs',
```

openreview/conference/builder.py
```diff
@@ -2398,51 +2398,53 @@ def __init__(self,
     allow_public_comments=False,
     anonymous=False,
     unsubmitted_reviewers=False,
+    submitted_reviewers=False,
```
melisabok, line 2401: I think it is time to refactor this a little bit, there are too many params that are too confusing. We could define the two parameters `participants` and `readers`. These two parameters are arrays that contain that elements from the Participants and Readers class, for example here: https://github.com/openreview/openreview-py/blob/master/openreview/conference/builder.py#L1887


You can check the SubmissionStage or ReviewStage classes to see how we handle the reviewers. 

From the participants array we are going to define the invitees and build the regex for signatures. 

From the readers array we are going to define the list of the dropdown. 

the parameters: anonymous, unsubmitted_reviewers, submitted_reviewers, authors, area_chairs, senior_area_chairs, additional_readers should be mapped to these two parameters.

openreview/conference/builder.py
```diff
@@ -2391,57 +2391,62 @@ def get_readers(self, conference, number, review_signature):
 
 class CommentStage(object):
 
+    class Readers(Enum):
+        EVERYONE = 0
+        SENIOR_AREA_CHAIRS = 1
```
melisabok, line 2396:
```suggestion
        SENIOR_AREA_CHAIRS_ASSIGNED = 1
```

Let's use assigned so it doesn't get confused with the general senior area chairs group

openreview/conference/builder.py
```diff
@@ -2391,57 +2391,62 @@ def get_readers(self, conference, number, review_signature):
...
+        AREA_CHAIRS = 2
```
melisabok, line 2397: same here, AREA_CHAIRS_ASSIGNED

openreview/conference/builder.py
```diff
@@ -2458,13 +2463,31 @@ def get_signatures_regex(self, conference, number):
...
 
-        if self.authors:
+        if self.Readers.AUTHORS in self.invitees:
             committee.append(conference.get_authors_id(number))
 
         return '|'.join(committee)
 
     def get_invitees(self, conference, number):
-        return conference.get_committee(number=number, with_authors=self.authors) + [conference.support_user]
+        invitees = [conference.get_program_chairs_id(), conference.support_user]
```
melisabok, line 2472: the function `get_signatures_regex` must also check the list of invitees to build the regex.

melisabok (approved):

PR #1337, opened 2022-06-29
Post Decision Stage: use active submissions to get decision notes

purujitgoyal:
Also, check if the authors have already been notified to avoid multiple messages

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -1759,7 +1759,7 @@ Program Chairs
         def send_notification(note):
             decision_note = None
             for reply in note.details['directReplies']:
-                if reply['invitation'].endswith(self.decision_stage.name):
+                if reply['invitation'].endswith('/-/' + self.decision_stage.name):
                     decision_note = reply
                     break
             subject = "[{SHORT_NAME}] Decision notification for your submission {submission_number}: {submission_title}".format(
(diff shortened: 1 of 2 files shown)


openreview/conference/builder.py
```diff
@@ -1754,24 +1754,27 @@ def is_release_authors(is_note_accepted):
         self.expire_recruitment_invitations()
 
     def send_decision_notifications(self, decision_options, messages):
-        decision_notes = self.client.get_all_notes(
-            invitation=self.get_invitation_id(self.decision_stage.name, '.*'),
-        )
-        paper_notes = {n.forum: n for n in self.get_submissions()}
+        paper_notes = self.get_submissions(details='directReplies')
```
melisabok, line 1757: the change looks good to me but it seems this line is not covered by the tests.

openreview/conference/builder.py
```diff
@@ -1754,24 +1754,27 @@ def is_release_authors(is_note_accepted):
...
 
         def send_notification(note):
-            paper_note = paper_notes[note.forum]
-            message = messages[note.content['decision']]
+            decision_note = None
+            for reply in note.details['directReplies']:
+                if reply['invitation'].endswith(self.decision_stage.name):
```
melisabok, line 1762:
```suggestion
                if reply['invitation'].endswith('/-/' + self.decision_stage.name):
```

in case there are other invitations that end with the same name

melisabok (approved):

melisabok:
@carlosmondra do you want to take a look?

carlosmondra (approved):

PR #1341, opened 2022-07-06
Adds option to hide fields in post decision stage

purujitgoyal:


openreview/venue_request/process/revisionProcess.py
```diff
@@ -229,6 +229,12 @@ def process(client, note, invitation):
                 'required': False
             }
 
+            content['hide_fields'] = {
```
melisabok, line 232: @celestemartinez do you think the Revision invitation should have this field "hide_fields"? 

they can use the post submission stage, right? or this stage is expired after the decisions are posted?

Another option is to show the field in the Post Decision Stage form.

celestemartinez: The Post Submisison Stage has no expiration date, so this invitation can be used at any point, but I am not sure if every venue uses it. It might not be a bad idea to also include it in the Post Decision Stage form

openreview/conference/builder.py
```diff
@@ -1719,6 +1719,9 @@ def is_release_authors(is_note_accepted):
                     anonymous=False
                 )
 
+            for field in hide_fields:
```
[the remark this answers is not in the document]
melisabok: @purujitgoyal could you check this?

openreview/conference/builder.py
```diff
@@ -1678,7 +1678,7 @@ def get_attachment_file(submission):
         for future in futures:
             result = future.result()
 
-    def post_decision_stage(self, reveal_all_authors=False, reveal_authors_accepted=False, decision_heading_map=None, submission_readers=None):
+    def post_decision_stage(self, reveal_all_authors=False, reveal_authors_accepted=False, decision_heading_map=None, submission_readers=None, hide_fields=[]):
```
melisabok, line 1681: instead of passing `hide_fields` as a parameter, can you add this field to the SubmissionStage class and get the value from there?

I would like to do the same for `setup_post_submission_stage`

openreview/conference/builder.py
```diff
@@ -1069,12 +1069,14 @@ def setup_final_deadline_stage(self, force=False, hide_fields=[]):
         if self.use_area_chairs:
             self.set_area_chairs()
 
-    def setup_post_submission_stage(self, force=False, hide_fields=[]):
+    def setup_post_submission_stage(self, force=False, hide_fields=None):
```
melisabok, line 1072: is it possible to remove `hide_fields` from here? I think the value should always come from the submission stage object.

openreview/conference/builder.py
```diff
@@ -1678,7 +1680,10 @@ def get_attachment_file(submission):
         for future in futures:
             result = future.result()
 
-    def post_decision_stage(self, reveal_all_authors=False, reveal_authors_accepted=False, decision_heading_map=None, submission_readers=None):
+    def post_decision_stage(self, reveal_all_authors=False, reveal_authors_accepted=False, decision_heading_map=None, submission_readers=None, hide_fields=None):
```
melisabok, line 1683: same here, remove parameter. The submission stage should always have the value.

openreview/conference/builder.py
purujitgoyal: I've left this because there are a bunch of tests that directly use this instead of setting the submission stage.

melisabok:
hide_fields was already added to the revision invitation

PR #1344, opened 2022-07-07
removed all examples

nlbahy:


docs/source/get_all_venues.rst
```diff
@@ -1,23 +0,0 @@
-Getting All Venues
-==========================
-
-::
-
-    >>> import openreview
-    >>> c = openreview.Client(baseurl='https://api.openreview.net')
-    >>> venues = openreview.tools.get_all_venues(c)
```
melisabok, line 8: there is another way of doing this, client.get_group(id='venues'), I'm not sure if the documentation has this.

docs/source/get_submission_invitations.rst
```diff
@@ -1,101 +0,0 @@
-Getting Submissions
-========================================
-
-All the Invitation Ids for Submissions can be retrieved like this::
-
-    >>> import openreview
-    >>> c = openreview.Client(baseurl='https://api.openreview.net')
-    >>> invitations = openreview.tools.get_submission_invitations(c)
```
melisabok, line 8: @nlbahy do you mind removing the function `get_submission_invitations` from the tools file?

docs/source/get_all_venues.rst
nlbahy: I did copy this into the docs, I can change it to get_group(id='venues')

docs/source/multiprocessing.rst
```diff
@@ -1,9 +0,0 @@
-Parallel Execution
-==================
-
-The tools package includes utilities that makes it easy to run a set of requests in parallel using a multiprocessing pool:
-
-    >>> notes = list(tools.iterget_notes(client, invitation='ICLR.cc/2019/Conference/-/Blind_Submission'))
-    >>> openreview.tools.parallel_exec(notes, do_work)
```
melisabok, line 7: we could keep this file but change the documentation and mention that `tools.concurrent_get` should do the trick

melisabok: cc @carlosmondra

docs/source/profiles.rst
```diff
@@ -1,142 +0,0 @@
...
-     'mccallum@cs.umass.edu',
-     'cs.umass.edu']
-
-Conflicts of interest are represented by direct relationships or by common domain affiliations.
-
-
-Add evidence to a profile
-----------------------------
-
-Add DBLP link:
```
melisabok, line 73: do we have these examples in the new doc?

docs/source/get_all_venues.rst
melisabok: venues contains all the venue ids. The host group only contain the root groups.

docs/source/profiles.rst
nlbahy: No, I figured we would not want people to use those examples

docs/source/multiprocessing.rst
carlosmondra: There is no section in the docs for this. If we were to add it, it should go under the Reference section. I would suggest having an openreview-py subsection under Reference. We could also link readthedocs from this subsection.

PR #1348, opened 2022-07-12
update regex for titles

purujitgoyal:


openreview/conference/matching.py
```diff
@@ -769,9 +769,9 @@ def _build_config_invitation(self, scores_specification):
                 'signatures': {'values': [self.conference.get_id()]},
                 'content': {
                     'title': {
-                        'value-regex': '.{1,250}',
+                        'value-regex': '(?!^ +$)^[a-zA-Z0-9-_ ]{1,250}$',
```
melisabok, line 772: could you try to add a test, probably in test_matching.py, create an Assignment Configuration note after calling `conference.setup_committee_matching`.

melisabok: @purujitgoyal could you add a test here?

carlosmondra, line 772: what is the negative lookahead here for?

purujitgoyal: To check for only spaces.

carlosmondra: I see, I think it would be better to have a regex like this instead:
```
^[a-zA-Z0-9-_][a-zA-Z0-9-_ ]{0,250}$
```
This achieves the same thing in a simpler way. It also disallows spaces at the beginning.

carlosmondra (approved):
Thanks!

melisabok (approved):

PR #1353, opened 2022-07-19
Adds SACs as AC group writers

purujitgoyal:


melisabok:
@nlbahy please confirm but I think the assigned SAC needs to be writer of the assigned AC group, right? not the general AC group.

nlbahy:
> @nlbahy please confirm but I think the assigned SAC needs to be writer of the assigned AC group, right? not the general AC group.

I have been fixing this by adding the overall SAC group as writers of the overall AC group, but technically this would allow SACs to edit the AC group. So it is probably safer to add each paper SAC as a writer of each paper AC.

melisabok:
I'm nervous that SAC users can edit the AC group :/

melisabok:
Fixed it in the API 2

PR #1356, opened 2022-07-25
Adds PC only Revision Invitation for single-blind venues

purujitgoyal:


openreview/conference/builder.py
```diff
@@ -1037,8 +1037,7 @@ def setup_first_deadline_stage(self, force=False, hide_fields=[], allow_author_r
             force=True
         )
 
-
-    def setup_final_deadline_stage(self, force=False, hide_fields=[]):
+    def setup_final_deadline_stage(self, force=False, hide_fields=[], pc_revision_stage=False):
```
melisabok, line 1040: what does `pc_revision_stage` stand for?

openreview/conference/builder.py
```diff
@@ -1064,6 +1063,15 @@ def setup_final_deadline_stage(self, force=False, hide_fields=[]):
             hide_fields=hide_fields
         )
 
+        if not self.submission_stage.double_blind and pc_revision_stage:
+            self.submission_revision_stage = SubmissionRevisionStage(
+                name='Revision',
+                start_date=None if force else self.submission_stage.due_date,
+                due_date=None,
+                multiReply=True
```
melisabok, line 1071: could you add a property to the SubmissionRevisionStage class to represent the invitees? it could be called: `program_chairs_only=False`, the default value indicates that program chairs and authors can make edits.

celestemartinez (commented):
I wonder if maybe we should call this revision by another name? Calling it 'Revision' means it can be overwritten if the PCs decide to open revisions to authors right after the deadline. It might not be a bad idea to have any changes done by PCs under a different invitation name. 

Also, should we allow PCs to use PaperX/Authors as the siganture? Right now they can choose between PCs and Authors to sign.

melisabok:
Good idea, how should we call the invitation? 

PCs should sign the revision using the PC group only.

melisabok:
This is added to the API 2 venues

PR #1361, opened 2022-07-28
Feature/request name removal

melisabok:
- Request a name removal posting a note using the invitation: Profile_Name_Removal
- Support team evaluates the requests and accepts or rejects the request posting a reference with the invitation Profile_Name_Removal_Decision
- Process function of Profile_Name_Removal_Decision does the following actions:

1.  Check if the name to delete is the profile id then rename it. 
2. Post a reference changing the authorid of all the publications using the invitation Author_Rename
3. Change the signatures of all the notes when the signature is the name to delete
4. Rename all the edges where the name to delete is either a head or a tail
5. Replace the members for all the group when the name to delete is a member
6. Remove the name to delete from the profile
7. Remove the name to delete group
8. Send a confirmation email to the user

- Add Archive invitation so we can test some publications. 

Fixes https://github.com/openreview/openreview-api-v1/issues/1349


xkopenreview:
how about edge tail, should it also be replaced

melisabok:
> how about edge tail, should it also be replaced

Good point! I will add it

xkopenreview:
i think the preferred id should be stored in the note too
otherwise access to user's profile will be lost when the id is deleted

melisabok:
What about the signatures of the note? it should always contain a valid profile id, right?

xkopenreview:
some profiles has name mapped to multiple tilde ids
should it be handled by the process function
or have ui to create multiple notes when user request to delete such name

tests/test_profile_management.py
```diff
@@ -0,0 +1,386 @@
...
+The OpenReview Team.
+'''
+
+        ## Accept the request
+        decision_note = client.post_note(openreview.Note(
+            referent=request_note.id,
+            invitation='openreview.net/Support/-/Profile_Name_Removal_Decision',
+            readers=['openreview.net/Support'],
+            writers=['openreview.net/Support'],
+            signatures=['openreview.net/Support'],
```
xkopenreview, line 126: should the signature be tilde id of su?

melisabok:
> some profiles has name mapped to multiple tilde ids
should it be handled by the process function
or have ui to create multiple notes when user request to delete such name

I thought about this. When someone wants to remove the name "Joe Doe" and there is more than one tilde id mapped to that name: Joe_Doe1, Joe_Doe2, etc. 

What about supporting an array of requested usernames? 

The concern about this is that the array may contain different names and the moderator needs to approve all of them. We can validate that all the tilde ids correspond to the same name.

tests/test_profile_management.py
melisabok: I think only super user is signatory of the support group so it is fine. I would like to stop using su for everything specially moderation that we would like to delegate the task to someone else.

melisabok:
@xkopenreview I made the following changes:

- Added new field "name" to the Profile Name Removal invitation.
- Convert field "username" into "usernames" so you can remove all the usernames pointing to the same name.

openreview/profile/process/request_remove_name_decision_process.py
```diff
@@ -0,0 +1,141 @@
...
+        print('Remove tilde id group')
+        client.delete_group(username)
+
+    client.post_message(subject='Profile name removal request has been accepted', 
+    recipients=[profile.id], 
+    message=f'''Hi {{{{fullname}}}},
+
+We have received your request to remove the name "{request_note.content['name']}" from your profile: https://openreview.net/profile?id={profile.id}.
+
+The name has been removed from your profile. Please check the information listed in your profile is correct.
```
nlbahy, line 136:
```suggestion
The name has been removed from your profile. Please check that the information listed in your profile is correct.
```

nlbahy:
If I submit a request for name removal I don’t need to save my profile before the request is sent. If SuperUser then accepts the request before I save the changes to my profile, I get an error when trying to submit the changes: Error: Can not create profile, username is already associated with [preferred id]

nlbahy:
Testing on the test data works for me, but when I try to test with real data, accepting a name change request causes this process function error: 

"error": {
"name": "Error",
"message": "AttributeError: 'Note' object has no attribute 'referent'",
"status": 500,
"stack": "Error: AttributeError: 'Note' object has no attribute 'referent'\n    at PythonShell.parseError (/Users/nadia/Code/openreview-api/node_modules/python-shell/index.js:295:21)\n    at terminateIfNeeded (/Users/nadia/Code/openreview-api/node_modules/python-shell/index.js:190:32)\n    at ChildProcess.<anonymous> (/Users/nadia/Code/openreview-api/node_modules/python-shell/index.js:182:13)\n    at ChildProcess.emit (node:events:526:28)\n    at Process.ChildProcess._handle.onexit (node:internal/child_process:291:12)\n    ----- Python Traceback -----\n    File \"/var/folders/8z/z7dqcjjd1pb_8_9ktghhm2nm0000gn/T/pythonShellFile7577647843.py\", line 165, in <module>\n      process(client, note, invitation)\n    File \"/var/folders/8z/z7dqcjjd1pb_8_9ktghhm2nm0000gn/T/pythonShellFile7577647843.py\", line 14, in process\n      request_note = client.get_note(note.referent)"
}

melisabok:
> If I submit a request for name removal I don’t need to save my profile before the request is sent. If SuperUser then accepts the request before I save the changes to my profile, I get an error when trying to submit the changes: Error: Can not create profile, username is already associated with [preferred id]

What did you change in the profile that you didn't save?

> Testing on the test data works for me, but when I try to test with real data, accepting a name change request causes this process function error:

Can you share the profile id that you are using to test? there must be something wrong when updating the author ids of publications.

nlbahy:
> What did you change in the profile that you didn't save?
I submitted the name change request, i didn't change anything in the profile. 

> Can you share the profile id that you are using to test? there must be something wrong when updating the author ids of publications.

It happened with every profile ID I tried; I tried it on ~Nadia_L'Bahy1 for instance

carlosmondra:
@melisabok do you think this fixes https://github.com/openreview/openreview-api-v1/issues/1349?

melisabok:
> @melisabok do you think this fixes https://github.com/openreview/openreview-api-v1/issues/1349?

Yes! I forgot to unlink it.

melisabok:
@nlbahy @xkopenreview 

>  "AttributeError: 'Note' object has no attribute 'referent'",

This error is being solved changing the configuration of development.json in the new API. Use another number for the database number:

```
    "jobQueue" : {
      "db": 1,
      "jobActive": 24
    },
```

for example: 10.

openreview/profile/process/request_remove_name_decision_process.py
```diff
@@ -0,0 +1,222 @@
...
+                    if publication.content.get('authors'):
+                        authors.append(publication.content['authors'][index])
+                    authorids.append(publication.content['authorids'][index])
+            if needs_change:
+                content = {
+                    'authors': authors,
+                    'authorids': authorids
+                }
+                if '_bibtex' in publication.content:
+                    content['_bibtex'] = publication.content['_bibtex']['value'].replace(openreview.tools.pretty_id(username), preferred_name)
```
nlbahy, line 53: This line threw an error for me when I tested it on real profile data:
TypeError: string indices must be integers

melisabok: sorry about this, I tested with the API 2 but API 1. I pushed the fix, please try again.

nlbahy: That worked!

nlbahy:
One thing I just realized is that the mdate of the submissions are changing. Should we set the pdate in this pr when submissions are updated?

nlbahy:
I noticed that clicking the link "View all # publications" on the profile does not work after the name is changed

melisabok:
> One thing I just realized is that the mdate of the submissions are changing. Should we set the pdate in this pr when submissions are updated?

I don't think this PR should change the pdate but we should start using pdate in the request form during the post decision stage.

melisabok:
> I noticed that clicking the link "View all # publications" on the profile does not work after the name is changed

Could you share the profile id and which name you removed?

nlbahy:
> > I noticed that clicking the link "View all # publications" on the profile does not work after the name is changed
> 
> Could you share the profile id and which name you removed?

Sorry, I'm not able to reproduce the error. You can ignore my comment!

melisabok:
@carlosmondra @xkopenreview I think this PR is ready to go!

nlbahy (approved):

carlosmondra (approved):

xkopenreview (approved):

openreview/profile/webfield/archiveWebfield.js
```diff
@@ -0,0 +1,220 @@
+// ------------------------------------
+// Advanced venue homepage template
+//
+// This webfield displays the conference header (#header),
+// important instructions for this phase of the conference,
+// and a tabbed interface for viewing various types of notes.
+// ------------------------------------
+
+// Constants
+var CONFERENCE_ID = 'openreview.net/Archive';
```
xkopenreview, line 10: i think it should be OpenReview.net/Archive? @melisabok

melisabok: yes! please could you fix it in the live site and create a PR here? thanks

openreview/profile/process/request_remove_name_decision_process.py
```diff
@@ -0,0 +1,222 @@
...
+        group.signatures = ['~Super_User1']
+        client.post_group(group)
+        
+        client.post_profile(openreview.Profile(
+            referent = profile.id, 
+            invitation = '~/-/invitation',
+            signatures = ['~Super_User1'],
+            content = {},
+            metaContent = {
+                'names': {
```
xkopenreview, line 200: @carlosmondra this is the line where names is an obj

carlosmondra: I see thanks. I'm not sure if this is causing the issue. I'll try to reproduce it. It would be interesting to see if all the profiles that have requested name deletion have this issue.

PR #1370, opened 2022-08-17
fix review stage for single blind venues

melisabok:


nlbahy:
Everything works, but while testing it I found an error in the Ethics Chair webfield: line 299 where it has "ETHICS_REVIEW_CHAIR_NAME" I think it should be "ETHICS_CHAIRS_NAME"

melisabok:
could you make the fix and push it here?

PR #1377, opened 2022-08-22
Features/cvpr requests

melisabok:


tests/test_cvpr_conference.py
```diff
@@ -0,0 +1,232 @@
...
+            referent= request_form.id,
+            replyto= request_form.id,
+            signatures= ['~Program_CVPRChair1'],
+            writers= [],
+        ))
+
+        helpers.await_queue()
+
+
+    def test_rebuttal_stage(self, conference, helpers, test_client, client):
```
melisabok, line 158: @nlbahy rebuttal stage with a single rebuttal and pdf supported.

nlbahy (approved):

openreview/venue_request/venue_request.py
```diff
@@ -1135,6 +1135,13 @@ def setup_request_form(self):
                 'order': 37,
                 'required': False,
                 'hidden': True # Change this value on exception request from the PCs.
+            },
+            'include_expertise_selection': {
```
melisabok, line 1139: @haroldrubio now the venues can select if they can include or exclude their expertise selection. 
I see the expertise API has a `exclusion_inv` parameter. We have two options to support the inclusion:

1. Add a new parameter `inclusion_inv`
2. Rename the parameter to `expertise_selection_inv` and the expertise server can inspect the invitation and check if the edges represent exclusions or inclusions.

What do you think? I like option 2.

PR #1383, opened 2022-09-02
add function to get journal object

celestemartinez:


celestemartinez:
Could we move this outside of the JournalRequest() class?

Right now we would need to do:
`journal_request = JournalRequest(client, 'OpenReview.net/Support')`
`journal = journal_request.get_journal(client, 'xLdUNpcuxA')`

openreview/journal/invitation.py
```diff
@@ -116,7 +116,11 @@ def get_process_content(self, file_path):
         process = None
         with open(os.path.join(os.path.dirname(__file__), file_path)) as f:
             process = f.read()
-            return process.replace('openreview.journal.Journal()', f'openreview.journal.Journal(client, "{self.journal.venue_id}", "{self.journal.secret_key}", contact_info="{self.journal.contact_info}", full_name="{self.journal.full_name}", short_name="{self.journal.short_name}", website="{self.journal.website}", submission_name="{self.journal.submission_name}")')
+            if self.journal.request_form_id:
+                return process.replace('openreview.journal.Journal()', f'openreview.journal.JournalRequest.get_journal(client, "{self.journal.request_form_id}")')
```
melisabok, line 120: this is the reason I made a class method so it is easier to get the journal object with all the settings.

PR #1394, opened 2022-09-09
moderate profile function

melisabok:


nlbahy:
I think we need a way to post a reason when limiting the profile so that the email that is sent to the user is correct

melisabok:
I don't think the API sends email to the user when the profile is limited. 

The user screening script is the one sending the email. Do you think we need to have a reason in the profile because we may use this status for other purposes?

nlbahy:
when I do client.moderate_profile and set it to limited, this email is sent to the user:
subject: OpenReview profile has been limited
content: Your OpenReview profile has been limited for the following reason:
undefined

melisabok:
ohhh I see, @carlosmondra should the backend send an email when the status of the profile changes? I don't think we should do that, at least from active -> limited and limited -> active

melisabok:
@nlbahy can you post this comment in the API PR: https://github.com/openreview/openreview-py/pull/1394#issuecomment-1246869483

https://github.com/openreview/openreview-api-v1/pull/2706

I'm going to merge this PR given it is just the implementation of the call to the API

nlbahy (approved):

PR #1396, opened 2022-09-12
Journal: setup action editor matching for submitted submissions

melisabok:
- Add a function to setup AE matching data and deploy assignments using the matching system.
- Rename EIC console complete submissions to be all submissions and show all the submissions.
- Check if the invitation is not expired before expiring it.
- Add an EIC_Revision invitation so the EICs can still edit the submissions after the camera ready revision was approved
- Refactor how to setup the JournalRequest, through a pytest fixture.
- Rename _process fields to end with _script
- Add new more fields to the review, official recommendation and decision forms.
-

### DIFF
--- a/openreview/journal/assignment.py
+++ b/openreview/journal/assignment.py
@@ -172,7 +172,7 @@ class Assignment(object):
                 ## Get AE recommendations
                 ae_recommendations = self.client.get_edges(invitation=journal.get_ae_recommendation_id(), head=submitted_submission.id)
                 if len(ae_recommendations) >= 3:
-                    ## Mark the papers that needs assignments. use venue: "TMLR Assigning AE" and venueid: 'TMLR/Assign_AE_20220928'
+                    ## Mark the papers that needs assignments. use venue: "TMLR Assigning AE" and venueid: 'TMLR/Assign_AE'
                     self.client.post_note_edit(
                         invitation = journal.get_meta_invitation_id(),
                         signatures = [journal.venue_id],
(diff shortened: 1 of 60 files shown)


openreview/journal/assignment.py
```diff
@@ -160,3 +160,117 @@ def compute_affinity_scores(self, note, committee_id):
...
+        journal = self.journal
+        submitted_submissions = self.client.get_notes(invitation= journal.get_author_submission_id(), content = { 'venueid': journal.submitted_venue_id })
+
+        for submitted_submission in tqdm(submitted_submissions):
+            ## Get AE group is empty
+            if not self.client.get_group(journal.get_action_editors_id(submitted_submission.number)).members:
+                ## Get AE recommendations
+                ae_recommendations = self.client.get_edges(invitation=journal.get_ae_recommendation_id(), head=submitted_submission.id)
+                if len(ae_recommendations) >= 3:
+                    ## Mark the papers that needs assignments. use venue: "TMLR Assigning AE" and venueid: 'TMLR/Assign_AE_20220928'
```
celestemartinez, line 175: What does the number in the venueid mean? Is this supposed to just be `TMLR/Assigning_AE`?

melisabok: sorry, outdated comment :( I was trying to use an unique venueid for each batch of submissions but it is too hard to do it so I decided to use the same venueid `TMLR/Assigning_AE` every time we setup the matching.

celestemartinez (approved):

PR #1407, opened 2022-09-22
Fix: update support group webfield

celestemartinez:
- Update the support group webfield every time we update the request form
- **Note**: the webfield is slightly different from what is on the live site (live site stil has things like TPMS)
- I noticed that the support group created in the setup script of the API creates this group slightly different than how it is in the live site: https://github.com/openreview/openreview-api-v1/blob/214dcd1b1e9971afee20bce13351f7aecbab57c1/scripts/mongo_database.js#L316 (for example, it is not visible to everyone, not sure if we want to change this)

### DIFF
--- a/openreview/venue_request/venue_request.py
+++ b/openreview/venue_request/venue_request.py
@@ -791,19 +791,12 @@ class VenueRequest():
         self.client = client
         self.super_user = super_user
 
-        if not self.support_group:
+        if self.support_group:
             with open(os.path.join(os.path.dirname(__file__), 'webfield/supportRequestsWeb.js')) as f:
                 file_content = f.read()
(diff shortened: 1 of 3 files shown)


nlbahy (approved):

melisabok (approved):

melisabok:
> Note: the webfield is slightly different from what is on the live site (live site stil has things like TPMS)

Could you update the live site with the version without TPMS?

> I noticed that the support group created in the setup script of the API creates this group slightly different than how it is in the live site

I don't have a strong opinion about this, we used to have our emails before but no anymore. I think having it private is ok for now. At some point we want the support group to do more things and remove the super user dependency.

celestemartinez:
The webfield has been updated on the live site

PR #1409, opened 2022-09-23
Feature/venue class

celestemartinez:


openreview/stages/venue_stages.py
```diff
@@ -95,18 +95,15 @@ def __init__(
         self.public = self.Readers.EVERYONE in self.readers
         self.author_reorder_after_first_deadline = author_reorder_after_first_deadline
 
-    def get_readers(self, conference, number, decision_note=None):
+    def get_readers(self, conference, number, decision=None):
```
celestemartinez, line 98: @melisabok I made this change here because both APIs are using this function. Let me know if something doesn't look good or if there is a better way to handle this

melisabok (approved):

PR #1416, opened 2022-09-28
Fix/invitation cdate process

melisabok:
- Move the process function to create paper invitations to the meta invitation so it can be used by several super invitations.
- Refactor tests, organize it in stages

### DIFF
--- a/openreview/venue/venue.py
+++ b/openreview/venue/venue.py
@@ -200,9 +200,9 @@ class Venue(object):
         options['contact'] = self.contact
         return options
 
-    def get_submission_venue_id(self, submission_invitation_id=None):
-        if submission_invitation_id:
-            return f'{self.venue_id}/{submission_invitation_id}'
+    def get_submission_venue_id(self, submission_invitation_name=None):
+        if submission_invitation_name:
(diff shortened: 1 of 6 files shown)


openreview/venue/venue.py
```diff
@@ -200,8 +200,15 @@ def get_homepage_options(self):
         options['contact'] = self.contact
         return options
 
+    def get_submission_venue_id(self, submission_invitation_id=None):
+        if submission_invitation_id:
+            return f'{self.venue_id}/{submission_invitation_id}'
```
celestemartinez, line 205: Why would we pass `submission_invitation_id` here? What would this venue id look like?

melisabok: I did it in case that in the future we may have multiple submission invitations.

celestemartinez: oh I see! I just got confused because it seems like the `submission_invitation_id` would already include the venue_id

melisabok: ohhh you are right! the parameter should be called "submission_name"! let me fix it.

celestemartinez (approved):

PR #1419, opened 2022-10-06
Feat/venue api2

celestemartinez:


tests/test_venue_request_v2.py
```diff
@@ -99,6 +99,12 @@ def venue(self, client, test_client, helpers):
             writers=[support_group_id]
         ))
 
+        helpers.await_queue()
+
+        submission_inv = openreview_client.get_invitation('V2.cc/2030/Conference/-/Submission')
+        assert submission_inv.duedate
```
melisabok, line 105: can you check that the duedate is the same than the variable `due_date`? and expdate is due_date + 30 min?

PR #1426, opened 2022-10-17
Feat: parametrize submission email template

celestemartinez:


celestemartinez:
@melisabok I didn't change the tokens we use in the process function since only we will be using this for now, but it shows like this in the request form:
Screen Shot 2022-10-17 at 3 48 07 PM

openreview/venue_request/venue_request.py
```diff
@@ -30,8 +30,15 @@ def setup_venue_revision(self):
             'values-dropdown':  ['abstract','keywords', 'pdf', 'TL;DR'],
             'description': 'Fields to remove from the default form: abstract, keywords, pdf, TL;DR'
         }
-        revision_content['homepage_override'] = {
+        revision_content['submission_email'] = {
             'order': 20,
+            'description': 'Please review the email sent to authors when a submission is posted. Make sure not to remove the parenthesized tokens.',
+            'default': 'Your submission to ${SHORT_PHRASE} has been ${action}.\n\nSubmission Number: ${note.number} \n\nTitle: ${note.content.title} ${noteAbstract} \n\nTo view your submission, click here: ${baseUrl}/forum?id=${note.forum}',
```
melisabok, line 36: let's replace it with {{}} to be consistent with the other tokens and all lowercase:

{{short_phrase}}
{{action}}
{{submission_number}}
{{submission_title}}
{{submission_id}}
{{submission_abstract}}

melisabok: we don't need to parametrize short_phrase.

celestemartinez: and then translate this for the process function? Should we rewrite the submission process function to python or is there a reason why we are still using js?

melisabok (approved):

PR #1434, opened 2022-10-25
translate review process to python

melisabok:


melisabok:
API 2 already uses Python

PR #1436, opened 2022-10-25
Support alternate group expertise selection

haroldrubio:
Also updates the schema for regular expertise selection


melisabok (approved):
it looks good, in the future we should rename the exclusion_invitation parameter name and have one for each group. We can do this later.

PR #1439, opened 2022-10-27
Fix: add recruitment preprocess

celestemartinez:
Add recruitment preprocess which checks:
- key is correct
- user is in invited group
- user has not been assigned to a paper

### DIFF
--- a/openreview/venue/matching.py
+++ b/openreview/venue/matching.py
@@ -41,6 +41,8 @@ class Matching(object):
         is_assignment_invitation=edge_id.endswith('Assignment') or edge_id.endswith('Aggregate_Score')
         paper_number = '${{2/head}/number}' if is_assignment_invitation else None
 
+        assignment_or_proposed = edge_id.endswith('Assignment')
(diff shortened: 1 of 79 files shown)


melisabok (approved):

openreview/venue/matching.py
```diff
@@ -995,7 +995,13 @@ def deploy_sac_assignments(self, assignment_title, overwrite):
                         if overwrite:
                             sac_group.members=[]
                         sac_group.members.append(sac)
-                        client.post_group(sac_group)
+                        client.post_group_edit(
```
melisabok, line 998: I think you can post the edit to override the members only:

```
invitation = venue.get_meta_invitation_id(),
readers = [venue.venue_id],
writers = [venue.venue_id],
signatures = [venue.venue_id],
group = openreview.api.Group(
     id = sac_group_id,
     members = [sac]
)
```

or add_members to add a sac. I think we will always override so no need to keep both choices.

PR #1441, opened 2022-10-28
add accepted email template

celestemartinez:


openreview/venue_request/venue_request.py
```diff
@@ -1405,6 +1405,17 @@ def setup_venue_recruitment(self):
 
         Program Chairs
         '''
+            },
+            'accepted_email_template': {
+                'value-regex': '[\\S\\s]{1,10000}',
+                'description': 'Please review the email sent to users when they accept a recruitment invitation. Make sure not to remove the parenthesized tokens.',
+                'order': 8,
+                'hidden': True,
+                'default': '''Thank you for accepting the invitation to be a {{reviewer_name}} for {SHORT_PHRASE}.
```
melisabok, line 1414: shouldn't be here `invitee_role` instead of reviewer_name?

melisabok: `SHORT_PHRASE` should be already replaced in the default template with the venue Abbreviated_Venue_Name?

celestemartinez: - well {reviewer_name} is replaced with `REVIEWER_NAME` in the process function

- {SHORT_PHRASE} is replaced with `conference.get_short_name()` in the deploy process function

PR #1448, opened 2022-11-02
fix bidding console configuration

melisabok:


zbialecki:
I tested this PR with the changes in https://github.com/openreview/openreview-web/pull/1169 and it's working. However I had to change the component config. I'm using the config below:

```js
// Webfield component
return {
  component: 'ExpertiseConsole',
  properties: {
    venueId: domain.id,
    apiVersion: 2
  }
}
```

I removed the description because I modified the default description to handle both "include" and "exclude" options.

melisabok:
What about overriding the description?

PCs should be able to edit the instructions same in the Bidding console.

zbialecki:
> What about overriding the description?

The description prop is still available and the PCs can use it to override the description at any time, I just thought that for the template it wasn't necessary since we have the default.

tests/test_icml_conference.py
```diff
@@ -442,4 +480,45 @@ def test_ac_bidding(self, client, openreview_client, helpers, test_client):
...
+        review_stage_note = openreview.Note(
+            content={
+                'review_start_date': start_date.strftime('%Y/%m/%d'),
+                'review_deadline': due_date.strftime('%Y/%m/%d'),
+                'make_reviews_public': 'No, reviews should NOT be revealed publicly when they are posted',
+                'release_reviews_to_authors': 'No, reviews should NOT be revealed when they are posted to the paper\'s authors',
+                'release_reviews_to_reviewers': 'Reviews should be immediately revealed to the paper\'s reviewers who have already submitted their review',
+                'remove_review_form_options': 'title',
+                'email_program_chairs_about_reviews': 'Yes, email program chairs for each review received',
+                'review_rating_field_name': 'review_rating'
```
xkopenreview, line 503: i think this does not match with the rating field name in official comment invitation which is called "rating"

melisabok: fixed!

xkopenreview:
the official review invitation of ICML has only optional:true for ddate and no deletable
it will disable deleted official review from being restored

https://github.com/openreview/openreview-api/pull/85#discussion_r913918014

melisabok:
>the official review invitation of ICML has only optional:true for ddate and no deletable
it will disable deleted official review from being restored

thanks for catching this, I added the deletable: true to all the ddates. However I still get the error deleting the note: "The field review cannot be empty or missing"

xkopenreview:
> > the official review invitation of ICML has only optional:true for ddate and no deletable
> > it will disable deleted official review from being restored
> 
> thanks for catching this, I added the deletable: true to all the ddates. However I still get the error deleting the note: "The field review cannot be empty or missing"

@celestemartinez is working on a ui fix for this.
adding deletable: true is to allow deleted comments to be restored

openreview/venue/webfield/paperBidWebfield.js
```diff
@@ -1,19 +1,19 @@
...
+const commitee_name = entity.content.commitee_name.value
 
 return {
   component: 'BidConsole',
   version: 1,
   properties: {
     header: {
-      "title": `${COMMITTEE_NAME} Bidding Console`,
+      "title": `${commitee_name.endsWith('s') ? commitee_name.slice(0, -1) : commitee_name} Bidding Console`,
       "instructions": BID_INSTRUCTIONS
```
zbialecki, line 11: I think the bid instructions should be included in this template and use varibles from the domain content. I don't think it needs to be set by python.

openreview/venue/webfield/seniorAreaChairsWebfield.js
```diff
@@ -0,0 +1,30 @@
+// Webfield component
+return {
+    component: 'SeniorAreaChairConsole',
```
zbialecki, line 3: I think this and the PC console template should be changed to use 2 spaces for indentation like all the other JS files.

xkopenreview (approved):

openreview/venue/group.py
```diff
@@ -84,15 +98,27 @@ def set_landing_page(self, group, parentGroup):
                     content = content.replace("var PARENT_GROUP_ID = '';", "var PARENT_GROUP_ID = '" + parentGroup.id + "';")
                 content = content.replace("var HEADER = {};", "var HEADER = " + json.dumps(header) + ";")
                 content = content.replace("var VENUE_LINKS = [];", "var VENUE_LINKS = " + json.dumps(links) + ";")
```
zbialecki, line 100: I would like to update the GroupDirectory webfield to use the parent property so that this VENUE_LINKS var isn't needed

melisabok: Now that all the groups have a parent, can this be implemented in the openreview-web side without having to set a component for every group?

if group.web then render group.web otherwise render the group directory

openreview/venue/webfield/seniorAreaChairsWebfield.js
melisabok: I will add this to another PR

openreview/venue/webfield/paperBidWebfield.js
melisabok: I will add this to another PR

openreview/conference/templates/paperRecruitResponseWebfield.js
```diff
@@ -7,8 +7,8 @@ return {
...
-    acceptMessage: "#### Thank you for accepting this invitation from " + HEADER.title + ".\n &nbsp; \n- Log in to your OpenReview account. If you do not already have an account, you can sign up [here](https://openreview.net/signup).\n\n- Ensure that the email address [TODO] that received this invitation is linked to your profile page and has been confirmed.\n\n- Complete your pending [tasks](https://openreview.net/tasks) (if any) for " + HEADER.subtitle + ".",
+    invitationMessage: "#### You were invited by {{inviter}} to serve as a *" + ROLE_NAME +"* for " + HEADER.title + ".\n ##### Please reply to this invitation clicking the response below:",
```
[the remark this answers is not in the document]
melisabok: I will add this to another PR

openreview/venue/group.py
zbialecki: Yeah I like that idea.

PR #1450, opened 2022-11-03
first version to request profile merges

melisabok:
- Use invitation openreview.net/Support/-/Profile_Merge to request a profile merge. 
- Logged users and guest users can make this request.
- They need to specify the left and right email/tilde id.
- When moderating the requests, the UI should show a link to the profile/compare with the left and right parameters.
- After the moderator merge the profiles, the moderator needs to click on "Accept" to complete the request and notify the user.
- The moderator can also rejects the request if one of the left/right is wrong or there is no evidence the user is the owner of the profiles. 

UPDATE:

- set dblp invitations
- fix author and authors matching the same name in the author name deletion process.

### DIFF
--- a/openreview/profile/management.py
+++ b/openreview/profile/management.py
@@ -15,6 +15,79 @@ class ProfileManagement():
         self.set_remove_name_invitations()
         self.set_archive_invitations()
         self.set_merge_profiles_invitations()
+        self.set_dblp_invitations()
+
+    def set_dblp_invitations(self):
+
+        dblp_group = openreview.tools.get_group(self.client, 'dblp.org')
+        if dblp_group is None:
+            self.client.post_group(
+                openreview.Group(
+                    id = 'dblp.org',
+                    readers = ['everyone'],
+                    writers = ['dblp.org'],
+                    nonreaders = [],
+                    signatures = ['~Super_User1'],
+                    signatories = ['dblp.org'],
+                    members = []
+                )
+            )
+
+        content = {
+            "dblp": {
+                "value-regex": "(.*\\n)+.*"
+            }
+        }
+
+        self.client.post_invitation(openreview.Invitation(
+            id='dblp.org/-/record',
+            readers=['everyone'],
+            writers=['dblp.org'],
+            signatures=['dblp.org'],
+            invitees=['~'],
+            transform=os.path.join(os.path.dirname(__file__), 'process/dblp_transform.js'),
+            reply={
+                'readers': {
+                    'values': ['everyone']
+                },
+                'writers': {
+                    'values':['dblp.org'],
+                },
+                'signatures': {
+                    'values-regex': "dblp.org|~.*"
+                },
+                'content': content
+            }
+        ))
+
+        self.client.post_invitation(openreview.Invitation(
+            id='dblp.org/-/author_coreference',
+            readers=['everyone'],
+            writers=['dblp.org'],
+            signatures=['dblp.org'],
+            invitees=['~'],
+            reply={
+                'referentInvitation': 'dblp.org/-/record',
+                'readers': {
+                    'values': ['everyone']
+                },
+                'writers': {
+                    'values':[],
+                },
+                'signatures': {
+                    'values-regex': "dblp.org|~.*"
+                },
+                "content": {
+                    "authorids": {
+                        "values-regex": ".*"
+                    },
+                    "authors": {
+                        "values-regex": ".*",
+                        "required": False
+                    }
+                }
+            }
+        ))                           
 
     def set_remove_name_invitations(self):
 
@@ -225,11 +298,11 @@ class ProfileManagement():
     def set_merge_profiles_invitations(self):
 
         content = {
-            'name': {
+            'email': {
                 'order': 1,
-                'description': 'Name that want to be removed.',
+                'description': 'email of the user making the request.',
                 'value-regex': '.*',
-                'required': True                
+                'required': False                
             },
             'left': {
                 'order': 2,
--- a/openreview/profile/management.py
+++ b/openreview/profile/management.py
@@ -233,13 +233,13 @@ class ProfileManagement():
             },
             'left': {
                 'order': 2,
-                'description': 'Usernames or email that want to be merged.',
+                'description': 'Username or email that want to be merged.',
                 'value-regex': '.*',
                 'required': True
             },
             'right': {
                 'order': 3,
-                'description': 'Usernames or email that want to be merged.',
+                'description': 'Username or email that want to be merged.',
                 'value-regex': '.*',
                 'required': True
             },
--- a/openreview/profile/management.py
+++ b/openreview/profile/management.py
@@ -307,13 +307,13 @@ class ProfileManagement():
             'left': {
                 'order': 2,
                 'description': 'Username or email that want to be merged.',
-                'value-regex': '^~[^\d\s]+_[^\d\s]+[0-9]+$|([a-z0-9_\-\.]{1,}@[a-z0-9_\-\.]{2,}\.[a-z]{2,},){0,}([a-z0-9_\-\.]{1,}@[a-z0-9_\-\.]{2,}\.[a-z]{2,})',
+                'value-regex': '^~[^\d\s]+[1-9][0-9]*$|([a-z0-9_\-\.]{1,}@[a-z0-9_\-\.]{2,}\.[a-z]{2,},){0,}([a-z0-9_\-\.]{1,}@[a-z0-9_\-\.]{2,}\.[a-z]{2,})',
                 'required': True
             },
             'right': {
                 'order': 3,
                 'description': 'Username or email that want to be merged.',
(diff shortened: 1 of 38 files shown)


openreview/profile/management.py
```diff
@@ -219,4 +220,105 @@ def set_archive_invitations(self):
...
+        content = {
+            'name': {
+                'order': 1,
+                'description': 'Name that want to be removed.',
+                'value-regex': '.*',
+                'required': True                
+            },
+            'left': {
+                'order': 2,
+                'description': 'Usernames or email that want to be merged.',
```
carlosmondra, line 236:
```suggestion
                'description': 'Username or email that want to be merged.',
```

openreview/profile/management.py
```diff
@@ -219,4 +220,105 @@ def set_archive_invitations(self):
...
+                'value-regex': '.*',
+                'required': True
+            },
+            'right': {
+                'order': 3,
+                'description': 'Usernames or email that want to be merged.',
```
carlosmondra, line 242:
```suggestion
                'description': 'Username or email that want to be merged.',
```

openreview/profile/management.py
```diff
@@ -219,4 +220,105 @@ def set_archive_invitations(self):
...
+
+    def set_merge_profiles_invitations(self):
+
+        content = {
+            'name': {
+                'order': 1,
+                'description': 'Name that want to be removed.',
+                'value-regex': '.*',
+                'required': True                
+            },
```
carlosmondra, line 233: Why are we including this field in this invitation?

openreview/profile/process/request_merge_profiles_decision_process.py
```diff
@@ -0,0 +1,34 @@
...
+
+We have received your request to merge the following profiles: {request_note.content['left']}, {request_note.content['right']}.
+
+We can not remove the name from the profile for the following reason:
+
+{request_note.content['support_comment']}
+
+Regards,
+
+The OpenReview Team.
```
carlosmondra, line 19: We need to change this line, right?

> We can not remove the name from the profile for the following reason:

openreview/profile/management.py
melisabok: sometimes guest users are requesting merge of profiles, and we would need to have a name.

openreview/profile/process/request_merge_profiles_decision_process.py
melisabok: fixed

carlosmondra:
> > Where are guest users going to do the request from ?
> 
> Some users from the signup page? i'm not sure who to do here but we receive emails with the merge profile request from guest users. Sometimes they lost the access to the profile?

What if we ask users to first register by either claiming a profile or creating a new one and then request a merge?

xkopenreview:
> > > Where are guest users going to do the request from ?
> > 
> > 
> > Some users from the signup page? i'm not sure who to do here but we receive emails with the merge profile request from guest users. Sometimes they lost the access to the profile?
> 
> What if we ask users to first register by either claiming a profile or creating a new one and then request a merge?

i think there are 3 places where user can request a merge:
1. sign up page there's a link at bottom to send feedback
2. when user add a email which belong to another profile
3. when user import dblp paper associated with another profile

i think we can check the title of the message to know whether it's profile merge or not

carlosmondra:
> i think there are 3 places where user can request a merge:
> 
> 1. sign up page there's a link at bottom to send feedback
> 2. when user add a email which belong to another profile
> 3. when user import dblp paper associated with another profile
> 
> i think we can check the title of the message to know whether it's profile merge or not

Right, I was hoping to remove option 1, so that only registered users can request merges.

melisabok:
Should guest users contact us if they want to merge profiles or recover a profile where they lost access to the email account?

xkopenreview:
> > i think there are 3 places where user can request a merge:
> > 
> > 1. sign up page there's a link at bottom to send feedback
> > 2. when user add a email which belong to another profile
> > 3. when user import dblp paper associated with another profile
> > 
> > i think we can check the title of the message to know whether it's profile merge or not
> 
> Right, I was hoping to remove option 1, so that only registered users can request merges.

i think the user action following or before 1 is to claim a profile/resend activation email/reset password
so it would save a tilde id to allow guest to request profile merge.

what's the rational to only allow logged in user to request merge?

carlosmondra:
> Should guest users contact us if they want to merge profiles or recover a profile where they lost access to the email account?

I think users should register with whatever profile they can claim and then request a merge.
Recovering profiles is a separate matter right? Or what am I missing?

> i think the user action following or before 1 is to claim a profile/resend activation email/reset password
> so it would save a tilde id to allow guest to request profile merge.

I think it would be better for users to claim a profile/resend activation email/reset password and then, after the profile is active, the user can request profile merges.

> what's the rational to only allow logged in user to request merge?

To have profile merge requests in a single place.

I'm not sure where the request to merge profiles is going to be created, but if, for example, we are going to use the profile page for that, then, guest users won't have a profile page to do it and I want to avoid that.

xkopenreview:
currently 1 and 2 are using the feedback modal 
i thought we can just change it to use a new profile merge request modal
but just realized that 3 just instruct user to contact us and it's already a modal and it's not good to have modal open new modal so maybe 3 will be in profile edit page

openreview/profile/management.py
xkopenreview: maybe this should be email instead of name
so that the user can be contacted if we need clarification

tests/test_profile_management.py
```diff
@@ -1006,4 +1006,102 @@ def test_rename_publications_from_api2(self, client, profile_management, test_cl
...
+
+We will evaluate your request and you will receive another email with the request status.
+
+Thanks,
+
+The OpenReview Team.
+'''
+
+        ## Accept the request
+        decision_note = client.post_note(openreview.Note(
```
xkopenreview, line 1061: i think instead of just accept or reject, the direction of merge also need to be specified

carlosmondra: Maybe the direction should be for us to decide

xkopenreview: discussed this in the daily meeting.
the actual profile merge will be done by the moderator in profile compare page(same as how it's done currently)

the user moderation page will have a link to profile compare page and after the profile is merged the moderator will go back to moderation page to click approve to indicate the profile merge has been done.(or don't merge and reject)

openreview/profile/management.py
xkopenreview: we discussed about combining signature and email column
so this field need to be renamed to email and required set to false

openreview/profile/management.py
```diff
@@ -219,4 +293,105 @@ def set_archive_invitations(self):
...
-        ))                       
+        ))
+
+    def set_merge_profiles_invitations(self):
+
+        content = {
+            'email': {
+                'order': 1,
+                'description': 'email of the user making the request.',
+                'value-regex': '.*',
```
xkopenreview, line 304: maybe this should be a regex for email

openreview/profile/management.py
```diff
@@ -219,4 +293,105 @@ def set_archive_invitations(self):
...
+            'right': {
+                'order': 3,
+                'description': 'Username or email that want to be merged.',
+                'value-regex': '.*',
+                'required': True
+            },
+            'comment': {
+                'order': 4,
+                'value-regex': '[\\S\\s]{1,5000}',
+                'description': 'Reason why you want to delete your name.',
```
celestemartinez, line 322: Should we change this to: `Reason why you want to merge your profiles.` (though I'm sure everyone wants to merge because they have multiple profiles). 
Maybe just `Any additional information about the merge.`

melisabok: good idea, @xkopenreview is the UI using this description?

xkopenreview: no currently the label of input is fixed.

openreview/profile/management.py
```diff
@@ -219,4 +293,105 @@ def set_archive_invitations(self):
...
+            'email': {
+                'order': 1,
+                'description': 'email of the user making the request.',
+                'value-regex': '([a-z0-9_\-\.]{1,}@[a-z0-9_\-\.]{2,}\.[a-z]{2,},){0,}([a-z0-9_\-\.]{1,}@[a-z0-9_\-\.]{2,}\.[a-z]{2,})',
+                'required': False                
+            },
+            'left': {
+                'order': 2,
+                'description': 'Username or email that want to be merged.',
+                'value-regex': '^~[^\d\s]+_[^\d\s]+[0-9]+$|([a-z0-9_\-\.]{1,}@[a-z0-9_\-\.]{2,}\.[a-z]{2,},){0,}([a-z0-9_\-\.]{1,}@[a-z0-9_\-\.]{2,}\.[a-z]{2,})',
```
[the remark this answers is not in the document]
melisabok: fixed!

xkopenreview (approved):

carlosmondra (approved):

celestemartinez (approved):

PR #1451, opened 2022-11-07
enable author profile status

melisabok:


zbialecki:
I don't understand why this template doesn't just use the vars and instead does ``${VAR_NAME}`` for every property.

melisabok:
when the venue request form is modified, it is always being synchronized with the properties of the templates. That means we always set the webfield  as template + venue properties. If the webfield is changed using the group editor, the next syncronization will override this change unless the line `// webfield_template` is removed. But this line is not supported in the components because they must start with `// Webfield component`. I'm not sure how to fix this in the API 1 without having to set a property in the request from to override in the template.

celestemartinez (approved):

PR #1458, opened 2022-11-15
increase package version

celestemartinez:


carlosmondra:
Does the fail test indicate that openreview-py was not deployed to pip?

celestemartinez:
well that happens after the tests pass in master, right? those are still running

melisabok:
it is live! https://pypi.org/project/openreview-py/1.13.0/

PR #1462, opened 2022-11-15
Fix/PC Console - add option to message reviewers/ACs with no assignments

xkopenreview:
this pr should add an option to message reviewers/ACs with no assignments in menu bar of Reviewer Status and Area Chair Status tab in PC Console 

similar function has been added in component based pc console pr


melisabok:
@haroldrubio this should work for ARR PC console if they want to remind them again.

haroldrubio (approved):
I tested it in my local with ARR October and it seems to work correctly

PR #1464, opened 2022-11-17
Feature/parallel tests

carlosmondra:


melisabok (approved):
let's keep those tests commented, the ones related to anonreviewer must be deleted we don't support that anymore. Thanks Carlos!

PR #1468, opened 2022-11-18
Feature/remove legacy code

melisabok:
- Remove use legacy anonids
- Remove use legacy invitation ids
- Remove all the deprecated methods
- Remove legacy webfields
- Remove tools.add_assignments (only works with legacy anonids)
- Remove other tools methods that are not being used
- Remove all the queries where '.*' is in the middle of the regex. With the exception of the GET /groups?regex=&signatory= that is being used in the note editor
- Fix tests.


zbialecki (approved):
Nice! I don't see any problem with removing these old webfields.

xkopenreview (approved):

PR #1479, opened 2022-12-06
Fix: add submission email to authors and PCs

celestemartinez:


openreview/venue/group.py
```diff
@@ -144,6 +144,8 @@ def create_venue_group(self):
             'public_submissions': { 'value': self.venue.submission_stage.public },
             'public_withdrawn_submissions': { 'value': self.venue.submission_stage.withdrawn_submission_public },
             'public_desk_rejected_submissions': { 'value': self.venue.submission_stage.desk_rejected_submission_public },
+            'submission_email_template': { 'value': '' },
+            'email_pcs_on_submission': { 'value': self.venue.submission_stage.email_pcs },
```
melisabok, line 148: rename the field name to `submission_email_pcs` so it follows the same format as `review_emails_pcs`, `comment_emails_pcs`, etc

openreview/venue/group.py
```diff
@@ -172,6 +174,16 @@ def create_venue_group(self):
             'desk_rejection_name': { 'value': 'Desk_Rejection'}
         }
 
+        if self.venue.submission_stage.submission_email:
+            email_template = self.venue.submission_stage.submission_email
+            email_template = email_template.replace('{{Abbreviated_Venue_Name}}', self.venue.get_short_name())
```
melisabok, line 179: what about doing this replacement in the process function directly so we replace the values only once at the moment of sending the email.

melisabok (approved):

PR #1481, opened 2022-12-06
Feature/zip multiple files

xkopenreview:
this pr should add a button in AC console which will download all note pdf field as single zip file


openreview/conference/templates/areachairWebfield.js
```diff
@@ -694,11 +694,12 @@ var renderStatusTable = function(conferenceStatusData, container) {
       '</ul>' +
     '</div>' +
     '<div class="btn-group"><button class="btn btn-export-data" type="button">Export</button></div>' +
+    '<div class="btn-group"><button class="btn btn-export-pdf" type="button">Download PDF</button></div>' +
```
melisabok, line 697: Should this say "Download PDFs" or "Download all PDFs"?

xkopenreview: the label is shorten so that it can fit in the row without taking too much space from search and sort

melisabok (approved):

celestemartinez (approved):

PR #1482, opened 2022-12-07
Add PC Console Changes from ARR

haroldrubio:
This PR contains some changes I've made to the ARR PC consoles that may be useful to other conferences in the CSV exports


openreview/conference/templates/programchairWebfield.js
```diff
@@ -518,12 +518,14 @@ var getUserProfiles = function(userIds, reviewerBidCounts, areaChairBidCounts, a
         profileMap[profile.id] = {
           id: profile.id,
           name: name,
+          gender: profile.content.gender,
```
melisabok, line 521: I'm not sure about exporting this data in all the PC consoles. I know this data is public but we don't show in the public profile so the owner of the profile doesn't know.

melisabok (approved):

xkopenreview:
i think the change for ac meta-review denominator issue in overview tab is missing?

xkopenreview:
discussed this with @haroldrubio. Harold will make a pr in web to port the changes in this pr over to pc console component.

melisabok:
@xkopenreview can this be merged?

xkopenreview (approved):

PR #1483, opened 2022-12-07
Fix/venue feedback

melisabok:
- Move bid instructions to the webfield and use markdown
- Fix indentation in PC and SAC console files.
- Remove landing page webfield.
- Make api_version field available through the UI.


openreview/venue_request/venue_request.py
```diff
@@ -1148,10 +1148,10 @@ def setup_request_form(self):
                 'hidden': True # Change this value on exception request from the PCs.
             },
             'api_version': {
+                'description': 'Which API version would you like to use? API 2 is still in a experimental face, contact us if you need more info.',
```
celestemartinez, line 1151:
```suggestion
                'description': 'Which API version would you like to use? API 2 is still in the experimental phase, contact us if you would like more information.',
```

zbialecki (approved):
Looks good to me. Only small thing that I would change is that the field names in the config object like title and instructions don't need quotes.

celestemartinez (approved):

PR #1486, opened 2022-12-08
Fix/journal features

melisabok:
- Parametrize the journal: allow certifications, allow show conflicts, allow publication chairs
- fix issue when the ACK invitation is created with an email address instead of tilde id
- implement get_all_messages


openreview/journal/invitation.py
```diff
@@ -4817,6 +4811,13 @@ def set_camera_ready_verification_invitation(self):
             'dateprocesses': [self.ae_reminder_process]
         }
 
+        if self.journal.has_publication_chairs():
```
celestemartinez, line 4814: Does this mean if the journal has publication chairs, then they submit the camera ready verification instead of AEs?

melisabok: don't merge it, I will fix this.

melisabok: fixed

celestemartinez (approved):

PR #1497, opened 2022-12-22
enable PC console component

melisabok:


melisabok:
Closing this because PC component supports API2 only

PR #1498, opened 2022-12-23
remove homepage from create_profile

racheljsmart:
Homepage is preferably empty when creating a profile.


melisabok:
it seems the url is mandatory, there is a test failing:

```MissingFieldsError: At least one link is required```

melisabok:
I would suggest to parametrize the url in the create_profile function and add this url to all the profiles "https://github.com/google/BIG-bench"

PR #1501, opened 2023-01-02
Fix/ Webfield component template formatting

zbialecki:
Make formatting webfield component templates consistent


melisabok (approved):
Thanks!

PR #1506, opened 2023-01-06
use authorids as readers before the submission deadline

melisabok:
Authors can see and edit their submission without depending on the process function to be executed with no errors.

### DIFF
--- a/openreview/venue/venue.py
+++ b/openreview/venue/venue.py
@@ -384,17 +384,36 @@ class Venue(object):
     def setup_post_submission_stage(self, force=False, hide_fields=[]):
         venue_id = self.venue_id
         submissions = self.get_submissions()
+        final_hide_fields = ['authors', 'authorids'] + hide_fields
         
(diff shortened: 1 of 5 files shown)


openreview/venue/venue.py
```diff
@@ -384,17 +384,36 @@ def create_registration_stages(self):
     def setup_post_submission_stage(self, force=False, hide_fields=[]):
         venue_id = self.venue_id
         submissions = self.get_submissions()
+        final_hide_fields = ['authors', 'authorids'] + hide_fields
```
celestemartinez, line 387: Are we always hiding authors and authorids?

melisabok: good catch I forgot to check if the venue is double blind.

melisabok: fixed.

carlosmondra (approved):

carlosmondra (approved):

PR #1514, opened 2023-01-12
propagate change to all the children

melisabok:
Each super invitation will run a date process function:
1. cdate
2. mdate + 10 seconds

in both cases it will try to create or re-create the child invitations. 

Every time the invitation is updated, it will wait until the date process function (mdate + 10 seconds) has finished.

If a user decides to use the invitation editor to make a change, this change will be propagated to all the children. 

This applies to all the stages (review, comment, metareview, decision, revision, etc) 

cdate, duedate and expdate are synchronized between the upper invitation and the inner invitation. NOT SUPPORTED: set null and propagate it to the children, they must be numbers.

In order to tell the process function which notes to use to create the child invitations we store a few parameters in the upper invitation content:

- public_notes_only: create child invitation on notes where readers is everyone (being used by the public comment)
- accepted_notes_only: create child invitation on notes where there is a decision with the "Accept" string (being used by the camera ready revision)

It is pending to implement more settings, like, "review_notes_only" or "metareview_notes_only", etc that can be used for rebuttal invitations o review rating invitations.


openreview/venue/invitation.py
```diff
@@ -328,10 +342,13 @@ def set_review_invitation(self):
...
-            expdate = review_expdate,
+            expdate = review_expdate,            
             date_processes=[{ 
                 'dates': ["#{4/cdate}"],
-                'script': self.cdate_invitation_process              
+                'script': self.invitation_edit_process              
+            }, { 
+                'dates': [self.update_date_string],
+                'script': self.invitation_edit_process             
             }],
```
carlosmondra, line 352: Not a necessary change, but you can achieve the same like this:

```python
date_processes=[{
    'dates': ["#{4/cdate}", self.update_date_string],
    'script': self.invitation_edit_process
}]
```

melisabok: ohhhhh, you are right. Let me change it.

melisabok:
ok, this is ready to merge!!!

celestemartinez (approved):

PR #1520, opened 2023-01-23
feature: venue setup invite assignment

celestemartinez:


openreview/venue/process/invite_assignment_post_process.py
```diff
@@ -0,0 +1,98 @@
+def process(client, edge, invitation):
+
+    SHORT_PHRASE = ''
```
melisabok, line 3: we should try reading the variables from the domain object content.

openreview/venue/process/simple_paper_recruitment_process.py
```diff
@@ -0,0 +1,150 @@
+def process(client, note, invitation):
+    from Crypto.Hash import HMAC, SHA256
+    import urllib.parse
+    from datetime import datetime
+    VENUE_ID = ''
```
melisabok, line 5: do we need this file? can we read the variable from the domain group?

celestemartinez: yes, we can, I haven't gotten around to updating this yet

openreview/venue/webfield/areachairsWebfield.js
```diff
@@ -2,7 +2,28 @@
 const reviewerAssignmentTitle = null
 const areaChairsId = domain.content.area_chairs_id.value
 const reviewerGroup = domain.content.reviewers_id.value
-// Add bid invitation if enabled
```
melisabok, line 5: @xkopenreview please review

melisabok:
@xkopenreview @zbialecki I wonder if there is a nicer way to show this, this is a paper recruitment invitation, we don't have the info of the paper, only the paper id. 

Screen Shot 2023-02-08 at 9 08 37 PM

zbialecki:
@melisabok I'm not sure what you're suggesting we show here, are the submission notes even visible to non-authenticated users?

I do think we could remove the "Please see venue website" text

melisabok (approved):

PR #1523, opened 2023-01-23
sync pcs

melisabok:


openreview/venue/group.py
```diff
@@ -247,7 +247,9 @@ def create_program_chairs_group(self, program_chair_ids=[]):
                 self.post_group(pc_group)
 
             ## Add pcs to have all the permissions
-            self.client.add_members_to_group(venue_id, pc_group_id)        
+            self.client.add_members_to_group(venue_id, pc_group_id)
+        elif pc_group.members != program_chair_ids:
+            self.client.add_members_to_group(pc_group_id, list(set(program_chair_ids) - set(pc_group.members)))
```
celestemartinez, line 252: What if a PC is removed from the list of PCs in the request form?

melisabok: I was too lazy to do it but you are right, let me implement it :)

melisabok: done

celestemartinez (approved):

PR #1525, opened 2023-01-25
Fix/icml pc revision

melisabok:
- Create a "PC Revision" button so the PCs can edit the submission after the submission deadline
- Use [short name] for all the subjects

### DIFF
--- a/openreview/venue/venue.py
+++ b/openreview/venue/venue.py
@@ -418,6 +418,9 @@ class Venue(object):
                     )
         ## Release the submissions to specified readers if venueid is still submission
         openreview.tools.concurrent_requests(update_submission_readers, submissions, desc='update_submission_readers')
+
+        ## Open PC Revision
+        self.invitation_builder.set_pc_submission_revision_invitation()
              
         ## Create revision invitation if there is a second deadline?
         ## Create withdraw and desk reject invitations
--- a/openreview/venue/venue.py
+++ b/openreview/venue/venue.py
@@ -102,6 +102,9 @@ class Venue(object):
     def get_submission_id(self):
         return self.submission_stage.get_submission_id(self)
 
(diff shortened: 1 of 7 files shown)


openreview/venue/venue.py
```diff
@@ -418,6 +421,9 @@ def update_submission_readers(submission):
                     )
         ## Release the submissions to specified readers if venueid is still submission
         openreview.tools.concurrent_requests(update_submission_readers, submissions, desc='update_submission_readers')
+
```
celestemartinez, line 424: can we add adding this venue to `active_venues` somewhere in the `post_submission_stage()`?

celestemartinez:
I noticed that when the PCs are adding a revision after the deadline, the checkbox is not already checked, even though it is in the original submission:
Screen Shot 2023-01-25 at 12 23 31 PM

Is this a bug?

openreview/venue/venue.py
melisabok: good point, I will add it

melisabok:
>I noticed that when the PCs are adding a revision after the deadline, the checkbox is not already checked, even though it is in the original submission:

I noticed the same, I think it is an issue in the UI. Do you mind to take a look?

openreview/venue/venue.py
melisabok: Done.

celestemartinez:
> > I noticed that when the PCs are adding a revision after the deadline, the checkbox is not already checked, even though it is in the original submission:
> 
> I noticed the same, I think it is an issue in the UI. Do you mind to take a look?

I checked this and it is not an issue in the UI, the checkbox text was edited in the invitation and so it doesn't match with what is in the note already.

celestemartinez (approved):

Issue #1368, opened 2022-08-05
Submission Revision stage: Add pdate when the revision is a Camera Ready Revision


melisabok:
I suggest to add a checkbox to indicate if the revision stage is a camera ready stage. 

When this checkbox is selected the camera ready process function will update the bibtex and the pdate of the submission. 

Thoughts @nlbahy , @celestemartinez ?

celestemartinez:
This sound good. Should the pdate be updated everytime a new camera ready revision is posted or only when the first revision is posted?

PR #1532, opened 2023-01-27
add pdate and odate to accepted papers

melisabok:
Set odate and pdate for submissions hosted in API 1 and API2. 

UPDATE: add small fix to the journal get late invitees.

Fixes: https://github.com/openreview/openreview-py/issues/1368

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -25,6 +25,26 @@ Client
 .. autoclass:: Profile
    :members:
 
+API 2 Client
+-------
+
+.. automodule:: openreview.api
+
+.. autoclass:: OpenReviewClient
+   :members:
+
+.. autoclass:: Group
+   :members:
+
+.. autoclass:: Invitation
+   :members:
+
+.. autoclass:: Note
+   :members:
+
+.. autoclass:: Edit
+   :members:
+
 
 Tools
 -------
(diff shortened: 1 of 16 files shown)


openreview/conference/builder.py
```diff
@@ -953,6 +954,7 @@ def create_blind_submissions(self, hide_fields=[], number=None):
 
                 revision_note = self.client.post_note(openreview.Note(
                     invitation = f'{self.support_user}/-/{self.venue_revision_name}',
+                    odate = online_date if note.odate is None else None,
```
carlosmondra, line 957: Shouldn't this be
```suggestion
                    odate = online_date if note.odate is not None else None,
```

carlosmondra: or
```suggestion
                    odate = online_date if note.odate is None else note.odate,
```

melisabok: I don't want to override the odate if already exists. So, if I pass a null in the reference, the value will be overwritten to null?  (I should write a test)

carlosmondra: Oh I think that's why I was checking for undefined and not undefined and null. So that you could remove the pdate/odate if you passed null to the original note

melisabok: ok, so I will accept your suggestion and pass the odate.

docs/source/api.rst
```diff
@@ -25,6 +25,26 @@ Client
 .. autoclass:: Profile
    :members:
 
+API 2 Client
```
melisabok, line 28: @carlosmondra could you review this?

carlosmondra: Do you mean the entire api.rst?

melisabok: the new section for the API 2 client. I realized they are not being shown in readthedocs. I guess this is correct. I want you to double check.

celestemartinez (approved):

PR #1538, opened 2023-01-31
Feature/chat example

melisabok:
- Define two tabs in the ICML submission invitation.
- Create a chat invitation
- Post chat messages:
Screen Shot 2023-01-31 at 1 46 37 PM

User paper number 1 to test this.


tests/test_icml_conference.py
```diff
@@ -1309,3 +1328,114 @@ def test_decision_stage(self, openreview_client, helpers):
...
+            invitations='ICML.cc/2023/Conference/-/Edit',
+            readers = ['ICML.cc/2023/Conference'],
+            writers = ['ICML.cc/2023/Conference'],
+            signatures = ['ICML.cc/2023/Conference'],
+            invitation = openreview.api.Invitation(
+                id = 'ICML.cc/2023/Conference/Submission1/-/Chat',
+                readers = ['everyone'],
+                writers = ['ICML.cc/2023/Conference'],
+                signatures = ['ICML.cc/2023/Conference'],
+                invitees = ['ICML.cc/2023/Conference/Program_Chairs', 'ICML.cc/2023/Conference/Submission1/Reviewers'],
```
zbialecki, line 1375: Can we add Area Chairs to invitees as well?

tests/test_icml_conference.py
```diff
@@ -1309,3 +1328,114 @@ def test_decision_stage(self, openreview_client, helpers):
...
+                                'ICML.cc/2023/Conference/Submission1/Reviewer_.*'
+                            ]
+                        }
+                    },
+                    'note': {
+                        'readers': ['ICML.cc/2023/Conference/Program_Chairs', 'ICML.cc/2023/Conference/Submission1/Reviewers'],
+                        'writers': ['ICML.cc/2023/Conference'],
+                        'signatures': ['${3/signatures}'],
+                        'forum': submission.id,
+                        "replyto": {
```
zbialecki, line 1392: Does this mean that the replyto field of the note will always be set to the forum id? I was thinking that users would still be able to reply to chat messages, even though they wouldn't appear nested like in the regular view.

melisabok: replyto is a param and it means that any value (note id) must have a forum that matches with the submisison.id.

This is exactly how it works with comment invitations.

zbialecki: I see

tests/test_icml_conference.py
```diff
@@ -1343,6 +1343,15 @@ def test_forum_chat(self, openreview_client, helpers):
                         "id": "all",
                         "label": "All"
                     },
+                    {
+                        "id": "discussion",
+                        "label": "Discussion",
+                        "filter": "-invitation:ICML.cc/2023/Conference/Submission${note.number}/-/Chat",
```
zbialecki, line 1349: I wanted to add this because I think it will be a common setup - to have 1 tab for the chat messages and one standard layout tab with other types of messages.

tests/test_icml_conference.py
melisabok: done.

tests/test_icml_conference.py
```diff
@@ -1433,3 +1452,124 @@ def test_decision_stage(self, openreview_client, helpers):
...
+                id = 'ICML.cc/2023/Conference/-/Submission',
+                reply_forum_views = [
+                    {
+                        'id': 'all',
+                        'label': 'All'
+                    },
+                    {
+                        'id': 'discussion',
+                        'label': 'Discussion',
+                        'filter': '-invitations:ICML.cc/2023/Conference/Submission${note.number}/-/Chat',
```
xkopenreview, line 1473: what does the "-" mean here?

zbialecki: This means that the view should hide notes that are replies of the listed invitations.

tests/test_icml_conference.py
```diff
@@ -2413,3 +2414,123 @@ def test_decision_stage(self, openreview_client, helpers):
...
+                        'param': {
+                            'enum': [
+                                'ICML.cc/2023/Conference/Program_Chairs',
+                                'ICML.cc/2023/Conference/Submission1/Area_Chair_.*',
+                                'ICML.cc/2023/Conference/Submission1/Reviewer_.*',
+                            ]
+                        }
+                    },
+                    'note': {
+                        'readers': ['ICML.cc/2023/Conference/Program_Chairs', 'ICML.cc/2023/Conference/Submission1/Area_Chairs', 'ICML.cc/2023/Conference/Submission1/Reviewers'],
```
melisabok, line 2484: @zbialecki you need to declare the field ddate here:

```
                    'ddate': {
                        'param': {
                            'range': [ 0, 9999999999999 ],
                            'optional': True,
                            'deletable': True
                        }
                    },

```

melisabok: to make it editable:

```
                    'id': {
                        'param': {
                            'withInvitation': 'ICML.cc/2023/Conference/Submission1/-/Chat',
                            'optional': True
                        }
                    },
```

zbialecki:
@melisabok is this what you mean? https://github.com/openreview/openreview-py/pull/1538/commits/757173daf0a97930117ec8ad546734bfb43ce4eb

melisabok:
@zbialecki should we merge this?

zbialecki:
Yes, I think it's ready to be merged. I don't think we'll need to make any other changes to invitations or other config.

zbialecki (approved):

zbialecki:
It would be helpful if we could merge this PR soon. It has a chance necessary for the enable chat script.

Issue #1537, opened 2023-01-31
Venue class: recruitment notes must be visible to the invited user


melisabok:
Otherwise the user won't see the reduce quota in the reviewers console.

PR #1541, opened 2023-02-01
Fix/api 2 assignment deployment

### LINKED ISSUE #1540: Venue class: make all the submission edits visible to the authors
Otherwise they see the modification date but they don't see what happened.

melisabok:
- Transfer SAC conflicts to assigned ACs when setting up the matching for ACs.
- Fix maxReplies on proposed assignments from TMLR to avoid too many error.
- Test set_assignments for SACs, ACs and reviewers.
- Make submission edits visible to authors. Fix #1540 
- Add invited user to be reader of recruitment notes. Fix https://github.com/openreview/openreview-py/issues/1537


openreview/venue/matching.py
```diff
@@ -275,6 +284,18 @@ def _build_note_conflicts(self, submissions, user_profiles, get_profile_info):
                 conflicts.update(author_emails.intersection(user_info['emails']))
                 conflicts.update(author_publications.intersection(user_info['publications']))
 
+                ## Transfer SAC conflicts
+                if len(conflicts) == 0 and self.is_area_chair:
```
celestemartinez, line 288: if self.is_area_chair, why would len(conflicts)==0?

melisabok: because if there are already conflicts between the AC and the paper we don't need to compute the SAC conflicts.

celestemartinez (approved):

PR #1547, opened 2023-02-04
Feat/ builder: create metareview revision stage

celestemartinez:


celestemartinez:
CVPR PCs haven't decided if they will be enabling revisions for metareviews orif they'll just use the same metareview invitation.

melisabok (approved):

PR #1557, opened 2023-02-09
Fix/ builder: restore assignment edges when submission is restored

celestemartinez:
- Restores assignment edges or desk-reject/withdrawal reversion
- Not sure how we can revert the `/Proposed_Assignment` edges. For `/Assignment` edges, I am checking the paper groups to make sure I am not restoring old assignment edges.


melisabok:
I like the solution, I'm not sure what to do with proposed assignments.

melisabok (approved):

PR #1564, opened 2023-02-16
Feature/AC console - show multiple SACs

xkopenreview:
there's a section at top of AC console to show the assigned SAC
currently only the first SAC will be shown.

this pr should show multiple SACs if they are assigned to the same AC


xkopenreview:
component based AC console will require changes similar to this PR too

melisabok (approved):

PR #1570, opened 2023-02-21
Fix/ venue recruitment: convert emails to lowercase

### LINKED ISSUE #1567: Venue recruitment: "Invalid group id" when the invitee is an email address with uppercase
This issue happened with NeurIPS 2023 SAC recruitment, the PCs sent the invitation to an email address with uppercase, the user was invited and added to the group correctly but the url contained the email in uppercase. 

I think the bug is when creating the url and the hash key we should send the email all in lowercase.

celestemartinez:
- Fixes #1567
- The lowercase email was being correctly added to the invited group, but I was getting an error with sending the email. Now, I am making all emails lowercase before starting the recruitment.

### DIFF
--- a/openreview/venue/recruitment.py
+++ b/openreview/venue/recruitment.py
@@ -55,6 +55,8 @@ class Recruitment(object):
         invitation_id = invitation.id
         hash_seed = invitation.content['hash_seed']['value']
 
+        invitees = [invitee.lower() if not invitee.startswith('~') else invitee for invitee in invitees]
+
         if remind:
             committee_invited_group = self.client.get_group(committee_invited_id)
             invited_committee = committee_invited_group.members
--- a/tests/test_venue_request_v2.py
+++ b/tests/test_venue_request_v2.py
@@ -485,7 +485,7 @@ class TestVenueRequest():
         # assert reply_row
         # buttons = reply_row.find_elements_by_class_name('btn-xs')
         # assert [btn for btn in buttons if btn.text == 'Recruitment']
-        reviewer_details = '''reviewer_candidate1_v2@mail.com, Reviewer One\nreviewer_candidate2_v2@mail.com, Reviewer Two'''
+        reviewer_details = '''reviewer_candidate1_v2@mail.com, Reviewer One\nReviewer_Candidate2_v2@mail.com, Reviewer Two'''
         recruitment_note = test_client.post_note(openreview.Note(
             content={
                 'title': 'Recruitment',


openreview/venue/recruitment.py
```diff
@@ -55,6 +55,8 @@ def invite_committee(self,
         invitation_id = invitation.id
         hash_seed = invitation.content['hash_seed']['value']
 
+        invitees = [invitee.lower() if not invitee.startswith('~') else invitee for invitee in invitees]
```
melisabok, line 58: Can you check by "@" instead for tilde id, like the API 1:

```
invitees = [e.lower() if '@' in e else e for e in invitees if len(e) > 0]
```

tests/test_venue_request_v2.py
```diff
@@ -485,7 +485,7 @@ def test_venue_recruitment(self, client, test_client, selenium, request_page, ve
         # assert reply_row
         # buttons = reply_row.find_elements_by_class_name('btn-xs')
         # assert [btn for btn in buttons if btn.text == 'Recruitment']
-        reviewer_details = '''reviewer_candidate1_v2@mail.com, Reviewer One\nreviewer_candidate2_v2@mail.com, Reviewer Two'''
+        reviewer_details = '''reviewer_candidate1_v2@mail.com, Reviewer One\nReviewer_Candidate2_v2@mail.com, Reviewer Two'''
```
melisabok, line 488: I don't understand why the email was added to the invited group with no error but the invitation link still had the email in capital letter, any ideas?

celestemartinez: It was because we were not changing the email to lowercase at any point, but for some reason `add_members_to_group()` add the email in lowercase. Not really sure why this is the case.

openreview/venue/recruitment.py
celestemartinez: oh I hadn't noticed we were doing something similar for API1! Changed.

tests/test_venue_request_v2.py
celestemartinez: if I do: 
`openreview_client.add_members_to_group('NeurIPS.cc/2023/Conference/Senior_Area_Chairs/Invited', ['Celeste.Martinez@gmail.com'])`, the email `celeste.martinez@gmail.com` is added to the group, but everywhere else in the recruitment function we were still using `Celeste.Martinez@gmail.com`

melisabok: @carlosmondra is the backend converting emails to lowercase when they are added as members of a group? 

this is the code for `add_members_to_group`: https://github.com/openreview/openreview-py/blob/master/openreview/api/client.py#L1470

melisabok (approved):

PR #1577, opened 2023-02-23
set author console component only once

melisabok:


melisabok:
@xkopenreview we are deploying this too so CVPR author console is not overwritten

celestemartinez (approved):

PR #1581, opened 2023-02-27
Fix/ invite assignment: use enum for invite assignment label

### LINKED ISSUE #1569: Venue paper recruitment: "User can not reply to this invitation, invalid status"
It seems the process function is contemplating all the possible statuses of the invite assignment edge:

```
if edge.label not in [invited_label, accepted_label, declined_label, 'Pending Sign Up']:
```

Declined label now saves the reason of the decline so it is not going to match, we should check the startswith. I would like to try to reproduce this error and then try to fix it. 

Here are some errors in the live site:

https://api2.openreview.net/logs/process?id=DCOuuVfzLa

https://api2.openreview.net/logs/process?id=xCDyX2AovA

To fix this in the live site, I ended up removing the validation so I'm not sure if we should keep it or not.

celestemartinez:
- Fixes #1569
- Uses https://github.com/openreview/openreview-api/pull/174

### DIFF
--- a/tests/test_icml_conference.py
+++ b/tests/test_icml_conference.py
@@ -1677,16 +1677,16 @@ ICML 2023 Conference Program Chairs'''
         client.rename_edges(new_id='~Rachel_ICML2', current_id='~Rachel_ICML1')
         client.merge_profiles(profileTo='~Rachel_ICML2', profileFrom='~Rachel_ICML1')
 
-        helpers.respond_invitation(selenium, request_page, invitation_url, accept=False)
+        helpers.respond_invitation(selenium, request_page, invitation_url, accept=False, comment='I am too busy.')
 
         helpers.await_queue(openreview_client)
 
         messages = openreview_client.get_messages(to='rachel_bis@icml.cc', subject='[ICML 2023] Reviewer Invitation declined for paper 1')
-        assert len(messages) == 1
+        assert len(messages) == 2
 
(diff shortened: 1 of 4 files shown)


tests/test_icml_conference.py
```diff
@@ -1684,6 +1684,10 @@ def test_reviewer_reassingment(self, client, openreview_client, helpers, seleniu
         messages = openreview_client.get_messages(to='rachel_bis@icml.cc', subject='[ICML 2023] Reviewer Invitation declined for paper 1')
         assert len(messages) == 1
 
+        invite_edges=openreview_client.get_edges(invitation='ICML.cc/2023/Conference/Reviewers/-/Invite_Assignment', head=submissions[0].id, tail='~Rachel_ICML2')
+        assert len(invite_edges) == 1
+        assert invite_edges[0].label == 'Declined'
```
celestemartinez, line 1689: I tested adding a comment in my local, but not sure how to add this to test

melisabok: I think with Selenium you can select the textarea and enter a comment. You can add a new parameter in `def respond_invitation(selenium, request_page, url, accept, quota=None):
` to pass a comment when accept is False. You can check in the implementation how we select a quota from the dropdown, you can do something similar with the textarea.

melisabok (approved):

PR #1590, opened 2023-03-02
Fix/AE console - modify reviewers assignment link is wrong

xkopenreview:
fix modify reviewers assignment link so that edge browser does not break


xkopenreview:
TMLR AE console in live has been fixed

celestemartinez (approved):

melisabok (approved):
:O thanks for the catch!

PR #1596, opened 2023-03-06
Fix/ Venue: assignment process function

### LINKED ISSUE #1595: Venue: AC assignment validation should check meta reviews are not posted
It seems the AC assignment pre-process function is checking the anon reviewers group and review invitations. 

We should parametrize it and use area chair anon groups and meta review invitations. 

Also we should check if review_name or metareview_name exists before checking if there is already a reply.

I fixed the code in the live site:

https://openreview.net/invitation/edit?id=ICML.cc/2023/Conference/Area_Chairs/-/Assignment

How to reproduce it:

- assign reviewer to a paper
- assign AC to a paper
- reviewer submit the review
- unassign AC from a paper

Error got: `“error: Can not remove assignment, the user ~Emma_Brunskill2 already posted a review.”`

The AC can see all the reviewer anon groups so the query got a group and that reviewer already posted a review

celestemartinez:
- Fixes #1595

### DIFF
--- a/openreview/venue/process/assignment_pre_process.py
+++ b/openreview/venue/process/assignment_pre_process.py
@@ -3,9 +3,11 @@ def process_update(client, edge, invitation, existing_edge):
     domain = client.get_group(edge.domain)
     venue_id = domain.id
     submission_name = domain.content['submission_name']['value']
-    is_reviewer = True
-    review_name = domain.content.get('review_name', {}).get('value') if is_reviewer else domain.content.get('meta_review_name', {}).get('value')
-    reviewers_anon_name = domain.content['reviewers_anon_name']['value'] if is_reviewer else domain.content['area_chairs_anon_name']['value']
+
+    REVIEW_NAME_STRING = 'review_name'
+    REVIEWERS_ANON_NAME_STRING = 'reviewers_anon_name'
+    review_name = domain.content.get(REVIEW_NAME_STRING, {}).get('value')
+    reviewers_anon_name = domain.content[REVIEWERS_ANON_NAME_STRING]['value']
     
     if edge.ddate:
         paper=client.get_note(edge.head)
@@ -28,5 +30,5 @@ def process_update(client, edge, invitation, existing_edge):
 
         for review in reviews:
             if review.signatures[0] == groups[0].id:
-                raise openreview.OpenReviewException(f'Can not remove assignment, the user {edge.tail} already posted a review.')
+                raise openreview.OpenReviewException(f'Can not remove assignment, the user {edge.tail} already posted a {review_name}')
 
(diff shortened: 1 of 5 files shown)


melisabok:
I would like you to try to reproduce the issue that Sherry had this morning and then fix it. 

I described the steps in the issue.

openreview/venue/process/assignment_pre_process.py
```diff
@@ -3,14 +3,18 @@ def process_update(client, edge, invitation, existing_edge):
     domain = client.get_group(edge.domain)
     venue_id = domain.id
     submission_name = domain.content['submission_name']['value']
-    review_name = domain.content['review_name']['value']
```
melisabok, line 6: I would prefer to replace the name of the variable in the file instead of introducing a new variable `is_reviewer`

melisabok:
There are three scenarios to test:

1. unassign reviewer when the review stage hasn't configured yet. Because the variable "review_name" is set only when the review_stage is configured. You should get a Key error. 
2. unassign AC when the reviewer of the same paper posted a review. You should get "error: Can not remove assignment, the user ~Emma_Brunskill2 already posted a review."
3. unassign AC when the meta review stage hasn't configured yet. Variable "metareview_name" is set when there metareview_stage is configured. You should get a Key error.

celestemartinez:
> 1. unassign reviewer when the review stage hasn't configured yet. Because the variable "review_name" is set only when the review_stage is configured. You should get a Key error.
> 3. unassign AC when the meta review stage hasn't configured yet. Variable "metareview_name" is set when there metareview_stage is configured. You should get a Key error.

I am not able to reproduce these two scenarios. I believe it is because these two stages are always set when we do `get_conference()`; we always set them with the default parameters?

```
venue.review_stage = get_review_stage(note)
venue.meta_review_stage = get_meta_review_stage(note)
```

I am not sure we should do this all the time?

melisabok:
I see, we set the stage but we don't call to create the stage. Ok, please keep the tests even if you didn't make them fail. If we decide to change this in the feature we can check we don't break anything.

celestemartinez:
The second scenario was happening only because Emma is a PC. ACs don't get any groups when doing 
`groups=client.get_groups(prefix='ICML.cc/2023/Conference/PaperX/Reviewers_', signatory=ac_id)`, so they would get the error: `Can not remove assignment, signatory groups not found for {edge.tail}.`

This has been fixed by changing the anon_name from Reviewers_ to Area_Chairs_ for ACs

melisabok (approved):
looks good!

melisabok:
> The second scenario was happening only because Emma is a PC

The issue can also happen when ACs can see the reviewer identities, so they are readers of the reviewer anon groups. Thanks for adding the tests!

PR #1597, opened 2023-03-08
Fix/ Venue: invite assignment fixes

### LINKED ISSUE #1576: Venue: check if the invite_assignment edge exist when accepting/declining invitation
To avoid not found errors in post process functions, I suggest to check if the invitation assignment edge for the user and submission id exists. 

We can get the user and the submission id from the url. If no edge found for that user, we can try to search the edge by profile.id. Sometimes there is a profile merge and edges are renamed.

How to reproduce this issue?

- AC invites reviewer.
- AC removes invite assignment edge.
- Reviewer accepts the invitaiton.

Expected:

- Reviewer should get an error after clicking the button saying "invitation no longer exists"

Another improvements:

- if the user clicks on "Accept" and the invitation edge says "Accepted" throw an error "You already accepted this invitation"
- if the user clicks on "Declined",  no comment is written and the invitation edge says "Declined" throw an error "You already declined the invitation".
- if the user click on "Declined and there is a comment, I would let the user decline again so the reason can be shared with the AC.

This way we avoid too many recruitment notes and emails.

celestemartinez:
- Fixes #1576

### DIFF
--- a/openreview/venue/process/paper_recruitment_pre_process.py
+++ b/openreview/venue/process/paper_recruitment_pre_process.py
@@ -41,10 +41,15 @@ def process(client, edit, invitation):
             raise openreview.OpenReviewException('You have already accepted this invitation.')
         if invite_edges[0].label == 'Pending Sign Up':
             raise openreview.OpenReviewException('You have already accepted this invitation, but the assignment is pending until you create a profile and no conflict are detected.')
+        if invite_edges[0].label == 'Conflict Detected':
+            raise openreview.OpenReviewException('You have already accepted this invitation, but a conflict was detected and the assignment cannot be made.')
 
-    if note.content['response']['value'] == 'No' and check_decline:
-        groups = client.get_groups(prefix=venue_id, member=user)
-        regex = venue_id + '/.*[0-9]/' + committee_name
-        for group in groups:
-            if re.match(regex, group.id):
-                raise openreview.OpenReviewException('You have already been assigned to a paper. Please contact the paper area chair or program chairs to be unassigned.')
+    if note.content['response']['value'] == 'No':
+        if not note.content.get('comment') and 'Declined' in invite_edges[0].label:
+            raise openreview.OpenReviewException('You have already declined this invitation.')
+        if check_decline:
+            groups = client.get_groups(prefix=venue_id, member=user)
+            regex = venue_id + '/.*[0-9]/' + committee_name
+            for group in groups:
+                if re.match(regex, group.id):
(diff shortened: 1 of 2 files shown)


tests/test_icml_conference.py
```diff
@@ -1509,7 +1510,7 @@ def test_assignment(self, client, openreview_client, helpers, request_page, sele
     def test_reviewer_reassingment(self, client, openreview_client, helpers, selenium, request_page):
 
         pc_client = openreview.api.OpenReviewClient(username='pc@icml.cc', password='1234')
-        ac_client = openreview.api.OpenReviewClient(username='ac1@icml.cc', password='1234')
+        ac_client = openreview.api.OpenReviewClient(username='ac2@icml.cc', password='1234')
```
celestemartinez, line 1513: I changed all these tests to use ac2@icml.cc since this is the AC for paper 1.

openreview/venue/process/paper_recruitment_pre_process.py
```diff
@@ -9,6 +9,7 @@ def process(client, edit, invitation):
     committee_name = invitation.content['committee_id']['value'].split('/')[-1]
     submission_venue_id = domain.content['submission_venue_id']['value']
     check_decline = False
+    invite_assignment_invitation = domain.content['reviewers_invite_assignment_id']['value']
```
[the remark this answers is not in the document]
melisabok: you can leave it for now.

tests/test_icml_conference.py
melisabok: ok

openreview/venue/process/paper_recruitment_pre_process.py
```diff
@@ -17,6 +18,15 @@ def process(client, edit, invitation):
 
     submission = client.get_notes(note.content['submission_id']['value'])[0]
 
+    invite_edges = client.get_edges(invitation=invite_assignment_invitation, head=submission.id, tail=user)
+    if not invite_edges:
+        profile = openreview.tools.get_profile(client, user)
+        if profile:
+            invite_edges = client.get_edges(invitation=invite_assignment_invitation, head=submission.id, tail=profile.id)
```
melisabok, line 25: why do you search the invitation edge by profile?

celestemartinez: From the issue: 
> We can get the user and the submission id from the url. If no edge found for that user, we can try to search the edge by profile.id. Sometimes there is a profile merge and edges are renamed.

PR #1598, opened 2023-03-09
Feature/ Venue: rebuttal stage

celestemartinez:


openreview/venue/invitation.py
```diff
@@ -100,6 +100,40 @@ def post_invitation(note):
...
+                'noteId': {
+                    'value': note.forum
+                },
+                'noteNumber': {
+                    'value': note.number
+                }
+            }
+            if not single_rebuttal and not unlimited_rebuttals:
+                paper_number = note.signatures[0].split(self.venue.submission_stage.name)[-1].split('/')[0]
+                content['noteNumber']['value'] = int(paper_number)
```
melisabok, line 118: can you use `note.number` here?

melisabok: I see the variable notes can be either submissions or reviews.

openreview/venue/invitation.py
```diff
@@ -378,6 +410,143 @@ def set_review_invitation(self):
...
+                }
+            }
+        )
+
+        if review_rebuttal_duedate:
+            invitation.edit['invitation']['duedate'] = review_rebuttal_duedate
+            invitation.edit['invitation']['expdate'] = review_rebuttal_expdate
+
+        if not review_rebuttal_stage.unlimited_rebuttals:
+            invitation.edit['invitation']['maxReplies'] = 1
```
melisabok, line 545: can we set `minReplies:1` for all the three types, so the task can appear as complete when there is at least one reply.

openreview/venue/venue.py
```diff
@@ -138,6 +139,13 @@ def get_recommendation_id(self, committee_id=None):
             committee_id = self.get_reviewers_id()
         return self.get_invitation_id('Recommendation', prefix=committee_id)
 
+    def get_rebuttal_invitation_id(self, name, number, signature = None):
```
melisabok, line 142: why don't you use `get_invitation_id` with the signature as a prefix?

melisabok:
Looks good! I'm confused with the flags, let me try to understand:

single_rebuttal: False, unlimited_rebuttal: False => One rebuttal per review
single_rebuttal: True, unlimited_rebuttal: False => One rebuttal per submission
single_rebuttal: False, unlimited_rebuttal: True => Multiple rebuttals per submissions
single_rebuttal: True, unlimited_rebuttal: True => ( not supported)

openreview/venue/venue.py
celestemartinez: `get_invitation_id()` expects either prefix or number I believe, because it overwrites the invitation_id everytime:

```
if prefix:
    invitation_id = prefix
if number:
    invitation_id = f'{self.get_paper_group_prefix(number)}/-/'
```

celestemartinez:
> Looks good! I'm confused with the flags, let me try to understand:
> 
> single_rebuttal: False, unlimited_rebuttal: False => One rebuttal per review single_rebuttal: True, unlimited_rebuttal: False => One rebuttal per submission single_rebuttal: False, unlimited_rebuttal: True => Multiple rebuttals per submissions single_rebuttal: True, unlimited_rebuttal: True => ( not supported)

The plan was that `unlimited_rebuttals` would override `single_rebuttal`, so in any case where `unlimited_rebuttals` is True, we treat it as Multiple rebuttals per submission.

I do:
`if not review_rebuttal_stage.single_rebuttal and not review_rebuttal_stage.unlimited_rebuttals:`  => One rebuttal per review

and whenever `unlimited_rebuttals=True` => Multiple rebuttals per submissions

Let me know if this is confusing or if it doesn't look right!

openreview/venue/venue.py
melisabok: isn't this the same you are doing?

celestemartinez: similar, but I am giving priority to the prefix, because when it is a signature, it includes the paper number as well. 

`get_rebuttal_invitation_id()` => '${2/content/noteSignatures/value}/-/Rebuttal'
`get_invitation_id()` => 'ICML.cc/2023/Conference/Submission${2/content/noteNumber/value}/-/Rebuttal' (it ignores the prefix because there is a paper number)

celestemartinez: nvm I see it now, I need to pass prefix only and not number! will update

celestemartinez:
I realize `unlimited_rebuttals` was not implemented for first API but the option will show in the request form so adding that now. It is easier because we already have separate invitations for paper vs review rebuttals. 

Another point, we are not checking if the signatures belongs to a reviewer in the API2, should I add this too?
We check this in the conference class: `conference.review_rebuttal_stage.unlimited_rebuttals:`
@melisabok

melisabok:
>I realize unlimited_rebuttals was not implemented for first API but the option will show in the request form so adding that now. It is easier because we already have separate invitations for paper vs review rebuttals.

Why don't we just create the rebuttal stage invitation for the venue of the API 2 only?

melisabok:
>Another point, we are not checking if the signatures belongs to a reviewer in the API2, should I add this too?

Sounds good, let's check it but I believe the Review Invitation doesn't allow other signature than the reviewer.

tests/test_icml_conference.py
```diff
@@ -2961,7 +2967,136 @@ def test_rebuttal_stage(self, openreview_client, helpers):
...
+        # post a rebuttal stage note
+        now = datetime.datetime.utcnow()
+        start_date = now - datetime.timedelta(days=2)
+        due_date = now + datetime.timedelta(days=3)
+        pc_client.post_note(openreview.Note(
+            content={
+                'rebuttal_start_date': start_date.strftime('%Y/%m/%d'),
+                'rebuttal_deadline': due_date.strftime('%Y/%m/%d'),
+                'number_of_rebuttals': 'Multiple author rebuttals per paper',
+                'email_program_chairs_about_rebuttals': 'No, do not email program chairs about received rebuttals'
```
melisabok, line 2985: you are not selecting any readers here? are the readers optional?

celestemartinez: Yes, because PCs and authors are added always. This is in the description but I do think it might be a bit confusing. I thought about adding PCs and authors as default but we don't really need them to be there for the stage because they are always added by default.

melisabok (approved):

Issue #1586, opened 2023-03-01
Venue: support custom "Stage Setup"


melisabok:
Give the Program Chairs the possibility of defining their own stages. Like "Author Ranking", "Review Rating", "MetaReview Confirmation" , etc. 

Sometime PCs decide to create different tasks for their workflow that require to create the invitation outside of the venue request form workflow. I suggest we help them to do this by their own. 

There are different types of stages: one reply per forum, multiple replies per forum, one reply per review, one reply per meta review, etc.

These is the proposal of fields to support in the stage:

1. Reply type: forum, review, metareview. (in case they decide to go for review/metareview, the process function should create these invitation when the review/metareview is posted, let's discuss this)
2. Source: all the submissions, range number of submissions, etc
3. invitees: SACs, ACs, reviewers or authors. (when authors is selected, we can have an option to sign using the authors group or the tilde id in case they want one reply per author) 
4. reply readers: SACs, ACs, reviewers or authors
5. content params: to describe the form.
6. notification template.

PR #1608, opened 2023-03-21
Feat/ Venue: Custom Stage Setup

celestemartinez:
Custom stage issue: #1586


openreview/venue/invitation.py
```diff
@@ -1818,6 +1818,144 @@ def set_submission_revision_invitation(self, submission_revision_stage=None):
...
+                        }
+                    },
+                    'noteId': {
+                        'value': {
+                            'param': {
+                                'regex': '.*', 'type': 'string'
+                            }
+                        }
+                    },
+                    'replytoSignatures': {
```
melisabok, line 1884: I would prefer to add these fields only they are needed instead of making them optional. 

If in the future there will be a UI to create these invitations I don't want to show fields that are not needed.

openreview/venue/invitation.py
```diff
@@ -1818,6 +1818,144 @@ def set_submission_revision_invitation(self, submission_revision_stage=None):
...
+        custom_stage_source = custom_stage.get_source_submissions()
+
+        if custom_stage_replyto == 'forum':
+            paper_invitation_id = self.venue.get_invitation_id(name=custom_stage.name, number='${2/content/noteNumber/value}')
+            with_invitation = self.venue.get_invitation_id(name=custom_stage.name, number='${6/content/noteNumber/value}')
+            reply_to = '${4/content/noteId/value}'
+        else:
+            paper_invitation_id = self.venue.get_invitation_id(name=custom_stage.name, prefix='${2/content/replytoSignatures/value}')
+            with_invitation = self.venue.get_invitation_id(name=custom_stage.name, prefix='${6/content/replytoSignatures/value}')
+            reply_to = '${4/content/replyto/value}'
```
melisabok, line 1844: so you are going to create one invitation per reply instead one invitation per forum?

what happens if there are reviews created after that?

openreview/venue/process/custom_stage_process.py
```diff
@@ -0,0 +1,128 @@
...
+        message=f'''We have received your {invitation_name} on a submission to {short_name}.
+
+{content}
+''' if not email_template else email_template
+    )
+
+    senior_area_chairs_name = domain.get_content_value('senior_area_chairs_name')
+    senior_area_chairs_id = domain.get_content_value('senior_area_chairs_id')
+    paper_senior_area_chairs_id = f'{paper_group_id}/{senior_area_chairs_name}'
+    if senior_area_chairs_name and ('everyone' in note.readers or senior_area_chairs_id in note.readers or paper_senior_area_chairs_id in note.readers):
```
melisabok, line 65: do we want to notify the SACs?

openreview/venue/invitation.py
celestemartinez: Yes, I just remembered we spoke about this; this would be very heplful for SAC agreement. I will make this change. What will we do about this showing as completed in tasks when there are more tasks to be completed? For example, review rating?

openreview/stages/venue_stages.py
```diff
@@ -971,3 +971,131 @@ def __init__(self, committee_id, name='Registration', start_date=None, due_date=
...
+        REVIEWERS_SUBMITTED = 7
+        AUTHORS = 8
+
+    class Source(Enum):
+        ALL_SUBMISSIONS = 0
+        ACCEPTED_SUBMISSIONS = 1
+        PUBLIC_SUBMISSIONS = 2
+
+    class ReplyTo(Enum):
+        SUBMISSION = 0
```
melisabok, line 998: what is the difference between submission and forum?

celestemartinez: Well, `submission` can also be multireply, but it means the replyto is the submission id. 
`Forum` means withForum, so the reply can be made to the submission or to any of it's replies as well.

melisabok: what if we rename them submission -> forum and forum -> withForum?

openreview/venue/process/metareview_process.py
```diff
@@ -0,0 +1,28 @@
+def process(client, edit, invitation):
+
+    domain = client.get_group(edit.domain)
+    venue_id = domain.id
+    submission_name = domain.content['submission_name']['value']
+
+    metareview = client.get_note(edit.note.id)
+
+    #create children invitation if applicable
+    venue_invitations = [i for i in client.get_all_invitations(prefix=venue_id + '/-/') if i.is_active()]
```
melisabok, line 10: you can use `type=invitation` so you don't need to check for the line 14

openreview/venue/process/review_process.py
```diff
@@ -110,5 +110,24 @@ def process(client, edit, invitation):
 
     if paper_reviewers_submitted_id:
         client.add_members_to_group(paper_reviewers_submitted_id, review.signatures[0])
-    
 
+    #create children invitation if applicable
+    venue_invitations = [i for i in client.get_all_invitations(prefix=venue_id + '/-/') if i.is_active()]
```
melisabok, line 115: same here

openreview/venue/process/metareview_process.py
```diff
@@ -0,0 +1,28 @@
...
+
+    for invitation in venue_invitations:
+        print('processing invitation: ', invitation.id)
+        if isinstance(invitation.edit, dict) and 'invitation' in invitation.edit:
+            metareview_reply = invitation.content.get('reply_to', {}).get('value', False) if invitation.content else False
+            content_keys = invitation.edit.get('content', {}).keys()
+            if 'metareviews' == metareview_reply and 'replytoSignatures' in content_keys and 'replyto' in content_keys and len(content_keys) == 4:
+                print('create invitation: ', invitation.id)
+                paper_number = metareview.signatures[0].split(submission_name)[-1].split('/')[0]
```
melisabok, line 19: I prefer to get the number from the submission, what about doing:
```
submission = client.get_note(edit.note.forum)
```

melisabok (approved):

PR #1618, opened 2023-03-30
check profile has email

### LINKED ISSUE #1615: Venue class: recruitment process should check if a profile contains an email otherwise save it in the error message
Instead of showing the exceptions.

celestemartinez:
Fixes #1615

### DIFF
--- a/openreview/venue/recruitment.py
+++ b/openreview/venue/recruitment.py
@@ -94,7 +94,17 @@ class Recruitment(object):
             invited_group_ids=list(set(invited_roles) & set(memberships))
             member_group_ids=list(set(member_roles) & set(memberships))
 
-            if invited_group_ids:
+            profile_emails = []
+            profile = None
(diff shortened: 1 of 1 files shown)


openreview/venue/recruitment.py
```diff
@@ -94,7 +94,17 @@ def invite_committee(self,
             invited_group_ids=list(set(invited_roles) & set(memberships))
             member_group_ids=list(set(member_roles) & set(memberships))
 
-            if invited_group_ids:
+            profile_emails = []
+            profile = None
+            if email.startswith('~'):
```
melisabok, line 99: what happens if the profile id doesn't exist?

celestemartinez: then it won't enter the first if because there is no profile, and the error will be caught later on. I guess it would be better to just catch it here as well?

melisabok (approved):

PR #1625, opened 2023-04-03
Fix/ Venue recruitment: check for validation error

### LINKED ISSUE #1622: Venue recruitment: validate that profiles are valid
NeurIPS is sending invitations to invitee where the profile id is missing the number and the backend fails with this error:

Screen Shot 2023-04-03 at 9 32 28 AM

I suggest to first catch this error in the recruitment status and then add a validation that profile id must be valid.

more errors:

https://api.openreview.net/logs/process?id=07m88SDyTV

celestemartinez:
- The error is actually in `tools.get_profile(client, profile_id)`, not sure if this is the best way to handle this error
- Fixes #1622

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -2184,12 +2184,6 @@ class Invitation(object):
         edate = self.expdate if self.expdate else now
         return cdate <= now and now <= edate
 
-    def is_active(self):
-        now = tools.datetime_millis(datetime.datetime.utcnow())
-        cdate = self.cdate if self.cdate else now
-        edate = self.expdate if self.expdate else now
-        return cdate <= now and now <= edate
-
     def pretty_id(self):
         tokens = self.id.split('/')[-2:]
         filtered_tokens = []
--- a/openreview/journal/assignment.py
+++ b/openreview/journal/assignment.py
@@ -131,24 +131,6 @@ class Assignment(object):
         self.post_submission_edges(conflict_edges)
         print('Finished setup Reviewer assignment.')
 
-
-    def assign_reviewer(self, note, reviewer, solicit=False):
-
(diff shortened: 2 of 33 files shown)


openreview/venue/recruitment.py
```diff
@@ -87,25 +87,38 @@ def invite_committee(self,
...
+                    error_string = repr(e)
+                    if 'ValidationError' in error_string:
+                        invalid_profile_id = True
+                    else:
+                        if error_string not in recruitment_status['errors']:
+                            recruitment_status['errors'][error_string] = []
+                        recruitment_status['errors'][error_string].append(email)
+                        continue
+                if not profile:
+                    invalid_profile_id = True
```
melisabok, line 107: should we distinguish between invalid_profile_id and not_found_profile_id? I think we could make that difference.

openreview/venue/recruitment.py
```diff
@@ -87,25 +87,38 @@ def invite_committee(self,
...
+            invalid_profile_id = False
+            if is_profile_id:
+                try:
+                    profile = tools.get_profile(self.client, email)
+                except openreview.OpenReviewException as e:
+                    error_string = repr(e)
+                    if 'ValidationError' in error_string:
+                        invalid_profile_id = True
+                    else:
+                        if error_string not in recruitment_status['errors']:
```
melisabok, line 102: in which case we get this error? like an upstream error or something like that?

celestemartinez: I am not sure, we haven't gotten any error here yet, but I was wondering if we should still catch errors other than ValidationError in case we get any

melisabok: yes, we should catch them.

PR #1629, opened 2023-04-04
Removed unused attribute from OpenReviewClient constructor.

miselico:


openreview/api/client.py
```diff
@@ -65,7 +65,6 @@ def __init__(self, baseurl = None, username = None, password = None, token= None
         self.group_edits_url = self.baseurl + '/groups/edits'
         self.user_agent = 'OpenReviewPy/v' + str(sys.version_info[0])
 
-        self.limit = 1000
```
melisabok, line 68: this property is being used inside the tools.py module to make concurrent requests:

https://github.com/openreview/openreview-py/blob/master/openreview/tools.py#L686

miselico: I see, I missed that. Closing.

PR #1634, opened 2023-04-05
Fix/ Venue matching: change order of entries when using expertise model to compute SAC scores

celestemartinez:


melisabok (approved):
thanks for the fix!

PR #1635, opened 2023-04-06
Fix/Builder + Venue: add profiles_only property

### LINKED ISSUE #1624: Venue request: add a field to decide if the submission accepts profile only or profile and email addresses
This will change the regex in the submission invitation. 

If NeurIPS wants to force having only profiles during the abstract submission we need to have this property implemented.

celestemartinez:
Fixes #1624


melisabok (approved):
Nice!

PR #1637, opened 2023-04-06
translate from python to JS pre process functions

melisabok:


melisabok:
There is one pre process missing that is the one that compute the conflicts before sending the recruitment invitation to the user. 

Please take a look and post the feedback about the JS code.

openreview/venue/invitation.py
```diff
@@ -1925,6 +1924,14 @@ def set_custom_stage_invitation(self):
     def set_assignment_invitation(self, committee_id):
         client = self.client
         venue = self.venue
+        content = {
```
melisabok, line 1927: @celestemartinez moving this to the content instead of having a variable.

openreview/venue/process/assignment_pre_process.js
```diff
@@ -0,0 +1,36 @@
+async function process(client, edge, invitation) {
+  client.throwErrors = true
+
+  const { groups } = await client.getGroups({ id: edge.domain })
+  const domain = groups[0]
+  const venue_id = domain.id
+  const submission_name = domain.content.submission_name.value
+  const review_name = invitation.content.review_name.value
+  const reviewers_anon_name = invitation.content.reviewers_anon_name.value
```
carlosmondra, line 9: In case some of these properties are not present in the content use optional chaining to avoid unintended errors.

```suggestion
  const submission_name = domain.content.submission_name?.value
  const review_name = invitation.content.review_name?.value
  const reviewers_anon_name = invitation.content.reviewers_anon_name?.value
```

openreview/venue/process/assignment_pre_process.js
```diff
@@ -0,0 +1,36 @@
...
+
+  if (!review_name) {
+    return
+  }
+
+  if (edge.ddate) {
+
+    const { notes } = await client.getNotes({ id: edge.head })
+    const submission = notes[0]
+    const submissionGroupId = `${venue_id}/${submission_name}${submission.number}`
```
carlosmondra, line 19: Using `${}` here does not cause any errors? Wouldn't it be better to do:
```suggestion
    const submissionGroupId = venue_id + '/' +submission_name + submission.number
```

openreview/venue/process/paper_recruitment_pre_process.js
```diff
@@ -0,0 +1,69 @@
+async function process(client, edit, invitation) {
+
+  const committeeInvitedId = invitation.content.committee_invited_id.value
+  const committeeName = invitation.content.committee_name.value
+  const inviteAssignmentInvitation = invitation.content.invite_assignment_invitation_id.value
+  const hash_seed = invitation.content.hash_seed.value
```
carlosmondra, line 6: Not sure if these properties will always be in `content`, but if they aren't then I suggest using optional chaining too.

```suggestion
  const committeeInvitedId = invitation.content.committee_invited_id?.value
  const committeeName = invitation.content.committee_name?.value
  const inviteAssignmentInvitation = invitation.content.invite_assignment_invitation_id?.value
  const hash_seed = invitation.content.hash_seed?.value
```

openreview/venue/process/paper_recruitment_pre_process.js
```diff
@@ -0,0 +1,69 @@
...
+    if (profiles.length > 0) {
+      const { count: edgeCount, edges } = await client.getEdges({ invitation: inviteAssignmentInvitation, head: submission.id, tail: profiles[0].id })
+      inviteEdges = edges
+      if (edgeCount == 0) {
+        return Promise.reject(new OpenReviewError({ name: 'Error', message: 'Invitation no longer exists. No action is required from your end.' }))
+      }
+    }
+  }
+
+  const { count: groupCount } = await client.getGroups({ id: committeeInvitedId, member: user })
```
carlosmondra, line 40: If you want to check that `user` is a member of the `committeeInvitedId` group, you can use `members` instead which does not calculate transitive members and is therefore more efficient.

```suggestion
  const { count: groupCount } = await client.getGroups({ id: committeeInvitedId, members: user })
```

carlosmondra (commented):
I think there are several places where optional chaining can be used before `.value` 😬

openreview/venue/process/assignment_pre_process.js
zbialecki: Using `+` for string concatenation is discouraged by most linters these days as it can lead to some type conversion weirdness. Like for example if submissionName and submissionNumber were both ints then that code would add the numbers together.

openreview/venue/process/paper_recruitment_pre_process.js
```diff
@@ -0,0 +1,69 @@
...
+    return Promise.reject(new OpenReviewError({ name: 'Error', message: 'Wrong key, please refer back to the recruitment email' }))
+  }
+
+  const { notes } = await client.getNotes({ id: note.content.submission_id.value })
+  const submission = notes[0]
+
+  const { count: edgeCount, edges } = await client.getEdges({ invitation: inviteAssignmentInvitation, head: submission.id, tail: note.content.user.value })
+  let inviteEdges = edges
+
+  if (edgeCount == 0) {
```
zbialecki, line 29:
```suggestion
  if (edgeCount === 0) {
```

openreview/venue/process/proposed_assignment_pre_process.js
```diff
@@ -0,0 +1,40 @@
...
+
+  if (edge.ddate) {
+    return
+  }
+
+  if (edge.tcdate !== edge.tmdate) {
+    return
+  }
+
+  if (customMaxPapersId) {
```
zbialecki, line 18: This is another case I would invert

openreview/venue/process/assignment_pre_process.js
melisabok: Thanks, I wonder if the variables should be named "submissionName", "reviewName", etc

openreview/venue/process/assignment_pre_process.js
melisabok: > Using ${} here does not cause any errors? Wouldn't it be better to do

It doesn't because this code is in the process function and not in the content. The API skips the $ replacement on process fields.

openreview/venue/process/paper_recruitment_pre_process.js
melisabok: I do want to check the transitive members

openreview/venue/process/recruitment_pre_process.js
```diff
@@ -0,0 +1,35 @@
...
+
+  const hashkey = crypto.createHmac('sha256', hash_seed)
+                      .update(user)
+                      .digest('hex');
+
+  if (hashkey != note.content.key.value) {
+    return Promise.reject(new OpenReviewError({ name: 'Error', message: 'Wrong key, please refer back to the recruitment email' }))
+  }
+
+  const { groups } = await client.getGroups({ id: committeeInvitedId, member: user })
```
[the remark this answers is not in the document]
melisabok: we always check with transitive members because the users could have been invited with different emails or the invited group could have been modified, it is safer to check the membership with transitive members.

openreview/venue/invitation.py
```diff
@@ -865,15 +865,14 @@ def set_official_comment_invitation(self):
...
-    exec(script, funcs)
-    funcs['process'](client, edit, invitation)
-''' if comment_stage.check_mandatory_readers and comment_stage.reader_selection else None,
+                    'preprocess': '''async function process(client, edit, invitation) {
+  client.throwErrors = true;
+  const { invitations } = await client.getInvitations({ id: invitation.invitations[0] })
+  const metaInvitation = invitations[0];
+  const script = metaInvitation.content.comment_preprocess_script.value;
+  eval(`var process = ${script}`);
+  await process(client, edit, invitation);
```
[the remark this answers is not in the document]
melisabok: I'm not sure about this question. The reason for a second eval is because the source code of the process function is stored in the content of an object so all the invitations that need it can just access to the code in the content and then execute it. I'm not sure if eval is the best way to do it.

openreview/venue/process/assignment_pre_process.js
```diff
@@ -0,0 +1,36 @@
+async function process(client, edge, invitation) {
+  client.throwErrors = true
+
+  const { groups } = await client.getGroups({ id: edge.domain })
+  const domain = groups[0]
+  const venue_id = domain.id
```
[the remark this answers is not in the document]
melisabok: Yeah, copilot did that for me so I was wondering if I should use camelCase everywhere.

melisabok:
> Using + for string concatenation is discouraged by most linters these days as it can lead to some type conversion weirdness. Like for example if submissionName and submissionNumber were both ints then that code would add the numbers together.

The reason we use + is because using ${} the backend tries to replace it with some other field in the object and there is no way to escape it.

openreview/venue/process/assignment_pre_process.js
carlosmondra: I would change them, but I don't think it's too bad.

melisabok: Done!

openreview/venue/process/assignment_pre_process.js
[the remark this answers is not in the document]
melisabok: Done!

openreview/venue/process/assignment_pre_process.js
```diff
@@ -0,0 +1,36 @@
...
+  const submission_name = domain.content.submission_name.value
+  const review_name = invitation.content.review_name.value
+  const reviewers_anon_name = invitation.content.reviewers_anon_name.value
+
+  if (!review_name) {
+    return
+  }
+
+  if (edge.ddate) {
```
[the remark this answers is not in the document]
melisabok: Done!

openreview/venue/process/proposed_assignment_pre_process.js
melisabok: Done!

melisabok:
TMLR pre process function can not be translated right now because I need the Journal object. Once I move all the configuration to the venue.content then I can migrate them.

openreview/venue/invitation.py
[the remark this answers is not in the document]
zbialecki: OK, never mind I understand now.

zbialecki (approved):

openreview/venue/process/paper_recruitment_pre_process.js
```diff
@@ -0,0 +1,67 @@
+async function process(client, edit, invitation) {
+
```
carlosmondra, line 2:
```suggestion
  client.throwErrors = true
 
```

openreview/venue/process/recruitment_pre_process.js
```diff
@@ -0,0 +1,37 @@
+async function process(client, edit, invitation) {
+
```
carlosmondra, line 2: Not sure if you want to add this:
```suggestion
  client.throwErrors = true

```

carlosmondra (approved):

openreview/venue/process/comment_pre_process.js
```diff
@@ -0,0 +1,24 @@
+async function process(client, edit, invitation) {
+  client.throwErrors = true
+
+  const { note } = edit
+  
+  if (note.ddate || note.readers === ['everyone']) {
```
xkopenreview, line 6: i think the everyone comparison logic should be changed to use includes
current logic will always be false

melisabok: like this? `note.readers.includes('everyone')`?

xkopenreview: yes
or readers[0]==='everyone'

openreview/venue/process/assignment_pre_process.js
```diff
@@ -0,0 +1,37 @@
...
+    return
+  }
+
+  const { notes } = await client.getNotes({ id: edge.head })
+  const submission = notes[0]
+  const submissionGroupId = `${venueId}/${submissionName}${submission.number}`
+
+  const { groups: anonGroups, count: groupCount } = await client.getGroups({ prefix: `${submissionGroupId}/${reviewersAnonName}`, signatory: edge.tail })
+
+  if (groupCount.length == 0) {
```
xkopenreview, line 25: i think this should be anonGroups.length=== or groupCount===

celestemartinez (approved):

openreview/venue/process/comment_pre_process.js
melisabok: fixed!

openreview/venue/process/assignment_pre_process.js
melisabok: fixed

xkopenreview (approved):

openreview/venue/process/comment_pre_process.js
```diff
@@ -0,0 +1,24 @@
...
+    client.getGroups({ id: invitation.domain })
+  ])
+  const forum = res1.notes[0]
+  const domain = res2.groups[0]
+
+  const commentMandatoryReaders = domain.content?.comment_mandatory_readers?.value || []
+  for (const m of commentMandatoryReaders) {
+    const reader = m.replace('{number}', forum.number)
+    if (!note.readers.includes(reader)) {
+      return Promise.reject(new OpenReviewError({ name: 'Error', message: reader + 'must be readers of the comment' }));
```
xkopenreview, line 21: i think it's missing a space here

xkopenreview: image

PR #1643, opened 2023-04-12
Feat/Venue request: registration stage

celestemartinez:
Add reviewer and AC registration to request form:
- the forum is created everytime we run the stage. I am now checking if a note already exists with that invitation and I edit it instead of creating a new one everytime
- either button will run `create_registration_stages()`, updating both registration stages if they exist
- added expdate
- added to venue only


openreview/venue_request/venue_request.py
```diff
@@ -964,6 +969,11 @@ def setup_registration_stages(self):
                 'required': False,
                 'order': 12
             },
+            'ac_registration_name': {
```
melisabok, line 972: should we call it "area_chair_registration_name" or "AC_regristation_name"?  so the UI shows them as "Area Chair Registration Name" or "AC Registration Nname"

melisabok (approved):

PR #1645, opened 2023-04-12
Feature/improve get conflicts

carlosmondra:
This PR:

- Adds `n_years` validation for conflict detection
- Adds `get_institutions`, `post_institution` and `delete_institution`

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -1247,7 +1247,7 @@ def get_all_venues(client):
     """
     return client.get_group("host").members
 
-def get_conflicts(author_profiles, user_profile, policy='default', n_years=5):
+def get_conflicts(author_profiles, user_profile, policy='default', n_years=None):
     """
     Finds conflicts between the passed user Profile and the author Profiles passed as arguments
 
@@ -1264,8 +1264,6 @@ def get_conflicts(author_profiles, user_profile, policy='default', n_years=5):
     author_relations = set()
     author_publications = set()
     info_function = get_neurips_profile_info if policy == 'neurips' else get_profile_info
-    if policy == 'neurips':
-        n_years = 3
 
     for profile in author_profiles:
         author_info = info_function(profile, n_years)
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -4,9 +4,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import json
 import os
-import string
 
-import sys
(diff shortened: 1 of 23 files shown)


openreview/tools.py
```diff
@@ -1319,10 +1319,17 @@ def get_profile_info(profile, n_years=3):
     ## Relations section
     relations.update([r['email'] for r in profile.content.get('relations', [])])
 
-    ## TODO:: Parameterize the number of years for publications to consider from
     ## Publications section: get publications within last n years, default is all publications from previous years
-    for pub in profile.content.get('publications', []):
-        publications.add(pub.id)
+    if n_years:
```
melisabok, line 1323: I think n_years should be used to filter all the data of the profile that contains a date like: history and relations.

carlosmondra: Ah, just like neurips. I think I can simply copy it over.

carlosmondra: OK, @melisabok  I added it to relations and history

melisabok: looks good!

"default" policy checks the whole profile content from the last N years. 
"neurips" policy checks the profile content ignoring intern positions, only including co-worker and co-auhtors relations and use email domains only when there is history record from the last N years.

melisabok: should we rename these policies? what do you think?

openreview/tools.py
```diff
@@ -1380,6 +1408,7 @@ def get_neurips_profile_info(profile, n_years=3):
             domains.update(openreview.tools.subdomains(email))
 
     ## Publications section: get publications within last n years
+    curr_year = datetime.datetime.now().year
```
melisabok, line 1411: what do you need to re-set curr_year? I see the filter is using `>` instead of `>=`:

```
        if year > cut_off_year:
            publications.add(pub.id)
```

carlosmondra: What do you mean re-set? This is the first that the variable is declared.

> I see the filter is using > instead of >=

What filter are you talking about?

openreview/tools.py
carlosmondra: Are you asking this because you want a name that is more descriptive of what it does?

openreview/tools.py
melisabok: sorry, I misread the variable, I thought it was "cut_off_year"

melisabok (approved):

PR #1651, opened 2023-04-13
Feat: Enhance get conflicts

carlosmondra:
This PR:

- Allows passing an info_function
- Calculates subdomains outside the info_function
- Adds a cache for domain to subdomains to speed up part of the process

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -353,11 +353,11 @@ class TestTools():
     def test_subdomains(self):
         # ensure that two part top-level domains are handled appropriately
         # e.g. "edu.cn", "ac.uk"
-        assert openreview.tools.subdomains('michael@mails.tsinghua.edu.cn') == ['mails.tsinghua.edu.cn', 'tsinghua.edu']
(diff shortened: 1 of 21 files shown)


tests/test_tools.py
```diff
@@ -353,11 +353,11 @@ def test_create_authorid_profiles(self, client):
     def test_subdomains(self):
         # ensure that two part top-level domains are handled appropriately
         # e.g. "edu.cn", "ac.uk"
-        assert openreview.tools.subdomains('michael@mails.tsinghua.edu.cn') == ['mails.tsinghua.edu.cn', 'tsinghua.edu']
```
melisabok, line 356: why did you change the assertions here?

carlosmondra: Because the function `openreview.tools.subdomains` was doing two things" splitting the emails and then getting the subdomains for the domains of the emails. I extracted that logic and split the emails before calling the `subdomains` function

melisabok (approved):

celestemartinez (approved):

PR #1653, opened 2023-04-17
Feat/ API2 client: implement get_group_edit and edit post_edit

celestemartinez:


melisabok (approved):
thanks for fixing this!

PR #1654, opened 2023-04-18
Feature/use strong password

carlosmondra:
Requires: https://github.com/openreview/openreview-api-v1/pull/2859


melisabok (approved):

xkopenreview:
can we use a strong password that is easier to remember?

Issue #1662, opened 2023-04-27
Venue: make the authorids writers of the submission edits


melisabok:
So they can delete their own edits during submission time

PR #1667, opened 2023-05-01
Fix/ Submission invitation: run with PCs permissions and allow authors to delete edits

celestemartinez:
- Fixes: #1662
- Fixes: #1663
- Needed to run NeurIPS pre-process function


melisabok:
Could you add to this PR a change that set the api_version: 2 as default? I think we should change the text saying that new venues are starting to run in the API 2 with the exception of a few like ARR

melisabok (approved):

PR #1673, opened 2023-05-03
TMLR stats for reviewers and action editors

melisabok:
- Create stats function for reviewers and action editors. 
- Allow changing the `cdate` of an edge.
- Let the EICs to approve the submission to be under review.
- Let the EICs to be readers of the solicit review and solicit review approval.
- Add new optional field to the decision invitation.
- Setup expertise selection task(optional) for AEs and reviewers.
- Don't send reminders to the EICs that are authors of the submission related to the reminder.
- Send a warning emails to the AE when the submission contains an author that is an EIC.
- Show the email address of the solicit reviewer in the email message so the AE can contact them.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, division, print_function, unicode_literals
+import inspect
 
 import json
 import os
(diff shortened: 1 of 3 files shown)


openreview/journal/journal.py
```diff
@@ -1001,4 +1005,474 @@ def archive_assignments(self):
...
+                    ack_invitation = invitations_by_id.get(self.get_reviewer_responsibility_id(name["username"]))
+                    if ack_invitation:
+                        return ack_invitation
+                    
+            for email in profile.content['emailsConfirmed']:
+                ack_invitation = invitations_by_id.get(self.get_reviewer_responsibility_id(email))
+                if ack_invitation:
+                    return ack_invitation
+                
+            print('no responisibility invitation for', profile.id)
```
celestemartinez, line 1045:
```suggestion
            print('no responsibility invitation for', profile.id)
```

celestemartinez:
Looks good! Why do we want to change the `cdate` of an edge?

celestemartinez (approved):

melisabok:
> Why do we want to change the cdate of an edge?

When archiving the edges for TMLR I need to keep the cdate(assignment date) for the stats otherwise we loose that data so I need to post an Archived_Assignment edge with the cdate of the Assignment edge.

Issue #1671, opened 2023-05-03
Paper Matching Setup Status: update message


racheljsmart:
I think the last line of the paper matching setup status message should be updated.

<img width="1072" alt="Screen Shot 2023-05-03 at 3 33 58 PM" src="https://user-images.githubusercontent.com/27733190/236026383-123a4cd0-01c0-44b8-9d9c-7a562f54e936.png">

`Please refer to the FAQ for pointers on how to run the matcher: https://openreview.net/faq#question-edge-browswer`

Issues:
- Browser is misspelled in the link
- None of the questions include the edge browser by name so this will be confusing for users who don't know what it is

The documentation has grown quite a lot, I think this link would be better to share https://docs.openreview.net/how-to-guides/paper-matching-and-assignment/how-to-do-automatic-assignments

Suggested line:
Please refer to the documentation for instructions on how to run the matcher: https://docs.openreview.net/how-to-guides/paper-matching-and-assignment/how-to-do-automatic-assignments

PR #1674, opened 2023-05-04
Updating profiles_status message

racheljsmart:
Reasons for change are outlined in this issue https://github.com/openreview/openreview-py/issues/1671

Issues:
Browser is misspelled in the link
None of the questions include the edge browser by name so this will be confusing for users who don't know what it is

The documentation has grown quite a lot, I think this link would be better to share https://docs.openreview.net/how-to-guides/paper-matching-and-assignment/how-to-do-automatic-assignments


melisabok (approved):

melisabok:
it seems the tests are failing because they are checking the text that you changed, could you make the changes in the test too?

https://github.com/openreview/openreview-py/blob/master/tests/test_venue_request.py#L1226

https://github.com/openreview/openreview-py/blob/master/tests/test_venue_request_v2.py#L1215

There should be other places too

PR #1676, opened 2023-05-08
Feat/ Venue: SAC releaxed conflict computation

celestemartinez:
- NeurIPS asked to use only current SAC submissions when transfering conflicts to ACs
- Run as follows:
```
venue_matching = openreview.venue.matching.Matching(venue, openreview_client.get_group(venue.get_area_chairs_id()))
venue_matching.sac_profile_info = openreview.tools.get_sac_profile_info
venue_matching.setup(compute_affinity_scores=True, compute_conflicts='NeurIPS')
```

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -1454,6 +1454,33 @@ def get_neurips_profile_info(profile, n_years=None):
         'publications': publications
     }
 
+def get_sac_profile_info(profile, submission_venue_id):
+    """
+    Gets only submissions submitted to the current venue
+
+    :param profile: Profile from which all publications will be obtained
+    :type profile: Profile
+    :param submission_venue_id: venue_id of submissions we want to obtain
+    :type submission_venue_id: str
+
+    :return: Dictionary with the current publications associated with the passed Profile
+    :rtype: dict
+    """
+    publications = set()
+
+    ## Get publications
+    for publication in profile.content.get('publications', []):
+        print(publication.content)
+        if isinstance(publication.content.get('venueid'), dict) and publication.content['venueid']['value'] == submission_venue_id:
+            publications.add(publication.id)
+
+    return {
+        'id': profile.id,
+        'domains': set(),
+        'emails': set(),
+        'relations': set(),
+        'publications': publications
+    }
 
 def post_bulk_edges (client, edges, batch_size = 50000):
     num_edges = len(edges)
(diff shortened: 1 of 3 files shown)


openreview/tools.py
```diff
@@ -1454,6 +1454,33 @@ def get_neurips_profile_info(profile, n_years=None):
         'publications': publications
     }
 
+def get_sac_profile_info(profile, submission_venue_id):
```
melisabok, line 1457: I would call it with a name related to the implementation because it can be used in other places. 

What about? `get_current_submissions_profile_info`

openreview/venue/matching.py
```diff
@@ -250,12 +251,14 @@ def _build_note_conflicts(self, submissions, user_profiles, get_profile_info, co
...
         ## for AC conflicts, check SAC conflicts too
         sac_user_info_by_id = {}
-        sac_assignment_edges = {}
         if self.is_area_chair:
             sacs_by_ac =  { g['id']['head']: [v['tail'] for v in g['values']] for g in self.client.get_grouped_edges(invitation=self.venue.get_assignment_id(self.venue.get_senior_area_chairs_id()), groupby='head', select=None)}
             if sacs_by_ac:
                 sac_user_profiles = openreview.tools.get_profiles(self.client, self.client.get_group(self.venue.get_senior_area_chairs_id()).members, with_publications=True)
-                sac_user_info_by_id = { p.id: info_function(p, compute_conflicts_n_years) for p in sac_user_profiles }
+                if self.sac_profile_info:
+                    sac_user_info_by_id = { p.id: self.sac_profile_info(p, self.venue.get_submission_venue_id()) for p in sac_user_profiles }
```
melisabok, line 259: I would like all the profile_info functions to have the same signatures so they can be used everywhere conflicts need to be computed.

melisabok: we could use the module `inspect` to check if the profile_info function expects `submission_venue_id` as a parameter. 

Example from the chat:

 
Screen Shot 2023-05-08 at 10 48 49 AM

celestemartinez:
NeurIPS decided to use current submissions and also any relations and history with no end date (I think we should also include anything with an end date in the future). Now this would be run as follows:
```
venue_matching = openreview.venue.matching.Matching(venue, openreview_client.get_group(venue.get_area_chairs_id()))
venue_matching.sac_profile_info = openreview.tools.get_current_submissions_profile_info
venue_matching.sac_n_years = 0   ## if we use n_years = 1, the cutoff year would be 2022, not 2023
venue_matching.setup(compute_affinity_scores=Yes, compute_conflicts='NeurIPS')
```

melisabok (approved):

Issue #1616, opened 2023-03-28
Venue class: move the homepage override options to the home page settings


melisabok:
The field instructions must be stored in the home page options. For the API 2 must be mark down and not HTML.

PR #1679, opened 2023-05-09
Fix/ Venue: homepage override

celestemartinez:
#1616


tests/test_neurips_conference_v2.py
```diff
@@ -101,11 +101,55 @@ def test_create_conference(self, client, openreview_client, helpers):
...
+                }
+            }
+        ))
+        helpers.await_queue()
+
+        request_page(selenium, 'http://localhost:3030/group?id=NeurIPS.cc/2023/Conference', pc_client.token, wait_for_element='header')
+        header_div = selenium.find_element_by_id('header')
+        assert header_div
+        location_tag = header_div.find_element_by_class_name('venue-location')
+        assert location_tag and location_tag.text == request_form.content['Location']
```
melisabok, line 151: can you add an assertion checking the header contains the `**Authors**` in HTML?

melisabok:
there is one test failing because it is using the old password "1234"

melisabok (approved):

Issue #1589, opened 2023-03-02
Venue: add "Publication Chair" as reader of all the accepted papers during the post decision stage


melisabok:
In order to solve the permission issue of publication chairs having access to the accepted papers, camera ready revisions and accepted authors, I proposed to include this new role when the post decision stage runs:

1. Add the "Publication Chair" group as reader of the accepted submission if they are not public.
2. Add the "Publication Chair" as writers of the Authors/Accepted group so they can message the authors.
3. Add the "Publication Chair" as readers of the request form and invitee of the "Submission Revision Stage" so they can configure the camera ready revision.
4. Create a "Publication Chair" webfield component that shows the accepted papers and if they submitted the camera ready revision or not.

This role can be optional in the request form, if it is selected then execute the steps described above.

PR #1680, opened 2023-05-09
Feat/ Venue: publication chair

celestemartinez:
Addresses #1589

Needs https://github.com/openreview/openreview-api-v1/pull/2906 and https://github.com/openreview/openreview-web/pull/1530


openreview/venue_request/venue_request.py
```diff
@@ -1449,14 +1455,16 @@ def setup_request_form(self):
                             'values-copied': [
                                 self.support_group.id,
                                 '{signatures}',
-                                '{content["program_chair_emails"]}'
+                                '{content["program_chair_emails"]}',
+                                '{content["publication_chair_email"]}'
```
melisabok, line 1459: can you debug the UI and see what happens when we post the note?

is it being sent a "null" value? if that is the case, we should check in view.js where we are getting the copied values and if the value is null then don't send it.

openreview/venue/webfield/publicationChairWebfield.js
```diff
@@ -0,0 +1,30 @@
+// Webfield component
```
celestemartinez, line 1: @haroldrubio Here I am adding the webfield, it only has an 'Accepted Submissions' tab, so if publication chairs are readers of accepted submissions they will see them here.

melisabok (approved):

melisabok:
Can I merge this?

celestemartinez:
> Can I merge this?

Yes, this is ready!

openreview/venue/group.py
```diff
@@ -433,6 +433,29 @@ def create_ethics_chairs_group(self):
                 ethics_chairs_group.web = content
                 self.post_group(ethics_chairs_group)
 
+    def create_publication_chair_group(self):
+        venue_id = self.venue_id
+
+        publication_chair_group_id = self.venue.get_committee_id('Publication_Chair')
```
melisabok, line 439: I think the group name should be in plural, there can be more than one publication chair.

PR #1683, opened 2023-05-11
first implementation of accept decision options

celestemartinez:


melisabok:
I think this is implemented in Emilia's PR, can we close it?

PR #1685, opened 2023-05-11
Name deletion: use super user signature to get hidden profiles

melisabok:


melisabok:
> the support group contains has the venues group as a transitive member

how does it happen?

There is a test here that fails if the signature is the Support group. Is there anything missing in the setup test?

melisabok:
Closing this PR, this is not really needed

PR #1693, opened 2023-05-15
Fix/ Venue: update revision invitations when PC uses PC_Revision

celestemartinez:


openreview/venue/process/pc_submission_revision_process.py
```diff
@@ -38,4 +38,19 @@ def process(client, edit, invitation):
...
                     members = submission_authors
                 )
+            )
+
+    invitation_invitations = [i for i in client.get_all_invitations(prefix=f'{venue_id}/-/', type='invitation') if i.is_active()]
+
+    for venue_invitation in invitation_invitations:
+        print('processing invitation: ', venue_invitation.id)
+        authorids = venue_invitation.edit['invitation'].get('edit', {}).get('note', {}).get('content', {}).get('authorids', {}).get('value', [])
+        if '${{4/id}/content/authorids/value}' in authorids:
```
celestemartinez, line 48: This should always be the path, correct? Or should I just check that `content/authorids/value` is in authorids[0]?

melisabok: I guess this should always be the path

melisabok (approved):

PR #1698, opened 2023-05-18
Fix/ Request form: limit error string to 200000 chars

### LINKED ISSUE #1690: Error stage status: limit the error string to less than 200k
That is the limit to post the error in the note.

celestemartinez:
Fixes #1690


melisabok (approved):
thanks 🙏

PR #1708, opened 2023-05-24
Chore: Update duplicate domains list

carlosmondra:


melisabok:
Can I use this PR to make the changes where we find duplicate domains for email addresses? like melisa@fb.com

tests/test_tools.py
```diff
@@ -353,11 +353,13 @@ def test_create_authorid_profiles(self, client):
     def test_subdomains(self):
         # ensure that two part top-level domains are handled appropriately
         # e.g. "edu.cn", "ac.uk"
-        assert openreview.tools.subdomains('mails.tsinghua.edu.cn') == ['mails.tsinghua.edu.cn', 'tsinghua.edu']
-        assert openreview.tools.subdomains('robots.ox.ac.uk') == ['oxford.ac.uk', 'robots.ox.ac.uk']
-        assert openreview.tools.subdomains('eng.ox.ac.uk') == ['eng.ox.ac.uk', 'oxford.ac.uk']
+        assert openreview.tools.subdomains('mails.tsinghua.edu.cn') == ['tsinghua.edu.cn']
```
melisabok, line 356: the logic to get the subdomains have changed here? I'm confused about the difference between

```
assert openreview.tools.subdomains('robots.ox.ac.uk') == ['oxford.ac.uk', 'robots.ox.ac.uk']

```

and

```
assert openreview.tools.subdomains('robots.ox.ac.uk') == ['ox.ac.uk']
```

carlosmondra: > the logic to get the subdomains have changed here?

No, the logic did not change. The difference is that the new list maps `robots.ox.ac.uk` and `oxford.ac.uk` to `ox.ac.uk`: 
https://api.openreview.net/settings/institutions?domain=ox.ac.uk

melisabok: it seems that previously the subdomains call was returning the parameter plus the subdomains and now that is not the case, I'm not sure why this is the difference and I'm worried that can affect the conflict detection. 

Can you explain why 

```
('cs.umass.edu') == ['cs.umass.edu', 'umass.edu']
```

now it is without cs domain?:

```
('cs.umass.edu') == ['umass.edu']
```

carlosmondra: If you feel uneasy about merging this PR. I would suggest changing how openreview-js loads duplicate domains and add the list that openreview-py currently has to avoid different results 😬

melisabok: I think the difference is in the file, the previous document contains only the domain mapping for duplicates like: fb.com and facebook.com.

The new file is mapping subdomains too:

```
  "umass.edu": "umass.edu",
  "cs.umass.edu": "umass.edu",
```

and this is not the same. But I'm not sure if this will affect the conflict detection.

carlosmondra: So, are you saying that `cs.umass.edu` should not map to `umass.edu`? How do we decide that? Should `robots.ox.ac.uk` have its own institution definition as well? Maybe we want to have a discussion about this because I thought that the important thing was to map all the domains and subdomains of an institution to the same institution.

carlosmondra: I do see your point and I agree, but if that's the case, then, the current file we use for duplicates has some errors too haha. For example:
```
 "cs.toronto": "toronto.edu",
  "toronto": "toronto.edu",
```

That's why I did it that way. What do we want to do then?

carlosmondra: Another example is this:

```
"yahoo-corp.jp": "yahoo-inc.com",
  "yahoo-inc.com": "yahoo-inc.com",
```
Why are we mapping this to the same? Maybe we should only have duplicate domains for a very small number of institutions that we know are actually the same.

melisabok: ohhhh no!

What we can do is to create the file from the mongo collection but removing the entries where:

- domain: domain (we don't really need this)

or 

- domain contains another domain that is already in the file. (I'm not sure how easy is to do this)

carlosmondra:
Closed in favor of this: https://github.com/openreview/openreview-js/pull/17

PR #1723, opened 2023-06-01
Let support users remove emails from profiles

melisabok:
- Create a note invitation to request the removal of an email in a profile.
- The process function of this invitation performs all the actions needed to remove the email from a profile.

@xkopenreview there are two options to implement this:

1. add a new tab in the moderation page where we can post a new note to delete the email
2. add the button next to each email in the profile that is only visible by openreview.net/Support users

I think option 2 is easier but I would like to see somewhere the list of all the requests.

### DIFF
--- a/openreview/profile/process/request_remove_email_process.py
+++ b/openreview/profile/process/request_remove_email_process.py
@@ -114,7 +114,7 @@ def process(client, note, invitation):
     if head_edges or tail_edges:
         client.rename_edges(email, profile.id)
         
-    print('Replace all the group members that contain the name to remove')
+    print('Replace all the group members that contain the email to remove')
     memberships = [ g for g in client.get_all_groups(member=email) if email in g.members ]
 
     anon_groups = [g for g in memberships if g.anonids == True ]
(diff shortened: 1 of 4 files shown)


xkopenreview:
so the source of email removal requests is still user sending an email to info right?

melisabok:
> so the source of email removal requests is still user sending an email to info right?

correct and then we will delete the email using this invitation

openreview/profile/process/request_remove_email_process.py
```diff
@@ -0,0 +1,158 @@
...
+        
+              
+
+    print('Rename all the edges')
+    head_edges = client.get_edges(head=email, limit=1)
+    tail_edges = client.get_edges(tail=email, limit=1)
+    if head_edges or tail_edges:
+        client.rename_edges(email, profile.id)
+        
+    print('Replace all the group members that contain the name to remove')
```
xkopenreview, line 117: should be email
same for ln136

melisabok: fixed

xkopenreview (approved):
i think the process function works fine

PR #1730, opened 2023-06-06
create users using the fullname instead of first and last name

melisabok:
Remove all the references to first, middle and last name.


carlosmondra (approved):

celestemartinez (approved):
Looks good!

PR #1734, opened 2023-06-13
fix/ Venue matching: expire Proposed_Assignment_Recruitment after assignments are deployed

### LINKED ISSUE #1732: Venue matching: expire Proposed_Assignment_Recruitment afer deployment
If a user is invited before assignments are deployed, they can still accept after deployment but the assignment is not made. 
We should expire the Proposed_Assignment_Recruitment invitation right before starting the deployment.

celestemartinez:
Fixes #1732

### DIFF
--- a/openreview/venue/matching.py
+++ b/openreview/venue/matching.py
@@ -1157,6 +1157,7 @@ class Matching(object):
         self.venue.invitation_builder.set_assignment_invitation(self.match_group.id)
         recruitment_invitation_id=self.venue.get_invitation_id('Proposed_Assignment_Recruitment', prefix=self.match_group.id)
         self.venue.invitation_builder.expire_invitation(recruitment_invitation_id)
+        self.venue.invitation_builder.expire_invitation(self.venue.get_assignment_id(self.match_group.id))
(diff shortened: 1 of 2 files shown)


openreview/venue/matching.py
```diff
@@ -1155,6 +1155,8 @@ def deploy_sac_assignments(self, assignment_title, overwrite):
     def deploy(self, assignment_title, overwrite=False, enable_reviewer_reassignment=False):
 
         self.venue.invitation_builder.set_assignment_invitation(self.match_group.id)
+        recruitment_invitation_id=self.venue.get_invitation_id('Proposed_Assignment_Recruitment', prefix=self.match_group.id)
+        self.venue.invitation_builder.expire_invitation(recruitment_invitation_id)
```
melisabok, line 1159: do you mind to move this line here?

```
self.venue.invitation_builder.expire_invitation(self.venue.get_assignment_id(self.match_group.id))
```

in case that someone is editing the proposed assignments while we are deploying the edges

melisabok (approved):

PR #1737, opened 2023-06-14
Fix/ Recruitment: update overlap committee if it has changed

### LINKED ISSUE #1731: Venue Recruitment: updating the invitation content
We only update the invitation content when the reduced load has changed, but even then we only repost the new reduced load. This means that the fields `overlap_committee_id` and `overlap_committee_name` are never updated, even if the PCs change this in the request form recruitment:

https://github.com/openreview/openreview-py/blob/c448f3cc285675879f6576abcc71ece9dd035ddb/openreview/venue/invitation.py#L769

celestemartinez:
Fixes #1731

### DIFF
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -741,7 +741,7 @@ class InvitationBuilder(object):
         invitation_id=venue.get_recruitment_id(venue.get_committee_id(name=committee_name))
         current_invitation=tools.get_invitation(self.client, id = invitation_id)
 
-        #if reduced_load hasn't change, no need to repost invitation
+        #if invitation does not exist, post it
         if not current_invitation:
             recruitment_invitation = Invitation(
                 id = invitation_id,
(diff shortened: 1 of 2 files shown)


openreview/venue/invitation.py
```diff
@@ -769,10 +769,14 @@ def set_recruitment_invitation(self, committee_name, options):
         else:
             print('current invitation', current_invitation.edit['note']['content'].get('reduced_load', {}))
             print('new invitation', reduced_load_dict)
-            if current_invitation.edit['note']['content'].get('reduced_load', {}) != reduced_load_dict:
-                print('update reduce load')
+            if ( current_invitation.edit['note']['content'].get('reduced_load', {}) != reduced_load_dict or
```
melisabok, line 772: can you split this in two different ifs?

if only one field has changed then I would like the edit to contain that single change instead of both. 

I would create the invitation object and set the value if each if sentence is true.

melisabok (approved):

PR #1745, opened 2023-06-16
Fix: Venue: post note edit if readers in content field changes

### LINKED ISSUE #1733: Invitation edit script: make sure the list of readers doesn't contain duplicates
We should make this change:

```
final_invitation_readers = list(set([note.signatures[0] if 'signatures' in r else r for r in invitation_readers]))
```

A PC can sign an Official Review and when releasing the review to the authors, the PC group appears twice

### LINKED ISSUE #1735: Venue: update note if readers of a content field changes
A venue is trying to update review notes and delete the readers of the rating field. Passing `{ "delete": True }`  works, but this does not update the reviews because the note readers have not changed. 

We should also update notes when the readers of content fields change.

celestemartinez:
Fixes #1733
Fixes #1735

### DIFF
--- a/openreview/stages/venue_stages.py
+++ b/openreview/stages/venue_stages.py
@@ -604,9 +604,11 @@ class ReviewStage(object):
 
     def get_signatures(self, conference, number):
         if self.allow_de_anonymization:
-            return '~.*|' + conference.get_program_chairs_id()
+            # return '~.*|' + conference.get_program_chairs_id()
+            return '~.*|'
 
-        return conference.get_anon_reviewer_id(number=number, anon_id='.*') + '|' +  conference.get_program_chairs_id()
+        # return conference.get_anon_reviewer_id(number=number, anon_id='.*') + '|' +  conference.get_program_chairs_id()
+        return conference.get_anon_reviewer_id(number=number, anon_id='.*')
     
     def get_content(self, api_version='2', conference=None):
 
--- a/openreview/stages/venue_stages.py
+++ b/openreview/stages/venue_stages.py
@@ -604,10 +604,8 @@ class ReviewStage(object):
 
     def get_signatures(self, conference, number):
         if self.allow_de_anonymization:
-            # return '~.*|' + conference.get_program_chairs_id()
             return '~.*|'
 
-        # return conference.get_anon_reviewer_id(number=number, anon_id='.*') + '|' +  conference.get_program_chairs_id()
         return conference.get_anon_reviewer_id(number=number, anon_id='.*')
     
     def get_content(self, api_version='2', conference=None):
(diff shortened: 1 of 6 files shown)


openreview/venue/process/invitation_edit_process.py
```diff
@@ -90,7 +101,8 @@ def update_note_readers(submission, paper_invitation):
                         note = openreview.api.Note(
                             id = note.id,
                             readers = final_invitation_readers,
-                            nonreaders = paper_invitation.edit['note'].get('nonreaders')
```
celestemartinez, line 93: @melisabok should I post the edit only with what is changing here too?
For example, if the readers of the note are not changing, should the edit not contain the readers?

melisabok: that's correct, we shouldn't pass the values that don't change.

openreview/venue/process/invitation_edit_process.py
```diff
@@ -78,20 +78,36 @@ def update_note_readers(submission, paper_invitation):
...
+                if note.content[key].get('readers', []) != invitation_readers:
+                    updated_content[key] = {
+                        'readers': invitation_readers if invitation_readers else { 'delete': True }
+                    }
+            return updated_content
+
         if type(invitation_readers) is list:
             for note in notes:
-                final_invitation_readers = [note.signatures[0] if 'signatures' in r else r for r in invitation_readers]
+                final_invitation_readers = list(set([note.signatures[0] if 'signatures' in r else r for r in invitation_readers]))
```
celestemartinez, line 93: Do you think this is still necessary if we don't allow PCs to sign using the PC group? The order of readers does not match the order in the tests now because of `set()`

openreview/stages/venue_stages.py
```diff
@@ -604,9 +604,11 @@ def get_nonreaders(self, conference, number):
 
     def get_signatures(self, conference, number):
         if self.allow_de_anonymization:
-            return '~.*|' + conference.get_program_chairs_id()
+            # return '~.*|' + conference.get_program_chairs_id()
```
melisabok, line 607: could you remove this line?

openreview/stages/venue_stages.py
```diff
@@ -604,9 +604,11 @@ def get_nonreaders(self, conference, number):
...
+            return '~.*|'
 
-        return conference.get_anon_reviewer_id(number=number, anon_id='.*') + '|' +  conference.get_program_chairs_id()
+        # return conference.get_anon_reviewer_id(number=number, anon_id='.*') + '|' +  conference.get_program_chairs_id()
```
melisabok, line 610: same here

melisabok (approved):

PR #1746, opened 2023-06-16
Feat: Use after keyword to get all notes

carlosmondra:


tests/test_tools.py
```diff
@@ -236,65 +236,6 @@ def post_note(number):
         assert notes
         assert len(notes) == 1333
 
-        notes = client.get_all_notes(invitation='GetAllNotes/-/Submission', limit=1200)
```
melisabok, line 239: why do you need to remove these tests?

melisabok, line 239: why do you need to remove these tests?

carlosmondra: Because it doesn't make sense to add a limit when you're getting ALL the notes. Also, the logic can become quite complicated.

melisabok: I see what you mean, it makes sense. 

The changes look good to me but I'm going to put this on hold until we have the `after` parameter deployed in both APIs.

carlosmondra: The parameter after has been deployed in both APIs.

melisabok (approved):
Thanks!

PR #1749, opened 2023-06-21
Fix/ Venue Withdrawl: check groups exist before sending email

### LINKED ISSUE #1689: Withdrawal reversion: make sure all the committee group exists before sending email notification
This is fixed in the NeurIPS 2023 venue.

celestemartinez:
Fixes #1689

### DIFF
--- a/openreview/venue/process/desk_rejected_submission_process.py
+++ b/openreview/venue/process/desk_rejected_submission_process.py
@@ -8,7 +8,14 @@ def process(client, edit, invitation):
     desk_reject_committee = domain.content['desk_reject_committee']['value']
     desk_rejection_name = domain.content['desk_rejection_name']['value']
     submission_name = domain.content['submission_name']['value']
-    authors_name = domain.content['authors_name']['value'] 
+    authors_name = domain.content['authors_name']['value']
+    reviewers_name = domain.content['reviewers_name']['value']
+    area_chairs_name = domain.content.get('area_chairs_name', {}).get('value')
+    senior_area_chairs_name = domain.content.get('senior_area_chairs_name', {}).get('value')
+    reviewers_id = domain.content['reviewers_id']['value']
+    area_chairs_id = domain.content.get('area_chairs_id', {}).get('value')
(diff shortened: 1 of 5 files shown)


celestemartinez:
I realized desk-rejections have this issue as well, I am fixing it!

openreview/venue/process/desk_rejected_submission_process.py
```diff
@@ -51,4 +49,32 @@ def process(client, edit, invitation):
         )
 
     print(f'Remove {paper_group_id}/{authors_name} from {venue_id}/{authors_name}')
-    client.remove_members_from_group(f'{venue_id}/{authors_name}', f'{paper_group_id}/{authors_name}')
\ No newline at end of file
+    client.remove_members_from_group(f'{venue_id}/{authors_name}', f'{paper_group_id}/{authors_name}')
+
+    recipients = [f'{paper_group_id}/{authors_name}']
```
celestemartinez, line 54: @melisabok here we were always emailing all committee for a desk-rejection; I changed it to emailing only the readers like we do in the withdrawal process function.

celestemartinez:
Should we use the `withdraw_committee`/`desk_reject_committee` to send emails in the reversal or should we just see who the readers of the notes are? I am not sure why we keep these variables.

openreview/venue/process/desk_rejected_submission_process.py
melisabok: I think we still need to check if the group exists. because the readers can be SubmissionX/Senior_Area_Chairs and the group hasn't been created yet

melisabok:
I think we still need to keep the committee variable. I would like to use the readers of the withdrawal note only but currently we can not decide the readers of the note during submission time. It has to be defined when the invitation is created and the readers of the submission may change....

openreview/venue/process/desk_rejected_submission_process.py
celestemartinez: yes! I am still checking if the groups exist

celestemartinez:
I have updated all process functions to use the variables `withdraw_committee/desk_reject_committee`

melisabok (approved):
Thanks!

PR #1750, opened 2023-06-21
Fix/Update selector of comment text area

xkopenreview:
the comment text area has the following class 
form-control TextareaWidget_textarea__{some random string}
in https://github.com/openreview/openreview-web/pull/1318

so update the selector to use a partial class match

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -129,7 +129,7 @@ class Helpers:
         elif comment:
             buttons[1].click()
             time.sleep(1)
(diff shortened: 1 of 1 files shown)


melisabok:
I don't think it is backward compatible. Could you please try getting the element using both strategies? if one is not found then try the other. Otherwise we can not merge this PR until the new note editor is merged into master

melisabok (approved):

PR #1754, opened 2023-06-23
Fix/ Venue Post Decision: add bibtex to submissions

celestemartinez:


openreview/venue_request/process/postSubmissionProcess.py
```diff
@@ -1,7 +1,7 @@
 def process(client, note, invitation):
     GROUP_PREFIX = ''
     SUPPORT_GROUP = GROUP_PREFIX + '/Support'
-    conference = openreview.helpers.get_conference(client, note.forum, SUPPORT_GROUP, setup=False)
+    conference = openreview.helpers.get_conference(client, note.forum, SUPPORT_GROUP, setup=True)
```
celestemartinez, line 4: @melisabok is there a reason why setup is `False` here? I noticed that if, for example, the submissions readers are changed to `everyone` using the post submission button, the content of the venue group is not updated (for example `public_submissions`)

Or should I just update `public_submissions` in the group content instead?

melisabok: I didn't want to setup the whole submission stage again but I'm realizing that is not happening in the `setup` function, the call to create submission stage is being done in the "deploy" and "revision" process functions, so you can change it to true.

openreview/venue/invitation.py
```diff
@@ -1357,6 +1357,16 @@ def set_withdrawal_invitation(self):
             },
             'venueid': {
                 'value': self.venue.get_withdrawn_submission_venue_id()
+            },
+            '_bibtex': {
```
melisabok, line 1361: I don't think you need to define the bibtex field, or are you getting an error?

celestemartinez: yes, I was getting a `ValidationError` I believe, so I checked the journal and saw it was defined there in some places. I can double check and see exactly what error I was getting.

celestemartinez: Screen Shot 2023-06-27 at 10 26 30 AM

celestemartinez:
This should be ready for review! Let me know if I missed anything

openreview/venue/process/post_submission_process.py
```diff
@@ -3,22 +3,41 @@ def process(client, invitation):
...
-        return    
+        return
 
     def post_submission_edit(submission):
 
+        updated_note = openreview.api.Note(
+            id=submission.id
+        )
+
+        if public and submission.odate is None:
```
melisabok, line 23: instead of using the `public` variable, can we check if the readers of the submission is everyone?

melisabok (approved):

PR #1757, opened 2023-06-27
test different widgets

melisabok:


melisabok:
you need to run API 2 using this branch: https://github.com/openreview/openreview-api/pull/188

tests/test_icml_conference.py
```diff
@@ -2205,18 +2205,18 @@ def test_review_stage(self, client, openreview_client, helpers):
...
-                                    "4: Borderline reject: Technically solid paper where reasons to reject, e.g., limited evaluation, outweigh reasons to accept, e.g., good evaluation. Please use sparingly.",
-                                    "3: Reject: For instance, a paper with technical flaws, weak evaluation, inadequate reproducibility and incompletely addressed ethical considerations.",
-                                    "2: Strong Reject: For instance, a paper with major technical flaws, and/or poor evaluation, limited impact, poor reproducibility and mostly unaddressed ethical considerations.",
-                                    "1: Very Strong Reject: For instance, a paper with trivial results or unaddressed ethical considerations"
+                                    { 'value': 10, 'description': "10: Award quality: Technically flawless paper with groundbreaking impact, with exceptionally strong evaluation, reproducibility, and resources, and no unaddressed ethical considerations." },
```
xkopenreview, line 2210: maybe the description here should remove "10:" and keep only the text
because this description will  be returned in details.presentation

in order to combine 10(rating) and "10: Award..." (presentation) with no duplication we will have to assume the format of description which i think is not ideal

melisabok: Yes. We discussed this with Carlos and I have to remove it. I hope it is clear for the user there is no number. I wonder if the drop-down can build the description joining both values.

xkopenreview: maybe items require another field to have a template string to indicate how to combine the value and description to be shown in note editor

openreview/journal/invitation.py
```diff
@@ -3768,7 +3838,11 @@ def set_solicit_review_invitation(self):
             'process': self.process_script,
             'preprocess': self.preprocess_script,
             'edit': {
-                'signatures': { 'param': { 'regex': f'~.*' }},
+                'signatures': { 
+                    'param': { 
+                        'items': [{ 'prefix': f'~.*' }]
```
xkopenreview, line 3843: i think it does not need f here

openreview/journal/invitation.py
```diff
@@ -4278,7 +4368,15 @@ def set_comment_invitation(self):
             'writers': [venue_id],
             'signatures': [venue_id],
             'edit': {
-                'signatures': { 'param': { 'regex': f"{editors_in_chief_id}|{self.journal.get_action_editors_id(number='${5/content/noteNumber/value}')}|{self.journal.get_reviewers_id(number='${5/content/noteNumber/value}', anon=True)}.*|{self.journal.get_authors_id(number='${5/content/noteNumber/value}')}" }},
+                'signatures': { 
+                    'param': { 
+                        'items': [ 
+                            { 'value': editors_in_chief_id, 'optional': True },
+                            { 'value': self.journal.get_action_editors_id(number='${7/content/noteNumber/value}'), 'optional': True },
+                            { 'prefix': self.journal.get_reviewers_id(number='${7/content/noteNumber/value}', anon=True), 'optional': True },
```
celestemartinez, line 4376: Should the reviewers anon id have '.*' at the end?

I do not understand why some appear as:
`{ 'prefix': f"{self.journal.get_reviewers_id(number='${7/content/noteNumber/value}', anon=True)}.*", 'optional': True },`
and some as:
`{ 'prefix': self.journal.get_reviewers_id(number='${7/content/noteNumber/value}', anon=True), 'optional': True }`

openreview/journal/invitation.py
melisabok: fixed

openreview/journal/invitation.py
carlosmondra: Just FYI: it shouldn't matter. The API removes the `.*`, but I think it's cleaner if it doesn't show up.

carlosmondra (approved):

celestemartinez (approved):

xkopenreview (approved):

enrubio (approved):

PR #1758, opened 2023-06-27
Feat/ Venue: allow SACs and ACs to desk-reject papers

celestemartinez:
SACs and ACs can be allowed to desk-reject papers. However, only PCs can revert a desk-rejection.

### DIFF
--- a/tests/test_emnlp_conference.py
+++ b/tests/test_emnlp_conference.py
@@ -1,3 +1,6 @@
+import csv
+import os
+import random
 import openreview
 import datetime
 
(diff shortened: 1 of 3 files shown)


tests/test_emnlp_conference.py
```diff
@@ -557,4 +560,195 @@ def test_submit_papers(self, test_client, client, openreview_client, helpers):
...
+                            'EMNLP/2023/Conference/Submission${3/content/noteNumber/value}/Senior_Area_Chairs',
+                            'EMNLP/2023/Conference/Submission${3/content/noteNumber/value}/Area_Chairs'
+                            ],
+                        'edit': {
+                            'signatures': {
+                                'param': {
+                                    'regex': 'EMNLP/2023/Conference/Submission${5/content/noteNumber/value}/Senior_Area_Chairs|EMNLP/2023/Conference/Submission${5/content/noteNumber/value}/Area_Chair_.*|EMNLP/2023/Conference/Program_Chairs'
+                                }
+                            },
+                            'note': {
```
celestemartinez, line 588: Maybe I should just remove the title from the content in the venue class. 

Also, not sure if it is worth it to add this option of allowing SACs and ACs to desk-reject to the venue request.

melisabok: I think it is a good idea to add a field in the request form asking who can desk reject papers and by the default is just the PCs.

melisabok (approved):

PR #1763, opened 2023-06-29
Feature/venue ethics stage

celestemartinez:
Uses https://github.com/openreview/openreview-web/pull/1497


tests/test_icml_conference.py
```diff
@@ -3083,6 +3085,224 @@ def test_delete_assignments(self, openreview_client, helpers):
...
+                            'param': {
+                                'type': 'string',
+                                'maxLength': 200000,
+                                'markdown': True,
+                                'input': 'textarea'
+                            }
+                        }
+                    }
+                },
+                'ethics_review_submissions': '1,5,6,7,8,100',
```
melisabok, line 3241: I think we should remove this field in the API 2 venues. can you try flagging the papers posting the note edit with the flag?

celestemartinez: This was added to the test already. If this is how we'd rather do it, I can remove this field from the ethics stage

PR #1765, opened 2023-06-29
Fix/PC console of non-anonymous venue

xkopenreview:
pc console of non-anonymous venues (not public) is not counting reviews correctly because reviewers are signing the review with profile id instead of the anonymous reviewer group id.

this pr should fix this issue by looking up the anonymous group id using the profile id in signature.


openreview/conference/templates/programchairWebfield.js
```diff
@@ -717,13 +718,23 @@ var buildSeniorAreaChairGroupMaps = function(noteNumbers, seniorAreaChairGroups)
...
 
   var reviewByAnonId = {};
 
   _.forEach(notes, function(n) {
-    var anonId = getNumberfromGroup(n.signatures[0], ANON_REVIEWER_NAME);
+    var anonId;
+    if (IS_NON_ANONYMOUS_VENUE) {
+      var reviewerProfileId = n.signatures[0];
+      anonId = getNumberfromGroup(
+        (anonReviewerGroups && anonReviewerGroups[submissionNumber] && anonReviewerGroups[submissionNumber][reviewerProfileId]) || '',
```
melisabok, line 731: can you pass `anonReviewerGroups[submissionNumber]` as a parameter directly? so you don't need to pass two variables?

openreview/conference/templates/programchairWebfield.js
```diff
@@ -159,7 +160,7 @@ var main = function() {
       officialReviews = officialReviews.concat(paperOfficialReviews);
       metaReviews = metaReviews.concat(paperMetaReviews);
       decisions = decisions.concat(decisions);
-      submission.details.reviews = getOfficialReviews(paperOfficialReviews);
+      getOfficialReviews(paperOfficialReviews, groups.anonReviewerGroups, submission.number);
```
melisabok, line 163: just pass one parameter only here: `groups.anonReviewerGroups[submissionNumber]`

openreview/conference/templates/programchairWebfield.js
```diff
@@ -717,13 +718,23 @@ var buildSeniorAreaChairGroupMaps = function(noteNumbers, seniorAreaChairGroups)
...
-var getOfficialReviews = function(notes) {
+var getOfficialReviews = function (notes, anonReviewerGroups, submissionNumber) {
   var ratingExp = /^(\d+): .*/;
 
   var reviewByAnonId = {};
 
   _.forEach(notes, function(n) {
-    var anonId = getNumberfromGroup(n.signatures[0], ANON_REVIEWER_NAME);
+    var anonId;
+    if (IS_NON_ANONYMOUS_VENUE) {
```
melisabok, line 728: instead of having a variable I would check if the signature is a tilde id:

```
anonId = getNumberfromGroup(n.signatures[0].startsWith('~') ? submissionAnonReviewerGroups[reviewerProfileId]) : n.signatures[0], ANON_REVIEWER_NAME);
```

xkopenreview: the variable is also used in other parts of the webfield.
the approach of having a local var and update it here when it first see a signature starts with ~ 
may fail if all existing reviews are signed using for example the pc group

melisabok (approved):

PR #1773, opened 2023-07-13
set SAC assignments based on the track

melisabok:
- Venue class readers the reviewer_roles, area_chair_roles and senior_area_chairs_roles from the request form. **All roles must have the same length if we want to split the permissions between the different groups**. Example: Track1_Area_Chairs can see Track1_Reviewers.
- For each venue role:
  - Create group with a webfield
  - Add the role to the recruitment invitation
  - Add the role to the setup matching invitation
- Venue class detects if the submission has tracks and use them to setup the matching for a specific role. 
- Venue class sends the track value when computing the affinity scores for a specific role. 
- Venue class sends the track value into the assignment configuration to run the paper matching with a subset of submissions(only the ones that match with the track value)
- Venue class support a new method `set_track_sac_assignments` that can be run to assign SAC to submissions. You need to pass a file that maps each track to a SAC role. (SAC roles must be defined in the request form). Optionally you can compute conflicts between the track SAC members and the track submissions. Optionally you can also assign ACs of the same track to the SACs of the same track. **Note: using this option, there is no need to run the SAC paper matching and assignment, set_track_sac_assignments replaces that.**
- Add tests

Pending for next iteration: parametrize edge browser urls so SACs and/or ACs can edit the reviewer assignments.


openreview/venue/webfield/reviewersWebfield.js
```diff
@@ -1,10 +1,11 @@
 // Webfield component
+const committee_name = entity.id.split('/').slice(-1)[0].replaceAll('_', ' ')
 return {
   component: 'ReviewerConsole',
   version: 1,
   properties: {
     header: {
-      title: 'Reviewer Console',
+      title: `${committee_name} Console`,
       instructions: '<div><p>some instructions</p></div>'
```
celestemartinez, line 9: Can we add some default instructions here:

``<p class="dark">This page provides information and status updates for ${domain.content.subtitle?.value}. It will be regularly updated as the conference progresses, so please check back frequently.</p>``

openreview/venue/group.py
```diff
@@ -340,62 +340,66 @@ def create_authors_group(self):
...
-                            readers=[venue_id, senior_area_chairs_id, area_chairs_id, reviewers_id],
-                            writers=[venue_id],
-                            signatures=[venue_id],
-                            signatories=[venue_id],
-                            members=[]
-                        )
+        for index, role in enumerate(self.venue.reviewer_roles):
+
+            reviewers_id = self.venue.get_committee_id(role)
+            area_chairs_id = self.venue.get_committee_id(self.venue.area_chair_roles[index]) if index <= len(self.venue.area_chair_roles) else self.venue.get_area_chairs_id()
```
celestemartinez, line 346: I do not think we are validating anywhere that the arrays for the roles should be the same length. In this case, if the reviewer_roles array is longer than the area_chair_roles one, we will get an IndexError. Perhaps we should do if `index < len(self.venue.area_chair_roles)` so that if there are more reviewer roles, the last one gets the overall AC group as readers

melisabok: You're right! I forgot to mention that. Roles must be synced in all the arrays but I will make the change.

openreview/venue/webfield/reviewersWebfield.js
melisabok: Done.

openreview/venue/group.py
melisabok: Done.

celestemartinez (approved):

racheljsmart (approved):

PR #1776, opened 2023-07-17
Venue Ethics Stage: add private optional ethics comments field

celestemartinez:


openreview/venue/invitation.py
```diff
@@ -2878,7 +2878,20 @@ def set_ethics_stage_invitation(self):
                                     'type': 'boolean',
                                     'const': True
                                 }
-                            }
+                            },
+                            'readers': [venue_id, self.venue.get_ethics_chairs_id() ,self.venue.get_ethics_reviewers_id('${{4/id}/number}')]
```
melisabok, line 2882: I checked with Sherry and she was fine if the flag was shown to all the readers of the note. 

I agree that authors shouldn't see the flag if the reviewers are not released to them yet but should we should the flag to the reviewers, AC, SACs too?

celestemartinez: Perhaps we should. I do not think we absolutely have to but we might as well

melisabok (approved):

openreview/venue/invitation.py
```diff
@@ -1037,6 +1037,32 @@ def set_official_comment_invitation(self):
...
+            if comment_stage.reader_selection:
+                comment_readers = {
+                    'param': {
+                        'enum': ['${7/content/noteReaders/value}']
+                    }
+                }
+            invitation.edit['invitation']['edit']['note']['readers'] = comment_readers
+
+            invitation.edit['invitation']['invitees'].append(self.venue.get_ethics_reviewers_id('${3/content/noteNumber/value}'))
+            invitation.edit['invitation']['edit']['signatures']['param']['regex'] += '|' + self.venue.get_ethics_reviewers_id('${5/content/noteNumber/value}')
```
celestemartinez, line 1064: @melisabok for comment invitees and signatures, I am always adding the SubmissionX/Ethics_Reviewers even when the paper is not flagged. These groups won't exist for not flagged papers so it shouldn't be an issue. Is this ok? It would be more complicated to have to do the same with invitees and sigantures as we are doing with readers (saving them in the invitation content and passing them when creating the paper comment invitations)

melisabok: ok, I think that's fine. 

I'm worried that a paper is flagged, assigned ethics reviewers and then un-flagged. But we un-flag a paper we are not doing anything in the process function, right?

celestemartinez: not yet. When we unflag it, I can expire the ethics review invitation and perhaps empty or delete the ethics reviewers group?

celestemartinez: Can we unflag a paper by deleting the `ethics_flag` field from the submission or should I do it by using Yes/No?

melisabok: I think with the value "No" is easier to unflag? what do you think?

If we remove the field then we need to remove the other one with the comment and from the UI is not so clear that deleting the data will unflag the paper

melisabok (approved):

celestemartinez:
This PR is almost ready, just two questions:
1. When a paper is unflagged, I am expiring the ethics review invitation and emptying the paper ethics reviewers group. Should I check if there is already an ethics review before doing all of this?
2. How should ethics chairs sign comments? At the moment, they have access to the comment invitation because they are part of the senior area chairs group

melisabok:
1. I think expiring the ethics review invitation is enough, I don't think we should empty the group. What about removing the ethics_reviewers group from the comment invitees? If that is too much, then I would just expire the invitation.
2. Can you add the Ethics_Chairs group in the regex so they can sign the comments as "Ethics Chairs"?

celestemartinez:
> 1. I think expiring the ethics review invitation is enough, I don't think we should empty the group. What about removing the ethics_reviewers group from the comment invitees? If that is too much, then I would just expire the invitation.

I am now removing the ethics reviewers as comment invitees. The problem is that if the comment stage is re-run after, they will be added back. 
> 2. Can you add the Ethics_Chairs group in the regex so they can sign the comments as "Ethics Chairs"?

Done

PR #1788, opened 2023-07-28
Test/ Neurips rebuttal

celestemartinez:


tests/test_neurips_conference_v2.py
```diff
@@ -1178,1648 +1173,1017 @@ def test_post_submission_stage(self, helpers, openreview_client, client, request
...
+                    replyto = review.id,
+                    content={
+                        'rebuttal': { 'value': 'This is another rebuttal reply to the same official review.' }
+                    }
+                )
+            )
 
-#         now = datetime.datetime.utcnow()
-#         due_date = now + datetime.timedelta(days=3)
+        # use custom stage to enable one author rebuttal per paper
```
celestemartinez, line 2112: @melisabok I think it's a better idea to use the custom stage to enable one rebuttal per paper. If we re-use the rebuttal stage, then all rebuttal invitations will have '/-/Rebuttal' as the super invitation and that might cause confusion.

celestemartinez:
Should the custom stage and the rebuttal stage allow ethics reviewers as readers?

tests/test_neurips_conference_v2.py
melisabok: ohhh I see, ok, let's use the custom stage then.

melisabok: there is no way to customize the name, right? because if you use another name then there will be two different super invitation. Anyway is fine with me.

melisabok:
> Should the custom stage and the rebuttal stage allow ethics reviewers as readers?

I think this is up to the PCs. Can we add them as part of the readers selection option in the stage?

tests/test_neurips_conference_v2.py
celestemartinez: There isn't a way to customize it from the request form, but the stage does have a `name` param, so perhaps I should just add it to the request form. This way they can change the name and use the same button.

melisabok (approved):

PR #1790, opened 2023-07-31
Fix/ Venue: allow multiple publication chairs

celestemartinez:


openreview/venue/venue.py
```diff
@@ -69,7 +69,7 @@ def __init__(self, client, venue_id, support_user):
         self.senior_area_chair_identity_readers = []
         self.automatic_reviewer_assignment = False
         self.decision_heading_map = {}
-        self.publication_chair = None
+        self.publication_chair = []
```
melisabok, line 72: can you use `self.publication_chairs` instead?

melisabok: why do you need this variable here?

celestemartinez: I think the most important place I use it is in the post decision to decide if they should be added as readers or not. I can try removing it but I would have to pass this as a parameter in a couple of places I believe

melisabok: But you need to the group id only, right?

Not the email addresses?

I agree we should have a setting to indicate the venue has publication chairs and it can be stored in the venue.content object, I think saving the publication_chairs_id field is enough.

openreview/venue/venue.py
```diff
@@ -693,7 +693,7 @@ def update_note(submission):
                         decision_note = reply
                         break
             note_accepted = decision_note and 'Accept' in decision_note['content']['decision']['value']
-            submission_readers = self.submission_stage.get_readers(self, submission.number, decision_note['content']['decision']['value'] if decision_note else None, add_publication_chair=self.publication_chair)
+            submission_readers = self.submission_stage.get_readers(self, submission.number, decision_note['content']['decision']['value'] if decision_note else None, add_publication_chair=self.publication_chairs)
```
melisabok, line 696: Instead of adding a new parameter, the function get_readers can figure out if they need to add the publication chairs or note based on decision note and `conference.use_publication_chairs` or `conference.publication_chairs_id`

celestemartinez: done

melisabok (approved):

PR #1794, opened 2023-08-02
Fix: Venue: replace signatures in field readers

celestemartinez:


tests/test_venue_request_v2.py
```diff
@@ -1536,7 +1536,7 @@ def test_release_reviews_to_authors(self, client, test_client, selenium, request
                         },
                         'readers': [
                             "V2.cc/2030/Conference",
-                            "V2.cc/2030/Conference/Submission${7/content/noteNumber/value}/Reviewers"
```
melisabok, line 1539: I don't understand why this one work and the other doesn't. 

Anyway, can you keep both are readers? I want to make sure both cases work.

celestemartinez: This one is already replaced in the invitation, so when we get the readers from the invitation it already has the note number

melisabok (approved):

PR #1795, opened 2023-08-02
Track Paper Matching: configure edge browser urls based on committee role

melisabok:
- Parametrize the edge invitations using the track groups instead of the general committee groups.
- Parametrize the edge browser urls in the SAC and AC consoles to use the track groups.

### DIFF
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -2088,18 +2088,30 @@ class InvitationBuilder(object):
         venue = self.venue
         venue_id = venue.get_id()
         assignment_invitation_id = venue.get_assignment_id(committee_id, deployed=True)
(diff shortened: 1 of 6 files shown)


racheljsmart:
I've only checked the AC consoles and edge invitations. The invitations and edge browser URL are using the track groups.

racheljsmart:
You're right SACs are also included in the test, I'm not sure why I thought they weren't. It looks good, SACs are able to modify assignments. Are SACs only able to use the Invite Assignment and not Assign?

melisabok:
After the assignments are deployed. By default we only let users to invite emergency reviewers but we can change it if the PCs decide something different. 

It is just changing the url in the SAC console.

racheljsmart (approved):

celestemartinez (approved):

PR #1803, opened 2023-08-07
Anon Groups: get tilde id members when anonids are present

melisabok:
This [PR](https://github.com/openreview/openreview-api/pull/245) is starting to save the anon group ids as members of the SubmissionX/Reviewers group. 

Here we are changing the client object to get the deanonymized members and anonymized members in the same group object. This way all the code in openreview-py will keep working as always and no need to change any code in the Venue/Journal classes.

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -246,7 +246,7 @@ class OpenReviewClient(object):
         response = self.__handle_response(response)
         return response.json()
 
-    def get_group(self, id, anonids=False):
+    def get_group(self, id):
         """
         Get a single Group by id if available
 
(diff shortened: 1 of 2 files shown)


openreview/api/client.py
```diff
@@ -2813,3 +2835,13 @@ def post(self, client):
         client.post_group(self)
 
 
+    def transform_to_anon_ids(self, elements):
+        if self.anonids:
+            for index, element in enumerate(elements):
+                if element in self.members and self.anon_members:
```
carlosmondra, line 2841: Is it necessary to initialize self.members and self.anon_members to empty arrays in the constructor?

melisabok: I think you are right, I need to initialize them

celestemartinez (approved):

PR #1804, opened 2023-08-07
Venue: allow SACs to flag papers for ethics review

celestemartinez:


openreview/venue/venue.py
```diff
@@ -909,6 +911,8 @@ def set_track_sac_assignments(self, track_sac_file, conflict_policy=None, confli
         print('Builiding ac group')
         self.client.add_members_to_group(self.get_area_chairs_id(), all_acs)
 
+    def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
```
melisabok, line 914: this function should check if the ethics stage is enabled otherwise I get a none exception when trying to get the name of the stage

PR #1810, opened 2023-08-14
Fix: Get the correct anonymity for bibtex

carlosmondra:


openreview/venue/process/post_submission_process.py
```diff
@@ -29,7 +29,7 @@ def post_submission_edit(submission):
                         year=str(datetime.datetime.utcnow().year),
                         url_forum=submission.forum,
                         paper_status='under review',
-                        anonymous='readers' in submission.content['authors']
+                        anonymous='readers' in submission.content['authors'] or 'readers' in invitation.edit.get('note', {}).get('content', {}).get('authors', {})
```
melisabok, line 32: I don't understand which issue is fixing this.

carlosmondra: The problem happened in a venue that wanted to set the readers of the Note to 'everyone', but hide the authors and authorids field by setting a readers field.

The problem is that because the authors field when posting the change didn't have a readers field in the Submission, the process function "thinks" that the note is non-anonymous and adds the authors' names to the bibtex even if the post edit hides them by adding a readers field. So the fix is to check if the readers is a constant in the invitation that will hide authors to avoid revealing them in the bibtex.

melisabok: So this happens when you want to release the submission to the public and hide the authors identities at the same time using the same edit. 

Could you add a test to reproduce this issue? I think this is a common case for ICLR venues. I can help writing the test if you need.

carlosmondra: Yes

carlosmondra:
I don't know how to create a test for this one. Should I create a new file with setting a venue from scratch or should I add it to one of the existing test files?

melisabok:
I think we need to create a new test file to test the ICLR workflow. 

1. Setup ICLR venue
2. Submit a few submissions
3. Run post submission stage or wait until the deadline is over. Check that submissions are public, author are anonymized and the bibtex entry says "anonymous" instead of the author's names

tests/test_journal.py
```diff
@@ -807,7 +807,7 @@ def test_submission(self, journal, openreview_client, test_client, helpers):
         ))
 
          # wait for process function delay (5 seconds) and check email has been sent
-        time.sleep(6)
+        helpers.await_queue(jobNames=set(['emailQueueStatus', 'gatherEmailsQueueStatus', 'jsEntityQueueStatus', 'pyEntityQueueStatus']), api_version=2)
```
melisabok, line 810: I think here we should wait until the process function associated with the edge is done, I can fix it

melisabok:
I could reproduce the issue and added a test to confirm the issue is fixed!

tests/conftest.py
```diff
@@ -57,16 +57,32 @@ def get_user(email):
...
+                'gatherEmailsQueueStatus',
+                'jsDateProcessQueueStatus',
+                'pyDateProcessQueueStatus',
+                'fileUploaderQueueStatus',
+                'fileDeletionQueueStatus'
+            ])
         if super_client is None:
-            super_client = openreview.Client(baseurl='http://localhost:3000', username='openreview.net', password=Helpers.strong_password)
-            assert super_client is not None, 'Super Client is none'
+            if api_version == 1:
```
melisabok, line 73: I don't think this is needed, you can pass the api 2 client to this function, that is what I usually do.

melisabok: Anyway I think I fixed the tests without using these changes, can I revert them?

carlosmondra: Sure, if they are not needed we can revert them.

melisabok:
changes are reverted so it should be ready to be merged

celestemartinez (approved):

PR #1811, opened 2023-08-15
Fix/recruitment curly brace validation

### LINKED ISSUE #1594: Validation: add validation for curly braces in Recruitment
PCs often leave out the double curly braces when using Recruitment or sending Recruitment Reminders. Add validation for this.

enrubio:
- WIP to resolve #1594.

This PR adds a preprocess function to validate the format of the parenthesized tokens used in Recruitment and Recruitment Reminders. Examples of tokens we support are: `fullname`, `invitation_url`, `contact_info`, etc.

The purpose is to:
1. Not allow users to put double curly braces around non-tokens. E.g. `{{cat}}` is not valid.
2. Check that tokens are surrounded by double curly braces.

Examples of invalid tokens:
`{fullname}`
`{{{fullname}`
`fullname}`

Examples of valid tokens:
`{{fullname}}`
`{{{fullname}}}`
`fullname`*

*Because `fullname` may also be used as a plain word in emails, this shouldn't raise a curly brace error. To allow this, the validation only checks the format of tokens that have at least 1 curly brace around it. If there are none, the token will be printed plainly. This also means that if in an email a user wrote `invitation_url` with no brackets, it would not get caught by the validation. This is a potential problem to fix.


tests/test_venue_request_v2.py
```diff
@@ -550,7 +550,19 @@ def test_venue_recruitment(self, client, test_client, selenium, request_page, ve
             readers=['{}/Program_Chairs'.format(venue['venue_id']), venue['support_group_id']],
             signatures=['~SomeFirstName_User1'],
             writers=[]
-        ))
+        )
+
+        with pytest.raises(openreview.OpenReviewException, match=r'Invalid token: .* does not exist.'):
```
carlosmondra, line 555: Why do you use a .* here instead of the actual string that makes the validation fail?

openreview/venue_request/process/recruitment_pre_process.js
```diff
@@ -0,0 +1,32 @@
+function(){
+    // V1 invitation
+
+    const tokens = ['fullname', 'invitee_role', 'invitation_url', 'contact_info', 'reviewer_name', 'accept_url', 'decline_url'];
+    const fields = ['invitation_email_subject', 'invitation_email_content', 'accepted_email_template'];
+
+    
+    for (let field of fields) {
```
carlosmondra, line 8: This can be `const`
```suggestion
    for (const field of fields) {
```

openreview/venue_request/process/recruitment_pre_process.js
```diff
@@ -0,0 +1,32 @@
...
+            let match;
+            while ((match = regex.exec(note.content[field])) !== null) {
+                parenthesized_token = match[1];
+                if (!tokens.includes(parenthesized_token)) {
+                    done(`Invalid token: {{${parenthesized_token}}} does not exist.`);
+                }
+            }
+
+            // Check for tokens that don't have double curly braces, raise an error.
+            for (let token of tokens) {
```
carlosmondra, line 22: This can be `const`
```suggestion
            for (const token of tokens) {
```

openreview/venue_request/process/recruitment_pre_process.js
```diff
@@ -0,0 +1,32 @@
...
+
+    const tokens = ['fullname', 'invitee_role', 'invitation_url', 'contact_info', 'reviewer_name', 'accept_url', 'decline_url'];
+    const fields = ['invitation_email_subject', 'invitation_email_content', 'accepted_email_template'];
+
+    
+    for (let field of fields) {
+        if (field in note.content) {
+            // Find all words wrapped in double curly braces. If it's not a token, raise an error.
+            let regex = /{{([^{}]+)}}/g;
+            let parenthesized_token = '';
```
carlosmondra, line 12: Let's use camel case :)
```suggestion
            let parenthesizedToken = '';
```

carlosmondra:
Thanks, @enrubio ! It's coming out great. I think just a few more tweaks and it will be ready 😬

tests/test_venue_request_v2.py
enrubio: This is because a variable is used in the error message to display what the invalid token is.

https://github.com/openreview/openreview-py/blob/bbb1b017ee936a7ce90c8d72c6629fb965882c29/openreview/venue_request/process/recruitment_pre_process.js#L16-L18

openreview/venue_request/process/recruitment_pre_process.js
```diff
@@ -0,0 +1,32 @@
+function(){
+    // V1 invitation
+
+    const tokens = ['fullname', 'invitee_role', 'invitation_url', 'contact_info', 'reviewer_name', 'accept_url', 'decline_url'];
+    const fields = ['invitation_email_subject', 'invitation_email_content', 'accepted_email_template'];
```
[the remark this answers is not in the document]
enrubio: Okay good to know, I'll change my settings

enrubio:
Thanks for taking a look Carlos!

tests/test_venue_request_v2.py
carlosmondra: Can you know the value of the token at test time?

enrubio: I see what you mean now, I made the change.

https://github.com/openreview/openreview-py/blob/45be3dc8de183e2807aadaad98fe08455a3bc286/tests/test_venue_request_v2.py#L555-L561

melisabok:
@enrubio is this Pr ready to be merged?

tests/test_venue_request_v2.py
```diff
@@ -550,7 +550,19 @@ def test_venue_recruitment(self, client, test_client, selenium, request_page, ve
...
+        with pytest.raises(openreview.OpenReviewException, match=r'Invalid token: {{serve}} does not exist.'):
+           recruitment_note=test_client.post_note(recruitment_note)
+
+        recruitment_note.content['invitation_email_subject'] = '[' + venue['request_form_note'].content['Abbreviated Venue Name'] + '] Invitation to serve as {invitee_role'
+
+        with pytest.raises(openreview.OpenReviewException, match=r'Invalid token: {invitee_role. Tokens must be wrapped in double curly braces.'):
+           recruitment_note=test_client.post_note(recruitment_note)
+                                                               
+        recruitment_note.content['invitation_email_subject'] = '[' + venue['request_form_note'].content['Abbreviated Venue Name'] + '] Invitation to serve as {{invitee_role}}'
+        recruitment_note = test_client.post_note(recruitment_note)
```
melisabok, line 564: could you add a test where `{{contact_info}}` is in the subject? I believe the validation is passing but how is the email subject being formatted? is contact_info being replaced or not?

enrubio: Currently this function doesn't check for where these tokens are being used, it just checks the braces. From what I see in the code about how the replacements are being done, you're right, I don't think `{{contact_info}}` would get replaced in the subject. I can add this validation.

melisabok: I recommend having a dictionary to represent which tokens are being replaced for each field name and then to the validation based on that:

```
{
"invitation_email_subject": ["invitee_role"],
"invitation_email_content": ['fullname', 'invitee_role', 'invitation_url', 'contact_info', 'reviewer_name']
"accepted_email_template": ....
}
```

'accept_url', 'decline_url' are being deprecated so I wouldn't validate those.

openreview/venue_request/process/recruitment_pre_process.js
```diff
@@ -0,0 +1,34 @@
+function(){
+    // V1 invitation
+
+    const fieldTokens = {
```
melisabok, line 4: to get the request form:

```
const notes = await client.get_notes({id: note.forum})
const request_forum = notes && notes[0]
if (request_forum.content.api_version == 1) {
   //skip validation for api 1 venues
    return
}
```

enrubio: Fixed

openreview/venue_request/process/deployProcess.py
```diff
@@ -138,6 +139,7 @@ def process(client, note, invitation):
         id = SUPPORT_GROUP + '/-/Request' + str(forum.number) + '/Recruitment',
         super = SUPPORT_GROUP + '/-/Recruitment',
         invitees = readers,
+        preprocess=None,
```
melisabok, line 142: no need of this line, you can remove it

melisabok (approved):
changes looks good to me!

celestemartinez (approved):

PR #1812, opened 2023-08-16
Fix email to edit.tauthor

### LINKED ISSUE #1791: If edit tauthor is OpenReview.net, do not send email
In the submission process function, if the `edit.tauthor` is OpenReview.net, recipients will be empty and we will get an error: https://api2.openreview.net/logs/process?id=Co3InHa5LN

We should check if this is the case, and not send the email.

https://github.com/openreview/openreview-py/blob/44ab347d4dd6f2e8585cc45da47e483c18e3d18f/openreview/venue/process/submission_process.py#L43

enrubio:
- Fixes #1791


openreview/venue/process/submission_process.py
```diff
@@ -37,11 +37,12 @@ def process(client, edit, invitation):
 To view your submission, click here: https://openreview.net/forum?id={note.forum}'''
     
     #send tauthor email
-    client.post_message(
-        subject=author_subject,
-        message=author_message,
-        recipients=[edit.tauthor] if edit.tauthor != 'OpenReview.net' else []
-    )
+    if edit.tauthor != 'OpenReview.net':
```
melisabok, line 40:
```suggestion
    if edit.tauthor.lower() != 'openreview.net':
```

melisabok: super user for the test environment is `openreview.net` in lowercase. Using this change you can add a test and verify that no emails is being sent then the submission is being posted by the super user. Could you add this test?

melisabok (approved):
👍

celestemartinez (approved):
Looks good!

PR #1814, opened 2023-08-17
Allows PCs to select Accept with Reduced Load

racheljsmart:
Adds allowAcceptWithReducedLoad: false


tests/test_venue_request_v2.py
```diff
@@ -585,21 +586,21 @@ def test_venue_recruitment(self, client, test_client, selenium, request_page, ve
 
         invalid_accept_url = re.search('https://.*\n', messages[0]['content']['text']).group(0).replace('https://openreview.net', 'http://localhost:3030').replace('&amp;', '&')[:-1].replace('user=reviewer_candidate2_v2%40mail.com', 'user=reviewer_candidate2_v1%40mail.com')
         print(invalid_accept_url)
-        helpers.respond_invitation(selenium, request_page, invalid_accept_url, accept=True)
+        helpers.respond_invitation(selenium, request_page, invalid_accept_url, accept=True, allow_accept_with_reduced_load=True)
```
melisabok, line 589: I would prefer to add a new test instead of editing an existing one. 

When you pass `allow_accept_with_reduced_load=True` don't you need to pass the quota value?

I think we don't need to add a new parameter, I think the function respond_invitation should handle this automatically: 

if accept is True and quota is not None then we should try to find the button to accept with reduced load. If the button is not found then we should decline the invitation and accept the invitation with reduced load.

enrubio: Ah okay, I see what you mean. I didn't see `quota` being passed in anywhere to `respond_invitation()` so I didn't fully understand what it was, but I understand now. And basically all this test was doing was checking if the Accept with Reduced Load button gets displayed. I'll update the tests.

melisabok: `quota` is the reduced_load index that you want to select for that reviewer that is accepting the invitation. Previously users could only select the reduced load rejecting the invitation first and then accepting the invitation with a quota. That is what this code does. Now that we want to support the acceptance with the reduce load without having to reject first, this function should check if this option is available (check with selenium if the button exists) and then accept the invitation selecting the reduce load, if the button is not found then you do the existing behavior (reject first and then accept with reduce load)

openreview/venue/webfield/recruitResponseWebfield.js
```diff
@@ -29,6 +29,7 @@ return {
     reducedLoadMessage: entity.edit.note.content.reduced_load && `
 If you chose to decline the invitation because the paper load is too high, you can request to reduce your load.
 You can request a reduced reviewer load below:
-`
+`,
+    allowAcceptWithReducedLoad: entity.content.allow_accept_with_reduced_load.value
```
melisabok, line 33:
```suggestion
    allowAcceptWithReducedLoad: entity.content.allow_accept_with_reduced_load?.value
```

tests/conftest.py
```diff
@@ -111,23 +111,40 @@ def respond_invitation(selenium, request_page, url, accept, quota=None, comment=
         container = selenium.find_element(By.CLASS_NAME, 'note_editor')
 
         buttons = container.find_elements(By.TAG_NAME, "button")
-        assert len(buttons) == 2
+        
+        if accept and quota:
```
melisabok, line 115: I do think you need to check the length of the buttons in order to know if the invitation allows accept with reduce quota. 

Previous tests send accept=true and quota=true and that means, we need to decline first and then accept with reduce quota.

The logic would be:

```
if accept and quota:

     if len(buttons) == 3:
         buttons[1].click() ## accept with reduced quota, new implemented button that is shown in the middle
         ... select quota, etc

     if len(buttons) == 2:
         buttons[1].click(). ## decline first
         ...continue with the original code....


```

tests/test_venue_request_v2.py
```diff
@@ -754,11 +754,11 @@ def test_venue_AC_recruitment_(self, client, test_client, openreview_client, sel
         #accept AC invitation after having accepted reviewer invitation
         invitation_url = re.search('https://.*\n', messages[2]['content']['text']).group(0).replace('https://openreview.net', 'http://localhost:3030')[:-1]
         print('invitation_url', invitation_url)
-        helpers.respond_invitation(selenium, request_page, invitation_url, accept=True)
+        helpers.respond_invitation(selenium, request_page, invitation_url, accept=True, quota=1)
```
melisabok, line 757: instead of modifying a new test, can you add a new one?

tests/conftest.py
enrubio: Yeah I saw the previous tests pass `accept=true` even though they're actually declining with reduced load. What do you think about changing them to `accept=false`? I feel it would make sense since they're declining with quota, but it would mean changing previous tests.

tests/test_venue_request_v2.py
enrubio: oh yes, just realized I forgot to remove this part

melisabok (approved):
it looks great!

melisabok:
@celestemartinez please take a look and feel free to merge it.

celestemartinez (approved):

PR #1819, opened 2023-08-25
Feat: Set impersonators for API v2 venues/journals

carlosmondra:


openreview/venue/group.py
```diff
@@ -561,3 +561,9 @@ def create_paper_group(submission):
 
         if create_paper_groups:
             tools.concurrent_requests(create_paper_group, venue.get_submissions(sort='number:asc'), desc='Creating paper groups')
+
+    def set_impersonators(self, impersonators):
```
melisabok, line 565: do you mind to add a test where a PC client can impersonate another user?

carlosmondra: Yeah, no problem. I wanted to ask you if you think this is the right place to have this function?

melisabok: I think this place is ok for now.

carlosmondra:
I have added a test

melisabok (approved):

celestemartinez (approved):

PR #1822, opened 2023-08-29
Fix regex for submission keywords

### LINKED ISSUE #1799: Submission keywords field: don't let empty string to be allowed in the keywords array
Currently the default submission form defines the keyword field as: https://github.com/openreview/openreview-py/blob/master/openreview/stages/default_content.py#L467

This allows the value `[""]` to be valid for the API validation and we shouldn't allow it. My recommendation is to change the regex definition to be `.+` as non empty strings.

enrubio:
- Resolves #1799

### DIFF
--- a/openreview/stages/default_content.py
+++ b/openreview/stages/default_content.py
@@ -464,7 +464,7 @@ submission_v2 = {
         'value': {
             'param': {
                 'type': 'string[]',
-                'regex': '(^$)|[^;,\\n]+(,[^,\\n]+)*'
+                'regex': '.+'
             }
         }
     },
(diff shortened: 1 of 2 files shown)


openreview/stages/default_content.py
```diff
@@ -464,7 +464,7 @@
         'value': {
             'param': {
                 'type': 'string[]',
-                'regex': '(^$)|[^;,\\n]+(,[^,\\n]+)*'
+                'regex': '.+'
```
melisabok, line 467: Could you please add a test reproducing the issue?

Try posting a submission where the value "keywords" is `[""]` and you should expect an exception.

enrubio: Okay, I've added a test for this under `test_venue_matching_setup`.

melisabok (approved):
💯

PR #1831, opened 2023-09-12
Fix/homepage dates

melisabok:


openreview/conference/helpers.py
```diff
@@ -361,6 +361,7 @@ def get_submission_stage(request_forum, venue):
     abstract_due_date_str = ''
     submission_second_due_date = request_forum.content.get('Submission Deadline', '').strip()
     if submission_second_due_date:
+        submission_deadline_str = submission_second_due_date.strftime('%b %d %Y') + ' UTC-0'
```
melisabok, line 364: Could we take the opportunity to change the format here and show the time too? the string should be "%b %d %Y %I:%M%p" like the abstract string. You would need to fix the test too.

tests/test_iclr_conference_v2.py
```diff
@@ -90,7 +92,9 @@ def test_create_conference(self, client, openreview_client, helpers, profile_man
 
         helpers.await_queue()
 
-        assert openreview_client.get_group('ICLR.cc/2024/Conference')
+        venue_group = openreview_client.get_group('ICLR.cc/2024/Conference')
+        assert venue_group
+        assert venue_group.content['date']['value'] == f'Abstract Registration: {abstract_date.strftime("%b %d %Y")} 12:00AM UTC-0, Submission Deadline: {due_date.strftime("%b %d %Y")} UTC-0'
```
melisabok, line 97: make sure you fix the test, it expects to have 12:00AM in the submission deadline too.

enrubio: Oh yes, just realized this. Fixed now.

celestemartinez (approved):

PR #1839, opened 2023-09-18
Journal consoles: add domain parameter

melisabok:
New API feature lets you pass the domain value to filter out the transitive members and reduce the query time.


celestemartinez (approved):

melisabok:
Please don't merge it because the API must be deployed first.

PR #1840, opened 2023-09-20
Feature/group recruitment

celestemartinez:


tests/test_group_recruitment.py
```diff
@@ -0,0 +1,118 @@
...
+
+Program Chairs
+''' },
+                        'recruitment_subject': { 'value': '[V 24] Invitation to serve as Reviewer' },
+                        'allow_overlap': { 'value': False }
+                    }
+                )
+            )
+        
+        invitee_details = '''reviewer1@venue.cc, Reviewer VenueOne\nreviewer2@venue.cc, Reviewer VenueTwo\n~Reviewer_VenueThree1'''
```
zbialecki, line 99: So essentially you are expecting the users who are managing reviewer recruitment to provide a CSV file of users they want to invite. Do all the users in this list have to have OR profiles or is that optional? What about if the PC wants to remove someone from the invited group?

openreview/venue/invitation.py
```diff
@@ -3095,4 +3095,100 @@ def set_SAC_ethics_flag_invitation(self, sac_ethics_flag_duedate=None):
...
+                            'param': {
+                                'type': 'string',
+                                'maxLength': 200000,
+                                'input': 'textarea',
+                                'optional': True,
+                                'markdown': True
+                            }
+                        }
+                    }
+                },
```
zbialecki, line 3131: Why not make the invitation modify group.members? This is what the group editor currently does when you add and remove members from a group.

celestemartinez: We do not want to add them directly to the /Invited group, in case the email or profile ID is not valid. There are also checks we do in the process function (like if the user was already invited to another role, for example.) The process function checks the user can be invited, sends the email and adds the user to the group

celestemartinez: @melisabok correct me if I am wrong, but this field should be just a textarea like we have in the venue request recruitment, right? The preprocess/process functions should check that they are valid profile IDs or emails.

openreview/venue/invitation.py
```diff
@@ -3095,4 +3095,100 @@ def set_SAC_ethics_flag_invitation(self, sac_ethics_flag_duedate=None):
...
+            process=self.get_process_content('process/group_recruitment_process.py'),
+            content={
+                'committee_name': { 'value': committee_name },
+                'official_committee_roles': { 'value': venue.get_committee_names()}
+            },
+            edit={
+                'signatures': [venue_id],
+                'readers': [venue_id],
+                'writers': [venue_id],
+                'content': {
```
celestemartinez, line 3119: @melisabok I am adding the different recruitment status to the content here, so I can add them to the recruitment edit. However, how can we differentiate between the fields that the PCs have to fill out (`inviteeDetails`) and the ones I am using to store other things (like the recruitment status)

openreview/venue/invitation.py
melisabok: Right, I don't think we need a profile search widget here, because the PCs will invite the reviewers in bulk so they will provide the list of emails. 

I think we could have another invitation that let them invite a reviewer one by one and use the profile search widget there.

openreview/venue/invitation.py
melisabok: can recruitment status be hidden in the note editor?

openreview/venue/invitation.py
zbialecki: OK, I guess that makes sense. 1 invitation for bulk add, 1 invitation for adding individuals and 1 invitation for removing individuals.

openreview/venue/invitation.py
```diff
@@ -3095,4 +3095,132 @@ def set_SAC_ethics_flag_invitation(self, sac_ethics_flag_duedate=None):
...
+    def set_group_recruitment_invitations(self, committee_name):
+
+        venue_id = self.venue_id
+        venue = self.venue
+
+        invitation = Invitation(id=venue.get_group_recruitment_id(committee_name),
+            invitees=[venue_id],
+            readers=[venue_id],
+            writers=[venue_id],
+            signatures=['~Super_User1'],
```
celestemartinez, line 3109: @melisabok I remember we spoke about using Super User for now, and that using the venue id should work too like you are using it for the journal.

If I am remebering correctly, `memberships = [g.id for g in client.get_groups(member=email, prefix=venue_id)]` would give an error only if the `email` group does not exist in OR. However, I am already checking this:
`memberships = [g.id for g in client.get_groups(member=email, prefix=venue_id)] if openreview.tools.get_group(client, email) else []` and I am still getting the following error with a profile ID that exists: 
`User Venue is not reader of ~Reviewer_VenueThree1`

melisabok (approved):

PR #1847, opened 2023-09-26
Venue/ Publication chairs: make sure authors accepted group is not overwritten

celestemartinez:


tests/test_icml_conference.py
```diff
@@ -4620,6 +4618,10 @@ def test_post_decision_stage(self, client, openreview_client, helpers):
 
         assert '_bibtex' in rejected_submissions[0].content and rejected_submissions[0].content['_bibtex']['value'] == valid_bibtex
 
+        #make sure all decision process functions have finished
+        for number in range(1, 101):
+            helpers.await_queue_edit(openreview_client, invitation=f'ICML.cc/2023/Conference/Submission{number}/-/Decision')
```
melisabok, line 4623: The function has a count that is 1 by default, you can pass the count instead of calling the same functions 100 times.

melisabok: Disregard my comment, I see it is a different invitation per note.

melisabok (approved):

PR #1848, opened 2023-09-27
Feat/add submission license field

enrubio:
This adds a submission_license field to the venue request form. The license is also added to the submission form content.

depends on https://github.com/openreview/openreview-api/pull/266


melisabok:
I think the tests are failing because we use the submission stage to setup the "Revision" invitations. We should set the license field only in the submission submission when the note is created. 

Given that the license field is going to be implemented here: https://github.com/openreview/openreview-api/pull/266, can you add the license field in the function `set_submission_invitation` instead of the SubmissionStage class? this way you don't have test errors.

melisabok:
This PR should fix the issue with dot missing in the dropdown: https://github.com/openreview/openreview-web/pull/1640

zbialecki:
I thought that we decided that notes would have a `license` field at the top level, outside of `content`.

melisabok:
> I thought that we decided that notes would have a license field at the top level, outside of content.

Yes, I think @enrubio hasn't pushed all the changes yet.

openreview/venue_request/venue_request.py
```diff
@@ -1423,17 +1423,23 @@ def setup_request_form(self):
                 'default': '2',
                 'order': 41
             },
+            'submission_license': {
+                'value-radio': ['CC BY 4.0', 'CC BY-SA 4.0', 'CC BY-NC 4.0', 'CC BY-ND 4.0', 'CC BY-NC-SA 4.0', 'CC BY-NC-ND 4.0', 'CC0 1.0'],
+                'description': 'Which license would you like to use for each submission? If you are unsure, we recommend "CC BY 4.0". If your license is not listed, please let us know and we can add it.',
```
melisabok, line 1428: @celestemartinez should we mention the link to the "Article License" in openreview term of use?

melisabok:
@enrubio can you remove the field license from the venue request revision invitation? I wouldn't want program chairs be editing the license once there are submissions in the system. You can add it to `remove_fields` in venue_request.py

melisabok:
When trying to edit a submission from the forum, I get the following error:

Screen Shot 2023-10-11 at 3 20 11 PM

@xkopenreview can you take a look?

openreview/venue_request/venue_request.py
celestemartinez: Yes! I don't think we can link to the specific section, right? We can have something like:

> Please refer to https://openreview.net/legal/terms for more information.

enrubio: Updated. This is what it looks like:

Screenshot 2023-10-11 at 4 22 49 PM

xkopenreview:
> When trying to edit a submission from the forum, I get the following error:
> 
> Screen Shot 2023-10-11 at 3 20 11 PM
> @xkopenreview can you take a look?

which venue is this?
i can't really get iclr_v2 which is modified in this pr to pass

enrubio:
@xkopenreview  Have you checked out Carlos' PR? https://github.com/openreview/openreview-api/pull/266

celestemartinez:
Looks good! I am not getting any errors in my local even when trying to edit a submission from the forum page. 

One question: why are we adding the license to the edit?

melisabok:
> One question: why are we adding the license to the edit?

Edits have their own license. Some venues may not want to release the edit history.

melisabok:
>  I am not getting any errors in my local even when trying to edit a submission from the forum page.

It still happens to me using the master version of openreview-web 🤸‍♀️

xkopenreview:
> > I am not getting any errors in my local even when trying to edit a submission from the forum page.
> 
> It still happens to me using the master version of openreview-web 🤸‍♀️

should be fixed by latest commit in web 1663
should it be in a separate pr?

melisabok:
> should be fixed by latest commit in web 1663
I can confirm the issue has been fixed.

openreview/venue_request/venue_request.py
```diff
@@ -1423,17 +1423,23 @@ def setup_request_form(self):
                 'default': '2',
                 'order': 41
             },
+            'submission_license': {
```
xkopenreview, line 1426: how does organizer specify that the submission license should be a dropdown/checkbox/radio instead of string?

melisabok:
@celestemartinez I think the submission license field should go closer to the submission stage fields and not at the bottom of the form, maybe after location and before the reviewer assignment? what do you think?

celestemartinez:
> @celestemartinez I think the submission license field should go closer to the submission stage fields and not at the bottom of the form, maybe after location and before the reviewer assignment? what do you think?

Yes! I also thought the same. I agree, right after location is the best spot.

enrubio:
Updated now, it definitely looks a lot better.

melisabok (approved):

xkopenreview (approved):

PR #1853, opened 2023-10-03
Check that invitee details is not empty

### LINKED ISSUE #1849: Recruitment: line with only a comma
A PC tried recruiting  a bunch of reviewers, but one line consisted of only a comma. This gave error because after we split at the commas, we assume the length >0.

https://github.com/openreview/openreview-py/blob/7c297c243415da1c781ce6b891032931fe71e81d/openreview/venue_request/process/recruitmentProcess.py#L37-L43

I think we can just ignore the line if len(details)==0.

enrubio:
- Resolves #1849


celestemartinez:
Looks good! Could you add a test in one of the existing test files to make sure you don't get an error and all other recruitment emails are sent?

melisabok (approved):

celestemartinez (approved):

PR #1854, opened 2023-10-03
Fix: PC/SAC/AC console: metaReviewRecommendationName var

celestemartinez:
Depends on https://github.com/openreview/openreview-web/pull/1646


xkopenreview:
default value of 'recommendation' has been added in the component
so when domain.meta_review_recommendation is undefined it will get the value 'recommendation'

melisabok (approved):

PR #1855, opened 2023-10-04
Journal process functions: get the note before checking tcdate and tmdate

melisabok:


openreview/venue/process/ethics_flag_process.py
```diff
@@ -8,7 +8,7 @@ def process(client, edit, invitation):
     review_name = domain.content.get('review_name', {}).get('value')
     ethics_review_name = domain.content.get('ethics_review_name', {}).get('value')
 
-    submission = client.get_note(edit.note.forum)
+    submission = client.get_note(edit.note.id)
```
celestemartinez, line 11: Any particular reason why this was changed? I believe both work but is id preferred?

celestemartinez (approved):

openreview/venue/process/ethics_flag_process.py
melisabok: new changes in the API 2 don't store the `forum` value inside of the edit. This is intended so all the fields of the edit are represented in the invitation.

PR #1871, opened 2023-10-23
Venue: author reorder and pub chairs fix

celestemartinez:
1. Show `submission_deadline_author_reorder` for PCs to select author reorder after the abstract deadline
2. Always create the publication chairs group and add to accepted papers. PCs can still use the request form to add members to this group or add them to the group directly.
3. Add Publication Chairs as readers of authors and authorids fields of accepted papers if author names have not been revealed


openreview/venue/venue.py
```diff
@@ -69,7 +69,6 @@ def __init__(self, client, venue_id, support_user):
         self.senior_area_chair_identity_readers = []
         self.automatic_reviewer_assignment = False
         self.decision_heading_map = {}
-        self.use_publication_chairs = False
```
melisabok, line 72: shouldn't we keep this property and ask in the request form if the venue has publication chairs. We don't want to create the publication chairs group for venues that don't have it

melisabok: I was thinking to add a question with "yes" or "no" asking if the venue has publication chairs but don't ask the emails because they might not now the emails at the moment of submitting the request form.

openreview/venue_request/venue_request.py
```diff
@@ -1134,11 +1134,21 @@ def setup_request_form(self):
                 'required': True,
                 'order': 6
             },
+            'publication_chairs': {
+                'description': 'Will your venue have Publication Chairs? The Publication Chairs will only have access to accepted submissions (including author names) and the author accepted group in order to email authors of accepted submissions.',
+                'value-radio': [
+                    'Yes, our venue has Publication Chairs',
+                    'No, our venue does not have Publication Chairs'
+                ],
+                'required': False,
```
melisabok, line 1143: can you set this field to be required true?

celestemartinez: Yep, wanted to make sure it would not break anything on the live site. I need to fix all the tests that use the request form.

openreview/venue_request/venue_request.py
```diff
@@ -1134,11 +1134,21 @@ def setup_request_form(self):
...
                 'order': 6
             },
+            'publication_chairs': {
+                'description': 'Will your venue have Publication Chairs? The Publication Chairs will only have access to accepted submissions (including author names) and the author accepted group in order to email authors of accepted submissions.',
+                'value-radio': [
+                    'Yes, our venue has Publication Chairs',
+                    'No, our venue does not have Publication Chairs'
+                ],
+                'required': True,
+                'default': 'No, our venue does not have Publication Chairs',
```
melisabok, line 1144: I think we shouldn't use a default value here so we force the PCs to reply this question. They can miss it and request for publication chairs later.

celestemartinez: Yes, you are right! This is epecially helpful when the field is added when the PCs use the Revision button, so we force them to pick an option. Changed!

melisabok (approved):

PR #1894, opened 2023-10-27
fix optional value based on public comments or not

melisabok:
- PC and assigned SAC shouldn't be mandatory if 'everyone' is present.

### DIFF
--- a/openreview/stages/venue_stages.py
+++ b/openreview/stages/venue_stages.py
@@ -942,16 +942,18 @@ class CommentStage(object):
 
     def get_readers(self, conference, number, api_version='1'):
 
-        readers = [{ 'value': conference.get_program_chairs_id(), 'optional': False }]
         if api_version == '2' and self.reader_selection:
-            if conference.use_senior_area_chairs and self.Readers.SENIOR_AREA_CHAIRS_ASSIGNED in self.readers:
-                readers.append({ 'value': conference.get_senior_area_chairs_id(number), 'optional': False })
 
-            if self.allow_public_comments or self.Readers.EVERYONE in self.readers:
+            readers  = []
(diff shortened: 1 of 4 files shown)


melisabok:
@carlosmondra when everyone is present in the list of available readers we can not make the PCs mandatory. I don't think the API is taking everyone as a possible mandatory PCs.

carlosmondra:
> @carlosmondra when everyone is present in the list of available readers we can not make the PCs mandatory. I don't think the API is taking everyone as a possible mandatory PCs.

I'm sorry, I don't understand what the issue is

melisabok:
> I'm sorry, I don't understand what the issue is

if we want the PCs to be always readers of the comment and the invitation allows 'everyone' we don't need to force the PCs group to be in the readers list. This means that PCs shouldn't be `optional: False` because everyone can satisfy the condition of PCs reading the comments. My question is if the API is allowing to set everyone only when the PCs are optional false.

carlosmondra:
> > I'm sorry, I don't understand what the issue is
> 
> if we want the PCs to be always readers of the comment and the invitation allows 'everyone' we don't need to force the PCs group to be in the readers list. This means that PCs shouldn't be `optional: False` because everyone can satisfy the condition of PCs reading the comments. My question is if the API is allowing to set everyone only when the PCs are optional false.

The API does not know whether a group is a subset of another. I think that could be very complicated. Unless I'm still misunderstanding the issue.

melisabok:
> EDIT: Or do you want that everyone has a special behavior? Example: If everyone is selected the rest of the mandatory fields are not needed.

I'm not sure about this, I would like to keep the API simple.

carlosmondra (approved):

celestemartinez (approved):

Issue #1873, opened 2023-10-23
Archive import should use API 2 invitations and groups


melisabok:
Replace `profileManagement.set_archive_invitations` to use API 2 invitations and groups.

PR #1895, opened 2023-10-30
use archive invitations in API 2

melisabok:
It fixes https://github.com/openreview/openreview-py/issues/1873

- Setup "Archive" venue and "Direct Upload" invitation using API 2.
- Archived submissions will have the following properties:

1. venue and venueid to recognizer them as articles. the venue is asked to the user to complete, the venueid is OpenReview.net/Archive
2. year to indicate the year when the submission was published. For conflict detection.
3. license

- There is a new function to filter the publications by cut_off_year, I saw there were different implementations for neurips and default policy. This function unifies it. 

@zbialecki this PR needs the functionality to call API 1 archived submissions.


openreview/profile/management.py
```diff
@@ -287,73 +295,157 @@ def set_archive_invitations(self):
...
+                            'year': {
+                                'order': 8,
+                                'description': 'Enter the year where the paper was published.',
+                                'value': {
+                                    'param': {
+                                        'type': 'integer',
+                                        'optional': True
+                                    }
+                                }
+                            },
```
carlosmondra, line 434: We are going to ask for this here instead of using pdate? Is it because the note editor only allows to edit fields inside the note content?

openreview/tools.py
```diff
@@ -1566,6 +1542,49 @@ def get_current_submissions_profile_info(profile, n_years=None, submission_venue
...
+                        year = converted_year
+                except Exception as e:
+                    year = None
+            if isinstance(publication.content['year'], dict) and 'value' in publication.content['year']:
+                try:
+                    converted_year = int(publication.content['year']['value'])
+                    if converted_year <= current_year:
+                        year = converted_year
+                except Exception as e:
+                    year = None
```
carlosmondra, line 1579:
```suggestion
            unformatted_year = None
            if isinstance(publication.content['year'], dict) and 'value' in publication.content['year']:
                unformatted_year = publication.content['year']['value']
            elif isinstance(publication.content['year'], str):
                unformatted_year = publication.content['year']

            try:
                converted_year = int(unformatted_year)
                if converted_year <= current_year:
                    year = converted_year
            except Exception as e:
                year = None
```

openreview/profile/management.py
melisabok: and we don't have a date picker to choose the pdate but Xukun implemented one. Do you think we should ask a pdate instead?

melisabok: cc @xkopenreview to see if we could use the date picker in the note editor to select the pdate.

carlosmondra: > and we don't have a date picker to choose the pdate but Xukun implemented one. Do you think we should ask a pdate instead?

That's what we are asking in the content of the note right?

xkopenreview: is direct upload the only place where pdate is user entered?

melisabok: so far, yes, this is the first case when we need to ask the user to enter the pdate.

melisabok: pdate is implemented now.

carlosmondra (approved):

celestemartinez (approved):

xkopenreview (approved):

PR #1896, opened 2023-10-31
Feat/ Post decision stage: add parentGroup to decision emails

celestemartinez:
This PR adds the /Authors group ad parentGroup for decision notifications. It also adds a comment to the request form if the emails were sent successfully with a link to the messages UI for PCs to see emails that were sent.

### DIFF
--- a/openreview/venue/venue.py
+++ b/openreview/venue/venue.py
@@ -778,7 +778,7 @@ Total Errors: {len(errors)}
                 message = messages[decision_note['content']['decision']['value']]
                 final_message = message.replace("{{submission_title}}", note.content['title']['value'])
                 final_message = final_message.replace("{{forum_url}}", f'https://openreview.net/forum?id={note.id}')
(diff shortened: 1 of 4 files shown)


openreview/venue/venue.py
```diff
@@ -778,7 +778,7 @@ def send_notification(note):
                 message = messages[decision_note['content']['decision']['value']]
                 final_message = message.replace("{{submission_title}}", note.content['title']['value'])
                 final_message = final_message.replace("{{forum_url}}", f'https://openreview.net/forum?id={note.id}')
-                self.client.post_message(subject, recipients=note.content['authorids']['value'], message=final_message)
+                self.client.post_message(subject, recipients=note.content['authorids']['value'], message=final_message, parentGroup=self.get_authors_id())
```
melisabok, line 781: Does this work?

I think the parentGroup checks the recipient is a direct member of the group not a transitive member. We can use the submission author group instead.

melisabok (approved):

Issue #1872, opened 2023-10-23
DBLP import should use API 2 invitations


melisabok:
Replace `profileManagement.set_dblp_invitations` to use API 2 invitations.

xkopenreview:
i think for now it should be adding an invitation for v2 dblp instead of replacing the existing one

melisabok:
We are not replacing invitations because they are saved in different collections, also you need to keep updating api 1 notes to add authorids and abstract

melisabok:
Draf PR here: https://github.com/openreview/openreview-py/pull/1901

PR #1901, opened 2023-11-02
Feature/dblp invitations

melisabok:
Fixes https://github.com/openreview/openreview-py/issues/1872


openreview/profile/process/dblp_author_coreference_pre_process.js
```diff
@@ -0,0 +1,30 @@
...
+  client.throwErrors = true;
+  
+  const value = edit.note.content.authorids?.value;
+
+  if (!value?.replace) {
+    return Promise.reject(new OpenReviewError({ name: 'Error', message: `Only author replacement is allowed` }))
+  }
+  
+  const index = value.replace.index;
+  const authorId = value.replace.value;
```
xkopenreview, line 11: i think this process function need to allow null replace value so that imported papers can be unlinked

melisabok: Trying using an empty string instead of null, I added a test to verify that.

melisabok:
@xkopenreview I updated the coreference invitation to enter the author info in the edit.content. this way the UI will be able to render the widgets for the user to enter this info.

xkopenreview:
is there a case that the edit content field has param but note editor should not send a value?

melisabok:
yes, it is possible. Note content can be empty.

openreview/profile/management.py
```diff
@@ -28,75 +20,252 @@ def setup(self):
...
+                            'items': [
+                                { 'prefix': '~.*', 'optional': True },
+                                { 'value': self.support_group_id, 'optional': True },
+                                { 'value': dblp_uploader_group_id, 'optional': True } 
+                            ]
+                        } 
+                    },
+                    'writers':  [dblp_group_id, '${2/signatures}'],
+                    'content': {
+                        'author_index': {
```
xkopenreview, line 192: i think it's not clear that this index is 0 based

openreview/profile/management.py
```diff
@@ -28,75 +20,252 @@ def setup(self):
...
+                            'value': {
+                                'param': {
+                                    'type': 'integer'
+                                }
+                            }
+                        },
+                        'author_id' : {
```
xkopenreview, line 199: i think user may not be able to find this author id if it's not their preferred name

openreview/profile/management.py
melisabok: I can add a description.

openreview/profile/management.py
melisabok: I was thinking about using the signature as author_id but this means this invitation can only be used by the author of the paper.

openreview/profile/management.py
```diff
@@ -28,75 +20,252 @@ def setup(self):
...
+                        },
+                    },
+                    'note': {
+                        'id': {
+                            'param': {
+                                'withInvitation': record_invitation_id
+                            }
+                        },
+                        'content': {
+                            'authorids': {
```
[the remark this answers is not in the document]
melisabok: I don't think so because the inference has to be applied to the authorids field.

openreview/profile/management.py
xkopenreview: for example the user has 2 names ABC and XYZ where ABC is the preferred name so when user go to profile page, the url show id=~ABC1

when the name XYZ is used in the DBLP publication, the user need to put the id ~XYZ1 in this field but currently the user can't find this id anywhere unless changing the preferred name to XYZ (and go to profile page) or query api directly and check the username of the name XYZ

melisabok: I think the preprocess validation lets the user use any username, it doesn't need to match with the name listed in the DBLP publication.

xkopenreview: i think the preprocess function fails if the name does not match
invitation validation also fails when author id is changed by editing the author coreference edit

openreview/profile/process/dblp_author_coreference_pre_process.js
xkopenreview: now the author conreference get author index and author id from edit content i think it's not possible to unlink a publication

openreview/profile/process/dblp_author_coreference_pre_process.js
```diff
@@ -0,0 +1,36 @@
...
+  const authorId = edit.content.author_id.value;
+
+  const { notes } = await client.getNotes({ id: edit.note.id });
+  const publication = notes[0];
+
+  if (index >= publication.content.authorids.value.length || index < 0) {
+    return Promise.reject(new OpenReviewError({ name: 'Error', message: `Invalid author index` }));
+  }
+
+  const { profiles } = await client.getProfiles({ id: edit.signatures[0] });
```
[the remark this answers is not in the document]
melisabok: can they import dblp papers if the profile is not active?

I can fix this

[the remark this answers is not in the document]
melisabok: can you share an example when the profile is not public?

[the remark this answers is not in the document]
xkopenreview: just need to uncheck the public profile checkbox in profile edit page of any profile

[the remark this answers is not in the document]
melisabok: I understand the issue now, waiting for @carlosmondra to fix this.

openreview/profile/management.py
```diff
@@ -30,75 +22,256 @@ def setup(self):
...
+                    'signatures': { 
+                        'param': { 
+                            'items': [
+                                { 'prefix': '~.*', 'optional': True },
+                                { 'value': self.support_group_id, 'optional': True },
+                                { 'value': dblp_uploader_group_id, 'optional': True } 
+                            ]
+                        } 
+                    },
+                    'writers':  [dblp_group_id, '${2/signatures}'],
```
xkopenreview, line 192: i think we discussed about editing author coreference edits and user may encounter an error when author id is changed because it does not match invitation value

so i think edit.writers need to be updated to disallow editing of author coreference edits

melisabok: good catch! let me fix it

melisabok: fixed!

melisabok:
Last commit enabled commenting on Archive submissions, this is going to happen on demand only. We need to discuss how to handle the moderation.

openreview/profile/process/dblp_author_coreference_pre_process.js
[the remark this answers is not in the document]
carlosmondra: Should be fixed with https://github.com/openreview/openreview-api-v1/pull/2984

openreview/profile/process/dblp_author_coreference_pre_process.js
```diff
@@ -0,0 +1,39 @@
...
+  const usernames = userProfile.content.names.map(name => name.username);
+  const names = userProfile.content.names.map(name => name.fullname);
+
+  if (authorId === '') {
+    const authorName = publication.content.authorids.value[authorIndex];
+    if (!usernames.some(username => username === authorName)) {
+      return Promise.reject(new OpenReviewError({ name: 'Error', message: `The author name ${authorName} from index ${authorIndex} doesn't match with the names listed in your profile` }));
+    }
+    return;
+  }
```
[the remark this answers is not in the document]
carlosmondra: Oh I see, this is preprocess. Disregard my comment.

carlosmondra (approved):

openreview/profile/process/dblp_author_coreference_pre_process.js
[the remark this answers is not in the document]
melisabok: To remove the link of the profile with the DBLP paper, I'm passing an empty string.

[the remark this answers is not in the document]
carlosmondra: But in this case it's a preprocess function so the data changed within the function does not affect anything.

celestemartinez (approved):

xkopenreview (approved):
looks ok

xkopenreview:
it looks like meta extraction is using the following invitations

- dblp.org/-/pdf 
- dblp.org/-/abstract

to update dblp note
and this pr should also include a v2 version of them

melisabok:
The abstract invitation is already here: https://github.com/openreview/openreview-py/blob/master/openreview/profile/management.py#L239

it is missing the PDF one.

Remember that API 2 will use different invitation names.

PR #1904, opened 2023-11-07
Feat/ Custom Stage: create invitations to revise reviews

celestemartinez:
This PR allows us to use the CustomStage to create the Review_Revision stage. 

One issue: this stage adds a field to the review form. However, if the Review Stage is run again from the request form, any readers this field had will be removed unless we add them to the review additional fields. We would need to be careful when adding extra fields and when wanting to use the Review Stage button.

### DIFF
--- a/openreview/stages/venue_stages.py
+++ b/openreview/stages/venue_stages.py
@@ -1322,12 +1322,16 @@ class CustomStage(object):
         WITHFORUM = 1
         REVIEWS = 2
         METAREVIEWS = 3
-        REVIEW_REVISIONS = 4
 
-    def __init__(self, name, reply_to, source, start_date=None, due_date=None, exp_date=None, invitees=[], readers=[], content={}, multi_reply = False, email_pcs = False, email_sacs = False, notify_readers=False, email_template=None, allow_de_anonymization=False):
+    class ReplyType(Enum):
+        REPLY = 0
+        REVISION = 1
+
+    def __init__(self, name, reply_to, source, reply_type=ReplyType.REPLY, start_date=None, due_date=None, exp_date=None, invitees=[], readers=[], content={}, multi_reply = False, email_pcs = False, email_sacs = False, notify_readers=False, email_template=None, allow_de_anonymization=False):
         self.name = name
         self.reply_to = reply_to
         self.source = source
+        self.reply_type = reply_type
         self.start_date = start_date
         self.due_date = due_date
         self.exp_date = exp_date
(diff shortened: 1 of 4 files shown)


openreview/venue/invitation.py
```diff
@@ -2024,6 +2027,13 @@ def set_custom_stage_invitation(self):
                     'withForum': '${6/content/noteId/value}'
                 }
             }
+        elif custom_stage_replyto == 'review_revisions':
+            paper_invitation_id = self.venue.get_invitation_id(name=custom_stage.name, prefix='${2/content/replytoSignatures/value}')
+            with_invitation = self.venue.get_invitation_id(name=self.venue.review_stage.name, number='${6/content/noteNumber/value}')
+            reply_to = None
+            edit_readers = [venue_id]
```
celestemartinez, line 2034: edit readers are usually just the note readers, but in this case we are not setting note readers in the invitation (since we want to keep the review readers). I set the venueid as the readers of the edit but I am not sure this is correct.

melisabok: can you add the signatures of the edit as reader too?

openreview/stages/venue_stages.py
```diff
@@ -1322,6 +1322,7 @@ class ReplyTo(Enum):
         WITHFORUM = 1
         REVIEWS = 2
         METAREVIEWS = 3
+        REVIEW_REVISIONS = 4
```
melisabok, line 1325: I'm wondering if this should be another parameter to specify if the custom stage should "replies" or "edits" on the ReplyTo selected value.
This way you can use the custom stage to create edits for reviews, metareviews and forums. I know that the SubmissionRevisionStage already works for forums...

celestemartinez: Yes, I thought about this. I think the CustomStage is getting more and more complicated and it might not be easy for PCs to know exactly how to do what they need. But it does make sense to do this. I'll make this change.

melisabok: I agree that is becoming more complex, maybe we should have a CustomStage for making replies and another CustomStage for editing existing notes.

celestemartinez: just saw this message! If we think the current code is too complex, I can separate the CustomStage into two.

melisabok (approved):

celestemartinez:
Please don't merge yet, I would like to test revisions to metareviews!

tests/test_venue_request_v2.py
```diff
@@ -2131,6 +2258,9 @@ def test_release_meta_reviews_to_authors_and_reviewers(self, test_client, helper
         assert 'V2.cc/2030/Conference/Submission1/Authors' in meta_reviews[0].readers
         assert len(meta_reviews[0].nonreaders) == 0
 
+        assert 'final_recommendation' in meta_reviews[0].content
+        assert 'readers' in meta_reviews[0].content['final_recommendation']
```
celestemartinez, line 2262: @melisabok this is what I meant about the issue with re-running the stage from the request form. 

In this case, a PC could re-run the metareview stage to release metareviews to authors. Since `final_recommendation` was added to metareviews in a metareview revision, this field is not in the metareview additional fields in the request form. This means that the readers is removed from this field when re-running the metareview. 
Not sure if this is an issue but something we would have to keep in mind.

melisabok:
Is this PR ready to merge?

PR #1915, opened 2023-11-16
Name deletion: delete old relation and add new one when deleting a name

melisabok:


carlosmondra (approved):

openreview/profile/process/request_remove_name_decision_process.py
```diff
@@ -256,20 +256,25 @@ def replace_group_members(group, current_member, new_member):
...
+                if old_relations or new_relations:
+                    client.post_profile(openreview.Profile(
+                        referent = related_profile.id, 
+                        invitation = '~/-/invitation',
+                        signatures = ['~Super_User1'],
+                        content = {},
+                        metaContent = {
+                            'relations': { 
+                                'values': new_relations + old_relations,
+                                'weights': ([1] * len(new_relations)) + ([-1] * len(old_relations))
```
celestemartinez, line 274: I am trying to understand this: won't len(new_relations) always be equal to len(old_relations) here, making weights 0? is this right? Am I missing something?

melisabok: new_retation has a different username, the one we are replacing.

For example: you want to delete the name "Ana Martinez" from your profile and my profile has "Ana Martinez" as a co-worker.

When we delete Ana Martinez, we need to delete the relation "Ana Martinez" and add the one one "Celeste Martinez" in my profile.

celestemartinez: yeah that makes sense, but then wouldn't the weight just be 0 always? Since we are always appending to both lists at the same time, their lenghts will be the same, right?

Since len(new_relations) == len(old_relations)
`'weights': ([1] * len(new_relations)) + ([-1] * len(old_relations))` == 0

Just making sure I understand!

melisabok: Let me add a test that verifies that weight arrays are built correctly.

celestemartinez (approved):

Issue #1874, opened 2023-10-23
Profile Management: merge profiles, delete names and emails should use API 2


melisabok:
Replace all invitations to use API 2.

melisabok:
Fixed here: https://github.com/openreview/openreview-py/pull/1916

PR #1916, opened 2023-11-20
Feature/migrate profile invitations

melisabok:
- Name deletion request using API 2 invitations
- Email deletion request using API 2 invitations
- Profile merge request using API 2 invitations

Fixes https://github.com/openreview/openreview-py/issues/1874


xkopenreview:
the process log id is the entity.id which is the edit.id so currently the moderation page won't find matching log by comparing the log id with note id
@carlosmondra is it possible to add a note.id field?
another way is to get all the edit ids when query the email/name deletion notes

melisabok:
> another way is to get all the edit ids when query the email/name deletion notes

what about querying the process logs by invitation instead of id?

xkopenreview:
> > another way is to get all the edit ids when query the email/name deletion notes
> 
> what about querying the process logs by invitation instead of id?

process logs are already queried by invitation
but still need to find which log is for which note by comparing the log id

openreview/api/client.py
```diff
@@ -588,6 +589,27 @@ def post_profile(self, profile):
...
+    def rename_profile(self, current_id, new_id):
+        """
+        Updates a Profile
+
+        :param profile: Profile object
+        :type profile: Profile
+
+        :return: The new updated Profile
+        :rtype: Profile
+        """
```
carlosmondra, line 601:
```suggestion
        """
        Updates a the profile id of a Profile

        :param current_id: Current profile id
        :type profile: str
        :param new_id: New profile id
        :type profile: str

        :return: The new updated Profile
        :rtype: Profile
        """
```

openreview/profile/management.py
```diff
@@ -104,26 +104,38 @@ def set_remove_name_invitations(self):
...
+                        'optional': True
+                    }
+                }
             },
             'status': {
-                'order': 4,
-                'value-dropdown': ['Pending', 'Accepted', 'Rejected'],
-                'description': 'Request status.',
-                'required': True
+                'value': 'Pending'
```
carlosmondra, line 138: Why was this changed to a constant value?
Is it OK to remove description and order from the invitation?

melisabok: it is a constant value because there is another invitation that changes the status.

carlosmondra (approved):

openreview/profile/process/request_remove_name_process.py
```diff
@@ -1,30 +1,31 @@
-def process(client, note, invitation):
+def process(client, edit, invitation):
 
     SUPPORT_USER_ID = ''
     REMOVAL_DECISION_INVITATION_ID = ''
-    baseurl_v2 = 'http://localhost:3001'
+    baseurl_v1 = 'http://localhost:3000'
+    print(edit.note)
```
xkopenreview, line 6: there's a print

openreview/profile/management.py
```diff
@@ -236,20 +273,35 @@ def set_remove_email_invitations(self):
             'email': {
                 'order': 1,
                 'description': 'email that want to be removed.',
-                'value-regex': '.*',
-                'required': True                
+                'value': {
+                    'param': {
+                        'type': 'string',
+                        'regex': r"^[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)"
```
xkopenreview, line 279: i think this regex is incomplete

xkopenreview (approved):

celestemartinez (approved):

PR #1921, opened 2023-11-22
Fix/ tools.get_profiles with_relations 

celestemartinez:
In line 226, we were overriding the `emails` array we initialize [here](https://github.com/openreview/openreview-py/blob/cd366c20ce08d84fe1a3fdb8049b09e33643273d/openreview/tools.py#L147) so we were missing some profiles in the output.


carlosmondra (approved):

melisabok:
Thanks for the fix!

PR #1925, opened 2023-11-28
Feature/Date picker widget

xkopenreview:
related to https://github.com/openreview/openreview-web/pull/1734

add date picker in test

### DIFF
--- a/tests/test_react_note_editor.py
+++ b/tests/test_react_note_editor.py
@@ -202,6 +202,18 @@ class TestReactNoteEditor():
                             'deletable': True
                         }
                     }
+                },
+                'datetime_picker_field': {
+                    'description': 'this is a datetime picker widget',
+                    'order': 1,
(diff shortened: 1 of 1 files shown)


tests/test_react_note_editor.py
```diff
@@ -202,6 +202,18 @@ def venue(self, openreview_client):
...
+                    'description': 'this is a datetime picker widget',
+                    'order': 1,
+                    'value': {
+                        'param': {
+                            'type': 'date',
+                            'optional': True,
+                            'deletable': True,
+                            'range': [0, 9999999999999]
+                        }
+                    }
```
melisabok, line 216: so we don't need to specify an input?

xkopenreview: based on the widget rendering logic discussion type=date or date[] will render date picker
if this is not accurate any more we can discuss again

PR #1926, opened 2023-11-28
Fix/Comment preprocess function is missing a space

xkopenreview:
image

this pr should add a space reader value and "must be readers..."

### DIFF
--- a/openreview/venue/process/comment_pre_process.js
+++ b/openreview/venue/process/comment_pre_process.js
@@ -17,12 +17,12 @@ async function process(client, edit, invitation) {
   const commentMandatoryReaders =
     domain.content?.comment_mandatory_readers?.value || []
   for (const m of commentMandatoryReaders) {
-    const mandatoryReader = m.replace("{number}", forum.number)
-    if (!note.readers.includes(mandatoryReader)) {
+    const reader = m.replace("{number}", forum.number)
+    if (!note.readers.includes(reader)) {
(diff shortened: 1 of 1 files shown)


melisabok:
some tests are failing here , can you check if it is related to these changes?

xkopenreview:
it looks like in process function 
>${reader} 

can't be used because it will be recognized as json pointer
the tests are failing because i used ${reader} in string interpolation

need to take note of this if it's not fixable

melisabok:
that's right, you need to use + to merge string.

openreview/venue/process/comment_pre_process.js
```diff
@@ -2,23 +2,29 @@ async function process(client, edit, invitation) {
...
   const domain = res2.groups[0]
 
-  const commentMandatoryReaders = domain.content?.comment_mandatory_readers?.value || []
+  const commentMandatoryReaders =
+    domain.content?.comment_mandatory_readers?.value || []
   for (const m of commentMandatoryReaders) {
-    const reader = m.replace('{number}', forum.number)
-    if (!note.readers.includes(reader)) {
-      return Promise.reject(new OpenReviewError({ name: 'Error', message: reader + 'must be readers of the comment' }));
+    const mandatoryReader = m.replace("{number}", forum.number)
```
melisabok, line 20: can you revert this change and just add a space to the line 21?

PR #1933, opened 2023-11-29
Update deployProcess.py with new profile moderation message

racheljsmart:
Informing PCs that new profiles will take up to two weeks to be activated, but also letting them know that institutional emails will automatically activate their profile so they may choose to encourage this.


openreview/venue_request/process/deployProcess.py
```diff
@@ -78,6 +78,8 @@ def process(client, note, invitation):
 
 Thank you for choosing OpenReview to host your upcoming venue.
 
+Users of new profiles that sign up with non institutional emails will go through a moderation process that **can take up to two weeks**. Users that signup for a profile with their institutional email will be activated automatically.
```
melisabok, line 81: Sounds great! I think we should suggest the PCs to share this information in the "Calls for papers". @celestemartinez do you have any suggestions for the wording?

racheljsmart: "To encourage early profile registration, we recommend alerting authors of this extended moderation period in the Call for Papers."

How about something like this?

racheljsmart: The bullet points might be easier to read. I'll make the change.

melisabok:
There are some conflicts here.

I'm waiting for the signup page to be released in the live before merging these changes.

melisabok (approved):

PR #1934, opened 2023-12-01
Add API2 venues to active venues group when deployed

### LINKED ISSUE #1729: Updating the active venues group for API 2 venues
Currently, members of the active venues group are added when the Post Submission stage is run, but this does not work for API 2 venues because everything is run automatically.

One option is to add the member when the venue is deployed.

enrubio:
- Fixes #1729

### DIFF
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -1696,7 +1696,7 @@ Program Chairs
 
         if venue_heading_map:
             self.set_homepage_decisions(decision_heading_map=venue_heading_map)
-        self.client.remove_members_from_group('active_venues', self.id)
+        self.client_v2.remove_members_from_group('active_venues', self.id)
 
         # expire recruitment invitations
         self.expire_recruitment_invitations()
--- a/openreview/conference/builder.py
+++ b/openreview/conference/builder.py
@@ -1843,6 +1843,7 @@ class ConferenceBuilder(object):
 
     def __init__(self, client, support_user=None):
         self.client = client
+        self.client_v2 = openreview.Client(baseurl=openreview.tools.get_base_urls(client)[1], token=client.token)
         self.conference = Conference(client)
         self.webfield_builder = webfield.WebfieldBuilder(client)
         self.submission_stage = None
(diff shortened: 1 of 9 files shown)


openreview/venue_request/process/deployProcess.py
```diff
@@ -615,4 +615,8 @@ def process(client, note, invitation):
                 }
             },
             signatures=['~Super_User1']
-        ))
\ No newline at end of file
+        ))
+
+        # Add API 2 venues to active venues group
+        active_venues = client.get_group('active_venues')
```
melisabok, line 621: I think this change should be done inside of the Venue class so the venue is added to the group even if we don't use the venue request form.

please remove the code where we are adding in the post submission stage.

enrubio: > so the venue is added to the group even if we don't use the venue request form.

what does this mean? don't all venues have to submit a venue request form?

melisabok: you can create a venue using the Venue class directly, no need to create a request form. The Venue class should be cohesive and have all the behavior inside, if you check the code we are already adding the venue to active_venues inside of the Venue class but only when post_submission_stage is run manually. I think you should move that line and run it in the setup function.

openreview/conference/builder.py
```diff
@@ -2109,7 +2110,7 @@ def get_result(self):
 
         venues = self.client.get_group(id = 'venues', details='writable')
         if venues.details.get('writable'):
-            self.client.add_members_to_group('venues', home_group.id)
+            self.client_v2.add_members_to_group('venues', home_group.id)
```
melisabok, line 2113: This is code is part of the Builder class not the conference so you can create the client_v2 in the Builder class too.

melisabok: some tests are failing because of this: https://github.com/openreview/openreview-py/pull/1934/files#diff-8d0f7e15bc90c791f490e1aef09fa75304f454f2e25fb186568515716e68c2b7L1844

melisabok (approved):

celestemartinez (approved):

PR #1936, opened 2023-12-04
Archive Edges: set responseArchiveDate for all edge invitations

melisabok:
Use the venue date + 1 year to set the archive date


openreview/venue/venue.py
```diff
@@ -79,6 +79,17 @@ def get_id(self):
 
     def get_short_name(self):
         return self.short_name
+    
+    def get_edges_archive_date(self):
```
melisabok, line 83: @carlosmondra I'm using the venue.date (when the conference is happening) plus 1 year to archive the edges, let me know what you think about it.

carlosmondra: Seems reasonable to me

carlosmondra (approved):

celestemartinez (approved):

PR #1937, opened 2023-12-04
Fix/ Venue: area chairs assignment

celestemartinez:
Fixes two issues:
- when an AC assignment is deleted, the corresponding SAC is removed from the group, instead of posting the group with empty members
- when an assignment is deleted, check if there exists another edge for the same assignment. If there is, do not remove member from group


melisabok (approved):
Thanks for the fixes!

PR #1939, opened 2023-12-05
Update venue_request.py to include new fields

racheljsmart:
Asking for more information about the venue helps us define the venue id.

### DIFF
--- a/openreview/venue_request/venue_request.py
+++ b/openreview/venue_request/venue_request.py
@@ -1122,17 +1122,35 @@ class VenueRequest():
                 'required': True,
                 'order': 4
             },
+            'venue_type': {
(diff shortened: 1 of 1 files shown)


openreview/venue_request/venue_request.py
```diff
@@ -1122,17 +1122,35 @@ def setup_request_form(self):
                 'required': True,
                 'order': 4
             },
+            'venue_type': {
+                'description': 'Please select the option that best describes the type of venue you\'re hosting. If neither option is appropriate, type your anwser into the text box.',
```
celestemartinez, line 1126: If this is a dropdown, then users can only pick from the dropdown and they can't type in their own answer

melisabok:
@racheljsmart do we keep this open or should we close it?

I think we can add these field to the new venue configuration request.

melisabok:
Closing this, it is open for too long and I think we should support this in the new venue configuration.

PR #1945, opened 2023-12-07
Fix/Venue: use withVenueid for submission invitation

celestemartinez:
Create submission invitation with `withVenueid` so this invitation to edit the paper only appears when the paper is still under submission. 

Requires:
https://github.com/openreview/openreview-api/pull/311
https://github.com/openreview/openreview-web/pull/1750


melisabok (approved):
💯

melisabok:
`find_elements_by_tag_name` was removed from the latest versions of selenium, you should use `find_elements(By.TAG_NAME,"a")` instead

PR #1950, opened 2023-12-12
Conference: Allow SAC-paper assignments deployment before AC-paper assignments

haroldrubio:
This PR is for ARR - it checks the assignment invitation to see if the edges are pointing to profiles. If it is not, then allow deployment before the AC assignments and use different logic for overwriting and populating the SAC groups.


melisabok:
I wonder if we would need this changes for API 2 venues too.

melisabok (approved):

celestemartinez (approved):

PR #1965, opened 2024-01-03
Edge invitations: replace regex with items keyword

melisabok:


melisabok:
@xkopenreview does the edge browser support `items` for signatures? I believe we not but we can plan to implement it soon.

melisabok:
I think we should add this to your TODO list. We want to remove all the dependencies to API 1 as soon as possible.

melisabok:
yes, that could work. If you search `browse=` in the tests, you could find other tests.

openreview/journal/invitation.py
```diff
@@ -1877,7 +1883,11 @@ def set_reviewer_assignment(self, assignment_delay):
                 'writers': [venue_id, self.journal.get_action_editors_id(number='${{2/head}/number}')],
                 'signatures': {
                     'param': {
-                        'regex': venue_id + '|' + editor_in_chief_id + '|' + self.journal.get_action_editors_id(number='.*', anon=True)
+                        'items': [
+                            { 'value': venue_id, 'optional': True },
+                            { 'value': editor_in_chief_id, 'optional': True },
+                            { 'prefix': self.journal.get_action_editors_id(number='${{3/head}/number}', anon=True), 'optional': True }
```
xkopenreview, line 1967: is this 
${{3/head}/number}
or
${3/head/number}

melisabok: double curly braces so the API get the note first and then gets the number

openreview/journal/journal_request/journal_request.py
```diff
@@ -166,7 +166,14 @@ def setup_journal_request(self):
                 writers = [],
                 signatures = ['~Super_User1'],
                 edit = {
-                    'signatures': { 'param': { 'regex': f'~.*|{self.support_group_id}' }},
+                    'signatures': { 
+                        'param': { 
+                            'items': [
+                                { 'prefix': '~', 'optional': True },
```
celestemartinez: should this (and some other prefixes below) be `"~.*"`?

melisabok: I don't think that changes the regex because the API add the ^ at the beginning but we can use `~.*` to be consistent with other places we use the same pattern.

PR #1974, opened 2024-01-10
Fix/ Rebuttal Stage: change naming convention of rebuttal invitations

celestemartinez:
Rebuttal invitations will now be named as follows (the last two remain unchanged):
- one rebuttal per review: `MICCAI.org/2023/STS/Submission1/Official_Review1/-/Rebuttal`
- one rebuttal per paper: `MICCAI.org/2023/STS/Submission1/-/Rebuttal`
- multiple rebuttals per paper: `MICCAI.org/2023/STS/Submission1/-/Rebuttal`

### DIFF
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -439,10 +439,10 @@ class InvitationBuilder(object):
         reply_to = '${4/content/noteId/value}'
 
         if not review_rebuttal_stage.single_rebuttal and not review_rebuttal_stage.unlimited_rebuttals:
-            submission_prefix = self.venue_id + '/' + self.venue.submission_stage.name + '${2/content/noteNumber/value}/'
+            submission_prefix = venue_id + '/' + self.venue.submission_stage.name + '${2/content/noteNumber/value}/'
             review_prefix = self.venue.review_stage.name + '${2/content/replyNumber/value}'
             paper_invitation_id = self.venue.get_invitation_id(name=review_rebuttal_stage.name, prefix=submission_prefix+review_prefix)
-            submission_prefix = self.venue_id + '/' + self.venue.submission_stage.name + '${6/content/noteNumber/value}/'
+            submission_prefix = venue_id + '/' + self.venue.submission_stage.name + '${6/content/noteNumber/value}/'
             review_prefix = self.venue.review_stage.name + '${6/content/replyNumber/value}'
             with_invitation = self.venue.get_invitation_id(name=review_rebuttal_stage.name, prefix=submission_prefix+review_prefix)
             reply_to = '${4/content/replyto/value}'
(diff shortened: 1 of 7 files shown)


celestemartinez:
We use a similar naming convention for the custom stage, so I still need to make the change for the CustomStage so we can get rid of `replytoSignatures` everywhere.

openreview/venue/process/invitation_edit_process.py
```diff
@@ -135,6 +135,13 @@ def post_invitation(note):
             content['replyto'] = { 'value': note.id }
             content['replytoSignatures'] = { 'value': note.signatures[0] }
 
+        if 'replyto' in invitation.edit['content'] and 'reviewNumber' in invitation.edit['content']:
+            paper_number = note.invitations[0].split(submission_name)[-1].split('/')[0]
+            content['noteId'] = { 'value': note.forum }
+            content['noteNumber'] = { 'value': int(paper_number) }
+            content['replyto'] = { 'value': note.id }
+            content['reviewNumber'] = { 'value': note.number }
```
melisabok, line 143: I think we should use a more general name here, like `replyNumber` because we can create invitations are replies of reviews, metareviews, etc

celestemartinez: yes! I realized this when I started to update the CustomStage

openreview/venue/invitation.py
```diff
@@ -439,8 +439,12 @@ def set_review_rebuttal_invitation(self):
         reply_to = '${4/content/noteId/value}'
 
         if not review_rebuttal_stage.single_rebuttal and not review_rebuttal_stage.unlimited_rebuttals:
-            paper_invitation_id = self.venue.get_invitation_id(name=review_rebuttal_stage.name, prefix='${2/content/replytoSignatures/value}')
-            with_invitation = self.venue.get_invitation_id(name=review_rebuttal_stage.name, prefix='${6/content/replytoSignatures/value}')
+            submission_prefix = self.venue_id + '/' + self.venue.submission_stage.name + '${2/content/noteNumber/value}/'
+            review_prefix = self.venue.review_stage.name + '${2/content/replyNumber/value}'
+            paper_invitation_id = self.venue.get_invitation_id(name=review_rebuttal_stage.name, prefix=submission_prefix+review_prefix)
+            submission_prefix = self.venue_id + '/' + self.venue.submission_stage.name + '${6/content/noteNumber/value}/'
```
melisabok, line 445: it seems there are two lines repeated?

openreview/venue/process/invitation_edit_process.py
```diff
@@ -135,6 +135,13 @@ def post_invitation(note):
             content['replyto'] = { 'value': note.id }
             content['replytoSignatures'] = { 'value': note.signatures[0] }
```
melisabok, line 136: we don't delete this because it is still being used by the custom stage, right?

openreview/venue/process/review_process.py
```diff
@@ -144,14 +144,18 @@ def create_group(group_id, members=[]):
...
         content_keys = invitation.edit.get('content', {}).keys()
-        if 'reviews' == review_reply and 'replytoSignatures' in content_keys and 'replyto' in content_keys and len(content_keys) == 4:
+        if 'reviews' == review_reply and 'replyto' in content_keys and len(content_keys) == 4:
             print('create invitation: ', invitation.id)
+            content  = {
+                'noteId': { 'value': review.forum },
+                'noteNumber': { 'value': submission.number },
+                'replyto': { 'value': review.id }
+            }
+            if 'replytoSignatures' in content_keys:
```
melisabok, line 154: why do you need to keep this?  for backward compatibility?

openreview/venue/invitation.py
celestemartinez: the paths are different since one is for the invitation id and one is for the with invitation field. I can try to unify this later

openreview/venue/process/review_process.py
celestemartinez: I think we still need this because of the one `replytoSignatures` I still have in the CustomStage

melisabok (approved):

openreview/venue/invitation.py
```diff
@@ -2034,24 +2040,27 @@ def set_custom_stage_invitation(self):
...
+            reply_prefix = stage_name + '${6/content/replyNumber/value}'
+            with_invitation = self.venue.get_invitation_id(name=custom_stage.name, prefix=submission_prefix+reply_prefix)
+            reply_to = '${4/content/replyto/value}'
+
+            if custom_stage_reply_type == 'revision':
+                with_invitation = self.venue.get_invitation_id(name=stage_name, number='${6/content/noteNumber/value}')
+                reply_to = None
+                edit_readers = [venue_id, '${2/signatures}']
+                note_readers = None
+                invitees = ['${3/content/replytoSignatures/value}']
```
celestemartinez, line 2063: @melisabok This is the only place where we are using `replytoSignatures` and I don't think we can remove it. For a review revision for example, the invitee of the invitation has to be the signatures of the review, unless you see some other way?

melisabok: Do we have review revision invitations in API 2?

PR #1980, opened 2024-01-16
Fix/ Update SAC Console config

xkopenreview:
This pr should add withdrawn venue id and desk-rejected venue id to SAC console config so that withdrawn papers and desk rejected papers are filtered out

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -9,7 +9,8 @@ else:
 
 from .. import tools
 import requests
-from requests.adapters import HTTPAdapter, Retry
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 import pprint
 import os
 import re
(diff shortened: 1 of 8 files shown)


melisabok:
Does it mean they are not being shown in the paper status tab?
 is this already fixed in the live site for CVPR?

celestemartinez:
I think they were being shown in the Paper Status tab and with this PR they are not longer being shown. Is this correct @xkopenreview?

xkopenreview:
> I think they were being shown in the Paper Status tab and with this PR they are not longer being shown. Is this correct @xkopenreview?

yes

ui has the logic to filter out withdrawn and deskrejected papers
but if there's no withdrawn venueid and desk rejected venue id config variables, the filtering will not work

CVPR does not have these config vars so withdrawn and deskrejected papers are still displayed.

with this pr, newly created sac consoles should have the 2 config variables so withdrawn and desk rejected notes are filtered out

celestemartinez (approved):

PR #1986, opened 2024-01-18
Feature: distinguish commitment venues in the request form

celestemartinez:
This PR makes the following changes to the submission invitation if the venue is a commitment venue:
- set signatures to Super User
- adds paper_link to content
- add preprocess


openreview/conference/helpers.py
```diff
@@ -434,6 +434,8 @@ def get_submission_stage(request_forum, venue):
     else:
         withdraw_submission_exp_date = None
 
+    commitments_venue = request_forum.content.get('commitments_venue', 'No') == 'Yes'
```
melisabok, line 437: can we call it `ARR_commitments_venue`?

openreview/stages/venue_stages.py
```diff
@@ -293,6 +295,20 @@ def get_content(self, api_version='1', conference=None, venue_id=None):
                     }
                 }
 
+            if self.commitments_venue:
```
melisabok, line 298: I would check if paper_link is already in the content and don't overwrite it, the PCs might want to change the description.

openreview/venue/process/submission_commitments_preprocess.py
```diff
@@ -0,0 +1,17 @@
...
+    paper_link = edit.note.content['paper_link']['value']
+    paper_forum = paper_link.split('=')[-1]
+
+    try:
+        client_v1=openreview.Client(baseurl=openreview.tools.get_base_urls(client)[0], token=client.token)
+        arr_submission = client_v1.get_note(paper_forum)
+    except openreview.OpenReviewException as e:
+        raise openreview.OpenReviewException('Provided paper link does not correspond to a submission in OpenReview')
+
+    valid_prefixes = ['aclweb.org/ACL/ARR/2021', 'aclweb.org/ACL/ARR/2022', 'aclweb.org/ACL/ARR/2023', 'aclweb.org/ACL/ARR/2024']
```
melisabok, line 12: what if we have one single prefix `aclweb.org/ACL/ARR`?

openreview/conference/helpers.py
melisabok: is it possible that we would receive commitments from other venues?

melisabok:
The changes look good! thanks Celeste for making this PR so quickly. 

I would suggest to put the migration code here so we can have an invitation that runs after the submission deadline and migrates the papers and reviews. What do you think?

openreview/venue/process/submission_commitments_preprocess.py
celestemartinez: Yes! Not sure why we were using the year here...I don't see any other venue ids that would start with this.

openreview/conference/helpers.py
racheljsmart: possibly from EACL (which themselves are ARR commitments)

openreview/stages/venue_stages.py
```diff
@@ -293,6 +295,20 @@ def get_content(self, api_version='1', conference=None, venue_id=None):
...
+                content['paper_link'] = {
+                    'value': {
+                        'param': {
+                            'type': 'string',
+                            'regex': '(http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\\(\\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+)',
+                            'mismatchError': 'must be a valid link to an OpenrReview submission: https://openreview.net/forum?id=...'
+                        }
+                    },
+                    'description': 'Please provide the link to your ARR submission.',
```
racheljsmart, line 307: Can "The link should have the following format: https://openreview.net/forum?id=<PAPER_ID>" be added to the description?

racheljsmart: Does <PAPER_ID> make sense?

celestemartinez:
> The changes look good! thanks Celeste for making this PR so quickly.
> 
> I would suggest to put the migration code here so we can have an invitation that runs after the submission deadline and migrates the papers and reviews. What do you think?

Yes, I agree. I can add this.

openreview/conference/helpers.py
celestemartinez: should we change this to `ARR_commitments_venue`? It does seem like all commitments we are getting so far come from ARR.

melisabok (approved):

openreview/conference/helpers.py
melisabok: sorry, I saw this too late. We can keep it as it is for now.

PR #1988, opened 2024-01-19
Fix/venue add reply to param

### LINKED ISSUE #1553: Venue class: use domain.content.contact.value as replyTo for all the email notifications
For all the email notification being sent on behalf of a venue, we should add the replyTo parameter and use the contact info. So users can contact the organizers clicking on reply. 

With the exceptions of the AC invite assignments?

enrubio:
- Resolves #1553 

This adds the domain contact email as the `replyTo` for `post_message` calls. Not affected are the emails that tell the user to contact us directly, that are signed off with "OpenReview Team", or emails that are already sent to the PCs.

### DIFF
--- a/openreview/venue/venue.py
+++ b/openreview/venue/venue.py
@@ -80,9 +80,6 @@ class Venue(object):
     def get_short_name(self):
         return self.short_name
     
-    def get_contact(self):
-        return self.contact
-    
     def get_edges_archive_date(self):
         archive_date = datetime.datetime.utcnow()
         if self.date:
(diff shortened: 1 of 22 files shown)


openreview/venue/venue.py
```diff
@@ -80,6 +80,9 @@ def get_id(self):
     def get_short_name(self):
         return self.short_name
     
+    def get_contact(self):
```
melisabok, line 83: do you need to implement the get function?

enrubio: I don't think it's necessary. I was just following `short_name` as an example, but it can be removed

melisabok: you can remove it, get_short_name was implemented so it could be polymorph to the conference class. The venue request scripts call this method for both classes.

enrubio: Makes sense, fixed now

melisabok (approved):

PR #1991, opened 2024-01-24
Client: log client retries

melisabok:
I would like to see if the client is retrying the request when there are upstream errors.


melisabok:
I would like to include this in the new release.

carlosmondra (approved):

celestemartinez (approved):

PR #1998, opened 2024-01-29
Fix/ Venue: withdrawn and desk-rejected visibility

### LINKED ISSUE #1992: Withdrawn/desk-rejected submission request form value
Currently, the Withdrawn Submissions Author Anonymity and Desk Rejected Submissions Author Anonymity values are not being retrieved from the venue request.

celestemartinez:
Fixes #1992 

I am always adding readers to authors/authorids in the content and readers to withdrawn/desk-rejected submissions. 
I then check `withdrawn_submission_public`/`desk_rejected_submission_public` to see if the readers should be `everyone` instead. I don't think we should check what the readers are in the submission_stage because these can change at any time. 

This PR also creates a `/-/Deletion` invitation so authors can delete their submissions between the abstract and the submission deadline.


tests/test_iclr_conference_v2.py
```diff
@@ -73,6 +73,10 @@ def test_create_conference(self, client, openreview_client, helpers, profile_man
                 'senior_area_chair_identity': ['Program Chairs', 'Assigned Senior Area Chair', 'Assigned Area Chair', 'Assigned Reviewers'],
                 'Open Reviewing Policy': 'Submissions and reviews should both be public.',
                 'submission_readers': 'Everyone (submissions are public)',
+                'withdrawn_submissions_visibility': 'Yes, withdrawn submissions should be made public.',
+                'withdrawn_submissions_author_anonymity': 'Yes, author identities of withdrawn submissions should be revealed.',
+                'desk_rejected_submissions_visibility':'No, desk rejected submissions should not be made public.',
```
melisabok, line 78: shouldn't ICLR reveal the author identities?

melisabok:
it seems the tests don't finish because one test is waiting for the process `NeurIPS.cc/2023/Track/Datasets_and_Benchmarks/-/Withdrawal-0-0` to finish, probably this is not being run.

celestemartinez:
> it seems the tests don't finish because one test is waiting for the process `NeurIPS.cc/2023/Track/Datasets_and_Benchmarks/-/Withdrawal-0-0` to finish, probably this is not being run.

Thanks for the heads-up! This should be fixed. 

Note that now the default will be to have withdrawn/desk-rejected and their author names hidden (so I adding readers to notes and authors/authorids by default). Only if PCs explicitly choose in the request form to make public will they be public. This means that papers that were initially single blind could potentially have their author names hidden (this was not allowed in API1 because in this case we wouldn't create blind submissions). If this sounds good, then I can remove this check from the request form preprocess:

```
if 'Double-blind' not in request_form.content['Author and Reviewer Anonymity']:
    if 'No' in note.content.get('withdrawn_submissions_author_anonymity', ''):
        raise openreview.OpenReviewException('Author identities of withdrawn submissions can only be anonymized for double-blind submissions')

    if 'No' in note.content.get('desk_rejected_submissions_author_anonymity', ''):
        raise openreview.OpenReviewException('Author identities of desk-rejected submissions can only be anonymized for double-blind submissions')
```

melisabok:
I agree that the default option should be hiding the data including single blind venues. Let's remove that check.

openreview/venue/process/submission_revision_process.py
```diff
@@ -9,6 +9,7 @@ def process(client, edit, invitation):
     submission_name = domain.content['submission_name']['value']
 
     submission = client.get_note(edit.note.id)
+    print(edit.id)
```
melisabok, line 12: should we distinguish between a revision that deletes the note and another one that edits it?

otherwise the other authors won't know the paper was deleted.

tests/test_emnlp_conference.py
```diff
@@ -341,91 +340,6 @@ def test_submit_papers(self, test_client, client, openreview_client, helpers):
             'EMNLP/2023/Conference/Submission5/Authors'
         ]
 
-        ## withdraw submission
```
melisabok, line 344: instead of deleting this case, can you withdraw that paper using the delete edit and revert the change?

I would also like to test the email notifications.

tests/test_neurips_conference_v2.py
```diff
@@ -892,8 +892,8 @@ def test_submit_papers(self, test_client, client, helpers, openreview_client):
 
         helpers.await_queue()
         helpers.await_queue_edit(openreview_client, 'NeurIPS.cc/2023/Conference/-/Post_Submission-0-0')
-        helpers.await_queue_edit(openreview_client, 'NeurIPS.cc/2023/Conference/-/Withdrawal-0-0')
-        helpers.await_queue_edit(openreview_client, 'NeurIPS.cc/2023/Conference/-/Desk_Rejection-0-0')
+        # helpers.await_queue_edit(openreview_client, 'NeurIPS.cc/2023/Conference/-/Withdrawal-0-0')
```
melisabok, line 895: remove this code if it is not needed

tests/test_emnlp_conference.py
```diff
@@ -341,90 +341,70 @@ def test_submit_papers(self, test_client, client, openreview_client, helpers):
...
-                    'withdrawal_confirmation': { 'value': 'I have read and agree with the venue\'s withdrawal policy on behalf of myself and my co-authors.' },
+                ddate = openreview.tools.datetime_millis(datetime.datetime.utcnow()),
+                content = {
+                    'title': { 'value': 'Test Paper Title 5' },
+                    'authors': submissions[4].content['authors'],
+                    'authorids': submissions[4].content['authorids'],
+                    'abstract': { 'value': 'This is a test abstract 5' },
+                    'keywords': { 'value': ['machine learning', 'nlp'] },
+                    'submission_type': { 'value': 'Regular Long Paper' },
+                    'pdf': { 'value': '/pdf/' + 'p' * 40 +'.pdf' } #fields not in the submission need to be added to delete submission?
```
celestemartinez, line 356: I found this issue yesterday. EMNLP added fields to the revision form that were required (that were not in the submission). In order to delete a submission, the authors need to fill out all these new, required fields, but it seems weird to ask them to if they are deleting their submission.

melisabok: @celestemartinez I think we would need to create an invitation that let authors delete their submissions between the abstract and full paper deadline. 

The invitation can specify the ddate field only, what do you think?

celestemartinez: Sounds good, I can try this! This means I would remove the ddate field from the submission invitation or would we keep that as well?

melisabok: let's remove the ddate from the revision invitation so we don't have the issue that EMNLP has. So if the invitation has the ddate field only, it should render the trash icon and it shouldn't appear in the edit dropdown, can you verify that?

openreview/venue/invitation.py
```diff
@@ -194,6 +194,145 @@ def set_submission_invitation(self):
...
+        if deletion_duedate:
+            invitation.edit['invitation']['duedate'] = deletion_duedate
+
+        if deletion_expdate:
+            invitation.edit['invitation']['expdate'] = deletion_expdate
+
+        self.save_invitation(invitation, replacement=False)
+
+        expire_invitation = Invitation (
+            id=self.venue.get_invitation_id('Deletion_Expiration'),
```
melisabok, line 301: why do we need this invitation?

melisabok: ohhh I see, to expire and unexpire the existing invitations

melisabok: I'm not sure if we need to do this because the forum is marked a "deleted" so there won't be any pending tasks or buttons shown in the forum page.

celestemartinez: I copied this from the Withdrawal process function. We use this to expire active invitations when a paper is deleted, and then we use it to find these invitations we expired and unexpire them if the submission is restored. 
I guess we could use the Meta invitation, but I think we do this so it's easier to get these invitations if we have to unexpire them?

melisabok: I don't think we need to expire any invitation because the forum note is marked a deleted so no invitations will be valid anymore.

celestemartinez: Makes sense! I wasn't sure we needed it but then the tests were not passing so I thought we needed it. I'll make the change.

openreview/venue/invitation.py
```diff
@@ -194,6 +194,110 @@ def set_submission_invitation(self):
...
+                        },
+                        'signatures': { 
+                            'param': { 
+                                'items': [
+                                    { 'value': self.venue.get_authors_id(number='${7/content/noteNumber/value}'), 'optional': True }, 
+                                    { 'value': self.venue.get_program_chairs_id(), 'optional': True }
+                                ] 
+                            }
+                        },
+                        'readers': ['${{2/note/id}/readers}'],
```
celestemartinez, line 275: @melisabok the readers of this edit will be all note readers. Is this correct or should it be PCs and authors only?

melisabok: can you just not edit the note.readers?

I think the invitation should only change the property `ddate` of the note, right?

celestemartinez: Well this is not changing the note readers, it's just setting the note readers as the readers of this specific edit, right? The note itself in the invitation only contains the id of the note and the ddate.

melisabok: ohhhh I see, then yes, let's set the PCs and authors

openreview/venue/invitation.py
```diff
@@ -2096,7 +2221,16 @@ def set_submission_revision_invitation(self, submission_revision_stage=None):
             invitation.edit['invitation']['duedate'] = revision_duedate
 
         if revision_expdate:
-            invitation.edit['invitation']['expdate'] = revision_expdate        
+            invitation.edit['invitation']['expdate'] = revision_expdate
+
+        if revision_stage.allow_deletion:
```
melisabok, line 2226: do we still need this change?

openreview/venue/process/invitation_edit_process.py
```diff
@@ -95,6 +95,8 @@ def update_note_readers(submission, paper_invitation):
 
         def updated_content_readers(note, paper_inv):
             updated_content = {}
+            if 'content' not in paper_inv.edit['note']:
```
celestemartinez, line 98: I am checking if the invitation has a content here, if not then the date process does not try to update the note.

openreview/venue/invitation.py
celestemartinez: Removed.

openreview/venue/process/invitation_edit_process.py
melisabok: looks good

melisabok:
Adding Emilia as a reviewer so she can be aware of these new changes.

enrubio (approved):

PR #1999, opened 2024-01-29
allow ACs to delete pending invite assignment edges

celestemartinez:


openreview/venue/process/invite_assignment_pre_process.js
```diff
@@ -8,10 +8,21 @@ async function process(client, edge, invitation) {
...
   const assignmentLabel = invitation.content.assignment_label?.value
   const inviteLabel = invitation.content.invite_label?.value
+  const acceptedLabel = invitation.content.accepted_label?.value
+  const declinedLabel = invitation.content.declined_label?.value
   const conflictPolicy = invitation.content.reviewers_conflict_policy?.value
   const conflictNYears = invitation.content.reviewers_conflict_n_years?.value
 
   if (edge.ddate) {
+    if (edge.label === acceptedLabel) {
+      return Promise.reject(new OpenReviewError({ name: 'Error', message: `Edge cannot be removed as the user has already accepted the invitation.` }))
```
melisabok, line 18: I would change the error message to something more use friendly: "Can not remove the invitation because the invitations was already accepted".

Instead of having an if for each status, can we show a parametrized one? there might be in the future another states so I would prefer to check if the edge.label != "Invitation Sent" then throw error

melisabok (approved):

PR #2000, opened 2024-01-29
Fix/paper matching setup activation date

### LINKED ISSUE #1829: Paper Matching Setup invitation activation date is not using the Abstract Registration deadline
The ML4H venue's abstract registration deadline has passed and they want to run the Paper Matching Setup however the invitation's activation date is reflecting the Submission Deadline and not the Abstract Registration Deadline. Venues should be able to run the Paper Matching Setup after the abstract registration deadline passes.

enrubio:
- Fixes #1829

Paper matching setup invitation activates on abstract deadline if it exists, otherwise uses submission deadline.


openreview/venue_request/process/revisionProcess.py
```diff
@@ -19,6 +19,13 @@ def process(client, note, invitation):
 
         if invitation_type == 'Revision':
             conference.create_submission_stage()
+            abstract_deadline = forum_note.content.get('abstract_registration_deadline', '').strip()
```
melisabok, line 22: instead of parsing the date again, you can get the value from the SubmissionStage object, right?

conference.submission_stage.second_due_date o conference.submission_stage.duedate

tests/test_iclr_conference_v2.py
```diff
@@ -270,6 +270,18 @@ def test_post_submission(self, client, openreview_client, helpers):
         assert submissions[0].readers == ['everyone']
         assert '_bibtex' in submissions[0].content
         assert 'author={Anonymous}' in submissions[0].content['_bibtex']['value']
+        
+        # Assert that activation date of paper matching setup inv == abstract deadline
+        request_form=pc_client.get_notes(invitation='openreview.net/Support/-/Request_Form')[0]
```
melisabok, line 275: you have the value of `abstract_date` in your test, you can use it to assert the matching_invitation.cdate. 

Otherwise you are running the same code you that are trying to test.

openreview/venue_request/process/revisionProcess.py
enrubio: I didn't know that. I will look into `second_due_date` and `due_date`, I'm a bit confused by them. I thought they were the submission deadline, how can it be the abstract?

tests/test_iclr_conference_v2.py
enrubio: Ah I see now, it's redundant. I'll fix it.

openreview/venue_request/process/revisionProcess.py
melisabok: @celestemartinez do you mind to clarify how is the mapping of the abstract deadline with the second_due_date and due_date values?

you can find the login here: https://github.com/openreview/openreview-py/blob/master/openreview/conference/helpers.py#L99

I think the matching invitation cdate should always match with the due_date that for venues with one single deadline that is the full paper deadline and for venue with two deadlines that is the abstract deadline.

celestemartinez: I know this is kind of confusing, but this is how we are getting the dates from the request form:

if the venue has abstract and submission deadline, then:
- `submission_stage.due_date = request_form.content['abstract_registration_deadline']`
- `submission_stage.second_due_date = request_form.content['Submission Deadline']`

if the venue only has a submission deadline, then:
- `submission_stage.due_date = request_form.content['Submission Deadline']`
- `submission_stage.second_due_date = None`

So @melisabok  is correct: we should always use `submission_stage.due_date`, because for venue with one deadline this represents the submission deadline, and for venues with two deadlines this represents the abstract deadline.

enrubio: Thank you Celeste, I understand these dates now!

tests/test_iclr_conference_v2.py
melisabok: You already have an instance of request_form in line 255: https://github.com/openreview/openreview-py/pull/2000/files#diff-28496bc067ac20c984a751f05b5894ae94a1323eb2aa4b23b7c12a60bda1e2a1L225

enrubio: @melisabok Yes but the request form is updated right after. I wanted to test the activation date after the abstract deadline had been set to the past, but I can move it?

melisabok (approved):
🥇

celestemartinez (approved):

PR #2003, opened 2024-01-31
Fix/ Custom Stage: set note id to id of note being revised

celestemartinez:
Only for the case where `reply_type=='revision`, set the id of the note to the id of the note (review or metareview) being edited. For all other cases, we still use `withInvitation`


celestemartinez:
I think I need to edit the Meta_Review_SAC_Revision as well to have the metareview as the id, right?https://github.com/openreview/openreview-py/blob/5774338bf2556eaf8ce9887d93db042fd5f91b1b/openreview/venue/invitation.py#L752

melisabok (approved):

PR #2010, opened 2024-02-07
Fix/disable reviewers proposed assignment title

### LINKED ISSUE #1864: Remove reviewers_proposed_assignment_title when assignments are deployed
This issue came up in the ICLR 2024 Conference. After the assignments were deployed, the reviewers_proposed_assignment_title was not deleted from group.content and the edge browser link was not updated.

enrubio:
- Fixes #1864


openreview/venue/matching.py
```diff
@@ -1165,6 +1165,23 @@ def deploy_assignments(self, assignment_title, overwrite):
             reviewer_name = venue.area_chairs_name
             review_name = 'Meta_Review'
 
+        # Remove reviewers_proposed_assignment_title if deploying reviewer assignments
```
melisabok, line 1168: I would add this change after we confirm the assignments are deployed, this means after the line `openreview.tools.post_bulk_edges(client=client, edges=assignment_edges)`

openreview/venue/matching.py
```diff
@@ -1165,6 +1165,23 @@ def deploy_assignments(self, assignment_title, overwrite):
...
+        if reviewer_name == venue.reviewers_name:
```
melisabok, line 1169: I think it is better to use `self.is_reviewer` to check if we are deploying reviewer assignments

openreview/venue/matching.py
```diff
@@ -1165,6 +1165,23 @@ def deploy_assignments(self, assignment_title, overwrite):
...
+            venue_group = client.get_group(venue.id)
+            if 'reviewers_proposed_assignment_title' in venue_group.content:
```
melisabok, line 1171: I don't think it is needed to check if the property exists to set the delete true, can you try just posting the edit?

melisabok (approved):
💯

celestemartinez (approved):

PR #2017, opened 2024-02-13
Test: add review rating to CVPR test

celestemartinez:
This PR:
- adds secondary ACs support to custom stage
- adds review rating to CVPR test

### DIFF
--- a/openreview/stages/venue_stages.py
+++ b/openreview/stages/venue_stages.py
@@ -1387,9 +1387,8 @@ class CustomStage(object):
     def get_noninvitees(self, conference, number):
         noninvitees = []
 
-        if conference.use_area_chairs and self.Participants.AREA_CHAIRS_ASSIGNED in self.invitees:
-            if conference.use_secondary_area_chairs and self.Participants.SECONDARY_AREA_CHAIRS not in self.invitees:
-                noninvitees.append(conference.get_secondary_area_chairs_id(number))
+        if conference.use_area_chairs and self.Participants.AREA_CHAIRS_ASSIGNED in self.invitees and conference.use_secondary_area_chairs and self.Participants.SECONDARY_AREA_CHAIRS not in self.invitees:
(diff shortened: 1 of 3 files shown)


openreview/stages/venue_stages.py
```diff
@@ -1379,6 +1383,15 @@ def get_invitees(self, conference, number):
             invitees.append(conference.get_ethics_reviewers_id(number))
 
         return invitees
+    
+    def get_noninvitees(self, conference, number):
+        noninvitees = []
+
+        if conference.use_area_chairs and self.Participants.AREA_CHAIRS_ASSIGNED in self.invitees:
```
melisabok, line 1390: this is can one single if, right?

openreview/stages/venue_stages.py
```diff
@@ -1411,6 +1424,15 @@ def get_readers(self, conference, number):
             readers.append('${3/signatures}')
 
         return readers
+    
+    def get_nonreaders(self, conference, number):
+        nonreaders = []
+
+        if conference.use_area_chairs and self.Participants.AREA_CHAIRS_ASSIGNED in self.readers:
```
melisabok, line 1431: so secondary area chairs won't see the ratings? are we sure about this?

celestemartinez: Well if we want Secondary ACs to be readers of ratings, we would only need to add them to readers in the CustomStage

melisabok:
I think you need to fix the test and use a string for the rating instead of 0

celestemartinez:
> I think you need to fix the test and use a string for the rating instead of 0

Yes! Sorry, I was using this to test Carlos' PR. This is fixed.

openreview/stages/venue_stages.py
melisabok: do you want to unify this in a single if?

melisabok (approved):

PR #2021, opened 2024-02-15
Feature/Venue: parametrize review stage

celestemartinez:
This PR adds two params to the Review Stage:
- `source_submissions_query`: the query by which to filter papers
- `child_invitations_name`: the name of the children invitations, since it can be different from the super invitation name

`invitation_edit_process.py` filters papers by source_submissions_query, and creates/updates official invitations only for these papers. 

Requires https://github.com/openreview/openreview-api/pull/362

### DIFF
--- a/openreview/stages/venue_stages.py
+++ b/openreview/stages/venue_stages.py
@@ -540,7 +540,7 @@ class ReviewStage(object):
         rating_field_name = 'rating',
         confidence_field_name = 'confidence',
         process_path = None,
-        content_query = {},
+        source_submissions_query = {},
         child_invitations_name = 'Official_Review'
     ):
 
@@ -560,7 +560,7 @@ class ReviewStage(object):
         self.rating_field_name = rating_field_name
         self.confidence_field_name = confidence_field_name
         self.process_path = process_path
-        self.content_query = content_query
+        self.source_submissions_query = source_submissions_query
         self.child_invitations_name = child_invitations_name
 
     def _get_reviewer_readers(self, conference, number, review_signature=None):
--- a/openreview/stages/venue_stages.py
+++ b/openreview/stages/venue_stages.py
@@ -540,7 +540,8 @@ class ReviewStage(object):
         rating_field_name = 'rating',
         confidence_field_name = 'confidence',
         process_path = None,
-        content_query = {}
+        content_query = {},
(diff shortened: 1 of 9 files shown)


openreview/stages/venue_stages.py
```diff
@@ -539,7 +539,9 @@ def __init__(self,
         remove_fields = [],
         rating_field_name = 'rating',
         confidence_field_name = 'confidence',
-        process_path = None
+        process_path = None,
+        content_query = {},
+        child_invitations_name = 'Official_Review'
```
melisabok, line 544: I think `name` should represent the child invitation name.

openreview/stages/venue_stages.py
```diff
@@ -539,7 +539,9 @@ def __init__(self,
         remove_fields = [],
         rating_field_name = 'rating',
         confidence_field_name = 'confidence',
-        process_path = None
+        process_path = None,
+        content_query = {},
```
melisabok, line 543: I'm not sure about calling `content_query`, it doesn't indicate that is the source of submissions that we want to create the review invitations.

what about:

- selected_submissions
- selected_submissions_query
- selected_submissions_content_query
- forums_query
- selected_forums
- replyto_query
...

openreview/stages/venue_stages.py
melisabok: we should have a parent_name that is optional and by default it is the name value

openreview/venue/process/invitation_edit_process.py
```diff
@@ -65,6 +67,10 @@ def get_children_notes():
             if source == 'flagged_for_ethics_review':
                 source_submissions = [s for s in source_submissions if s.content.get('flagged_for_ethics_review', {}).get('value', False)]
 
+        if content_query:
+            for key, value in content_query.items():
+                source_submissions = [s for s in source_submissions if value in s.content.get(key, {}).get('value', '')]
```
melisabok, line 72: should we use `in` or `==`?

melisabok:
The changes look good, it is very simple. I made some comments about naming conventions.

openreview/venue/process/invitation_edit_process.py
melisabok: I think in is better, it will work for strings and arrays.

PR #2038, opened 2024-03-05
Fix/ Venue: replace note readers

celestemartinez:
If the edit copies the readers from note.readers, replace this value in `invitation_edit_process.py`

### DIFF
--- a/openreview/venue/process/invitation_edit_process.py
+++ b/openreview/venue/process/invitation_edit_process.py
@@ -108,6 +108,8 @@ def process(client, invitation):
         for note in notes:
             final_invitation_readers = list(dict.fromkeys([note.signatures[0] if 'signatures' in r else r for r in invitation_readers]))
             edit_readers = list(dict.fromkeys([note.signatures[0] if 'signatures' in r else r for r in paper_invitation.edit.get('readers',[])]))
+            if '${2/note/readers}' in edit_readers:
+                edit_readers = final_invitation_readers
             updated_content = updated_content_readers(note, paper_invitation)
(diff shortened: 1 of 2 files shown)


openreview/venue/process/invitation_edit_process.py
```diff
@@ -108,6 +108,8 @@ def updated_content_readers(note, paper_inv):
         for note in notes:
             final_invitation_readers = list(dict.fromkeys([note.signatures[0] if 'signatures' in r else r for r in invitation_readers]))
             edit_readers = list(dict.fromkeys([note.signatures[0] if 'signatures' in r else r for r in paper_invitation.edit.get('readers',[])]))
+            if '${2/note/readers}' in edit_readers:
+                edit_readers = final_invitation_readers
```
melisabok, line 112: why is not `note.readers`?

melisabok: I would use the current note.readers instead of the invitation.note.readers, is there an issue with that?

celestemartinez: the readers of the note could change though, we do this in line 117. If they change, we set them to `final_invitation_readers`

```
if final_invitation_readers and note.readers != final_invitation_readers:
    updated_note.readers = final_invitation_readers
    updated_note.nonreaders = paper_invitation.edit['note'].get('nonreaders')
```

melisabok: can `final_invitation_readers` be empty?

melisabok (approved):

Issue #2015, opened 2024-02-13
Paper Matching Setup: check conflicts are not computed for SACs


celestemartinez:
We don't compute conflicts between ACs and SACs. Add this check to the paper matching setup preprocess.

Issue #2016, opened 2024-02-13
Paper Matching Setup: don't let run more than one setup per group at the same time


melisabok:
PCs should wait until the setup matching is complete to run a new one for the same group id. 

I think they get anxious because they don't see a status so they run it and this takes a lot of resources of our expertise API. Also there could be race condition issues when saving the edges.

I suggest to add a validation in the pre process function that checks if there is a setup matching note with no status and throw an error. @celestemartinez @enrubio thoughts?

celestemartinez:
This is a good idea. We would only need to check the last posted `Paper_Matching_Setup` for that `matching_group` and check if there is a `Paper_Matching_Setup_Status` note posted already.

PR #1843, opened 2023-09-20
Paper Matching Setup: let the user select the expertise model to be used to compute affinit…

melisabok:
…y scores

It depends on: https://github.com/openreview/openreview-expertise/pull/161

PR #2041, opened 2024-03-07
Don't allow PCs to compute affinity scores if there are over 2k submissions

### LINKED ISSUE #2022: Paper Matching Setup: don't let compute affinity scores to venues with more than 2k submissions
They should contact us and we will compute them for them. 

Only for computing scores between submissions and another group.

For SAC, we compute scores between groups so we don't need to validate the size, at least for now.

enrubio:
- Fixes #2022

Edit:
Decided to fix the other paper matching issues in separate PRs.

### DIFF
--- a/openreview/venue_request/process/matching_pre_process.py
+++ b/openreview/venue_request/process/matching_pre_process.py
@@ -1,4 +1,21 @@
 def process(client, note, invitation):
+    _, baseurl_v2 = openreview.tools.get_base_urls(client)
+    client_v2 = openreview.api.OpenReviewClient(baseurl=baseurl_v2, token=client.token)
+    forum_note = client.get_note(note.forum)
+    venue_id = forum_note.content.get('venue_id', '')
+    domain = client_v2.get_group(venue_id)
+    submission_name = domain.content.get('submission_name', '').get('value')
+    senior_area_chairs_name = domain.content.get('senior_area_chairs_name', {}).get('value')
 
-    if 'Yes' in note.content['compute_affinity_scores'] and 'upload_affinity_scores' in note.content:
+    submissions_over_2000 = client_v2.get_notes(invitation=f'{venue_id}/-/{submission_name}', offset=2000)
+    compute_affinity_scores = note.content.get('compute_affinity_scores') == 'Yes'
+
+    matching_group = note.content['matching_group']
+    role_name = matching_group.split('/')[-1]
+    pretty_role = role_name.replace('_', ' ')
+
+    if compute_affinity_scores and 'upload_affinity_scores' in note.content:
         raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
+
+    if compute_affinity_scores and len(submissions_over_2000) and role_name != senior_area_chairs_name:
+        raise openreview.OpenReviewException(f'Can not compute affinity scores between {pretty_role} and 2000+ papers. Please contact us at info@openreview.net to compute your scores.')
--- a/openreview/venue_request/process/matching_pre_process.py
+++ b/openreview/venue_request/process/matching_pre_process.py
@@ -12,9 +12,9 @@ def process(client, note, invitation):
     if forum_note.content.get('api_version', '1') == '2':
         matching_group = note.content['matching_group']
         senior_area_chairs_name = domain.get_content_value('senior_area_chairs_name')
-        if matching_group.endswith(senior_area_chairs_name):
+        if senior_area_chairs_name and matching_group.endswith(senior_area_chairs_name):
             return
-        
+
         submission_venue_id = domain.get_content_value('submission_venue_id')
         _, num_submissions = client_v2.get_notes(content={ 'venueid':submission_venue_id }, limit=1, with_count=True)
         if compute_affinity_scores and num_submissions >= 2000:
--- a/openreview/venue_request/process/matching_pre_process.py
+++ b/openreview/venue_request/process/matching_pre_process.py
@@ -4,18 +4,17 @@ def process(client, note, invitation):
     forum_note = client.get_note(note.forum)
     venue_id = forum_note.content.get('venue_id', '')
     domain = client_v2.get_group(venue_id)
-    submission_name = domain.content.get('submission_name', '').get('value')
+    submission_venue_id = domain.content['submission_venue_id']['value']
     senior_area_chairs_name = domain.content.get('senior_area_chairs_name', {}).get('value')
-
-    submissions_over_2000 = client_v2.get_notes(invitation=f'{venue_id}/-/{submission_name}', offset=2000)
     compute_affinity_scores = note.content.get('compute_affinity_scores') == 'Yes'
-
-    matching_group = note.content['matching_group']
-    role_name = matching_group.split('/')[-1]
-    pretty_role = role_name.replace('_', ' ')
 
     if compute_affinity_scores and 'upload_affinity_scores' in note.content:
         raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
 
-    if compute_affinity_scores and len(submissions_over_2000) and role_name != senior_area_chairs_name:
+    matching_group = note.content['matching_group']
+    role_name = matching_group.split('/')[-1]
+    pretty_role = role_name.replace('_', ' ')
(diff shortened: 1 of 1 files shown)


openreview/venue_request/process/matching_pre_process.py
```diff
@@ -1,4 +1,21 @@
...
+    forum_note = client.get_note(note.forum)
+    venue_id = forum_note.content.get('venue_id', '')
+    domain = client_v2.get_group(venue_id)
+    submission_name = domain.content.get('submission_name', '').get('value')
+    senior_area_chairs_name = domain.content.get('senior_area_chairs_name', {}).get('value')
 
-    if 'Yes' in note.content['compute_affinity_scores'] and 'upload_affinity_scores' in note.content:
-        raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
\ No newline at end of file
+    submissions_over_2000 = client_v2.get_notes(invitation=f'{venue_id}/-/{submission_name}', offset=2000)
```
melisabok, line 10: Check get_all_notes and the parameter `with_count`, it can help you to get the count without using the offset.

openreview/venue_request/process/matching_pre_process.py
```diff
@@ -1,4 +1,21 @@
...
+    compute_affinity_scores = note.content.get('compute_affinity_scores') == 'Yes'
+
+    matching_group = note.content['matching_group']
+    role_name = matching_group.split('/')[-1]
+    pretty_role = role_name.replace('_', ' ')
+
+    if compute_affinity_scores and 'upload_affinity_scores' in note.content:
```
melisabok, line 17: I would leave this if before checking the submission count that is needed later.

melisabok:
Could you please fix this issue in this PR too:https://github.com/openreview/openreview-py/issues/2016

it will require changing the preprocess function too.

openreview/venue_request/process/matching_pre_process.py
celestemartinez: I would use the venueid here to get the active submissions, using the invitation means we will get withdrawn/desk-rejected as well

enrubio:
Yes, I will also fix #2015 in this PR

openreview/venue_request/process/matching_pre_process.py
enrubio: @melisabok for this call, I'm worried about venues with a large number of submissions and that it could take several seconds before PCs receive any feedback. It seems like `offset` with `get_notes()` is significantly faster and gets a response in < 1s. What do you think?

melisabok: ok! I thought there was a call to get 1 only paper and get the count. 

For example: api2.openreview.net/notes?invitation=ICML.cc/2023/Conference/-/Submission&limit=1

you can see the count in the response. It seems the python function doesn't read that data. 

you can implement a new function called `get_notes_count` and read the count from the json response.

enrubio: Ahh okay I see what you mean now. Then I think we can do:
`client_v2.get_notes(content={ 'venueid':submission_venue_id }, limit=1, with_count=True)`
which returns 1 submission note & the total count. I didn't realize we could do it this way.

If we can get count like this, do you still think a `get_notes_count()` function is needed?

melisabok: you can use the first option, no need to implement a new function.

openreview/venue_request/process/matching_pre_process.py
```diff
@@ -1,4 +1,20 @@
 def process(client, note, invitation):
+    _, baseurl_v2 = openreview.tools.get_base_urls(client)
+    client_v2 = openreview.api.OpenReviewClient(baseurl=baseurl_v2, token=client.token)
+    forum_note = client.get_note(note.forum)
+    venue_id = forum_note.content.get('venue_id', '')
+    domain = client_v2.get_group(venue_id)
+    submission_venue_id = domain.content['submission_venue_id']['value']
+    senior_area_chairs_name = domain.content.get('senior_area_chairs_name', {}).get('value')
+    compute_affinity_scores = note.content.get('compute_affinity_scores') == 'Yes'
```
melisabok, line 9: can you check the value `No` instead? we are going to add more options that represent the affinity model but "No" will stay: https://github.com/openreview/openreview-py/pull/1843/files

openreview/venue_request/process/matching_pre_process.py
```diff
@@ -1,4 +1,20 @@
...
 
-    if 'Yes' in note.content['compute_affinity_scores'] and 'upload_affinity_scores' in note.content:
-        raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
\ No newline at end of file
+    if compute_affinity_scores and 'upload_affinity_scores' in note.content:
```
melisabok, line 11: I guess you are missing `not compute_affinity_scores` is it is reading the yes value.

openreview/venue_request/process/matching_pre_process.py
```diff
@@ -1,4 +1,20 @@
...
+        raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
+
+    matching_group = note.content['matching_group']
+    role_name = matching_group.split('/')[-1]
```
melisabok, line 15: instead of parsing the matching_group can you check `if matching.group.ends_with(senior_area_chairs_name)`?

openreview/venue_request/process/matching_pre_process.py
```diff
@@ -1,4 +1,20 @@
...
+    pretty_role = role_name.replace('_', ' ')
+    _, num_submissions = client_v2.get_notes(content={ 'venueid':submission_venue_id }, limit=1, with_count=True)
```
melisabok, line 17: you don't need to get the notes if the role is SAC, so you can check:

```
if matching.group.ends_with(senior_area_chairs_name):
  return
...get_notes()
```

openreview/venue_request/process/matching_pre_process.py
```diff
@@ -1,4 +1,20 @@
...
-    if 'Yes' in note.content['compute_affinity_scores'] and 'upload_affinity_scores' in note.content:
-        raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
\ No newline at end of file
+    if compute_affinity_scores and 'upload_affinity_scores' in note.content:
+        raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
+
+    if forum_note.content.get('api_version', '1') == '2':
+        matching_group = note.content['matching_group']
+        senior_area_chairs_name = domain.get_content_value('senior_area_chairs_name')
+        if senior_area_chairs_name and matching_group.endswith(senior_area_chairs_name): return
```
melisabok, line 15: There is no Python convention about this but our code put the `return` statement in a new line. We would like to keep consistency in the code style.

melisabok:
The changes look good!

openreview/venue_request/process/matching_pre_process.py
```diff
@@ -1,4 +1,21 @@
...
-    if 'Yes' in note.content['compute_affinity_scores'] and 'upload_affinity_scores' in note.content:
-        raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
\ No newline at end of file
+    if compute_affinity_scores and 'upload_affinity_scores' in note.content:
+        raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
+
+    if forum_note.content.get('api_version', '1') == '2':
+        matching_group = note.content['matching_group']
+        senior_area_chairs_name = domain.get_content_value('senior_area_chairs_name')
+        if matching_group.endswith(senior_area_chairs_name):
```
melisabok, line 15: I'm not sure why you deleted the check that `senior_area_chairs_name` is not None.

enrubio: Oops, I thought the senior_area_chairs_name was always a part of the domain and thought it was redundant. Fixed now.

melisabok: senior_area_chairs_name is only present if the venue has SAC.

melisabok (approved):

celestemartinez (approved):

PR #2042, opened 2024-03-08
Fix/ Venue: check fields are in content before removing

celestemartinez:
The Submission Revision Stage would give an error if the PCs select `authors` and `authorids` in the remove options dropdown and also select `Do not allow any changes to author lists`.

Now that I think about it, it might be best to just remove `authors` and `authorids` from the remove options dropdown and control these two fields solely from the `submission_author_edition` field.  This might be less confusing to PCs. 

What do you think, @melisabok , @enrubio ?


melisabok (approved):
I agree.

enrubio (approved):

PR #2043, opened 2024-03-12
Avoid passing null to ids

carlosmondra:
This PR does not pass the id to entities if they are None.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ parameters:
     default: "master"
   openreview-api-v2-branch:
     type: string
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ parameters:
     default: "master"
   openreview-api-v2-branch:
     type: string
-    default: "main"
+    default: "feature/update-openapi-docs"
```
carlosmondra, line 11:
```suggestion
    default: "main"
```

melisabok (approved):

PR #2044, opened 2024-03-12
Support type 'profile' for authorids

melisabok:
It depends on https://github.com/openreview/openreview-api/pull/380

This PR is ready to be merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ parameters:
     default: "master"
   openreview-api-v2-branch:
     type: string
-    default: "main"
+    default: "fix/validate-group-array"
 
 jobs:
   build:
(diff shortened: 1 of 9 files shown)


xkopenreview:
currently type group and group[] are rendered as profile search widget
for this to continue working i will need to change that check to profile and profile[]

does this change mean that type of a field can't be group and group[] any more?
or group and group[] should still be rendered as profile search widget or some other widget?

melisabok:
it can be both, group or profile. Having profile helps the API to validate the authorids are valid profileids or emails. 

This is done to fix the issue when ECCV submitted the papers and use a fake tilde id as authorids. 

I do think the UI could validate both types.

xkopenreview:
so both profile and group will render as profile search widget right?

melisabok:
> so both profile and group will render as profile search widget right?

Yes.

PR #2047, opened 2024-03-14
Paper matching: Don't allow multiple setup matchings per group if process is ongoing

### LINKED ISSUE #2016: Paper Matching Setup: don't let run more than one setup per group at the same time

enrubio:
- Fixes #2016

Tested in my local by running multiple paper matching setups in a large venue:
paper-matching

### DIFF
--- a/openreview/venue_request/process/matching_pre_process.py
+++ b/openreview/venue_request/process/matching_pre_process.py
@@ -2,3 +2,20 @@ def process(client, note, invitation):
 
     if 'Yes' in note.content['compute_affinity_scores'] and 'upload_affinity_scores' in note.content:
         raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
+
+    matching_group = note.content['matching_group']
+    matching_notes = client.get_all_notes(invitation=invitation.id, sort='tmdate:desc')
+
+    if matching_notes:
+        latest_matching_note = None
+        for i, matching_note in enumerate(matching_notes):
+            if matching_note.content['matching_group'] == matching_group:
+                latest_matching_note = matching_notes[i]
+                break
+
+        if latest_matching_note:
+            status_inv = invitation.id.replace('Paper_Matching_Setup', 'Paper_Matching_Setup_Status')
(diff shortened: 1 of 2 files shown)


openreview/venue_request/process/matching_pre_process.py
```diff
@@ -1,4 +1,21 @@
...
-        raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
\ No newline at end of file
+        raise openreview.OpenReviewException('Either upload your own affinity scores or select affinity scores computed by OpenReview')
+
+    matching_group = note.content['matching_group']
+    matching_notes = client.get_all_notes(invitation=invitation.id, sort='tmdate:desc')
+
+    if matching_notes:
+        latest_matching_note = None
+        for i, matching_note in enumerate(matching_notes):
```
melisabok, line 11: I don't think you need to use the index here:

```
for matching_note in matching_notes:
       if ....:
           latest_matching_note  =matching_note
```

A nicer way to do this and to avoid the `break` sentence is to use a lambda function:

```
matching_group_notes = [matching_note for matching_note in matching_notes if matching_note.content['matching_group'] == 'matching_group']

if matching_group_notes:
    status_note = client.get_all_notes(invitation= invitation.id.replace('Paper_Matching_Setup', 'Paper_Matching_Setup_Status'), replyto=matching_group_notes[0].id)
    if not status_note:
       ....
```

    

```

melisabok:
There are some conflicts now.

openreview/venue_request/process/matching_pre_process.py
enrubio: You're right, that is much better. Updated now.

melisabok (approved):
💯

celestemartinez (approved):

Issue #1664, opened 2023-04-27
Venue: decision stage should let the PCs decide which decision options are representing an acceptance


melisabok:
Currently the decision option that contains the work "Accept" is considered an acceptance. But some venues want to decide their own labels without the word "Accept".

@celestemartinez is suggesting that PCs should indicate this while configuring the decision stage or post decision stage? there should be a dropdown where they can select the decision options considered as acceptance.

enrubio:
Fixed in #2115

PR #2050, opened 2024-03-19
Decision stage: Update instructions

enrubio:
Update instructions to tell PCs to use "Accept" in the decision names.

### DIFF
--- a/openreview/venue_request/venue_request.py
+++ b/openreview/venue_request/venue_request.py
@@ -721,7 +721,7 @@ class VenueStages():
                 'required': True
             },
             'decision_options': {
-                'description': 'What are the decision options (provide comma separated values, e.g. Accept (Best Paper), Accept, Reject)? Leave empty for default options - "Accept (Oral)", "Accept (Poster)", "Reject"',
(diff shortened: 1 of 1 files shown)


melisabok:
I'm not sure if this is going to help, PCs may not read the instructions. 

I would like to implement this: https://github.com/openreview/openreview-py/issues/1664

If this issue can take too much time, I suggest adding a preprocess validation that checks that are least one decision option has the work "Accept".

racheljsmart:
> I'm not sure if this is going to help, PCs may not read the instructions.
> 
> I would like to implement this: #1664
> 
> If this issue can take too much time, I suggest adding a preprocess validation that checks that are least one decision option has the word "Accept".

Until that issue is implemented, I think adding a preprocess will help.

melisabok:
@enrubio what should we do with this PR?

enrubio:
I can add a preprocess here that checks if there's at least one "Accept" option. I still think we should update the instructions.

melisabok:
What about fixing the issue https://github.com/openreview/openreview-py/issues/1664 directly.

enrubio:
Okay, so no preprocess then. I can work on that issue. We can just close this PR in that case.

PR #2059, opened 2024-03-26
allow setup matching between SACs and submissions

celestemartinez:
We should allow SACs to be assigned to submissions as well, instead of only to ACs.

This are the changes to be made:

- [x] add this as an option in the request form. This should allow the PCs to setup the paper matching between SACs and submissions (compute conflicts and affinity scores)
- [x] add a property in the group.content about assigning ACs or submissions
- [x] change the SAC console to load the AC tab when the property says they are assigning ACs.
- [x] Create bid invitation for SACs to bid on papers (make this optional in bid stage form)
- [x] Fix the deploy when SACs should be assigned to papers
- [x] Fix any process functions that expect SACs to be assigned to ACs

@melisabok anything else missing?

Requires https://github.com/openreview/openreview-web/pull/1876


tests/test_sac_paper_matching.py
```diff
@@ -0,0 +1,215 @@
...
+        helpers.await_queue()
+
+        comment_invitation_id = f'openreview.net/Support/-/Request{request_form.number}/Paper_Matching_Setup_Status'
+        matching_status = client.get_notes(invitation=comment_invitation_id, replyto=matching_setup_note.id, forum=request_form.forum, sort='tmdate')[0]
+        assert matching_status
+        assert matching_status.content['comment'] == '''Affinity scores and/or conflicts were successfully computed. To run the matcher, click on the 'Senior Area Chairs Paper Assignment' link in the PC console: https://openreview.net/group?id=TSACM/2024/Conference/Program_Chairs
+
+Please refer to the documentation for instructions on how to run the matcher: https://docs.openreview.net/how-to-guides/paper-matching-and-assignment/how-to-do-automatic-assignments'''
+
+        scores_invitation = openreview_client.get_invitation('TSACM/2024/Conference/Senior_Area_Chairs/-/Affinity_Score')
```
melisabok, line 207: can you check the Assignment_Configuration invitation is also set properly?

celestemartinez:
> > Create bid invitation for SACs to bid on papers
> 
> I'm not sure if we should make this optional in the BidStage form.

I was planning on asking PCs in the bid stage form if they would like to enable it for SACs as well. I also don't think this should be the default. 

> 
> Missing:
> 
> 1. add a property in the group.content about assigning ACs or submissions
> 2. change the SAC console to load the AC tab when the property says they are assigning ACs.

I've added this two to the description. The first one is already done.

openreview/venue/process/sac_paper_assignment_pre_process.js
```diff
@@ -0,0 +1,21 @@
...
+    const domain = groups[0]
+    const venueId = domain.id
+    const submissionName = domain.content.submission_name?.value
+    const sacsName = invitation.content.reviewers_name?.value
+
+    const { notes } = await client.getNotes({ id: edge.head })
+    const submission = notes[0]
+    const submissionGroupId = `${venueId}/${submissionName}${submission.number}`
+
+    if (!edge.ddate) {
```
melisabok, line 14: do we have this validation with the other groups?

celestemartinez: Yes, the [default preprocess](https://github.com/openreview/openreview-py/blob/master/openreview/venue/process/assignment_pre_process.js) checks this but also checks if there is a review/metareview posted or if the signatory group does not exist, but neither of those apply to SACs. 
Do you think this should be the same preprocess for all and we check if it's an SAC?

openreview/venue/matching.py
```diff
@@ -307,9 +309,7 @@ def _create_edge_invitation(self, edge_id, any_tail=False, default_label=None):
 
     def _build_conflicts(self, submissions, user_profiles, get_profile_info, compute_conflicts_n_years):
         if self.alternate_matching_group:
-            other_matching_group = self.client.get_group(self.alternate_matching_group)
-            other_matching_profiles = tools.get_profiles(self.client, other_matching_group.members, with_publications=True, with_relations=True)
-            return self._build_profile_conflicts(other_matching_profiles, user_profiles, compute_conflicts_n_years)
+            return
```
haroldrubio, line 312: Sorry if I'm missing where this is happening now - how are profile conflicts being built now?

celestemartinez: Profile conflicts are not being computed anymore. We don't allow SAC-AC conflicts to be computed from the request form. 
Do you need conflicts to be computed between two groups?

haroldrubio: Ohhh no I don't I didn't know we were deprecating that from the request form, okay that makes sense!

celestemartinez: I guess we could leave support for this just in case, but I removed it for now because I don't think the code is working. I would need to go over the code and fix it.

openreview/venue/process/sac_paper_assignment_pre_process.js
```diff
@@ -0,0 +1,21 @@
+async function process(client, edge, invitation) {
```
melisabok, line 1: can we just use the same assignment_pre_process that we use for the other groups?

tests/test_arr_venue_v2.py
```diff
@@ -1983,30 +1985,31 @@ def test_setup_matching(self, client, openreview_client, helpers, test_client, r
...
+        helpers.await_queue()
 
         assert openreview_client.get_invitation('aclweb.org/ACL/ARR/2023/August/Senior_Area_Chairs/-/Conflict')
         assert openreview_client.get_invitation('aclweb.org/ACL/ARR/2023/August/Senior_Area_Chairs/-/Affinity_Score')
-        assert openreview_client.get_invitation('aclweb.org/ACL/ARR/2023/August/Senior_Area_Chairs/-/Proposed_Assignment')
-        assert openreview_client.get_invitation('aclweb.org/ACL/ARR/2023/August/Senior_Area_Chairs/-/Assignment')
+        proposed_inv = openreview_client.get_invitation('aclweb.org/ACL/ARR/2023/August/Senior_Area_Chairs/-/Proposed_Assignment')
+        assert proposed_inv
+        assert proposed_inv.edit['head']['param']['type'] == 'note'
+        assert proposed_inv.edit['head']['param']['withVenueid'] == 'aclweb.org/ACL/ARR/2023/August/Submission'
```
melisabok, line 2012: can the ARR test deploy the SAC-paper assignments?

tests/test_sac_paper_matching.py
```diff
@@ -0,0 +1,448 @@
...
+
+        openreview_client.add_members_to_group(
+            'TSACM/2024/Conference/Senior_Area_Chairs', 
+            ['~SAC_MatchingOne1', '~SAC_MatchingTwo1', '~SAC_MatchingThree1']
+        )
+
+        # enable bidding
+        now = datetime.datetime.utcnow()
+        due_date = now + datetime.timedelta(days=1)
+        bid_stage_note = pc_client.post_note(openreview.Note(
```
melisabok, line 183: is the bidding interface showing the papers sorted by affinity scores?

maybe we should upload the scores before starting the bidding to check this.

openreview/venue/invitation.py
```diff
@@ -2563,6 +2563,24 @@ def set_assignment_invitation(self, committee_id, submission_content=None):
...
+            edge_nonreaders = [venue.get_authors_id(number='${{2/head}/number}')]
+            preprocess=self.get_process_content('process/sac_paper_assignment_pre_process.js')
+            content = {
+                'reviewers_id': {
+                    'value': venue.get_senior_area_chairs_id()
+                },
+                'reviewers_name': {
+                    'value': venue.senior_area_chairs_name
+                },
+                'sync_sac_id': {
```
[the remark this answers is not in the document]
celestemartinez: The process function expects the content to have `sync_sac_id`, but I guess we can only add it when needed and instead do:

> sync_sac_id = invitation.content.get('sync_sac_id', {}).get('value')

openreview/venue/process/sac_paper_assignment_pre_process.js
celestemartinez: done

tests/test_arr_venue_v2.py
celestemartinez: It looks like Harold is already doing this: https://github.com/openreview/openreview-py/blob/6543c02305e5ef1a5084df17f46f3403da97d552/tests/test_arr_venue_v2.py#L2562

is this what you mean?

tests/test_sac_paper_matching.py
celestemartinez: I impersonate the SAC and it looks like they are being sorted by affinity scores. 

Should I add this to the test as well?

melisabok: no, that's fine. thanks for checking.

melisabok (approved):

haroldrubio (approved):

PR #2063, opened 2024-03-28
Fix/ARR fields

haroldrubio:
- Removes seconds field from the dates of the Configuration invitation
- Synchronizes configuration date fields with the other request form invitations
- Changes the blind submission reply invitation to a submission revision invitation
- Adds the update string to the process invitation dateprocess dates, only runs if cdate is in the past
- Fixes bug in the SAC assignment process function
- Changes the submission process to expiring and removing the duedate on the blind submission license


melisabok:
When I edit the expdate of the review stage, the date process function is still running twice. It seems we are post the Official_Review invitation twice, there are two edits with the same timestamp in expdate.

haroldrubio:
I think that works for the pre-process but the `review_process_script` will always get overwritten right?

melisabok:
Check this commit: https://github.com/openreview/openreview-py/pull/2063/commits/5fe3b4da885e416f53340bb6b3ae218f5e79fb59

I'm setting the values only once.

melisabok:
I think one solution for this is to store the file paths of the preprocess and process functions in the ReviewStage class. In the ARR class we overwrite them with the ARR values and then we call to the invitation builder to create the invitation. I can try to make this change later. We don’t need it right now so I will approve this PR so we can deploy it soon.

melisabok (approved):

haroldrubio:
Ok that sounds good - this will only really matter after the submission deadline passes and the child invitations are created anyway.

I still have one more change I want to push - it seems edits are still getting posted for the invitations with scripts even if the date is the same, so I'm adding a check to not post edits/notes if that is the case

haroldrubio:
I also think we can wait until Monday to deploy, ARR just needs their submission form to be updated and I don't anticipate many submissions right as submissions open (Monday morning), and I can manually handle the data for any submissions that come in before we deploy next week.

celestemartinez:
We can merge and deploy this today, but @haroldrubio mentioned in the meeting that we can wait until Monday instead. 
I am ok with either, but I guess it depends on if Harold needs it today.

melisabok:
My preference is to wait until Monday if it is possible. Thank you both for the big effort!

celestemartinez (approved):

PR #2065, opened 2024-03-29
Fix/ Venue: read review and comment readers from right invitations

celestemartinez:
Sherry reported an issue when a paper is flagged for Ethics Review. It looks like the we were reading the review readers from the wrong place. I am now reading it from the review invitation itself instead of the ethics flag invitation.


melisabok (approved):

celestemartinez:
There are new changes here:

I save the reviewer and AE checklist names in the ARR ethics flag invitation to process these invitations as well in the Ethics_Flag process function. I deleted the ARR specific process function since it's no longer needed.

cc @haroldrubio

openreview/arr/arr.py
```diff
@@ -422,9 +422,24 @@ def create_custom_stage(self):
...
         self.venue.ethics_review_stage = self.ethics_review_stage
-        self.venue.ethics_review_stage.flag_process_path = '../arr/process/ethics_flag_process.py'
-        
-        return self.venue.create_ethics_review_stage()
+        stage_value = self.venue.create_ethics_review_stage()
+        self.client.post_invitation_edit(
+            invitations=self.venue.get_meta_invitation_id(),
+            signatures=[self.venue_id],
+            replacement=False,
+            invitation=openreview.api.Invitation(
```
celestemartinez, line 430: I am reposting this invitation every time because I need to add some ARR specific values, but this invitation has no date process so nothing runs.

haroldrubio (approved):

PR #2069, opened 2024-04-02
ARR: Fix resubmission bug for fields with only readers

haroldrubio:
This PR adds an extra safety check for fetching resubmissions. There is some behavior with the preprint release invitation that adds readers to optional fields that may not have been in the original submission. The preprint release invitation adds these fields to the submission with only readers and no value.

The resubmission script was just checking for the previous URL to be in the content and then assumes that it has a value, this PR fixes that and also checks for a value

- Adds some extra tests for other fields that should be restricted to authors and PCs
- Moves resubmission logic to helpers.py and calls the function in the tests
- Resubmission logic now checks for 'value' key in the previous_URL field


tests/test_arr_venue_v2.py
```diff
@@ -2213,6 +2229,10 @@ def test_resubmission_and_track_matching_data(self, client, openreview_client, h
 
         helpers.await_queue()
 
+        # Getting resubmissions should pass
+        previous_url_field = 'previous_URL'
+        resubmissions = openreview.arr.helpers.get_resubmissions(submissions, previous_url_field)
```
melisabok, line 2234: using the code that you are testing in your test doesn't test anything :) 

Here in the test you can get the submissions that you know must be resubmissions, like `i == 6`

celestemartinez (approved):

openreview/stages/arr_content.py
```diff
@@ -1007,10 +1007,10 @@
     "preferred_venue",
     "consent_to_share_data",
     "consent_to_share_submission_details",
-    "preprint_status",
     "existing_preprints",
     "Association_for_Computational_Linguistics_-_Blind_Submission_License_Agreement",
-    "section_2_permission_to_publish_peer_reviewers_content_agreement"
+    "section_2_permission_to_publish_peer_reviewers_content_agreement",
+    "reviewing_volunteers"
```
melisabok, line 1013: got it

melisabok (approved):

PR #2070, opened 2024-04-02
ARR: Fix ethics review tests

haroldrubio:
This PR sets the ethics review stage to be active in the future instead of activating it as soon as the tests start. This fixes a bug where all submissions had an ethics review invitation created since the ethics review super invitation was active while the submission process functions ran.

Edited to add: I changed the test to start the ethics review stage from the beginning, to make sure no Ethics_Review invitatinos are created since no papers are flagged.

### DIFF
--- a/openreview/venue/process/submission_process.py
+++ b/openreview/venue/process/submission_process.py
@@ -79,9 +79,9 @@ To view your submission, click here: https://openreview.net/forum?id={note.forum
 
     for venue_invitation in invitation_invitations:
         print('processing invitation: ', venue_invitation.id)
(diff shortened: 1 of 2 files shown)


celestemartinez:
I am a little bit confused about why ethics review invitations were created for all submissions, if it should be created only for flagged submissions? It shouldn't matter when the ethics review super invitation is active.

melisabok:
@celestemartinez should we make that change in this PR? Only create invitations for `all_submissions` source.

celestemartinez:
> @celestemartinez should we make that change in this PR? Only create invitations for `all_submissions` source.

yes, I can work on this

celestemartinez:
This should be ready.

melisabok (approved):

celestemartinez (approved):

PR #2071, opened 2024-04-02
ARR: Allow fields for venue stages to have custom readers

haroldrubio:
The reviewer certification may end up deanonymizing the reviewers so that field is made only visible to the SACs and ACs

- Add a new enum type to the ARRStage class to decide the readers
- Uses the `FIELD_READERS` variable to define which fields get specific readers, this can't be defined in the `arr_content.py` file since it needs to be aware of the venue
- Sets the readers for the field on first time setup of a stage

### DIFF
--- a/openreview/arr/helpers.py
+++ b/openreview/arr/helpers.py
@@ -1143,12 +1143,37 @@ class ARRStage(object):
         PROCESS_INVITATION = 3
         SUBMISSION_REVISION_STAGE = 4
 
+    class Participants(Enum):
+        EVERYONE = 0
+        SENIOR_AREA_CHAIRS = 1
+        SENIOR_AREA_CHAIRS_ASSIGNED = 2
+        AREA_CHAIRS = 3
+        AREA_CHAIRS_ASSIGNED = 4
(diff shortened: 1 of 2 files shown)


openreview/arr/helpers.py
```diff
@@ -1143,13 +1143,38 @@ class Type(Enum):
...
         'Official_Comment': 'Comment_Stage',
         'Ethics_Review': 'Ethics_Review_Stage',
         'Blind_Submission_License_Agreement': 'Submission_Revision_Stage'
     }
+    FIELD_READERS = {
+        'Official_Review': {
+            'content_name': 'additional_review_form_options',
+            'fields': {
+                'reviewer_certification': [
+                    Participants.SENIOR_AREA_CHAIRS_ASSIGNED,
```
melisabok, line 1172: what about the VENUE_ID as reader?

melisabok: I see that is being added by default.

melisabok (approved):

celestemartinez (approved):

PR #2076, opened 2024-04-03
ARR: Update Submission and Max Load Fields for April

haroldrubio:
- Replace field name `maximum_load_resubmission` with `maximum_load_this_cycle_for_resubmissions` for the max load request logic
- Updates the submission fields, and the list of fields to be blinded and hidden from the public on anonymous pre-prints
- Fixes tests to use updated fields
- Removes test that requires non-empty reviewing volunteers (this was made optional)
- Updates test that non-resubmission is not fetched when resubmissions are fetched
  - This is because `reassignment_request_reviewers` and `reassignment_request_action_editor` are now required fields, non-resubmissions must select "This is not a resubmission"

### DIFF
--- a/openreview/stages/arr_content.py
+++ b/openreview/stages/arr_content.py
@@ -1790,7 +1790,7 @@ arr_max_load_task_forum = {
 }
 
 arr_max_load_task = {
-    "maximum_load": {
+    "maximum_load_this_cycle": {
         "value": {
             "param": {
                 "input": "radio",
@@ -1871,7 +1871,7 @@ arr_reviewer_max_load_task["meta_data_donation"] = {
 }
 
 arr_ac_max_load_task = deepcopy(arr_max_load_task)
-arr_ac_max_load_task["maximum_load"] = {
+arr_ac_max_load_task["maximum_load_this_cycle"] = {
         "value": {
             "param": {
                 "input": "radio",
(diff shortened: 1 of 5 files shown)


openreview/stages/arr_content.py
```diff
@@ -105,32 +124,43 @@
...
         "description": "Research Areas / Tracks. Select the most relevant research area / track for your paper. This will be used to inform the reviewer and action editor assignment.",
-        "order": 9
+        "order": 10
+    },
+    "research_area_keywords": {
+        "order": 11,
+        "description": "Area-specific keywords. Please provide a comma-separated list of keywords from this page: https://aclrollingreview.org/areas",
+        "value": {
+            "param": {
+                "type": "string",
```
melisabok, line 160: shouldn't it be `string[]`?

openreview/stages/arr_content.py
```diff
@@ -360,50 +373,38 @@
...
+    "A1_limitations_section": {
         "value": {
             "param": {
-                "input": "radio",
+                "input": "checkbox",
                 "enum": [
-                    "yes",
-                    "no",
-                    "n/a"
+                    "This paper has a limitations section."
```
melisabok, line 383: if this is the only option and it is not optional, does it mean all the submissions should say "this paper has a limitations section"?

melisabok: ok, They must select it otherwise it will be desk rejected.

melisabok (approved):

celestemartinez:
Should `maximum_load` be changed to `maximum_load_this_cycle`? Looks like it was changed in the preprocess and process functions here: https://openreview.net/invitation/edit?id=aclweb.org/ACL/ARR/2024/April/Reviewers/-/Max_Load_And_Unavailability_Request

openreview/stages/arr_content.py
haroldrubio: Yeah I guess they just want the authors to have to explicitly check that off

haroldrubio:
Thanks for pointing that out, I didn't realize that field changed as well

celestemartinez:
It looks like the `max_load_preprocess.py` was also changed by the PC on the live site?

It is kind of hard to keep track of all the changes they have made.

celestemartinez (approved):

PR #2088, opened 2024-04-10
Fix/ Venue: Custom Stage signatures and invitation_edit_process paper number

### LINKED ISSUE #2083: Custom Stage: adding PCs as signatures
Right now, we always add the PCs as invitees, readers and signatures of all CustomStage invitations.

There was an issue where the PC submitted the review revision of a review. This resulted in an error when `invitation_edit_process.py` tried to update all invitations and notes.

Two solutions:
1. Allow PCs to be signatures always in the CustomStage and fix `invitation_edit_process.py`. The only issue with this, is that if the field has readers `"${5/signatures}",`then the readers will include the PCs twice, since the PCs are the signatures.
2. Add PCs as Participants of the PC stage so we can decide when to add them as signatures and when not to. 

@melisabok thoughts?

celestemartinez:
Fixes #2083

This PR removes PCs from signatures of the CustomStage and edits how we get the paper number in `invitation_edit_process.py`
I am not sure what the best way to get the paper number is when we are deadling with reviews and metareviews, but this is a possible solution.


openreview/venue/process/invitation_edit_process.py
```diff
@@ -133,27 +133,34 @@ def updated_content_readers(note, paper_inv):
 
     def post_invitation(note):
 
-        content = {
-            'noteId': {
-                'value': note.id
-            },
-            'noteNumber': {
-                'value': note.number
+        if isinstance(note, tuple):
```
melisabok, line 136: I'm not fully convinced about checking if the note is a tuple or not.

melisabok: We could always return a tuple as [note, forumNote] where in the case of submissions they are the same note:

`source_submissions = [(n, n) for s in client.get_all_notes(content={ 'venueid': venue_id }, sort='number:asc')]`

and when we build the invitation edit content:

```
{
  noteId: forumNote.id,
  noteNumber: forumNote.number
}
 if 'replyto' in invitation.edit['content']:
        content['replyto'] = { 'value': note.id }

....
```

I think `noteId` and `noteNumber` should have been called `forumId` and `forumNumber` but I think it is too late.

melisabok (approved):

PR #2091, opened 2024-04-11
Fix/ Submission Revision Stage: do not add readers to any field

celestemartinez:
This PR removes `hide_fields` from the Revision invitation and allows to hide fields only using the Post_Submission. 

This PR also adds all possible content fields to the `hide_fields` dropdown (we were not adding the `second_deadline_additional_options`)

~This issue has happened twice, once to MIDL and once to AutoML.~
~I believe what happened was both venues had an abstract deadline, which created the `/Revision` invitations. At some~ ~point, the pdf was hidden for all submissions. Then the Post Submission was used to unhide the pdfs. Later on, they re-~ ~run the Revision stage to allow authors to revise their papers, but they used the default name `/Revision`. Since pdfs were~ ~hidden in these invitations, pdfs started being hidden after authors edited their submission.~

~This PR checks if the field is in `hide_fields` and if it isn't, it removes the readers.~

~**Edit: I have now also noticed that if we used `replacement=True` when posting revision invitations, we would not have~ ~this issue. I am unsure why we sometimes use `replacement=True` and sometimes we use `replacement=False`**~

### DIFF
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -2186,13 +2186,6 @@ class InvitationBuilder(object):
         only_accepted = revision_stage.only_accepted
         content = revision_stage.get_content(api_version='2', conference=self.venue)
 
-        hidden_field_names = self.venue.submission_stage.get_hidden_field_names()
-        for field in content:
-            if field in hidden_field_names:
-                content[field]['readers'] = [venue_id, self.venue.get_authors_id('${{4/id}/number}')]
-            if field not in hidden_field_names:
-                content[field]['readers'] = { 'delete': True }
-
         invitation = Invitation(id=revision_invitation_id,
             invitees=[venue_id],
             readers=[venue_id],
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -2186,6 +2186,16 @@ class InvitationBuilder(object):
         only_accepted = revision_stage.only_accepted
         content = revision_stage.get_content(api_version='2', conference=self.venue)
 
+        hidden_field_names = self.venue.submission_stage.get_hidden_field_names()
+        existing_invitation = tools.get_invitation(self.client, revision_invitation_id)
+        invitation_content = existing_invitation.edit['invitation']['edit']['note']['content'] if existing_invitation else {}
+
+        for field in content:
+            if field in hidden_field_names:
+                content[field]['readers'] = [venue_id, self.venue.get_authors_id('${{4/id}/number}')]
+            if field not in hidden_field_names and invitation_content.get(field, {}).get('readers', []):
+                content[field]['readers'] = { 'delete': True }
+
         invitation = Invitation(id=revision_invitation_id,
             invitees=[venue_id],
             readers=[venue_id],
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -2190,6 +2190,8 @@ class InvitationBuilder(object):
         for field in content:
             if field in hidden_field_names:
                 content[field]['readers'] = [venue_id, self.venue.get_authors_id('${{4/id}/number}')]
+            if field not in hidden_field_names:
+                content[field]['readers'] = { 'delete': True }
 
         invitation = Invitation(id=revision_invitation_id,
             invitees=[venue_id],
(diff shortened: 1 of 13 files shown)


melisabok:
We try to not use replacement=True because that will always post an edit and run the date process function that syncs all the child invitations.

Sometimes PCs make some changes that don't affect the child invitations so we don't want to be re-running the expensive date process function.

openreview/venue/invitation.py
```diff
@@ -2190,6 +2190,8 @@ def set_submission_revision_invitation(self, submission_revision_stage=None):
         for field in content:
             if field in hidden_field_names:
                 content[field]['readers'] = [venue_id, self.venue.get_authors_id('${{4/id}/number}')]
+            if field not in hidden_field_names:
```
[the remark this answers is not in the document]
celestemartinez: I think we would need to get the current invitation, because those fields in `content` shouldn't have readers.

openreview/venue_request/process/bid_stage_pre_process.py
```diff
@@ -6,4 +6,4 @@ def process(client, note, invitation):
         raise openreview.OpenReviewException('Papers should be visible to all program committee if bidding is enabled')
     
     if 'pdf' not in request_form.content.get('hide_fields', []):
-        raise openreview.OpenReviewException('The pdf field should be hidden during the bidding stage')
\ No newline at end of file
+        raise openreview.OpenReviewException('The pdf field should be hidden during the bidding stage. Please use the Post Submission button to hide pdfs.')
```
melisabok, line 9: Thanks for this 🙏

openreview/venue_request/process/revisionProcess.py
```diff
@@ -214,6 +214,12 @@ def process(client, note, invitation):
                 #update post submission hide_fields
                 submission_content = conference.submission_stage.get_content(api_version='2', conference=conference)
                 hide_fields = [key for key in submission_content.keys() if key not in ['title', 'authors', 'authorids', 'venue', 'venueid'] and 'delete' not in submission_content[key]]
+                second_deadline_fields = forum_note.content.get('second_deadline_additional_options', {})
+                if second_deadline_fields:
+                    fields = list(second_deadline_fields.keys())
+                    for field in fields:
+                        if field not in hide_fields and 'delete' not in second_deadline_fields[field]:
+                            hide_fields.append(field)
```
melisabok, line 222: do you need to do this because there is no way to set the hide_fields in the post submission button?

I'm worried that we are building the hide_fields on the fly and it is not persisted in the request note.

celestemartinez: I do this because if they add fields to the submissions using the `second_deadline_fields` field, then these fields will not appear in the `hide_fields` dropdown and PCs won't be able to hide them. The alternative is to have `hide_fields` be a textbox instead of a dropdown, and allow PCs to add any string there.

melisabok: is there a chance to build the dropdown values reading the fields from `additional_submission_options` and `second_deadline_fields`?

melisabok: it seems we are building the dropdown using the submission content:

https://github.com/openreview/openreview-py/blob/master/openreview/venue_request/process/revisionProcess.py#L216

I think we should include the `second_deadline_fields` or get the revision content?

celestemartinez: Well I am adding the `second_deadline_fields` here too, or is this not what you mean?

melisabok: ohhhh I see, we are talking about the same thing! sorry about that, I thought this code was creating the revision invitation 🤦‍♀️  sorry I'm too 🤰

openreview/venue/invitation.py
```diff
@@ -2187,9 +2187,14 @@ def set_submission_revision_invitation(self, submission_revision_stage=None):
         content = revision_stage.get_content(api_version='2', conference=self.venue)
 
         hidden_field_names = self.venue.submission_stage.get_hidden_field_names()
+        existing_invitation = tools.get_invitation(self.client, revision_invitation_id)
+        invitation_content = existing_invitation.edit['invitation']['edit']['note']['content'] if existing_invitation else {}
+
         for field in content:
             if field in hidden_field_names:
                 content[field]['readers'] = [venue_id, self.venue.get_authors_id('${{4/id}/number}')]
+            if field not in hidden_field_names and invitation_content.get(field, {}).get('readers', []):
```
celestemartinez, line 2196: @melisabok I reverted back to adding readers to hidden fields. However, I also now check if the invitation exists and if a field has readers that shouldn't be hidden anymore, I remove the readers.

celestemartinez: This would prevent the issue we had with NAACL today.

melisabok (approved):

enrubio (approved):

PR #2095, opened 2024-04-16
Client: add get archived edges call

melisabok:


openreview/api/client.py
```diff
@@ -1531,6 +1532,23 @@ def get_grouped_edges(self, invitation=None, head=None, tail=None, label=None, g
         json = response.json()
         return json['groupedEdges'] # a list of JSON objects holding information about an edge
 
+    def get_edges_archive(self, invitation):
```
celestemartinez, line 1535: could we call this `get_archived_edges()`. I think it's a bit more descriptive

carlosmondra: I agree

carlosmondra, line 1535:
```suggestion
    def get_archived_edges(self, invitation):
```

melisabok: Done!

celestemartinez (approved):

carlosmondra (approved):

PR #2099, opened 2024-04-17
Committee identity readers: let PCs edit the identity visibility

melisabok:
Changes should be propagated to all the anon groups at any stage. The submission group invitations have a date process function that runs after the invitation is updated. 

This is an issue found by Emilia reported by COLM and it was fixed in the API: https://github.com/openreview/openreview-api/pull/415


celestemartinez (approved):
This is very nice!

enrubio (approved):

PR #2101, opened 2024-04-18
ARR: Change how reassignment score edges are posted

haroldrubio:
Previously, the reassignment script did not allow affinity scores to be posted for resubmissions if the affinity score edge did not exist. We should enable the resubmission script to post edges that do not exist exist. The reviewer must also currently be in the reviewers group.


celestemartinez:
Can we remove the print statements? At least the line `print(list(resubmissions))`. This prints all resubmissions and it's hard to look at the log.

celestemartinez (approved):

PR #2102, opened 2024-04-18
Fix/add test confidential comment

enrubio:
Add test for Author-AC confidential comments.


tests/test_icml_conference.py
```diff
@@ -3830,6 +3831,97 @@ def test_comment_stage(self, openreview_client, helpers):
...
+            invitation='ICML.cc/2023/Conference/Submission1/-/Author_AC_Confidential_Comment',
+            signatures=['ICML.cc/2023/Conference/Submission1/Authors'],
+            note=openreview.api.Note(
+                replyto=submissions[0].id,
+                content={
+                    'comment': { 'value': 'Author confidential comment to AC' },
+                }
+            )
+        )
+        helpers.await_queue_edit(openreview_client, edit_id=confidential_comment_edit['id'])
```
melisabok, line 3899: could you add a test to verify the readers are getting notified of the new confidential comment?

enrubio: I noticed that PCs and the assigned SAC weren't notified even though they're readers. I think it's because `email_sacs` and `email_pcs` are false by default. I'm guessing this is how it's supposed to function?

melisabok: Yes, usually PCs and SACs don't receive emails by default.

tests/test_icml_conference.py
```diff
@@ -3830,6 +3831,113 @@ def test_comment_stage(self, openreview_client, helpers):
...
+        confidential_comment = openreview_client.get_notes(invitation='ICML.cc/2023/Conference/Submission1/-/Author_AC_Confidential_Comment')[0]
+        assert confidential_comment.readers == [
+            'ICML.cc/2023/Conference/Program_Chairs',
+            'ICML.cc/2023/Conference/Submission1/Senior_Area_Chairs',
+            'ICML.cc/2023/Conference/Submission1/Area_Chairs',
+            'ICML.cc/2023/Conference/Submission1/Authors'
+        ]
+
+        # Check messages sent to readers
+        messages = openreview_client.get_messages(subject = '[ICML 2023] A author ac confidential comment has been received on your.*')
```
melisabok, line 3910: Thanks for adding the test, I think we should add the logic to use "An" when the invitation name stats with a vowel.

melisabok: Here is an example from the Journal:

```
vowels = ['a', 'e', 'i', 'o', 'u']
before_invitation = 'An' if lower_formatted_invitation[0] in vowels else 'A'
```

enrubio: Fixed now

openreview/venue/process/custom_stage_process.py
```diff
@@ -49,7 +52,7 @@ def process(client, edit, invitation):
             sender=sender,
             recipients=[program_chairs_id],
             ignoreRecipients = ignore_groups,
-            subject=f'''[{short_name}] A {invitation_name} has been received on Paper Number: {submission.number}, Paper Title: "{submission.content['title']['value']}"''',
+            subject=f'''[{short_name}] {before_invitation} {invitation_name} has been received on Paper Number: {submission.number}, Paper Title: "{submission.content['title']['value']}"''',
             message=f'''We have received a {invitation_name} on a submission to {short_name} for which you are serving as Program Chair.
```
celestemartinez, line 56: Can we use `before_invitation` in the message text as well? 

> We have received a author ac confidential comment on a submission to ICML 2023 for which you are an official area chair.

openreview/venue/process/custom_stage_process.py
```diff
@@ -152,7 +155,7 @@ def process(client, edit, invitation):
             recipients=submission.content['authorids']['value'],
             ignoreRecipients=ignore_groups,
             replyTo=contact,
-            subject=f'''[{short_name}] A {invitation_name} has been received on your Paper Number: {submission.number}, Paper Title: "{submission.content['title']['value']}"''',
+            subject=f'''[{short_name}] {before_invitation} {invitation_name} has been received on your Paper Number: {submission.number}, Paper Title: "{submission.content['title']['value']}"''',
             message=f'''We have recieved a {invitation_name} on your submission to {short_name}
```
celestemartinez, line 159:
```suggestion
            message=f'''We have received a {invitation_name} on your submission to {short_name}
```

openreview/venue/process/custom_stage_process.py
enrubio: You're right, I missed that change. Fixed now.

openreview/venue/process/custom_stage_process.py
melisabok: please include a test for the message content.

celestemartinez (approved):

PR #2114, opened 2024-04-26
Venue: fix get content value if not present

melisabok:


celestemartinez (approved):

celestemartinez:
Can we increase the version here so we can deploy this fix?

enrubio (approved):

PR #2115, opened 2024-04-26
Add custom decision options

enrubio:
Allows custom decision options. Tests still need to be added here.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -19,7 +19,7 @@ import urllib.parse as urlparse
 from tqdm import tqdm
 from concurrent.futures import ThreadPoolExecutor
 
-def decision_to_venue(venue_id, decision_option):
+def decision_to_venue(venue_id, decision_option, accept_options=None):
     """
     Returns the venue for a submission based on its decision
 
@@ -29,11 +29,14 @@ def decision_to_venue(venue_id, decision_option):
     :type decision_option: string
     """
     venue = venue_id
-    if 'Accept' in decision_option:
-        decision = decision_option.replace('Accept', '')
-        decision = re.sub(r'[()\W]+', '', decision)
+    decision = decision_option
+    decision = re.sub(r'[()\W]+', '', decision)
+    if (not accept_options and 'Accept' in decision_option):
+        decision = decision.replace('Accept', '')
         if decision:
             venue += ' ' + decision.strip()
+    elif (accept_options and decision_option in accept_options):
+        venue += ' ' + decision.strip()
     else:
         venue = f'Submitted to {venue}'
     return venue
--- a/openreview/venue_request/process/revisionProcess.py
+++ b/openreview/venue_request/process/revisionProcess.py
@@ -426,11 +426,12 @@ def process(client, note, invitation):
                     'description': 'Select which submission fields should be hidden. Author names will be hidden for double-blind conferences unless otherwise specified in the "reveal_authors" field above. These fields will be hidden from all readers of the submissions, except for program chairs and paper authors.'
                 }
 
-            decision_options = forum_note.content.get('decision_options')
-            if decision_options:
-                decision_options = [s.translate(str.maketrans('', '', '"\'')).strip() for s in decision_options.split(',')]
-            else:
-                decision_options = ['Accept (Oral)', 'Accept (Poster)', 'Reject']
+            accept_decision_options = forum_note.content.get('accept_decision_options', '').strip()
+            reject_decision_options = forum_note.content.get('reject_decision_options', '').strip()
+
+            accept_decision_options = [s.translate(str.maketrans('', '', '"\'')).strip() for s in accept_decision_options.split(',')]
+            reject_decision_options = [s.translate(str.maketrans('', '', '"\'')).strip() for s in reject_decision_options.split(',')]
+            decision_options = accept_decision_options + reject_decision_options
 
             content['send_decision_notifications'] = {
                 'description': 'Would you like to notify the authors regarding the decision? If yes, please carefully review the template below for each decision option before you click submit to send out the emails. Note that you can only send email notifications from the OpenReview UI once. If you need to send additional emails, you can do so by using the python client.',
@@ -441,12 +442,11 @@ def process(client, note, invitation):
                 'required': True,
                 'default': 'No, I will send the emails to the authors'
             }
-            for decision in decision_options:
-                if 'Accept' in decision:
-                    content[f'{decision.lower().replace(" ", "_")}_email_content'] = {
-                        'value-regex': '[\\S\\s]{1,10000}',
-                        'description': 'Please carefully review the template below before you click submit to send out the emails. Make sure not to remove the parenthesized tokens.',
-                        'default': f'''Dear {{{{fullname}}}},
+            for decision in accept_decision_options:
+                content[f'{decision.lower().replace(" ", "_")}_email_content'] = {
+                    'value-regex': '[\\S\\s]{1,10000}',
+                    'description': 'Please carefully review the template below before you click submit to send out the emails. Make sure not to remove the parenthesized tokens.',
+                    'default': f'''Dear {{{{fullname}}}},
 
 Thank you for submitting your paper, {{{{submission_title}}}}, to {short_name}. We are delighted to inform you that your submission has been accepted. Congratulations!
 You can find the final reviews for your paper on the submission page in OpenReview at: {{{{forum_url}}}}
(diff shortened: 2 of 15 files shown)


melisabok:
Thanks for pushing the code, my recommendation is always starting with the test and then the implementation. This is known as TDD (test development driven) and let you build the feature while you are running it. 

1. Start the decision stage for a venue where the decision options don't contain Accept.
2. Expect accepted papers to be marked as accepted with venueid == venue group.id

From that you start making the changes.

openreview/venue_request/venue_request.py
```diff
@@ -714,10 +714,19 @@ def setup_decision_stage(self):
                 'order': 15,
                 'required': True
             },
-            'decision_options': {
```
melisabok, line 717: I would prefer to keep decision options and have a new field to enter the ones that are considered acceptance. This way you don't need to make so many changes. 

When you evaluate if a decision is acceptance you just check if the decision is in the accept_decision_options instead of checking the string has an "Accept" token.

openreview/venue_request/process/revisionProcess.py
```diff
@@ -426,11 +426,12 @@ def process(client, note, invitation):
                     'description': 'Select which submission fields should be hidden. Author names will be hidden for double-blind conferences unless otherwise specified in the "reveal_authors" field above. These fields will be hidden from all readers of the submissions, except for program chairs and paper authors.'
                 }
 
-            decision_options = forum_note.content.get('decision_options')
-            if decision_options:
-                decision_options = [s.translate(str.maketrans('', '', '"\'')).strip() for s in decision_options.split(',')]
-            else:
-                decision_options = ['Accept (Oral)', 'Accept (Poster)', 'Reject']
+            accept_decision_options = forum_note.content.get('accept_decision_options', '').strip()
```
melisabok, line 429: we don't need this change if we keep `decision_options`

openreview/venue_request/venue_request.py
enrubio: I think if we completely remove the "Accept" string checking logic then API 1 tests will fail, what do you think?

melisabok: yes, you should support both.

If accept_decision_options is present then check the decision is in the array.
if accept_decision_options is not present then check "Accept" is in the decision string.

enrubio: I pushed some updated changes from your feedback. I forgot to add the preprocess validation for the decisions, I'll add that next.

melisabok: Changes look good! it seems the build is failing, I'm not sure if it is related to these changes....

enrubio: Oops, I set the wrong default value in `invitation_edit_process.py`, should be fixed now.

melisabok (approved):

openreview/tools.py
```diff
@@ -19,21 +19,25 @@
...
-    if 'Accept' in decision_option:
-        decision = decision_option.replace('Accept', '')
-        decision = re.sub(r'[()\W]+', '', decision)
+    decision = re.sub(r'[()\W]+', '', decision_option)
+    if not accept_options and 'Accept' in decision_option:
+        decision = decision.replace('Accept', '')
         if decision:
             venue += ' ' + decision.strip()
+    elif accept_options and decision_option in accept_options:
+        venue += ' ' + decision.strip()
```
celestemartinez, line 40: I think we always should replace `Accept` right? Some of the `accept_options` might have the word Accept and we should treat those the same way we used to.

melisabok: ☝️

openreview/venue_request/venue_request.py
```diff
@@ -715,18 +715,24 @@ def setup_decision_stage(self):
...
-                'description': 'What are the decision options (provide comma separated values, e.g. Accept (Best Paper), Accept, Reject)? Leave empty for default options - "Accept (Oral)", "Accept (Poster)", "Reject"',
+                'description': 'List all decision options. By default, decisions containing "Accept" are considered Accept options. Otherwise, please specify the accept options in "Accept Decision Options". All other decisions are considered reject options. Provide comma separated values, e.g. "Accept (Best Paper), Invite to Archive, Reject". Leave empty for default options - "Accept (Oral)", "Accept (Poster)", "Reject"',
                 'value-regex': '.*',
                 'order': 30
             },
-            'make_decisions_public': {'description': 'Should the decisions be made public immediately upon posting? Default is "No, decisions should NOT be revealed publicly when they are posted".',
+            'accept_decision_options': {
+                'description': 'What are the accept decision options? If your accept options do not contain "Accept", please specify them here. Provide comma separated values, e.g. "Accept (Best Paper), Invite to Archive".',
```
celestemartinez, line 723: I know there is an example, but the wording of 

> If your accept options do not contain "Accept", please specify them here.

makes it sound like PCs should only specify those accept options that do not contain the word `Accept`, and it could be confusing. 

Maybe something like:

> What are the accept decision options? Please specify all decision options that signify acceptance to the venue. Any decision option not specified here will be treated as a rejection.

tests/test_workshop_v2.py
```diff
@@ -360,7 +360,8 @@ def test_publication_chair(self, client, openreview_client, helpers):
             content={
                 'decision_start_date': start_date.strftime('%Y/%m/%d'),
                 'decision_deadline': due_date.strftime('%Y/%m/%d'),
-                'decision_options': 'Accept, Reject',
+                'decision_options': 'Invite to Venue, Reject',
```
celestemartinez, line 363: Can you add an option `Accept` here and as an `accept_decision_option`? Also, could you add a check for what the venue value should be for every decision option?

celestemartinez: Also, it looks like the homepage is not showing the tab 'Invite to Venue' with the accepted papers. I think there is a `get_decision_heading_map()` function in helpers that you need to edit. Could you also add this to the test? There is an example here: https://github.com/openreview/openreview-py/blob/c4892b14855691dcafa109b4462e53a7f975286d/tests/test_arr_venue_v2.py#L1899

enrubio:
I'm unable to get the homepage tab tests working, no idea why. They show up in the browser. I've tried both `selenium.find_element(By.CLASS_NAME, 'nav-tabs').find_elements(By.TAG_NAME, 'li')` and `selenium.find_element(By.CLASS_NAME, 'tabs-container')` and neither give the correct tabs.

tests/test_workshop_v2.py
```diff
@@ -510,13 +566,46 @@ def test_publication_chair(self, client, openreview_client, helpers):
...
+
+        messages = openreview_client.get_messages(subject=f'[PRL ICAPS 2023] Decision notification for your submission 2:.*')
+        assert 'We are delighted to inform you that your submission has been invited to the venue.' in messages[0]['content']['text']
+
+        messages = openreview_client.get_messages(subject=f'[PRL ICAPS 2023] Decision notification for your submission 3:.*')
+        assert 'We regret to inform you that your submission was not accepted.' in messages[0]['content']['text']
+
+        # Check homepage tabs
+        request_page(selenium, 'http://localhost:3030/group?id=PRL/2023/ICAPS', by=By.CLASS_NAME, wait_for_element='nav-tabs')
+        tabs = selenium.find_element(By.CLASS_NAME, 'nav-tabs')
```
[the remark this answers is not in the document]
melisabok: I'm trying with this: `tabs = selenium.find_element(By.CLASS_NAME, 'nav-tabs').find_elements(By.TAG_NAME, 'li')`


but it still fails, I'm not sure why. We need to try other way to inspect the elements.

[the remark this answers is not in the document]
melisabok: it seems you are missing the token in the request_page so that's why you get one element.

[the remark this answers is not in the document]
enrubio: yes! that was the issue, should be fixed now.

openreview/venue_request/venue_request.py
```diff
@@ -715,18 +715,24 @@ def setup_decision_stage(self):
                 'required': True
             },
             'decision_options': {
-                'description': 'What are the decision options (provide comma separated values, e.g. Accept (Best Paper), Accept, Reject)? Leave empty for default options - "Accept (Oral)", "Accept (Poster)", "Reject"',
+                'description': 'List all decision options. By default, decisions containing "Accept" signify acceptance to the venue. Otherwise, please specify the accept options in "Accept Decision Options". All other decisions will be treated as a rejection. Provide comma separated values, e.g. "Accept (Best Paper), Invite to Archive, Reject". Leave empty for default options - "Accept (Oral)", "Accept (Poster)", "Reject"',
                 'value-regex': '.*',
                 'order': 30
             },
```
enrubio, line 721: What do you think about changing the description to just:

> List all decision options. Provide comma separated values, e.g. "Accept (Best Paper), Invite to Archive, Reject". Leave empty for default options - "Accept (Oral)", "Accept (Poster)", "Reject"

I'm worried it's too confusing and PCs would fill out the fields incorrectly. For example, if "Accept" and "Invite to Archive" are considered accept options, they may just put "Invite to Archive" in `accept_decision_options`. I think this would cause even "Accept" papers to be rejected since it's not listed as an accept option. We could also add to the `accept_decision_options` description something like:

> Leaving this empty will treat only decisions with "Accept" as accepted.

melisabok: can we make the field mandatory and have a default value?

celestemartinez: We do use ` ['Accept (Oral)', 'Accept (Poster)', 'Reject']` as the default options in the request form if they leave the field empty. Do you want this value to be added to the request form?

melisabok: yes, I think it is more clear for the users to know which are the default options how they can edit them.

enrubio: Yes I think with the simplified description we don't need to make the fields required. I fixed it and added the default value. I think it's much better now.

openreview/tools.py
celestemartinez: It looks like with this approach, we are doing a combination of checking for 'Accept' and using `accept_decision_options`? 
For example, if I have an option `Accept (Oral)` in the decision_options, but not in accept_decision_options, it is still treated as an accept?
Not sure if this approach is right or wrong, but want to double check this is the desired behavior?

enrubio: Yup, basically we only check for "Accept" if `accept_decision_options` is empty. This is to make it backwards compatible and not let API 1 tests fail and have to make lots of changes

melisabok: That's correct and that is what is implemented now.

if accept_decision_options then check if decision is in option then marked as accept
else if 'Accept' is in decision then marked as accept

We need to reflect that in the documentation. Correct me if I'm wrong.

celestemartinez: ok! I thought this should be the behavior, but from what I see here we are always checking if the decision has 'Accept', regardless of if we have `accept_decision_options` or not. 

Just to be clear @melisabok: a decision of `Accept (Oral)` that was not specified in `accept_decision_options` should **not** be treated as an accept, assuming `accept_decision_options` is non-empty, correct?

melisabok: yes, accept_decision_options overwrites the "Accept" string rule.

you are right, the tools implementation should check the accept_decision_options first and then the string rule like in other parts of the code. Emilia could you change it?

openreview/tools.py
```diff
@@ -19,20 +19,22 @@
...
     :type decision_option: string
+    :param accept_options: accept decisions (i.e., [ Accept (Best Paper), Invite to Archive ])
+    :type accept_options: list
     """
     venue = venue_id
-    if 'Accept' in decision_option:
-        decision = decision_option.replace('Accept', '')
-        decision = re.sub(r'[()\W]+', '', decision)
-        if decision:
+    decision = re.sub(r'[()\W]+', '', decision_option)
```
melisabok, line 34: I think this can be inside the if, right?

melisabok: it seems the condition `if (not accept_options and 'Accept' in decision_option) or (accept_options and decision_option in accept_options)` is used in several places, can we create a tools function that we use from the venue class, process function and this function too?

enrubio: Yes that would be great, updated to add the tools function

PR #2116, opened 2024-04-26
Venue: let PCs undeploy assignments

melisabok:
- New unset_assignment method.
- Make sure we use the review and meta review names instead of hardcoded names.
- Fix validation to set and unset assignments when there are already reviews/metareviews posted.
- Add tests
- Remove commented code


openreview/venue/matching.py
```diff
@@ -1309,3 +1366,15 @@ def deploy(self, assignment_title, overwrite=False, enable_reviewer_reassignment
         if self.is_reviewer and enable_reviewer_reassignment:
             hash_seed=''.join(random.choices(string.ascii_uppercase + string.digits, k = 8))
             self.setup_invite_assignment(hash_seed=hash_seed, invited_committee_name=f'''Emergency_{self.match_group_name}''')
+
+    def undeploy(self):
+
+        ## Undeploy assignments
+        if self.is_senior_area_chair and not self.venue.sac_paper_assignments:
+            self.undeploy_sac_assignments()
```
celestemartinez, line 1374: Has this been implemented already? I can't find it

melisabok: oops, I forgot it!

I think in this case it is only deleting the assignments edges?

celestemartinez: Correct! About changing the status of the configuration, will this be done in the matcher code?

melisabok:
UPDATE: I changed the strategy to undeploy the assignments related to the configuration title instead of removing all the assignments. This way we can deploy multiple assignments from the UI.

enrubio (approved):

melisabok:
@celestemartinez this PR is ready to be reviewed again, thanks!

openreview/journal/process/ae_assignment_process.py
```diff
@@ -51,9 +70,9 @@ def process_update(client, edge, invitation, existing_edge):
         ## expire AE recommendation
         journal.invitation_builder.expire_invitation(journal.get_ae_recommendation_id(number=note.number))
 
-        ## update assigned_action_editor if exists in the submission
+        ## add assigned_action_editor
         content = {}
-        if 'assigned_action_editor' in note.content:
+        if note.content['venueid']['value'] == journal.under_review_venue_id:
```
celestemartinez, line 75: Will an AE be added to the submission only when the `venueid == journal.under_review_venue_id`? Could the venue id be `journal.assigning_AE_venue_id` instead and need the `assigned_action_editor` added?

melisabok: usually `assigned_action_editor` is being set when the paper is under review.

I'm handling this case when the AE is unassigned and then another one is assigned when the paper is under review.

melisabok: BTW I will make a new PR where the assigned_action_editor will be visible as soon as the AE is assigned.

celestemartinez:
We don't need to do anything about the`Invite_Assignment` invitation for reviewers right, which is created when we deploy reviewer assignments? Since the edge browser is not visible to ACs unless we set `'enable_reviewers_reassignment': { 'value': True }` in the group content, ACs shouldn't be able to invite reviewers anyway

celestemartinez (approved):

melisabok:
That's true, I didn't change the invite invitation.

PR #2134, opened 2024-05-06
ARR: fix desk rejection verification process

melisabok:


openreview/arr/process/verification_process.py
```diff
@@ -24,7 +25,7 @@ def process(client, edit, invitation):
                     )
             )
         
-    elif not submission.content['flagged_for_desk_reject_verification']['value']:
+    else:
```
melisabok, line 28: I'm not sure if flagged is false we should do the same where the value doesn't exist.

haroldrubio (approved):

Issue #2104, opened 2024-04-19
ARR: Submission previous_url validation


melisabok:
The pre process validation that checks if the previous_url is valid doesn't fail when the user pass the forum id + plus comment id:

`id=[forum_id]&noteId=[note_id]`

To fix this we should check the note that we get has an invitation that correspond to a submission and it is forum note (id == forum).

I would also request the users to enter the full url, the UI knows how to render the link to the previous forum.

melisabok:
The submission form should have a regex validation:

```
'regex': 'https:\/\/openreview\.net\/forum\?id=.*'
```

and check the id is an ARR submission.

melisabok:
Fixed!

PR #2138, opened 2024-05-07
Fix: Commitments: improve regex and preprocess

### LINKED ISSUE #2136: ARR commitment: fix paper_link validation
Use a regex that check the link follow the format "https://openreview.net/forum?id=[note_id]"

No pdf link should be allowed and not extra parameters and no comma separated values. 

https://github.com/openreview/openreview-py/blob/master/openreview/venue/process/submission_commitments_preprocess.py

Add a descriptive error message.

celestemartinez:
Fixes #2136 

I tried multiple regexes to make sure authors don't add `&noteId=xxx` after the forum id, but none seemed to work. Instead, I take care of this in the preprocess.

### DIFF
--- a/openreview/venue/process/submission_commitments_preprocess.py
+++ b/openreview/venue/process/submission_commitments_preprocess.py
@@ -3,14 +3,29 @@ def process(client, edit, invitation):
     paper_link = edit.note.content['paper_link']['value']
     paper_forum = paper_link.split('?id=')[-1]
 
+    if '&' in paper_link:
+        raise openreview.OpenReviewException('Invalid paper link. Please make sure not to provide anything after the character "&" in the paper link.')
+
+    client_v1=openreview.Client(baseurl=openreview.tools.get_base_urls(client)[0], token=client.token)
+
     try:
-        client_v1=openreview.Client(baseurl=openreview.tools.get_base_urls(client)[0], token=client.token)
-        arr_submission = client_v1.get_note(paper_forum)
-    except openreview.OpenReviewException as e:
+        arr_submission_v1 = client_v1.get_note(paper_forum)
+    except:
+        arr_submission_v1 = None
+
+    try:
+        arr_submission_v2 = client.get_note(paper_forum)
+    except:
+        arr_submission_v2 = None
+
+    if not arr_submission_v1 and not arr_submission_v2:
         raise openreview.OpenReviewException('Provided paper link does not correspond to a submission in OpenReview')
 
-    if 'aclweb.org/ACL/ARR' not in arr_submission.invitation:
+    if (arr_submission_v1 and 'aclweb.org/ACL/ARR' not in arr_submission_v1.invitation) or (arr_submission_v2 and not any('aclweb.org/ACL/ARR' in inv for inv in arr_submission_v2.invitations)):
         raise openreview.OpenReviewException('Provided paper link does not correspond to an ARR submission')
 
-    if arr_submission.id != arr_submission.forum:
-        raise openreview.OpenReviewException('Provided paper link does not correspond to an ARR submission. Make sure the link points to a submission and not to a reply.')
+    if (arr_submission_v1 and arr_submission_v1.id != arr_submission_v1.forum) or (arr_submission_v2 and arr_submission_v2.id != arr_submission_v2.forum):
+            raise openreview.OpenReviewException('Provided paper link does not correspond to an ARR submission. Make sure the link points to a submission and not to a reply.')
+
+    if arr_submission_v1 and 'aclweb.org/ACL/ARR' in arr_submission_v1.invitation and not arr_submission_v1.invitation.endswith('Blind_Submission'):
+        raise openreview.OpenReviewException('Provided paper link does not point to a blind submission')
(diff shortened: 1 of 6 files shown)


tests/test_acl_commitment_v2.py
```diff
@@ -119,7 +119,7 @@ def test_create_conference(self, client, openreview_client, helpers):
                         'value': {
                         'param': {
                             'type': 'string',
-                            'regex': '(http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\\(\\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+)',
+                            'regex': 'https:\/\/openreview\.net\/forum\?id=.*',
```
melisabok, line 122: could you add a test trying to post a note using the urls:

https://openreview.net/pdf?id=1234
https://openreview.net/forum?id=1234&referrer=[Author Console](/group?id=ICML.cc/2024/Conference/Authors)
https://openreview.net/forum?id=1234&replyto=4567

melisabok:
@celestemartinez what is the status of this PR?

melisabok:
@celestemartinez could you also fix this issue that is related to this one?

https://github.com/openreview/openreview-py/issues/2104

celestemartinez:
@melisabok This should be ready for review. I added another check if submission.id != submission.forum

tests/test_acl_commitment_v2.py
melisabok: @celestemartinez could you add these tests cases?

melisabok:
@celestemartinez please try to update this PR so we can merge it soon, thanks.

celestemartinez:
> Thanks Celeste, can you add the same validation to the ARR venues? or it is the same code?

What do you mean enable it for ARR venues?

melisabok:
> What do you mean enable it for ARR venues?

The ARR cycles have a field called "previous_url" where the validation is the same for ARR commitment venues. There is an open issue here: https://github.com/openreview/openreview-py/issues/2104

Could you make the same fix you are doing here for the ARR commitment venues for ARR class too?

Also please add these test cases: https://github.com/openreview/openreview-py/pull/2138#discussion_r1593024249

celestemartinez:
@melisabok @haroldrubio sorry! I didn't realize this was also an ARR-specific issue.

I have made the fix for ARR and added the tests. Let me know if I missed something!

melisabok:
Thanks for the changes Celeste, there is one more case that I would like to test:

```
https://openreview.net/forum?id=1234&referrer=[Author%20Console](/group?id=aclweb.org/ACL/ARR/2023/June
```

when 1234 is a valid ARR note, I think in this case the validation will pass but it will be very difficult to parse this value when migrating data so I would like the users to enter the url like https://openreview.net/forum?id=1234 with no other parameters. Is it possible to describe this in the regex? I think instead of using .* we can see any character with the exception of the &, could you try that?

celestemartinez:
> Thanks for the changes Celeste, there is one more case that I would like to test:
> 
> ```
> https://openreview.net/forum?id=1234&referrer=[Author%20Console](/group?id=aclweb.org/ACL/ARR/2023/June
> ```
> 
> when 1234 is a valid ARR note, I think in this case the validation will pass but it will be very difficult to parse this value when migrating data so I would like the users to enter the url like https://openreview.net/forum?id=1234 with no other parameters. Is it possible to describe this in the regex? I think instead of using .* we can see any character with the exception of the &, could you try that?

If they enter this URL, the preprocess should fail with `Provided paper link does not correspond to a submission in OpenReview` (this is in the tests, doesn't matter if the note is valid or not because we take everything after the ?id= as the note id. )

I had previously tried having a more specific regex but I couldn't get it to work. I can try again.

melisabok:
ohhh I see, because you are splitting by `paper_forum = paper_link.split('?id=')[-1]` so the last element contains the forum id plus all the parameters and the note is not found.

Got it, I wonder if we should say in the message the url must not have extra parameters

celestemartinez:
I added another check, if the paper link contains '&' then we show a more descriptive error.

melisabok (approved):

openreview/venue/process/submission_commitments_preprocess.py
```diff
@@ -1,13 +1,31 @@
...
-        raise openreview.OpenReviewException('Provided paper link does not correspond to an ARR submission')
\ No newline at end of file
+    if (arr_submission_v1 and 'aclweb.org/ACL/ARR' not in arr_submission_v1.invitation) or (arr_submission_v2 and not any('aclweb.org/ACL/ARR' in inv for inv in arr_submission_v2.invitations)):
+        raise openreview.OpenReviewException('Provided paper link does not correspond to an ARR submission')
+
+    if (arr_submission_v1 and arr_submission_v1.id != arr_submission_v1.forum) or (arr_submission_v2 and arr_submission_v2.id != arr_submission_v2.forum):
+            raise openreview.OpenReviewException('Provided paper link does not correspond to an ARR submission. Make sure the link points to a submission and not to a reply.')
+
+    if arr_submission_v1 and 'aclweb.org/ACL/ARR' in arr_submission_v1.invitation and not arr_submission_v1.invitation.endswith('Blind_Submission'):
+        raise openreview.OpenReviewException('Provided paper link does not point to a blind submission')
```
haroldrubio, line 31: Do you think the authors will know what a blind submission is, do you think we should add instructions on how to get the right link?

But it might be too much for an error message

celestemartinez: Fixed.

haroldrubio (approved):

PR #2147, opened 2024-05-09
Fix Chat_Reaction invitation ddate

zbialecki:
Fix ddate for Chat_Reaction invitation. Also add additional emoji options.

### DIFF
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -1769,7 +1769,7 @@ class InvitationBuilder(object):
                         },
                         'ddate': {
                             'param': {
-                                'range': [ 0, 999999999999 ],
+                                'range': [ 0, 9999999999999 ],
                                 'optional': True,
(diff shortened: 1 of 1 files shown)


melisabok (approved):

openreview/venue/invitation.py
```diff
@@ -1783,7 +1783,7 @@ def set_chat_invitation(self):
                         'writers': [venue_id, '${2/signatures}'],
                         'tag': {
                             'param': {
-                                'enum': ['👍', '👎', '😄', '😂', '🔥']
+                                'enum': ['👍', '👎', '👌', '👆', '😄', '😂', '🔥', '🚀', '✅']
```
zbialecki, line 1786: I'm not sure what the correct number of options to offer is. What do you think?

melisabok: I think that is ok for now.

Issue #2117, opened 2024-05-01
ARR: Emergency load process function always expects emergency load


haroldrubio:
The emergency load process function tries to get `note.content['emergency_load']['value']` but the emergency load field is optional causing errors

melisabok:
Change the process function code to be (emergency_load_process.py):

```
note.content.get('emergency_load', {}).get('value')
```

melisabok:
Fixed!

PR #2170, opened 2024-05-30
Fix: Emergency load process not using get

haroldrubio:
Resolves
- #2117

### DIFF
--- a/openreview/arr/process/emergency_load_process.py
+++ b/openreview/arr/process/emergency_load_process.py
@@ -113,7 +113,7 @@ def process(client, edit, invitation):
         original_load = 0
     else:
         original_load = cmp_edges[0].weight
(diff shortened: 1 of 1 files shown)


openreview/arr/process/emergency_load_process.py
```diff
@@ -113,7 +113,7 @@ def process(client, edit, invitation):
         original_load = 0
     else:
         original_load = cmp_edges[0].weight
-    emergency_load = edit.note.content['emergency_load']['value']
+    emergency_load = edit.note.content.get('emergency_load', {}).get('value', 0)
```
melisabok, line 116: is there a way to reproduce this error in the tests?

haroldrubio: There are already cases where the emergency load is not present but they're being caught by the pre-process

Issue #2164, opened 2024-05-23
ARR: Add ethics chairs to ethics flagged submissions


haroldrubio:
Add ethics chairs to readers of flagged submission in the ethics flag process function

melisabok:
@celestemartinez could you help with this issue?

Issue #2126, opened 2024-05-03
ARR Workflow: Add Ethics Reviewer Matching step


haroldrubio:
The EMNLP and ARR ethics chairs are proposing to make the following changes to the ethics reviewing period:

1. Add a new unavailability form for ethics reviewers, including the long-term unavailability logic for reviewers and area chairs
2. Send messages to ethics chairs when ethics reviews are submitted
3. Some kind of commenting for ethics chairs and ethics reviewers, either a separate invitation or modifying the ethics flag process function to add/remove ethics chairs to the official comment invitation

melisabok:
@celestemartinez can you work on items 2 and 3?

melisabok:
I think bullet 1 is fixed here: https://github.com/openreview/openreview-py/pull/2184

melisabok:
@haroldrubio I think this issue is fixed now, right?

PR #2172, opened 2024-05-30
Fix/ethics review settings

celestemartinez:
The following issues are fixed here:
- add ethics chairs as readers of flagged submissions (since we are using append in the invitation, we add them even if the paper is public. Perhaps we should do this in the process function so we can check this?)
- send an email to ethics chair and to the writer of the ethics review when a ethics review is posted
- adds ethics chairs as invitees of official comment invitations (and removes them if the papers is unflagged)

Fixes:  #2164 and items (2) and (3) of #2126

### DIFF
--- a/openreview/arr/helpers.py
+++ b/openreview/arr/helpers.py
@@ -941,8 +941,11 @@ class ARRWorkflow(object):
                         'release_ethics_reviews_to_reviewers': 'Ethics reviews should be immediately revealed to the paper\'s reviewers and ethics reviewers',
                         'additional_ethics_review_form_options': arr_ethics_review_content,
                         'remove_ethics_review_form_options': 'ethics_review',
+                        'release_submissions_to_ethics_chairs': 'Yes, release flagged submissions to the ethics chairs.',
                         "release_submissions_to_ethics_reviewers": "We confirm we want to release the submissions and reviews to the ethics reviewers",
                         'enable_comments_for_ethics_reviewers': 'Yes, enable commenting for ethics reviewers.',
+                        'compute_affinity_scores': 'No'
+
                     },
                     'forum': request_form_id,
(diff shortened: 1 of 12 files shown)


melisabok:
> since we are using append in the invitation, we add them even if the paper is public

Does it mean the readers could be [everyone, ARR/Submission1/Ethic_Chairs]?

I think we should avoid this.

celestemartinez:
> > since we are using append in the invitation, we add them even if the paper is public
> 
> Does it mean the readers could be [everyone, ARR/Submission1/Ethic_Chairs]?
> 
> I think we should avoid this.

Yes, I am looking at the test and the readers of some papers end up being `eveyone, Submission1/Ethics_Reviewers and Ethics_Chairs`. I won't do this with the invitation then and add the readers in the process function instead.

melisabok:
I don't think we can use an invitation if not all the submissions have the same set the of readers, we would need to use the meta invitation is set the readers accordingly.

celestemartinez:
Two points:

1. Right now I am always adding the ethics chairs as readers of reviews. Should this be done only if the PC choose to make submissions visible to ethics chairs? I would need to check this in a couple of places.
2. The Ethics_Flag invitation now only adds the `flagged_for_ethics` field and the process function appends/removes readers depending on that value. Not sure if there is a better way to do this since this creates two edits on the submissions.

melisabok:
@celestemartinez do you have any update on this?

celestemartinez:
Sorry about that; this Pr should be ready.
I made a change so only if the PCs choose to make submissions visible to ethics chairs then ethics chairs are added as readers of comments. Otherwise, they would have to use the submission SAC group like CVPR does.

melisabok (approved):

PR #2191, opened 2024-06-10
ARR: Add integer validation to SAC maximum load 

### LINKED ISSUE #2190: ARR: Fix SAC max load cycle to use a number instead of free text string
Avoid process function errors, like this one: https://api2.openreview.net/logs/process?id=9jPQMQasc4

haroldrubio:
Resolves #2190

### DIFF
--- a/openreview/stages/arr_content.py
+++ b/openreview/stages/arr_content.py
@@ -1889,7 +1889,10 @@ arr_sac_max_load_task = deepcopy(arr_max_load_task)
 del arr_sac_max_load_task["maximum_load_this_cycle_for_resubmissions"]
 arr_sac_max_load_task["maximum_load_this_cycle"] = {
     "value": {
-        "param": {"regex": "[0-9]{0,3}", "optional": False, "type": "string"}
+        "param": {
+            "optional": False,
+            "type": "integer"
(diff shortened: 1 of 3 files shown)


openreview/stages/arr_content.py
```diff
@@ -1889,7 +1889,10 @@
 del arr_sac_max_load_task["maximum_load_this_cycle_for_resubmissions"]
 arr_sac_max_load_task["maximum_load_this_cycle"] = {
     "value": {
-        "param": {"regex": "[0-9]{0,3}", "optional": False, "type": "string"}
+        "param": {
```
melisabok, line 1892: is there a test that uses this invitation?

haroldrubio: Yes it gets tested with the other max load invitations:
https://github.com/openreview/openreview-py/blob/7f2eb6729a79325aed6349efc6cbacd30a6d5fba/tests/test_arr_venue_v2.py#L1463-L1471

melisabok: don't you need to pass 10 as number instead of a string?

melisabok (approved):

PR #2193, opened 2024-06-11
Fix/ Camera Ready Revisions: add pub chairs as readers of authors and authorids

### LINKED ISSUE #2177: Camera ready revisions overwrite author readers
After publication chairs have been added as readers of author names, enabling revision invitations for accepted papers will remove the chairs as readers. 

I think we can check if the revision stage is enabled for accepted papers only and the venue has publication chairs, then they should be added as readers of the authors and authorids fields in the revision invitations.

celestemartinez:
Fixes #2177 

If the venue has publication chairs, then add them as readers of the `authors` and `authorids` fields in revision invitations when the submission revision stage is enabled for accepted papers only. 

The issue this still has is that if author names are released after creating camera ready invitations, the invitations will still have readers for `authors` and `authorids`, and posting a camera ready revision will once again hide the author names from the public; this happened in ICML. I am not sure how to fix this, unless after running the post decision and releasing author names to the public, we check if there is a revision stage active and re-run it?

### DIFF
--- a/openreview/venue_request/process/revisionProcess.py
+++ b/openreview/venue_request/process/revisionProcess.py
@@ -591,6 +591,9 @@ Best,
                         )
                         client.post_note(comment_note)
 
+            if forum_note.content.get('accepted_submissions_only', '') == 'Enable revision for accepted submissions only':
+                conference.create_submission_revision_stage()
+
             # update Submission_Revision_stage invitation
             if forum_note.content.get('reveal_authors') == 'Reveal author identities of only accepted submissions to the public':
                 revision_stage_inv = openreview.tools.get_invitation(client, SUPPORT_GROUP + '/-/Request' + str(forum_note.number) + '/Submission_Revision_Stage')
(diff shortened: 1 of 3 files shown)


celestemartinez:
@melisabok any ideas on how to fix the issue mentioned above, where camera ready revisions are created (which contain readers on the author fields) and then author identities are released to the public)?

melisabok:
> unless after running the post decision and releasing author names to the public, we check if there is a revision stage active and re-run it?

Yes, I think we should re-run the stage in the post decision implementation.

openreview/venue_request/process/revisionProcess.py
```diff
@@ -591,6 +591,9 @@ def process(client, note, invitation):
                         )
                         client.post_note(comment_note)
 
+            if forum_note.content.get('accepted_submissions_only', '') == 'Enable revision for accepted submissions only':
+                conference.create_submission_revision_stage()
```
melisabok, line 595: shouldn't be this inside of the post_decision_stage implementation?

I think we should call to `create_submission_stage` to any existing submission revision invitation that is active, no matter is only for accepted papers, what do you think?

celestemartinez: I think I am worried that if only author names of accepted submissions are released, then re-running a submission revision stage that is enabled for all papers would remove readers from all revision invitations, even rejected papers. 

Also, once the post submission stage is run and accepted paper authors are revealed, we only allow the revision stage to be run for accepted papers: #2039

Issue #2189, opened 2024-06-07
ARR: PC console, enable more email options


melisabok:
- Let email SACs from the SAC status tab. Pending PR: https://github.com/openreview/openreview-web/pull/1942
- Let email reviewers with different filter options. For example: reviewers that haven't replied to a rebuttal or haven't updated the review after the rebuttal.

jkkummerfeld:
Note that this overlaps with https://github.com/openreview/openreview-py/issues/2200

melisabok:
Closing this issue because it is repeated.

PR #2194, opened 2024-06-12
ARR: Add custom emailing options to PC console

haroldrubio:
This PR depends on https://github.com/openreview/openreview-web/pull/1938

We need to fix how to get the registration notes for reviewers with no assignments, otherwise there will be reminders for all reviewers.

Fixes https://github.com/openreview/openreview-py/issues/2189


openreview/arr/webfield/programChairsWebfield.js
```diff
@@ -146,6 +146,193 @@ return {
...
+        console.log(row);
+        if (row.notesInfo.length <= 0){
+          return false;
+        }
+
+        return row.notesInfo.some(obj => {
+          const reviewId = obj?.officialReview?.id ?? ''
+          const anonId = obj?.officialReview?.anonymousId ?? ''
+          const replies = obj?.note?.details?.replies ?? []
+          const authorReplies = replies.filter(reply =>
```
celestemartinez, line 162: when getting authorReplies, is the invitation we are looking for those notes just Official_Comment? I am testing this emailing option. It looks like an email would be sent to `reviewer1@aclrollingreview.com` but their review does not have a rebuttal, just a Revie_Rating reply by the authors. I am assuming in this case no email should be sent to them?

openreview/arr/webfield/programChairsWebfield.js
```diff
@@ -146,6 +146,193 @@ return {
...
+        const registrationForm = registrationNotes.filter(note => {
+          const invitations = note?.invitations ?? []
+          return invitations.some(inv => inv.includes('Reviewers/-/Registration'))
+        })
+        const maxLoadForm = registrationNotes.filter(note => {
+          const invitations = note?.invitations ?? []
+          return invitations.some(inv => inv.includes('Reviewers/-/Max_Load_And_Unavailability_Request'))
+        })
+        
+        if (registrationForm.length <= 0 || maxLoadForm.length <= 0) {
```
celestemartinez, line 220: if a reviewer did not submit the registration note, but they selected 0 in "maximum_load_this_cycle", they would not be emailed?

openreview/arr/webfield/programChairsWebfield.js
melisabok: I agree,  `reviewId` mustn't be empty. If it is empty then don't check it.

openreview/arr/webfield/programChairsWebfield.js
melisabok: I understand that registration notes are always present but I don't know why we need to check if they exist.

Also I think reminder should be sent when there is no reply for Max_Load_And_Unavailability_Request not when the load is 0, right?

haroldrubio: Yeah I think I can remove this option, it was used one cycle to get more capacity out of the reviewers but there's probably a better way to do this. Also, this option is not on the list of email options requested.

melisabok (approved):

openreview/arr/webfield/programChairsWebfield.js
```diff
@@ -153,6 +153,195 @@ return {
...
+        const load = typeof maxLoadForm[0].content.maximum_load_this_cycle.value === 'number' ? 
+          maxLoadForm[0].content.maximum_load_this_cycle.value : 
+          parseInt(maxLoadForm[0].content.maximum_load_this_cycle.value, 10)
+        return load > 0
+        
+        `
+      },
+      {
+        label: 'Available Area Chairs with No Assignments', filterFunc: `
+        console.log(row);
```
celestemartinez, line 251: There's 3 `console.log(row);` we should remove

melisabok:
@haroldrubio what is the status of this PR? there are some conflicts here.

melisabok:
@haroldrubio please continue working with this PR or close it. 

Let's avoid having a PR open for so many weeks.

haroldrubio:
Closing and splitting the email options into separate PRs

PR #2201, opened 2024-06-14
Bid Stage: Add support for custom roles

enrubio:
This PR adds support for custom roles in the bid stage.


openreview/conference/helpers.py
```diff
@@ -607,15 +607,18 @@ def get_bid_stages(request_forum):
     else:
         bid_due_date = None
 
-    reviewer_bid_stage = openreview.stages.BidStage(request_forum.content['venue_id'] + '/Reviewers', start_date = bid_start_date, due_date = bid_due_date, request_count = int(request_forum.content.get('bid_count', 50)))
+    reviewers_name = venue.reviewers_name if venue else 'Reviewers'
+    reviewer_bid_stage = openreview.stages.BidStage(request_forum.content['venue_id'] + f'/{reviewers_name}', start_date = bid_start_date, due_date = bid_due_date, request_count = int(request_forum.content.get('bid_count', 50)))
```
melisabok, line 611: can we figure out the reviewers_name inside of the BidStage?

for example:

if committee_id == venue.reviewers_id:
       commmittee_name = reviewers_name

enrubio: Where would `committee_id` come from in this function?

melisabok: from here? https://github.com/openreview/openreview-py/blob/master/openreview/stages/venue_stages.py#L382

melisabok: ohhhh I see, you are building the BidStage parameters. 

In this case, I would do:

```
reviewer_bid_stage = openreview.stages.BidStage(reviewers_id, start_date = bid_start_date, due_date = bid_due_date, request_count = int(request_forum.content.get('bid_count', 50)))
```

and pass reviewers_id as a param instead of the venue object.

melisabok: same for area chairs, if area_chairs_id is None then you don't set the stage.

enrubio: To get the reviewers_id don't I still need the venue object? To access `venue.get_reviewers_id()`?

melisabok: yes but I was thinking that can pass the reviewers_id directly instead of passing the venue object because this call is being done from the API 1 conference object too.

```
get_bid_stages(request_forum, reviewers_id, area_chairs_id, senior_area_chairs_id):
```

enrubio: does this mean for API 1 support I should still use the request form?:
```
reviewer_bid_stage = openreview.stages.BidStage(
        reviewers_id if reviewers_id else request_forum.content['venue_id'] + '/Reviewers', 
        ...
```

melisabok: no, the API 1 can pass these parameters right? `reviewers_id, area_chairs_id, senior_area_chairs_id`?

enrubio: ah that makes sense now, I thought I shouldn't modify API 1 code. Fixed now.

openreview/venue/webfield/paperBidWebfield.js
```diff
@@ -1,6 +1,17 @@
 // Webfield component
 const committeeName = entity.content.committee_name?.value
-const affinityScoreId = domain.content[`${committeeName.toLowerCase()}_affinity_score_id`]?.value
+const reviewersName = domain.content['reviewers_name']?.value
```
melisabok, line 3: Thanks for doing this, I think there will be other places with the same issue:

- assignment_pre_process.js
- invite_assignment_pre_process.js
- proposed_assignment_pre_process.js


You can fix them in this PR or make a new one. But I would like to reproduce the issue with a test first.

enrubio: okay! I think I should make these changes in another PR, I'll add tests for each one

melisabok (approved):

Issue #2198, opened 2024-06-14
ARR: Change emergency reviewing form to allow multiple areas


jkkummerfeld:
For June onwards, the editors would like people volunteering to be emergency reviewers to say they can do so in multiple areas.

I believe this just requires editing this line (though maybe there is also post-processing logic that is affected?):

https://github.com/openreview/openreview-py/blob/e4d06e94984866d47051dc90cdba410cca6d2aab/openreview/stages/arr_content.py#L1990

melisabok:
input but must a checkbox and type must a string[] in order to allow multiple values

jkkummerfeld:
Yes, AC too, thanks for confirming

Issue #2197, opened 2024-06-14
ARR: AC and Reviewer checklist changes


jkkummerfeld:
The ARR editors have requested a few changes for the reviewer checklists in June (so needed once matching is complete - June 26th).

1. In the reviewer checklist, under "Need Ethics Review", please change the wording to:

"Should this paper be sent for an in-depth ethics review? Before you answer this question, please refer to  https://aclrollingreview.org/ethics-flagging-guidelines/ for guidelines on what papers should and shouldn't be flagged. If your answer is yes, then explain why below, and we will ry to ensure that it receives a separate ethics review."

2. The questions in the AC/AE checklist should have their wording modified to matching the reviewer checklist.

melisabok:
@jkkummerfeld for bullet 1 it says explain why below, which field should they explain the paper needs ethics review? `ethical_concerns`? that field is before `needs_ethics_review`. 

I updated the description for the field in the review and metareview forms: https://github.com/openreview/openreview-py/pull/2212/commits/28fbe1acefc45cc0d95f4126009105a55061b79a

jkkummerfeld:
Ah, in that case, let's make the text:

"Should this paper be sent for an in-depth ethics review? Before you answer this question, please refer to https://aclrollingreview.org/ethics-flagging-guidelines/ for guidelines on what papers should and shouldn't be flagged. If your answer is yes, then ensure you have explained why in the question above, and we will try to ensure that it receives a separate ethics review."

This is a small edit so I don't think I need to run it by the editors for confirmation.

PR #2212, opened 2024-06-24
ARR: fix several issues

melisabok:
- It fixes: https://github.com/openreview/openreview-py/issues/2197
- It fixes: https://github.com/openreview/openreview-py/issues/2198
- It fixes: https://github.com/openreview/openreview-py/issues/2216 
- Includes the "Edit reviewers assignments" and "Edit area chair assignments" in the Paper Status tab of the SAC console.
It requires: https://github.com/openreview/openreview-web/pull/1980
- Fixes some issues with the emergency load process function: aggregate scores are posted mote than once when the reply is edited. The Custom Max Papers edge is added the old emergency load if the reply is edited. The edges are not cleared if the reply is edited from Yes to No.
- Added tools function to get a note
- Fix request form process function to execute call when the venue is an instance of the class ARR
- Fix Tests

### DIFF
--- a/openreview/arr/arr.py
+++ b/openreview/arr/arr.py
@@ -19,6 +19,7 @@ from openreview.venue.recruitment import Recruitment
 from openreview.arr.helpers import (
     setup_arr_invitations
 )
+from openreview.stages.arr_content import hide_fields
 
 SHORT_BUFFER_MIN = 30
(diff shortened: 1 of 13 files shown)


celestemartinez (approved):
Looks good.

haroldrubio (approved):

PR #2213, opened 2024-06-25
feat: generate plagiarism reports for AAAI submissions

akshatchhabra:
Creating a script to generate plagiarism reports for AAAI submissions using iThenticate API 

It depends on 

https://github.com/openreview/openreview-web/pull/2023

https://github.com/openreview/openreview-api/pull/487


openreview/aaai/commands.py
```diff
@@ -0,0 +1,266 @@
+import os
+import json
+
+
+class TCACommands:
```
melisabok, line 5: Can we call this `iThenticateClient` or the name of the API we are calling too?

.gitignore
```diff
@@ -12,3 +12,4 @@ cov.xml
 .coverage*
 tests/drivers/geckodriver
 tests/data/*.csv
+openreview/aaai/TCA_key.json
```
melisabok, line 15: I would like this client to be used for other venues that provide the API key, can we use another filter name instead of aaai, what about iThenticate or plagiarism?

carlosmondra:
Is there a reason to use curl instead of the requests library that is already installed in the library?

akshatchhabra:
> Is there a reason to use curl instead of the requests library that is already installed in the library?

I had just copied the commands from the API documentation and was using them to run a sample script. I'll make the change and use the requests library instead.

melisabok:
Run `pytest tests/test_aaai_conference.py` to setup the plagiarism detection.

openreview/venue/venue.py
```diff
@@ -1009,6 +1014,32 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+        
+        self.invitation_builder.set_iThenticate_plagiarism_check_invitation()
+
+        iThenticate_client = openreview.api.iThenticateClient(self.iThenticatePlagiarismCheckApiKey)
+        
+        submissions = self.get_submissions()
+
+        for submission in submissions:
+
+            # TODO: Akshat to complete this
```
melisabok, line 1030: @akshatchhabra here is where you need to implement the calls to the iThenticate API to create the submission and upload the PDF.

openreview/api/iThenticate_client.py
```diff
@@ -0,0 +1,233 @@
+import os
+import requests
+import json
+from datetime import datetime
+
+
+class iThenticateClient:
+    def __init__(self, api_key):
```
melisabok, line 8: @akshatchhabra I moved the client to this folder and added the api_key as a parameter. The venue that is calling to detect plagiarism should provide the api key.

openreview/venue_request/venue_request.py
```diff
@@ -1623,6 +1623,19 @@ def setup_request_form(self):
...
+                'default': 'No',
+                'order': 49,
+                'required': False,
+                'hidden': True
+            },
+            'iThenticate_plagiarism_check_api_key': {
+                'value-regex': '.*',
+                'order': 50,
+                'required': False,
+                'hidden': True
```
melisabok, line 1638: @akshatchhabra the base url field is missing here.

melisabok: `iThenticate_plagiarism_check_api_base_url`

tests/test_aaai_conference.py
```diff
@@ -73,7 +73,9 @@ def test_create_conference(self, client, openreview_client, helpers, profile_man
                 'Expected Submissions': '10000',
                 'use_recruitment_template': 'Yes',
                 'api_version': '2',
-                'submission_license': ['CC BY 4.0']
+                'submission_license': ['CC BY 4.0'],
+                'iThenticate_plagiarism_check': 'Yes',
+                'iThenticate_plagiarism_check_api_key': '1234',
```
melisabok, line 78: @akshatchhabra add to the test the value for `iThenticate_plagiarism_check_api_base_url`

akshatchhabra: Done

openreview/venue_request/venue_request.py
akshatchhabra: Done

openreview/api/iThenticate_client.py
```diff
@@ -0,0 +1,233 @@
...
+        if viewer_permissions is not None:
+            data["author_metadata_override"] = author_metadata_override
+        if viewer_permissions is not None:
+            data["sidebar"] = sidebar
+        headers = self.headers.copy()
+        headers["Content-Type"] = "application/json"
+        response = requests.get(
+            f"https://{self.TCA_URL}/api/v1/submissions/{submission_id}/viewer-url",
+            headers=headers,
+            json=data,
```
melisabok, line 211: do you need to pass a JSON in a GET request?

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+        self.invitation_builder.set_iThenticate_plagiarism_check_invitation()
+
+        iThenticate_client = openreview.api.iThenticateClient(self.iThenticatePlagiarismCheckApiKey, self.iThenticatePlagiarismCheckApiBaseUrl)
+        
+        submissions = self.get_submissions()
+        for submission in submissions:
+            # Create a new submission ID in iThenticate
+            # TODO - Decide what should go in metadata.group_context.owners
+            res = iThenticate_client.create_submission(
+                owner=submission.content["authorids"]["value"][0],
```
melisabok, line 1031: instead of passing the first author, I would use the signature of the submission.

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+                title=submission.content["title"]["value"],
+                timestamp=datetime.datetime.fromtimestamp(submission.tcdate/1000, tz=datetime.timezone.utc).strftime('%Y-%m-%dT%H:%M:%SZ'),
+                owner_first_name=submission.content["authors"]["value"][0].split(
+                    " ", 1
+                )[0],
+                owner_last_name=submission.content["authors"]["value"][0].split(
```
melisabok, line 1037: you can use the pretty_id of the signature:

```
name = openreview.tools.pretty_id(submission.signatures[0])
first = name.split(' ')[0]
last = name.split(' ')[1:]
```

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+                    " ", 1
+                )[1],
+                owner_email=self.client.get_profile(submission.content["authorids"]["value"][0]).content["emailsConfirmed"][0],
```
melisabok, line 1040: let's use the property preferred email:

```
self.client.get_profile(submission.signatures[0]).get_preferred_email()
```

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+                group_id=f"{self.id}/-/Submission",
```
melisabok, line 1041: you can use: `self.get_submission_id()`

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+            submission_file_binary_data = self.client.get_attachment(
+                id=submission.id, field_name="pdf"
+            )
+
+            submission_file_object = io.BytesIO(submission_file_binary_data)
+
+
+            # update the edge with status file sent
+            iThenticate_edge.label = "File Sent"
+            iThenticate_edge = self.client.post_edge(iThenticate_edge)
```
melisabok, line 1087: I would post this edge after the `upload_submission` function is called.

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+
+
+            # Upload PDF to iThenticate
+            res = iThenticate_client.upload_submission(
+                submission_id=iThenticate_submission_id,
+                file_data=submission_file_object,
+                file_name=submission.content['title']['value']
```
melisabok, line 1094: @carlosmondra don't we need to format the title here? what is it too large or contains markdown/mathjax?

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+    def ithenticate_create_and_upload_submission(self):
+        if not self.iThenticatePlagiarismCheck:
+            raise openreview.OpenReviewException('iThenticatePlagiarismCheck is not enabled for this venue.')
+        
+        self.invitation_builder.set_iThenticate_plagiarism_check_invitation()
+
+        iThenticate_client = openreview.api.iThenticateClient(self.iThenticatePlagiarismCheckApiKey, self.iThenticatePlagiarismCheckApiBaseUrl)
+        
+        submissions = self.get_submissions()
+        for submission in submissions:
```
melisabok, line 1027: you can use `tqdm` to follow the progress of the submissions.

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+            
+
+    def ithenticate_request_similarity_report(self):
+        if not self.iThenticatePlagiarismCheck:
+            raise openreview.OpenReviewException('iThenticatePlagiarismCheck is not enabled for this venue.')
+        
+        iThenticate_client = openreview.api.iThenticateClient(self.iThenticatePlagiarismCheckApiKey, self.iThenticatePlagiarismCheckApiBaseUrl)
+
+        edges = self.client.get_grouped_edges(
+            invitation=self.get_iThenticate_plagiarism_check_invitation_id(), groupby='tail'
```
melisabok, line 1105: what about filtering by label='File Uploaded' here so you don't need to do an if later?

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+        )
+
+        for edge in edges:
```
melisabok, line 1108: use tqdm here.

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+            # change edge status to similarity requested
+            if edge["values"][0].label == "File Uploaded":
+                edge["values"][0].label = "Similarity Requested"
+                updated_edge = self.client.post_edge(edge["values"][0])
```
melisabok, line 1112: again, post the edge after the operation is performed with no errors.

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+                )
+
+    def check_ithenticate_status(self, type):
+        if not self.iThenticatePlagiarismCheck:
+            raise openreview.OpenReviewException('iThenticatePlagiarismCheck is not enabled for this venue.')
+
+        edges = self.client.get_grouped_edges(
+            invitation=self.get_iThenticate_plagiarism_check_invitation_id(), groupby='tail'
+        )
+        if type == 'Upload':
```
melisabok, line 1132: can the type match with the label so we can get different statuses?

openreview/api/iThenticate_client.py
akshatchhabra: This should be a post request. I've pushed the change.

openreview/venue/venue.py
```diff
@@ -1009,6 +1015,125 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+        
+        self.invitation_builder.set_iThenticate_plagiarism_check_invitation()
+
+        iThenticate_client = openreview.api.iThenticateClient(self.iThenticatePlagiarismCheckApiKey, self.iThenticatePlagiarismCheckApiBaseUrl)
+        
+        submissions = self.get_submissions()
+        for submission in submissions:
+            # Create a new submission ID in iThenticate
+            # TODO - Decide what should go in metadata.group_context.owners
+            res = iThenticate_client.create_submission(
```
melisabok, line 1030: I think before creating the submission, we need to check if there is already an edge associated with the openreview note id. Otherwise we can duplicate the requests, same head with multiple tails.

openreview/api/iThenticate_client.py
melisabok: ok, I fixed the conflicts.

openreview/venue/venue.py
akshatchhabra: I looked at a few pre-existing notes but I see that the signatures field has a group ID and not the name of the person submitting the paper.

openreview/venue/venue.py
akshatchhabra: I looked at some pre-existing notes and the signatures field doesn't have the user ID it only contains a group ID for the authors. I wonder if get_profile() will work with that.

openreview/venue/venue.py
akshatchhabra: I felt that posting it after the `upload_submission` function is called could cause a race condition

openreview/venue/venue.py
akshatchhabra: I looked at some pre-existing notes and the signatures field doesn't have the user ID/name of the author it only contains a group ID for the authors.

openreview/venue/venue.py
akshatchhabra: I looked at some pre-existing notes and the signatures field doesn't have the user ID/name of the author it only contains a group ID for the authors.

openreview/venue/venue.py
melisabok: but if upload_submission fails then the status of the edge is wrong. 

I suggest to use a try/except and rollback the status if an error occurs. What do you think?

openreview/venue/venue.py
melisabok: This depends on the status of the submission.

Ida's submissions has "~Ida_Camacho2" as the signature: https://devapi2.openreview.net/notes?id=lLRPAwGsd6

If this doesn't work, we should ask AAAI if using the first author as owner is ok. My preference is to use the signature that agreed the EULA license.

openreview/venue/venue.py
akshatchhabra: Yes, that will work. I'll make that change. Thanks!

openreview/venue/venue.py
akshatchhabra: In that case, would you suggest deleting the submission and edge? Or, should we skip creating a new edge and change the tail in the existing edge to the new submission ID (and delete the previous submission ID)?

openreview/venue/venue.py
akshatchhabra: I've added that change

openreview/venue/venue.py
akshatchhabra: Done

openreview/venue/venue.py
akshatchhabra: Done

openreview/venue/venue.py
akshatchhabra: I've added this change

openreview/venue/venue.py
akshatchhabra: Done

openreview/venue/venue.py
akshatchhabra: I've added a try-except block that rolls back the label value in case of an error

openreview/venue/venue.py
akshatchhabra: I've used the signature for now.

openreview/venue/venue.py
melisabok: That's depends on the edge.label. 

If the label is error then we can submit a new submission but if there is another state then I would just skip that submission.

openreview/venue/venue.py
```diff
@@ -1053,6 +1060,144 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+            name = openreview.tools.pretty_id(submission.signatures[0])
+
+            res = iThenticate_client.create_submission(
+                owner=submission.signatures[0],
+                title=submission.content["title"]["value"],
+                timestamp=datetime.datetime.fromtimestamp(
+                    submission.tcdate / 1000, tz=datetime.timezone.utc
+                ).strftime("%Y-%m-%dT%H:%M:%SZ"),
+                owner_first_name=name.split(" ", 1)[0],
+                owner_last_name=name.split(" ", 1)[1],
```
melisabok, line 1089: if the name has two last names, shouldn't we send both?

openreview/venue/venue.py
```diff
@@ -1053,6 +1060,144 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+            iThenticate_edge = self.client.post_edge(iThenticate_edge)
+
+            try:
+                # Upload PDF to iThenticate
+                res = iThenticate_client.upload_submission(
+                    submission_id=iThenticate_submission_id,
+                    file_data=submission_file_object,
+                    file_name=submission.content["title"]["value"],
+                )
+            except requests.exceptions.HTTPError as e:
```
melisabok, line 1145: shouldn't we catch any exception here?

openreview/venue/venue.py
```diff
@@ -1053,6 +1060,144 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+                    submission_id=updated_edge.tail,
+                    search_repositories=[
+                        "INTERNET",
+                        "SUBMITTED_WORK",
+                        "PUBLICATION",
+                        "CROSSREF",
+                        "CROSSREF_POSTED_CONTENT",
+                    ],
+                )
+            except requests.exceptions.HTTPError as e:
```
melisabok, line 1183: same here

openreview/venue/venue.py
akshatchhabra: This will send both since I am splitting it only on the first space
So after the split at index `0` we will have the first name and at index `1` we will have all the last names separated by space.

openreview/venue/venue.py
akshatchhabra: Changed it to catch all exceptions

openreview/venue/venue.py
akshatchhabra: Done

openreview/venue/venue.py
akshatchhabra: Added the change

openreview/venue/venue.py
```diff
@@ -1075,8 +1075,11 @@ def ithenticate_create_and_upload_submission(self):
 
         submissions = self.get_submissions()
         for submission in tqdm(submissions):
-            # Create a new submission ID in iThenticate
             # TODO - Decide what should go in metadata.group_context.owners
+            if (self.client.get_edges_count(head=submission.id, invitation=self.get_iThenticate_plagiarism_check_invitation_id()) > 0 and
```
melisabok, line 1079: I think making two API calls for every submissions can be too expensive and the process will be very slow for 10k.

You can load all the edges in a dictionary before starting the loop and check the label status using this dictionary instead of doing an API call.

Also my preference is to avoid the sentence `continue` in a loop.

openreview/venue/venue.py
```diff
@@ -1113,37 +1116,41 @@ def ithenticate_create_and_upload_submission(self):
...
+                    invitation=self.get_iThenticate_plagiarism_check_invitation_id(),
+                    head=submission.id,
+                    tail=iThenticate_submission_id,
+                    label="Created",
+                    weight=-1,
+                )
+
+            
+            else:
+                iThenticate_edge = self.client.get_edges(head=submission.id, invitation=self.get_iThenticate_plagiarism_check_invitation_id())[0]
```
melisabok, line 1130: why do you need to do this?

akshatchhabra: In the if statement before I checked if there was a pre-existing edge. If there exists an edge that means it was an edge with the label `'Error...'` So I retrieved the edge, reset its fields, and set the tail as the new submission ID.

openreview/venue/venue.py
akshatchhabra: Sure, I'll make that change

openreview/venue/venue.py
melisabok: My only concern here is that if Error is coming from a later stage, for example: computing the report. We don't need to delete the submission and upload it again, we might need to retry recomputing the report.

Maybe we could have a flag that resubmit the edge with errors if it is on but by default I would ignore them.

akshatchhabra: I think we can create submissions for notes that don't have an edge already, and we can have a separate function that handles edges with errors. In the function that handles edges with errors, 

1. We can make a call to `get_submission_info`, this will tell us if there was an error while uploading. 
2. We can repeat the upload request if an error is returned by `get_submission_info`. 
3. We submit a report generation request regardless of whether `get_submission_info` returned an error.

openreview/venue/venue.py
akshatchhabra: I've removed the multiple API calls, loaded the edges in a dictionary, and removed the `continue` as well

openreview/venue/venue.py
```diff
@@ -1073,50 +1073,49 @@ def ithenticate_create_and_upload_submission(self):
             self.iThenticatePlagiarismCheckApiBaseUrl,
         )
 
+        edges = self.client.get_grouped_edges(invitation=self.get_iThenticate_plagiarism_check_invitation_id(), groupby='head')
+        edges_dict = {edge["id"]["head"]: edge["value"] for edge in edges}
```
melisabok, line 1077: I think the keyword here is `values` instead of `value`

openreview/venue/venue.py
```diff
@@ -1073,50 +1073,49 @@ def ithenticate_create_and_upload_submission(self):
...
+                            #     "id": "7a62f070-c265-11e8-b568-0800200c9a66",
+                            #     "family_name": "test_instructor_2_first_name",
+                            #     "given_name": "test_instrutor_2_last_name",
+                            #     "email": "intructor_2_email@test.com"
+                            # }
+                        ],
+                    },
+                    group_type="ASSIGNMENT",
+                )
+                iThenticate_submission_id = res["id"]
```
melisabok, line 1117: I would suggest to add an else and print the submission that already have an edge so it can help us to know what is happening.

openreview/venue/venue.py
akshatchhabra: Done

openreview/venue/venue.py
akshatchhabra: Done

openreview/venue/venue.py
melisabok: Sounds good to me.

openreview/venue/venue.py
```diff
@@ -1169,8 +1169,9 @@ def ithenticate_request_similarity_report(self):
         )
 
         for edge in tqdm(edges):
-            edge["values"][0].label = "Similarity Requested"
-            updated_edge = self.client.post_edge(edge["values"][0])
+            e = self.client.get_edge(edge["values"][0]["id"])
```
melisabok, line 1172: instead of getting the edge again you can do `e = openreview.api.Edge.from_json(edge["values"][0])` to construct the edge object.

openreview/venue/venue.py
```diff
@@ -1144,6 +1144,9 @@ def ithenticate_create_and_upload_submission(self):
                 except Exception as err:
                     iThenticate_edge.label = "Created"
                     iThenticate_edge = self.client.post_edge(iThenticate_edge)
+
+            else:
+                print(f"Submission {submission.id} already has edges associated with it")
```
melisabok, line 1149: can you print the label of the edge too?

akshatchhabra: Done

openreview/venue/venue.py
akshatchhabra: I've made the change, thanks! 😄

openreview/venue/venue.py
akshatchhabra: I've added the function for error handling. I deviated slightly from this since it was difficult to determine how long I should wait to submit a similarity report request after uploading the file.
Now the steps look like this:
1. We make a call to `get_submission_info`, this will tell us if there was an error while uploading.
2. We repeat the upload request if an error is returned by `get_submission_info`.

Do steps 3 and 4 only if `get_submission_info` didn't return an error.

3. We make a call to `get_similarity_report_status`, this will tell us if there was an error while submitting the similarity report request.
4. We submit a report generation request if an error is returned by `get_similarity_report_status`.

We can use the `ithenticate_request_similarity_report` function to request the similarity report for cases where we re-uploaded the file.

openreview/venue/venue.py
```diff
@@ -1053,6 +1061,244 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+            self.iThenticate_plagiarism_check_api_key,
+            self.iThenticate_plagiarism_check_api_base_url,
+        )
+
+        edges = self.client.get_grouped_edges(invitation=self.get_iThenticate_plagiarism_check_invitation_id(), groupby='head')
+        edges_dict = {edge["id"]["head"]: edge["values"] for edge in edges}
+
+        submissions = self.get_submissions()
+        for submission in tqdm(submissions):
+            # TODO - Decide what should go in metadata.group_context.owners
```
melisabok, line 1082: what should we do with this TODO? do we need to send this data?

openreview/venue/venue.py
```diff
@@ -1053,6 +1061,244 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+                edge = self.client.post_edge(edge)
+
+                try:
+                    res = iThenticate_client.upload_submission(
+                        submission_id=edge.tail,
+                        file_data=submission_file_object,
+                        file_name=self.client.get_note(id=edge.head).content["title"]["value"],
+                    )
+                except Exception as err:
+                    edge.label = "Created"
```
melisabok, line 1194: should we go back to original error here?

openreview/venue/venue.py
```diff
@@ -1053,6 +1061,244 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+                        search_repositories=[
+                            "INTERNET",
+                            "SUBMITTED_WORK",
+                            "PUBLICATION",
+                            "CROSSREF",
+                            "CROSSREF_POSTED_CONTENT",
+                        ],
+                    )
+                except Exception as err:
+                    updated_edge.label = "File Uploaded"
```
melisabok, line 1214: same here

openreview/venue/venue.py
akshatchhabra: I believe this is the group's owners (which should be the PCs in this case?). We successfully made requests without it so we could leave it out as well.

openreview/venue/venue.py
```diff
@@ -1053,6 +1061,244 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+            )
+
+        iThenticate_client = openreview.api.iThenticateClient(
+            self.iThenticate_plagiarism_check_api_key,
+            self.iThenticate_plagiarism_check_api_base_url,
+        )
+
+        edges = self.client.get_grouped_edges(
+            invitation=self.get_iThenticate_plagiarism_check_invitation_id(),
+            label="Error_PROCESSING_ERROR",
```
melisabok, line 1168: PROCESSING_ERROR can come from any part of the workflow? either upload a submission or requesting a similarity report?

akshatchhabra: Yes

openreview/venue/venue.py
```diff
@@ -1053,6 +1061,244 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+            groupby="tail",
+        )
+        
+        for e in tqdm(edges):
+            edge = openreview.api.Edge.from_json(e["values"][0])
+
+            if iThenticate_client.get_submission_status(edge.tail) == "ERROR":
```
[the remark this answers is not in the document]
akshatchhabra: The function I wrote returns the status which would be `ERROR`,`PROCESSING_ERROR` is the error code. I append the error codes while saving the label of the edge.

openreview/venue/venue.py
melisabok: I do think we need to track when this process error happened, during submission upload or similarity report. Based on that we know how to retry.

Could it be two type of error: "Error_upload_PROCESSING_ERROR" and "Error_similarity_PROCESSING_ERROR". 

Then you can upload the submission in the first error or request the similarity report in the second.

openreview/venue/venue.py
akshatchhabra: This is for catching 400, or 500 errors. Errors like `Error_CANNOT_EXTRACT_TEXT` won't be caught here. I just realized I don't have something to handle these cases. I can add changes to handle these cases. I can check edges still in the `Created` or `File Uploaded` status and send requests for them again in the handle errors function.

openreview/venue/venue.py
akshatchhabra: According to the documentation, both parts can have PROCESSING_ERROR.

openreview/venue/venue.py
akshatchhabra: In the case of errors like `Error_CANNOT_EXTRACT_TEXT` an exception does not occur

akshatchhabra: You're right

akshatchhabra: I'll make the change

melisabok (approved):

PR #2215, opened 2024-06-26
Fix/retries and await

carlosmondra:
This PR fixes how retries are made in openreview-py and how the error is read.
I also migrated the test_edges file to use API 2. However, I could not make the polling strategy work because the count that I was getting was cached, so the polling timed out every time.

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -16,7 +16,7 @@ import os
 import re
 import time
 import jwt
-import traceback
+import json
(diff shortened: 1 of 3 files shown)


melisabok:
- Could you make the change in the client that when there is a rate limit, it waits to run it again? I think it is a parameter.

- Could you also disable the old client to use api2.openreview.net?

melisabok (approved):

PR #2223, opened 2024-07-01
Feature: Add ARR AC checklist options to PC console

haroldrubio:
This PR adds email options to the ARR PC console for ACs and their respective checklists


melisabok (approved):

tests/test_arr_venue_v2.py
```diff
@@ -4273,4 +4280,182 @@ def test_review_rating_forms(self, client, openreview_client, helpers, test_clie
...
+                groupby='tail',
+                select='head'
+            )
+        }
+
+        acs_with_missing_checklists = set()
+        # Check note data directly
+        for ac in area_chairs:
+            try:
+                assigned_ids = assignment_edges[ac]
```
melisabok, line 4403: to avoid the KeyError you can use `.get` that returns None if the key doesn't exist.

PR #2240, opened 2024-07-09
get profiles by current institution domain

melisabok:
It depends on https://github.com/openreview/openreview-api/pull/490


melisabok:
closing, it is not needed now.

PR #2242, opened 2024-07-09
Feature: ARR Reviewer Assignment Quotas

### LINKED ISSUE #2220: ARR: validate that no more than N assignments are being done per paper after deployment
Pasting the conversation from the slack channel:

"Melisa: After the reviewer proposed assignments are deployed, they become visible to the reviewers and the AEs can add/remove reviewers and invite external reviewers too. At this stage, the validation to restrict 3 reviewers becomes more complicated, here a few questions/scenarios:
How many reviewers can a AE invite?
If the AE needs a new reviewer and invites 3 and the 3 of them accepts the invitation. How do we tell the other 2 reviewers that they are not needed anymore?
If an AE invites a reviewer, the reviewer accepts that invitation but doesn't have a profile. The assignment is not being confirmed until the user signs up. In the meantime the AE could have assigned another reviewer so when the reviewer complete the registration they won't get the paper assigned. What should we do in this case?
The AE has 3 reviewers assigned to a paper, only 2 reviews are submitted, the late reviewer says that they can enter the review soon but they don't do it. So the AE needs to unassign the late reviewer in order to find another reviewer. What if the late reviewer finishes the review? they are not assigned anymore...."

"Vincent: the idea that was discussed with Harold at that time was that we would implement this as a parameter - before reviewer assignment is deployed, we will set it to 3. At a later stage (i.e., after reviewer deployment), when an AC needs additional reviewers we will increase its value (probably to 5 or 6). This restriction is most useful before reviewer deployment because we have seen all sorts of crazy things in previous cycles (there were ACs who assigned as many as 15 reviewers to a submission) even though we clearly asked them not to assign more than 3."

We still need to discuss what is the best way to address this. We don't want AEs to be assignments too many reviewers.

haroldrubio:
This PR adds pre-process checks to the Invite_Assignment and Assignment invitations to ensure that at most 3 reviewers are assigned/invited at any time

Resolves #2220


openreview/arr/arr.py
```diff
@@ -387,6 +387,20 @@ def setup(self, program_chair_ids=[], publication_chairs_ids=[]):
...
 
         setup_arr_invitations(self.invitation_builder)
+
+        self.client.post_group_edit(
+            invitation=self.get_meta_invitation_id(),
+            signatures=[self.venue_id],
+            group=openreview.api.Group(
+                id=self.venue_id,
+                content={
+                    'assignment_quota': {
```
melisabok, line 397: can we call it `reviewer_assignment_quota`?

melisabok: I think the value should be added to the request form otherwise it will be removed and added back every time we call to the setup function.

openreview/venue_request/venue_request.py
```diff
@@ -1617,6 +1617,12 @@ def setup_request_form(self):
                 'order': 47,
                 'required': False,
                 'hidden': True
+            },
+            'reviewer_assignment_quota': {
+                'value-radio': [3, 4, 5],
```
melisabok, line 1622: AFAIK the API 1 don't support numbers, the UI always send a string. You can check this posting a request form from the UI.

Also value from 3 to 5 is too restrictive. We can use a value-regex that allows any number but you need to convert the string to a number.

openreview/arr/process/proposed_assignment_pre_process.js
```diff
@@ -4,6 +4,7 @@ async function process(client, edge, invitation) {
   const committeeName = invitation.content.committee_name?.value;
   const { groups } = await client.getGroups({ id: invitation.domain });
   const domain = groups[0];
+  const quota = domain.content.reviewer_assignment_quota?.value
```
melisabok, line 7: what happens if this value is not being set?

openreview/arr/process/proposed_assignment_pre_process.js
```diff
@@ -39,7 +40,7 @@ async function process(client, edge, invitation) {
     return Promise.reject(new OpenReviewError({ name: 'Error', message: `Max Papers allowed reached for ${Tools.getPreferredName(profile)}` }));
   }
 
-  if ((submissionEdges.length + 1) > 3 && committeeName.includes('Reviewers')) {
+  if ((submissionEdges.length + 1) > quota && committeeName.includes('Reviewers')) {
     return Promise.reject(new OpenReviewError({ name: 'Error', message: `You cannot assign more than 3 reviewers to this paper` }));
```
melisabok, line 44: you may need to change the message error too

openreview/arr/process/proposed_assignment_pre_process.js
melisabok: yes, I think we should check if there is a quota available then do the validation.

haroldrubio:
I've added a test to ensure that the pre-process validation is skipped if it is not provided in the venue request form (in the mock June cycle), but I'm not sure if this is necessary

openreview/venue/group.py
```diff
@@ -291,6 +291,9 @@ def create_venue_group(self):
         if self.venue.source_submissions_query_mapping:
             content['source_submissions_query_mapping'] = { 'value': self.venue.source_submissions_query_mapping }    
 
+        if self.venue.reviewer_assignment_quota:
```
melisabok, line 294: If this property is going to be used in the ARR venue only then we don't need to save it in the venue code. You should do it for the ARR venues only.

If this property is going to be used by other venues too then we should implement it in the pre processes of the venue class too.

openreview/arr/process/invite_assignment_pre_process.js
```diff
@@ -0,0 +1,70 @@
...
+  if (edges.length) {
+    return Promise.reject(new OpenReviewError({ name: 'Error', message: `Can not invite ${userProfile.id}, the user is already assigned` }))
+  }
+
+  const [{ edges: inviteAssignmentEdges }, { edges: assignmentEdges }] = await Promise.all([
+      client.getEdges({ invitation: edge.invitation, head: edge.head }),
+      client.getEdges({ invitation: assignmentInvitationId, head: edge.head })
+  ])
+
+  if (quota && inviteAssignmentEdges.length + assignmentEdges.length >= quota) {
```
melisabok, line 46: can you avoid the getEdges calls if quota is not set?

openreview/arr/arr.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, client, venue_id, support_user, venue=None):
         self.use_publication_chairs = False
         self.source_submissions_query_mapping = {}
         self.sac_paper_assignments = False
+        self.reviewer_assignment_quota = None
```
melisabok, line 89: can we call this `submission_assignment_max_reviewers` so it matches with the name convention of the other quotas?

I also think that `reviewer_assignment_quota` can be interpreted as the max number of papers to assign to a reviewer.

openreview/venue/process/proposed_assignment_pre_process.js
```diff
@@ -37,4 +40,8 @@ async function process(client, edge, invitation) {
     return Promise.reject(new OpenReviewError({ name: 'Error', message: `Max Papers allowed reached for ${Tools.getPreferredName(profile)}` }));
   }
 
+  if (quota && (submissionEdges.length + 1) > quota && committeeName.includes('Reviewers')) {
+    return Promise.reject(new OpenReviewError({ name: 'Error', message: `You cannot assign more than 3 reviewers to this paper` }));
```
melisabok, line 44: make sure you parametrize the number 3 here.

openreview/venue/process/proposed_assignment_pre_process.js
```diff
@@ -4,6 +4,7 @@ async function process(client, edge, invitation) {
   const committeeName = invitation.content.committee_name?.value;
   const { groups } = await client.getGroups({ id: invitation.domain });
   const domain = groups[0];
+  const quota = domain.content.reviewer_assignment_quota?.value
```
melisabok, line 7: this quota must be retrieved when `committeeName` is a reviewer or there should be another . This process function is being used for other roles too. 

I would do:
```
domain.content['submission_assignment_max_' + committeeName.toLowerCase()]?.value
```

openreview/venue/process/proposed_assignment_pre_process.js
melisabok: remove this condition:  `committeeName.includes('Reviewers')`

the quota is already configured for reviewers only.

openreview/venue/process/assignment_pre_process.js
```diff
@@ -5,15 +5,25 @@ async function process(client, edge, invitation) {
   const domain = groups[0]
   const venueId = domain.id
   const submissionName = domain.content.submission_name?.value
+  const quota = domain.content.reviewer_assignment_quota?.value
```
melisabok, line 8: same here:
```
domain.content['submission_assignment_max_' + committeeName.toLowerCase()]?.value
```

openreview/venue/process/assignment_pre_process.js
```diff
@@ -5,15 +5,25 @@ async function process(client, edge, invitation) {
...
   const reviewName = invitation.content.review_name?.value
   const reviewersAnonName = invitation.content.reviewers_anon_name?.value
   const reviewersName = invitation.content.reviewers_name?.value
+  const inviteAssignmentId = invitation.id.replace('Assignment', 'Invite_Assignment')
```
melisabok, line 12: can you get this invitation from the domain?

openreview/venue/process/invite_assignment_pre_process.js
```diff
@@ -3,6 +3,7 @@ async function process(client, edge, invitation) {
 
   const { groups } = await client.getGroups({ id: edge.domain })
   const domain = groups[0]
+  const quota = domain.content.reviewer_assignment_quota?.value
```
melisabok, line 6:
```
domain.content['submission_assignment_max_' + committeeName.toLowerCase()]?.value
```

openreview/venue/process/invite_assignment_pre_process.js
```diff
@@ -3,14 +3,15 @@ async function process(client, edge, invitation) {
...
   const domain = groups[0]
-  const quota = domain.content.reviewer_assignment_quota?.value
   const reviewersId = invitation.content.match_group?.value
   const assignmentInvitationId = invitation.content.assignment_invitation_id?.value
   const conflictInvitationId = invitation.content.conflict_invitation_id?.value
   const assignmentLabel = invitation.content.assignment_label?.value
   const inviteLabel = invitation.content.invite_label?.value
   const conflictPolicy = domain.content.reviewers_conflict_policy?.value
   const conflictNYears = domain.content.reviewers_conflict_n_years?.value
+  const reviewersName = reviewersId.split('/').pop().toLowerCase()
```
melisabok, line 13: there is a property called `reviewers_name` that you can use, no need to split the group id.

openreview/venue/process/proposed_assignment_pre_process.js
```diff
@@ -40,8 +40,8 @@ async function process(client, edge, invitation) {
     return Promise.reject(new OpenReviewError({ name: 'Error', message: `Max Papers allowed reached for ${Tools.getPreferredName(profile)}` }));
   }
 
-  if (quota && (submissionEdges.length + 1) > quota && committeeName.includes('Reviewers')) {
-    return Promise.reject(new OpenReviewError({ name: 'Error', message: `You cannot assign more than 3 reviewers to this paper` }));
+  if (quota && (submissionEdges.length + 1) > quota) {
+    return Promise.reject(new OpenReviewError({ name: 'Error', message: `You cannot assign more than ${quota} reviewers to this paper` }));
```
melisabok, line 44: you missed the replacement of the `reviewers` string.

openreview/venue/process/invite_assignment_pre_process.js
haroldrubio: I can't call that directly right because this is used for other roles as well?

melisabok (approved):

celestemartinez (approved):

PR #2243, opened 2024-07-09
Feature: ARR emergency reviewer options

haroldrubio:
This PR adds some additional messaging options to recruit reviewers who have indicated themselves as available


openreview/arr/webfield/programChairsWebfield.js
```diff
@@ -169,6 +169,62 @@ return {
...
+        if (row.notesInfo.length > 0){
+          return false;
+        }
+
+        const registrationNotes = row.reviewerProfile?.registrationNotes ?? []
+        if (registrationNotes.length <= 0) {
+          return false
+        }
+
+        const maxLoadForm = registrationNotes.filter(note => {
```
melisabok, line 184: why do you check this? the label indicates reviewers with no assignments only.

melisabok: ohhh I see, you need to check if they are available.

openreview/arr/webfield/programChairsWebfield.js
```diff
@@ -169,6 +169,62 @@ return {
...
+          const invitations = note?.invitations ?? []
+          return invitations.some(inv => inv.includes('Reviewers/-/Max_Load_And_Unavailability_Request'))
```
melisabok, line 186: for future functions, no need to check all invitations, checking the first of the array should be enough. 

The first invitation is the one that is being used to create the note.

melisabok (approved):

PR #2245, opened 2024-07-10
support signatures as readers in the custom stage

melisabok:


celestemartinez (approved):

openreview/stages/venue_stages.py
```diff
@@ -1523,15 +1524,15 @@ def get_readers(self, conference, number):
...
+        if self.allow_de_anonymization or self.Participants.SIGNATURES in self.readers:
             readers.append('${3/signatures}')
 
         return readers
     
     def get_nonreaders(self, conference, number):
         nonreaders = []
 
-        if conference.use_area_chairs and self.Participants.AREA_CHAIRS_ASSIGNED in self.readers and conference.use_secondary_area_chairs and self.Participants.SECONDARY_AREA_CHAIRS not in self.readers:
+        if conference.use_area_chairs and self.Participants.AREA_CHAIRS_ASSIGNED in self.invitees and conference.use_secondary_area_chairs and self.Participants.SECONDARY_AREA_CHAIRS not in self.readers:
```
celestemartinez, line 1535: I am not sure I understand this change. What if the ACs are added as readers only (not invitees) but we want to hide these notes from the secondary ACs?

melisabok: this is to make the CVPR test to pass,  assigned AC is not readers anymore so the secondary ACs are not added as nonreaders.

I thought that changing this to invitees still works. 

another options is to check if: assigned AC in readers or (assigned AC in invitees and signatures in readers)?

I'm not sure how often we are using the secondary AC for the new venues.

celestemartinez: If the signatures is added as readers, then we don't need to add the Secondary ACs as nonreaders, because they already are not readers. (they are readers only when the paper AC group is added as readers, if it's only the signatures then they are not readers anyways)

celestemartinez: so I think we can remove this from the test?

melisabok: ohhhh you are right! let me remove them.

celestemartinez (approved):

PR #2254, opened 2024-07-15
give access to ARR submissions to commitment venues

melisabok:
ARR class function to give access to the commitment venues where their paper_link submission is pointing to an API 2 ARR submission. 

The permissions are: 

- Be member of ARR submission readers group
- Be reader of the ARR submission note
- Be reader of the ARR official review and meta review notes

For now the member is only the commitment venue id. In the future, I could add the SAC or AC if they want to give access to assigned SAC/AC.

They way to call it is, the client must be created with the super user credentials in order to see the ARR and commitment submissions :

```
openreview.arr.ARR.process_commitment_venue(client, 'aclweb.org/ACL/2024/Workshop/C3NLP_ARR_Commitment')
```


racheljsmart (approved):

openreview/arr/arr.py
```diff
@@ -558,4 +558,96 @@ def set_SAC_ethics_review_process(self, sac_ethics_flag_duedate=None):
...
+            print(f'Add group as reader of the submission')
+            if 'everyone' not in submission.readers:
+                add_readers_to_note(submission, [commitment_readers_group_id])
+
+            print(f'Add group as reader of the submission review and meta reviews')
+            replies = client.get_notes(forum = submission.id)
+
+            for reply in replies:
+
+                if 'Official_Review' in reply.invitations[0]:
```
haroldrubio, line 625: Do you think we should parameterize which replies get revealed?

melisabok: Yes, I can do that. This could be an optional field where the default value is "Official_Review" and "Meta_Review".

I would support invitation names only for now, I know there is some logic to release comments based on readers.

haroldrubio (approved):

PR #2260, opened 2024-07-16
Fix/invitaiton param

carlosmondra:


.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ parameters:
     default: "master"
   openreview-api-v2-branch:
     type: string
-    default: "main"
+    default: "feature/validate-param"
```
carlosmondra, line 11:
```suggestion
    default: "main"
```

melisabok:
Why do we need to remove the regex keyword?

carlosmondra:
> Why do we need to remove the regex keyword?

From which one?
This one? --> In this case it is superfluous

{ regex: '.*' , type: 'string' }

or this one? --> Do you want to use regex to validate integers? I would assume that using keywords like minimum and maximum or range or enum would be better, but I'm open to counter arguments, of course.

{ regex: '.*' , type: 'integer' }

celestemartinez (approved):

PR #2267, opened 2024-07-17
Custom Stage: add venue id as committee instead of PCs + Invite Assignment fix

### LINKED ISSUE #2261: ARR: Expire checklist invitations if a paper is desk rejected/withdrawn
It seems the invitations are still active when the paper is desk rejected. Check paper from ARR.

### LINKED ISSUE #2262: Invite Assignment: remove the user from the invited group only when there are no pending invitations
An invitation to recruit a user can be cancelled by the PCs/SAC/AC but only remove the user from the invited group if there are no other invitations sent to that user.

Another option is to never remove them, does it break anything?

celestemartinez:
Fixes #2261

The reason the desk-rejection process function is not getting some invitations to expire them is because those invitations have PCs as invitees instead of the venue id. This was being done by the custom stage, so I am changing it there.

Also fixes #2262

### DIFF
--- a/openreview/stages/venue_stages.py
+++ b/openreview/stages/venue_stages.py
@@ -1461,7 +1461,7 @@ class CustomStage(object):
         self.allow_de_anonymization = allow_de_anonymization
 
     def get_invitees(self, conference, number):
-        invitees = [conference.get_program_chairs_id()]
+        invitees = [conference.id]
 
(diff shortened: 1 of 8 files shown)


melisabok (approved):
💯

haroldrubio (approved):

PR #2276, opened 2024-07-19
ARR: Fix Checklist Flagging

### LINKED ISSUE #2275: ARR: Review and Checklist Flagging
The following changes need to be made to the flagging workflow:

1. Send flagging and unflagging emails whenever the flag is set, not just the first time
2. Make the flagging information global and avoid passing reply-specific arguments in the process function

haroldrubio:
Resolves #2275

This PR changes the `flag_submission` function to check all replies and no longer requires on reply-specific arguments

### DIFF
--- a/openreview/arr/helpers.py
+++ b/openreview/arr/helpers.py
@@ -1502,9 +1502,6 @@ def flag_submission(
         }
     }
 
-    dsv_flagged = forum.content.get('flagged_for_desk_reject_verification', {}).get('value', False)
-    ethics_flagged = forum.content.get('flagged_for_ethics_review', {}).get('value', False)
-
     def post_flag(invitation_name, value=False):
         print(f"posting {value} to {invitation_name}")
         return client.post_note_edit(
@@ -1593,6 +1590,11 @@ def flag_submission(
                 violation_field,
                 field_default
             )
+
+    # Fetch fresh statuses
+    forum = client.get_note(id=edit.note.forum)
+    dsv_flagged = forum.content.get('flagged_for_desk_reject_verification', {}).get('value', False)
+    ethics_flagged = forum.content.get('flagged_for_ethics_review', {}).get('value', False)
 
     # Check ethics flag
     ## False -> True
--- a/openreview/arr/helpers.py
+++ b/openreview/arr/helpers.py
@@ -1479,6 +1479,12 @@ def flag_submission(
             'anonymity': 'Yes',
             'responsible_checklist': 'Yes',
             'limitations': 'Yes'
+        },
+        'Official_Review': {
+            'Knowledge_of_or_educated_guess_at_author_identity': 'No'
+        },
+        'Meta_Review': {
+            'author_identity_guess': [4, 3, 2, 1]
         }
     }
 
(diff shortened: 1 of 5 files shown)


openreview/arr/helpers.py
```diff
@@ -1467,19 +1466,24 @@ def flag_submission(
...
+            'responsible_checklist': 'Yes',
+            'limitations': 'Yes'
+        }
+    }
 
-    dsv_flagged = forum.content.get('flagged_for_desk_reject_verification', {}).get('value')
-    ethics_flagged = forum.content.get('flagged_for_ethics_review', {}).get('value')
-    has_ethic_flag_history = len(ethics_flag_edits) > 0
-    has_dsv_flag_history = len(dsv_flag_edits) > 0
+    dsv_flagged = forum.content.get('flagged_for_desk_reject_verification', {}).get('value', False)
```
melisabok, line 1485: do we need a default value here? I'm not sure if we send a message the is paper is unflagged but it was never flagged.

melisabok:
I would add an assertion for every post of a checklist and a review and make sure the flag is correct and if an email is expected or not.

openreview/arr/helpers.py
haroldrubio: Action is only taken on status change, so if this value is False and none of the checklists are flag the submission then no message is sent

openreview/arr/helpers.py
```diff
@@ -1467,106 +1466,135 @@ def flag_submission(
...
+        return client.post_note_edit(
             invitation=f'{venue_id}/-/{invitation_name}_Flag',
             note=openreview.api.Note(
                 id=edit.note.forum,
                 content={f'flagged_for_{invitation_name.lower()}': {'value': value}}
             ),
             signatures=[venue_id]
         )
+
+    def check_field_violated(note, field, default_value):
```
celestemartinez, line 1505: @haroldrubio I am a little bit confused about what this function does and why we have to check this?

carlosmondra (approved):

openreview/arr/helpers.py
melisabok: I think this function is the one that checks in the content of the note if there is an indicator that the submission needs ethics reviews:

for official reviews, it checks the field `needs_ethics_review` violates the default value that is 'No'

haroldrubio: Yes that's correct it's just a small helper function for that

celestemartinez: I see, thanks! I was also confused because of the `violation_fields` but now I see those are to check for desk reject verification

celestemartinez (approved):

openreview/arr/helpers.py
```diff
@@ -1481,106 +1480,135 @@ def flag_submission(
...
+                field_default
             )
-    
-    # Ethics Flagging
-    if needs_ethics_review and not has_ethic_flag_history:
-        print('flagging ethics and emailing')
+
+    # Check ethics flag
+    ## False -> True
+    if not ethics_flagged and any([
```
melisabok, line 1599: Instead of getting `ethics_flagged` at the beginning of the process function. I would get it at the moment that we evaluate if we need to flag it or not to avoid race conditions.

melisabok (approved):

PR #2288, opened 2024-07-23
Update README.md

pmandler-umass:
Separate SUPERUSER and npm run to 2 lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,8 @@ npm run cleanStart
 
 Inside the OpenReview Web directory
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -90,7 +90,8 @@ npm run cleanStart
 
 Inside the OpenReview Web directory
 ```bash
-SUPER_USER=openreview.net npm run dev
+SUPER_USER=openreview.net
```
melisabok, line 93: Why do you want to run it in two lines? one line should work.

PR #2296, opened 2024-07-26
Feature: ARR email option for reminding reviewers about rebuttals

haroldrubio:
This PR adds a new option to the ARR PC console to email reviewers that have not responded to their rebuttals. This will send an email if any of their assigned papers is missing a reviewer response to an author rebuttal of their official review.

### DIFF
--- a/openreview/arr/webfield/programChairsWebfield.js
+++ b/openreview/arr/webfield/programChairsWebfield.js
@@ -171,6 +171,75 @@ return {
     },
     reviewerEmailFuncs: [
       {
+        label: 'Reviewers with No Rebuttal Responses', filterFunc: `
+        console.log(row);
+        if (row.notesInfo.length <= 0){
+          return false;
+        }
+        console.log('pass notes validation')
+
+        const buildReplyToMap = (note) => {
+          console.log('building map now');
+          const replyToMap = new Map();
+          console.log('map initialized');
+          (note?.details?.replies ?? []).forEach((reply) => {
+            console.log('in reply iteration');
+            if (!replyToMap.has(reply.replyto)) {
+              replyToMap.set(reply.replyto, []);
+            }
+            replyToMap.get(reply.replyto).push(reply);
+          });
+          console.log('map building complete');
+          return replyToMap;
+        }
+        console.log('pass declare build')
+
+        const childrenWithSignature = (reply, signature, replyToMap) => {
+          console.log(signature);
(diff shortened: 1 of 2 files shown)


openreview/arr/webfield/programChairsWebfield.js
```diff
@@ -170,6 +170,53 @@ return {
...
     },
     reviewerEmailFuncs: [
+      {
+        label: 'Reviewers with No Rebuttal Responses', filterFunc: `
+        if (row.notesInfo.length <= 0){
+          return false;
+        }
+
+        const buildReplyToMap = (note) => {
+          const replyToMap = new Map();
```
melisabok, line 180: do you really need to use a Map here? what about using `{}`?

openreview/arr/webfield/programChairsWebfield.js
```diff
@@ -170,6 +170,53 @@ return {
...
+          const officialReview = noteObj?.officialReview
+          if (!officialReview) {
+            return false;
+          }
+          const anonId = officialReview.anonymousId;
+          const authorReplies = childrenWithSignature(officialReview, '/Authors', noteReplyMap);
+          if (authorReplies.length <= 0) {
+            return false;
+          }
+          const reviewerResponses = childrenWithSignature(officialReview, anonId, noteReplyMap);
```
melisabok, line 215: A simpler way of doing this is to find all the replies in the forum where the signature is an author. For each reply check there is a direct reply made by the reviewer. This is tricky because some authors post 3 comments for their rebuttal given the character limit and the reviewer will reply to one of them only. How do you solve that?

openreview/arr/webfield/programChairsWebfield.js
```diff
@@ -186,14 +186,32 @@ return {
           });
           return replyToMap;
         }
+
+        const getAuthorRebuttals = (reply, note) => {
```
melisabok, line 190: I still don't understand why you are doing this?

Can we just iterate the array of replies and check the authors in readers/signatures? Why do you need to use while and update the queue object in the loop?

melisabok:
@haroldrubio can this be merged?

haroldrubio:
I think I need to make a change because it looks like it is not handling the case where an author only sets the reviewer as the reader, and not the entire `SubmissionX/Reviewers` or `SubmissionX/Reviewers/Submitted`

We can hold off on merging this - I'll convert it to draft and open it again once I've made the change

PR #2306, opened 2024-08-02
Update setup.py to require Python 3.9

emily-grabowski:
Add requirement for Python 3.9 in setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,9 @@ setup(
         'openreview/arr',
         'openreview.api'
     ],
+
+    python_requires='==3.9.*', 
+
     install_requires=[
         'pycryptodome',
         'requests>=2.18.4',
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
         'openreview.api'
     ],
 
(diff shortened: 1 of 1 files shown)


carlosmondra:
Is it possible to make the requirement accept versions 3.8 - 3.10?
circleci tests run with version 3.10 and my local installation works with 3.8. I suppose several users use these versions. I would also check if it works with versions 3.6 and 3.7 even if they are going to be deprecated soon.

An alternative would be to make these checks in circleci. We could add a job that tries to install openreview-py with every version that we want to support.

melisabok:
CircleCI is already failing because of this change so if Python 3.10 works then we accept this version too.

emily-grabowski:
I can change this to include a range of Python versions that are accepted - I think the function was removed in Python 3.12, so anything < Python 3.12 should work

emily-grabowski:
I updated setup.py to require <Python 3.12 and tested it on Python 3.8-3.12

melisabok (approved):

carlosmondra (approved):

PR #2310, opened 2024-08-06
iThenticate: get owner from edit signatures and eula version from submission.content

melisabok:


openreview/venue/venue.py
```diff
@@ -1088,10 +1088,21 @@ def ithenticate_create_and_upload_submission(self):
...
                     if submission.signatures[0].startswith("~")
-                    else submission.content["authorids"]["value"][0]
+                    else self.client.get_note_edits(note_id=submission.id, invitation=self.get_submission_id(), sort='tcdate:asc')[0].signatures[0]
                 )
+                print(f"Creating submission for {submission.id} with owner {owner}")
                 owner_profile = self.client.get_profile(owner)
 
+                eula_version = submission.content.get("iThenticate_agreement", {}).get("value", "v1beta").split(":")[-1].strip()
+                
+                iThenticate_client.accept_EULA(
```
melisabok, line 1098: is this a new request required?

openreview/api/iThenticate_client.py
```diff
@@ -1,7 +1,7 @@
 import os
 import requests
 import json
-from datetime import datetime
+import urllib.parse
```
melisabok, line 4: where do we use this new import?

openreview/venue/venue.py
melisabok: you are passing the profile.id as user_id but I don't see the user_id in the create_submission call, how iThenticate matches the agreement with the submission?

akshatchhabra: This is so that iThenticate has it in their record that the user accepted EULA, what version they accepted, and the timestamp.

openreview/api/iThenticate_client.py
akshatchhabra: We're using it to parse the filename we send when uploading files.

openreview/venue/venue.py
akshatchhabra: > you are passing the profile.id as user_id but I don't see the user_id in the create_submission call, how iThenticate matches the agreement with the submission?

I believe iThenticate uses the owner field from the create_submission call as the user_id

melisabok: the owner field might not be the same than the user_id.  An author can sign the edit using a tilde id but their profile.id can be different. 

In order to make sure they match, you need to pass `owner=owner_profile.id`

openreview/venue/venue.py
```diff
@@ -1088,10 +1088,21 @@ def ithenticate_create_and_upload_submission(self):
...
+                    user_id=owner_profile.id,
+                    eula_version=eula_version,
+                    timestamp=datetime.datetime.fromtimestamp(
```
melisabok, line 1101: this timestamp is computed twice, can you save it in variable and use it for both calls?

openreview/venue/venue.py
akshatchhabra: Made the change. Thanks for catching this.

PR #2323, opened 2024-08-15
Add autoapproval for some types of name deletion requests

emily-grabowski:


openreview/profile/process/request_remove_name_process.py
```diff
@@ -12,8 +12,29 @@ def process(client, edit, invitation):
...
+            invitation=REMOVAL_DECISION_INVITATION_ID,
+            signatures=[SUPPORT_USER_ID],
+            note=openreview.api.Note(
+                id=edit.note.id,
+                content={
+                    'status': { 'value': 'Accepted' }
+                }
+        ))        
+
+    print("Check if the username appears in any publications")
```
carlosmondra, line 37: Do we need to check publications after it has been approved? Shouldn't we just return in this case?

openreview/profile/process/request_remove_name_process.py
```diff
@@ -12,8 +12,29 @@ def process(client, edit, invitation):
...
+        print('Accepting the name removal request')
+        client.post_note_edit(
+            invitation=REMOVAL_DECISION_INVITATION_ID,
+            signatures=[SUPPORT_USER_ID],
+            note=openreview.api.Note(
+                id=edit.note.id,
+                content={
+                    'status': { 'value': 'Accepted' }
+                }
+        ))
```
carlosmondra, line 35: Maybe we should return after the name has been approved. I don't think we should check for publications.

emily-grabowski: Checking for publications was something that was already in the process function before my changes -  the name would be autoapproved if it did not appear in any publications, so I left the code in for that. I'm not sure whether we want to keep that system, since it means people can make pretty drastic changes to their name as long as its not associated with a submission.

carlosmondra: I'm not sure if I didn't understand or if I phrased my comment incorrectly.

So suppose that you auto approve the name change because it passes your validation. The name will be changed, but it will still send an email to the user if it has publications telling him that we are reviewing his request. Is this what we want?

celestemartinez: Should we do this validation then only if no publications are found that use this username?

emily-grabowski: @carlosmondra Got it that makes sense- with the way it is currently written it will accept the name deletion request, but still send an email saying that we are reviewing the changes if there are publications - I'll add a return to prevent that

emily-grabowski: Celeste, could you clarify? I'm not sure I follow

celestemartinez: I guess I am just wondering if we should add this check to automatically approve after we have checked if there are publications. It looks like if there are publications, we always send the email about evaluating the request. 
I am not sure if it's correct to automatically approve some cases before we check if the user has publications?

emily-grabowski: The original code would autoapprove all requests that had no publications and moderate all requests associated with publications. 

The change I proposed was to reduce moderation by also auto-approving certain simple changes (capitalization, spacing, name reversal) that are associated with publications. Does this seem like a reasonable change to you? I'm happy to discuss further if you think it'll cause any issues.

celestemartinez: ohh I understand now, thanks for clarifying! Your change looks good.

carlosmondra (approved):
Awesome!

PR #2325, opened 2024-08-16
Assignment preprocesses: Fix domain content keys

enrubio:
- Fixes the issue mentioned here: https://github.com/openreview/openreview-py/pull/2201#discussion_r1700795788

In some preprocess functions we retrieve fields from the domain using the committee name in the key. The problem is that these fields are hardcoded and don't use the committee name. It's an issue for venues that use custom role names.

### DIFF
--- a/openreview/venue/process/assignment_pre_process.js
+++ b/openreview/venue/process/assignment_pre_process.js
@@ -8,8 +8,8 @@ async function process(client, edge, invitation) {
   const reviewName = invitation.content.review_name?.value
   const reviewersAnonName = invitation.content.reviewers_anon_name?.value
   const reviewersName = invitation.content.reviewers_name?.value
-  const quota = domain.content?.['submission_assignment_max_' + reviewersName.toLowerCase()]?.value
-  const inviteAssignmentId = domain.content?.[reviewersName.toLowerCase() + '_invite_assignment_id']?.value
+  const quota = domain.content?.['submission_assignment_max_reviewers']?.value
+  const inviteAssignmentId = domain.content?.['reviewers_invite_assignment_id']?.value
 
   const { notes } = await client.getNotes({ id: edge.head })
   const submission = notes[0]
--- a/openreview/venue/process/proposed_assignment_pre_process.js
+++ b/openreview/venue/process/proposed_assignment_pre_process.js
@@ -4,9 +4,15 @@ async function process(client, edge, invitation) {
   const committeeName = invitation.content.committee_name?.value;
   const { groups } = await client.getGroups({ id: invitation.domain });
   const domain = groups[0];
-  const quota = domain.content?.['submission_assignment_max_' + committeeName.toLowerCase()]?.value
-
-  const customMaxPapersId = domain.content[committeeName.toLowerCase() + '_custom_max_papers_id']?.value;
+  const reviewersName = domain.content['reviewers_name']?.value
+  const areaChairsName = domain.content['area_chairs_name']?.value
+  const roleMap = {
+    [reviewersName]: 'reviewers',
+    [areaChairsName]: 'area_chairs',
+  };
+  const internalRole = roleMap[committeeName]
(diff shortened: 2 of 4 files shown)


openreview/venue/process/assignment_pre_process.js
```diff
@@ -8,8 +8,8 @@ async function process(client, edge, invitation) {
   const reviewName = invitation.content.review_name?.value
   const reviewersAnonName = invitation.content.reviewers_anon_name?.value
   const reviewersName = invitation.content.reviewers_name?.value
-  const quota = domain.content?.['submission_assignment_max_' + reviewersName.toLowerCase()]?.value
-  const inviteAssignmentId = domain.content?.[reviewersName.toLowerCase() + '_invite_assignment_id']?.value
+  const quota = domain.content?.['submission_assignment_max_reviewers']?.value
+  const inviteAssignmentId = domain.content?.['reviewers_invite_assignment_id']?.value
```
melisabok, line 12: I think this won't work for area chairs assignment. May be the invitation should have another variable here.

enrubio: @melisabok Do you know what you meant by this? Should I add a check for whether or not the assignment is for ACs or reviewers?

We don't have a `submission_assignment_max_area_chairs` field currently and it looks like we don't usually create an invite assignment invitation for ACs, but I could have missed something.

melisabok: I think we do have AC invite assignment, @celestemartinez can you confirm that?

enrubio: What I did was search for `get_assignment_id` in the code and see if we ever pass `invite=True` for ACs. I see `setup_invite_assignment` is called only for reviewers.

But venues can ask for SACs to reassign ACs (if SACs are assigned to papers). In which case, we would need to create the invite invitation. So I will make changes to make sure all the roles are covered.

openreview/venue/process/proposed_assignment_pre_process.js
```diff
@@ -4,9 +4,15 @@ async function process(client, edge, invitation) {
   const committeeName = invitation.content.committee_name?.value;
   const { groups } = await client.getGroups({ id: invitation.domain });
   const domain = groups[0];
-  const quota = domain.content?.['submission_assignment_max_' + committeeName.toLowerCase()]?.value
-
-  const customMaxPapersId = domain.content[committeeName.toLowerCase() + '_custom_max_papers_id']?.value;
+  const reviewersName = domain.content['reviewers_name']?.value
```
melisabok: to avoid doing all of this here, can you just save the `internalRole` as a property of the invitation content like the `committee_name`?

enrubio: I see, okay I can add an `internal_role` field to the content of the Assignment, Proposed Assignment, and Invite Assignment invitations

melisabok: Please call it `committee_role`, that is the name that we use in other invitations.

enrubio: I added a function in the venue class that normalizes the role name, then called it in the invitation content. Let me know if it should be done in a different way

openreview/venue/invitation.py
```diff
@@ -3161,6 +3161,9 @@ def set_assignment_invitation(self, committee_id, submission_content=None):
             },
             'reviewers_anon_name': {
                 'value': venue.get_anon_reviewers_name() if is_reviewer else venue.get_anon_area_chairs_name()
+            },
+            'committee_role': { 
+                'value':  venue.get_standard_committee_role(committee_id=committee_id)
```
melisabok: I don't think this function would work in the case the venue has different names for reviewers: Expertise_Reviewers, Senior_Reviewers, etc. 

You need to get the role by using the variables defined above: `is_reviewer`, `is_area_chair`, `is_senior_area_chair1`

enrubio: I added the checks for committee role

PR #2328, opened 2024-08-26
Fix/Venue: Chat process and date process

celestemartinez:
Check if the group exists before trying to send the email. 
We got many errors because the Reviewers/Submitted group for some papers might not exist at the time when the Chat invitation is enabled.


carlosmondra:
A test is failing here 🥲

enrubio (approved):

carlosmondra (approved):

PR #2332, opened 2024-08-29
Fix/ Remove DBLP Abstract process function

xkopenreview:
dblp_abstract_process process function was added initially to trigger the dblp abstract again in case of cloud function failure or the server specified in html is done (temporarily)

but it looks like when the server specified in html was done it's never restored again or the html url iteself does not resolve (or point to a post script file or the paper does not have abstract) and the paper abstract can only be fixed manually

so  dblp abstract process is only failing again without doing anything useful so remove it


celestemartinez (approved):
If we are sure triggering this process function again is not useful, I am ok with removing it.

carlosmondra (approved):

PR #2336, opened 2024-09-09
Feat/ API2 Client: add get_group_edits

celestemartinez:


openreview/api/client.py
```diff
@@ -1391,6 +1391,32 @@ def get_group_edit(self, id):
...
+        :rtype: list[Edit]
+        """
+        params = {}
+        if group_id:
+            params['group.id'] = group_id
+        if invitation:
+            params['invitation'] = invitation
+        if sort:
+            params['sort'] = sort
+        params['trash'] = 'true' if trash == True else 'false'
```
carlosmondra, line 1408: I thought that `tools.format_params` already did that for us.

celestemartinez: Yes, you're right. I've fixed this in both `get_group_edits()` and `get_note_edits()`

carlosmondra (approved):

PR #2340, opened 2024-09-12
Fix/tmlr requests

celestemartinez:
1. Adds default:0 to Pending Reviews invitation
2. Sends copy of invite assignment email sent to a reviewer to the AE
3. Sends a one-time reminder about invite assignment emails to reviewers who have not accepted/declines the invitation

### DIFF
--- a/openreview/journal/process/remind_invited_reviewer_process.py
+++ b/openreview/journal/process/remind_invited_reviewer_process.py
@@ -1,0 +1,50 @@
+def process(client, invitation):
+
+    print('Remind invited reviewers')
+    journal = openreview.journal.Journal()
+    grouped_edges = client.get_grouped_edges(invitation=journal.get_reviewer_invite_assignment_id(), label='Invitation Sent', groupby='id')
+    
+    if len(grouped_edges) == 0:
+        return
+    
+    edges = [openreview.api.Edge.from_json(edge['values'][0]) for edge in grouped_edges]
+
+    start_reminder_period = openreview.tools.datetime_millis(datetime.datetime.utcnow() - datetime.timedelta(weeks = journal.invite_assignment_reminder_period, days=1))
+    end_reminder_period = openreview.tools.datetime_millis(datetime.datetime.utcnow() - datetime.timedelta(weeks = journal.invite_assignment_reminder_period))
+    one_week = datetime.timedelta(weeks=1)
+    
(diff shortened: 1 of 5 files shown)


openreview/journal/process/remind_invited_reviewer_process.py
```diff
@@ -0,0 +1,50 @@
...
+                user_profile=openreview.Profile(id=user,
+                    content={
+                        'names': [],
+                        'emails': [user],
+                        'preferredEmail': user
+                    })
+                
+            last_modified = datetime.timedelta(milliseconds=edge.tmdate - edge.tcdate)
+
+            if last_modified < one_week:
```
carlosmondra, line 37: I'm not sure why you need to make this check. Can you please explain?

openreview/journal/process/reviewer_invitation_assignment_process.py
```diff
@@ -48,7 +51,7 @@ def process(client, edge, invitation):
 
         # format the message defined above
         subject=f'[{short_phrase}] Invitation to review paper titled "{submission.content["title"]["value"]}"'
-        message=f'''Hi {{{{fullname}}}},
+        message=f'''Hi {user_preferred_name},
```
carlosmondra, line 54: `fullname` doesn't return the preferred name?

openreview/journal/process/remind_invited_reviewer_process.py
celestemartinez: I agree that checking the 7 and 8 day interval is enough, but I am worried that if the date process fails in the middle and we re-run it, many reminders will be sent again. I am using the tmdate here as a tracker to know if the reminder has been sent or not.

carlosmondra: OK, I understand. Just one more question:
if this is true:
```py
if start_reminder_period < edge.tcdate and edge.tcdate < end_reminder_period:
```
isn't this false?
```py
if last_modified < one_week:
```

celestemartinez: `last_modified` is checking the difference between the tmdate and the tcdate, and I am updating the tmdate after I send the reminder. This should be happening about a week after the edge was created.

carlosmondra (approved):

PR #2354, opened 2024-09-24
Redo aaai tests

enrubio:
Updates the AAAI tests to match their workflow.


melisabok:
Merging this because AAAI is done already.

PR #2361, opened 2024-09-27
Fix/ DBLP record process error handling

xkopenreview:
currently the dblp record process assumes that 
Tools.extractAbstract
will always return OpenReviewError which may not hold
extractAbstract returns OpenReviewError only when cloud function returns a response

when extractAbstract does not fail with OpenReviewError the toJson() method is not available and it will cause the process function to fail immediately and DBLP.org/-/Edit edit is not posted (ln35).

this pr should avoid the failure at toJson when error is not OpenReivewError

### DIFF
--- a/openreview/profile/process/dblp_record_process.js
+++ b/openreview/profile/process/dblp_record_process.js
@@ -28,7 +28,7 @@ async function process(client, edit, invitation) {
       console.log('html field is empty');
     }
   } catch (error) {
-    console.log('error: ' + JSON.stringify(error.toJson()));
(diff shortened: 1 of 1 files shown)


carlosmondra:
Is the error returned by the cloud function in JSON format?

carlosmondra:
> > Is the error returned by the cloud function in JSON format?
> 
> yes if it's online. this is to handle the case the cloud function is offline

This change you are making will print `undefined` for the `console.log`. Is this intended?

celestemartinez (approved):

carlosmondra:
@xkopenreview do we still need this change?

PR #2365, opened 2024-09-30
fix: iThenticate fixes

akshatchhabra:
* Index out of range error for users with first name only
* Get correct owner when submission signature is not a tilde ID
* Print number of submissions that are not in the desired state in check_ithenticate_status()
* Handle Openreview Profile not found error in ithenticate_create_and_upload_submission()
* Use correct timestamp format in ithenticate_create_and_upload_submission()


carlosmondra (approved):
Thanks!

PR #2367, opened 2024-10-01
Test using different characters in role names

enrubio:
This is only for testing, not meant to be merged.

Changes the role names to use Japanese characters. Currently breaks around Line 1676 when asserting the edge browser url.

Update:
Commented out edge browser url assertions because the url encoding is different for these characters.

Now the tests break at Line 1681. Not sure why yet, I could have missed a name replacement or there could be another issue. I think it should be tested further.


carlosmondra:
Good idea!

melisabok:
@enrubio I would like to keep these tests but can you do it in a different test file?

enrubio:
@melisabok Okay, I'm thinking of doing that in another PR and closing this one?

melisabok:
Closing this PR, @enrubio is working on a new one to parametrize the role name in the email notifications.

PR #2370, opened 2024-10-08
Feat: Add index param

carlosmondra:
This PR uses concurrent requests for the recruitment to speed it up. 
I also included a retry functionality with backoff strategy in case one of the requests fails. It will retry up to five times, backing off every time it fails.


melisabok:
I tried to solve this in the new Celeste's workflow sending emails in bulk and posting one single edit to add members. 

I need to check these changes again but I think that could speed up the recruitment process.

carlosmondra:
> I tried to solve this in the new Celeste's workflow sending emails in bulk and posting one single edit to add members.
> 
> I need to check these changes again but I think that could speed up the recruitment process.

Closing 😬

PR #2374, opened 2024-10-10
Feature: ARR Rebuttal Comment Email Threshold

### LINKED ISSUE #2265: ARR:  add a threshold in the comment process function to stop notifying the SAC/AC when the threshold is reached
This is to avoid AC/SAC to follow a discussion between authors and reviewers that is taking more than N comments.

haroldrubio:
Resolves #2265

This PR adds a new hidden field `comment_notification_threshold` to the request form and stores the variable in the domain and the Venue class. It also parameterizes the process/pre-process function scripts.

The ARR class passes in its own Official_Comment process function that counts (each reply is only counted once):
1) the number of comments (including the newest one) written by the signature
2) the number of comments in the author-reviewer discussion that's visible to the signature

The process function only emails the SACs/ACs if:
1) the comment threshold wasn't set
2) the comment was written by an author/reviewer and the total count is below the threshold
3) the comment was written by an AC/SAC/PC

I also added a small delay between the stages to stabilize the tests when its creating the custom stages

### DIFF
--- a/openreview/arr/process/comment_process.py
+++ b/openreview/arr/process/comment_process.py
@@ -1,0 +1,148 @@
+def process(client, edit, invitation):
+
+    domain = client.get_group(edit.domain)
+    venue_id = domain.id
+    meta_invitation_id = domain.get_content_value('meta_invitation_id')
+    short_name = domain.get_content_value('subtitle')
+    contact = domain.get_content_value('contact')
+    authors_name = domain.get_content_value('authors_name')
+    submission_name = domain.get_content_value('submission_name')
+    reviewers_name = domain.get_content_value('reviewers_name')
+    reviewers_anon_name = domain.get_content_value('reviewers_anon_name')
+    reviewers_submitted_name = domain.get_content_value('reviewers_submitted_name')
+    sender = domain.get_content_value('message_sender')
+
+    submission = client.get_note(edit.note.forum)
+    comment = client.get_note(edit.note.id)
+    paper_group_id=f'{venue_id}/{submission_name}{submission.number}'
+
+    ### TODO: Fix this, we should notify the use when the review is updated
+    if comment.tcdate != comment.tmdate:
+        return    
+
+    ignore_groups = comment.nonreaders if comment.nonreaders else []
+    ignore_groups.append(edit.tauthor)
+
+    signature = comment.signatures[0].split('/')[-1]
+    pretty_signature = openreview.tools.pretty_id(signature)
+    pretty_signature = 'An author' if pretty_signature == 'Authors' else pretty_signature
+
+    content = f'''
+    
+Paper number: {submission.number}
+
+Paper title: {submission.content['title']['value']}
+
+Comment: {comment.content['comment']['value']}
+
+To view the comment, click here: https://openreview.net/forum?id={submission.id}&noteId={comment.id}'''
+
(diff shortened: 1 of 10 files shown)


openreview/arr/process/comment_process.py
```diff
@@ -0,0 +1,182 @@
...
+    signature = comment.signatures[0].split('/')[-1]
+    pretty_signature = openreview.tools.pretty_id(signature)
+    pretty_signature = 'An author' if pretty_signature == 'Authors' else pretty_signature
+
+    # Count comments between reviewer and authors
+    comment_count = 0
+    signed_by_reviewer = 'Reviewer' in signature
+    signed_by_author = 'Authors' in signature
+
+    for reply in submission.details['replies']:
```
celestemartinez, line 36: So here you are checking all replies to the submission, not just comments within a thread?
So after a while, no more emails will be sent to PC, SACs and ACs at all, even if a new thread is started?

openreview/arr/process/comment_process.py
```diff
@@ -0,0 +1,182 @@
...
+        comment_by_author = any('Authors' in r for r in reply['signatures'])
+        comment_by_reviewer = any('Reviewer' in r for r in reply['signatures'])
+        comment_by_signature = comment.signatures[0] == reply['signatures'][0]
+
+        if not(readable_by_authors and readable_by_reviewer):
+            continue
+        ## Comments are readable by both authors and at least 1 reviewer
+        
+        # if new comment by reviewer, and (reply from author is readable by reviewer) or reply written by reviewer
+        if signed_by_reviewer and ((comment_by_author and readable_by_signature) or comment_by_signature):
```
celestemartinez, line 55: why does `signed_by_signature` and `comment_by_signature` count as a comment? What if it's not visible to the authors?

openreview/arr/process/comment_process.py
```diff
@@ -0,0 +1,182 @@
...
+            comment_count += 1
+        # if new comment by author and reply written by author, or reply written by reviewer
+        elif signed_by_author and (comment_by_signature or comment_by_reviewer):
```
celestemartinez, line 58: I am confused, why don't we check who the readers are here?

celestemartinez:
I don't think the rules of when a reply counts given the latest comment are very clear, could you clarify that?

openreview/arr/process/comment_process.py
haroldrubio: There are some checks in the for loop that skip the reply if its not a comment, and checks that its readable by the Authors and some reviewer

haroldrubio:
With the new code that accounts for threads:

1) Filter the comments that are in the same thread as the new comment
2) Only look at comments that are readable by the authors and one of the reviewers
2a) If the authors made the new comment, replies written by the authors or a reviewer are counted
2b) If a reviewer made the new comment, replies written by that reviewer or (by the authors and readable by that reviewer) are counted

openreview/arr/process/comment_process.py
haroldrubio: Oh actually in the new changes to count based on thread, I did make the filtering more explicit with these list comprehensions:
https://github.com/openreview/openreview-py/blob/ee86eb371219f035019f6409a60c7bd5b233f2c6/openreview/arr/process/comment_process.py#L48-L54

celestemartinez (approved):

PR #2378, opened 2024-10-15
ARR: Separate matching scripts

### LINKED ISSUE #2350: ARR: Separate matching data setup scripts
The reassignment data script sets up all the data for all the roles at once, but we do the matching in stages so each role should have their own separate invitations

haroldrubio:
Resolves #2350 

This PR just takes the logic in the single `setup_reassignment_data.py` script and splits it into separate invitations and scripts per role.

These new invitations won't be run automatically, and only runs when an edit is posted to the invitation since the EiCs don't need to see this part of the workflow. The edits will be posted manually after the affinity scores for the roles are computed.

Added a new field `ignore_dates` to the ARRStage to skip checking these fields when looking to see if the ARRConfiguration note has a new date. It prevents the new invitations from running more than necessary.


celestemartinez:
Looks good! If possible, it would be nice to remove any unused variables as well as commented out code. This will make the scripts easier to understand/debug.

tests/test_arr_venue_v2.py
```diff
@@ -3055,46 +3055,50 @@ def test_resubmission_and_track_matching_data(self, client, openreview_client, h
...
+        for matching_invitation in matching_invitations:
+            openreview_client.post_invitation_edit(
+                invitations='aclweb.org/ACL/ARR/2023/August/-/Edit',
+                readers=['aclweb.org/ACL/ARR/2023/August'],
+                writers=['aclweb.org/ACL/ARR/2023/August'],
+                signatures=['aclweb.org/ACL/ARR/2023/August'],
+                invitation=openreview.api.Invitation(
+                    id = f"aclweb.org/ACL/ARR/2023/August/-/{matching_invitation}",
+                    content = {
+                        'count': {'value': 1}
```
celestemartinez, line 3068: What is this count for?

celestemartinez (approved):

PR #2385, opened 2024-10-17
Journal: only count assignments for current year

celestemartinez:
There is a shortage of available AEs. ~~The TMLR EICs recommended recalculation the assignment count of each AE and only consider assignments made in 2024.~~

~~I am checking when the assignment edge was created and count the assignment only if the assigment was made in the current year or if the assignment is still active.~~

Edit: I am archiving the ~300 assignments made in 2023, so the change described above is not necessary. 
**New change:** when looking at active assignments to compute `Local_Custom_Max_Papers`, only count assignments made in the last two months.


celestemartinez:
Not finished yet, looks like TMLR wants another change as well.

carlosmondra (approved):
Nice

celestemartinez:
@carlosmondra @enrubio last change! I added the inclusion date limit as an optional param in case other journal use this function as well.

carlosmondra:
> @carlosmondra @enrubio last change! I added the inclusion date limit as an optional param in case other journal use this function as well.

I saw it! Looks good. Thanks!

PR #2386, opened 2024-10-19
ARR: Allow Authors to Change Reviewer Nominations

haroldrubio:
This PR adds a submission revision stage to allow authors to only edit the fields related to reviewer nomination. 

Some extra changes were made to the Venue InvitationBuilder to support the custom pre-process function

Screenshot 2024-10-19 at 10 47 57 AM

Screenshot 2024-10-19 at 10 48 37 AM


openreview/arr/process/change_reviewer_nomination_process.py
```diff
@@ -0,0 +1,10 @@
+def process(client, edit, invitation):
+    # Get paper number
+    note = edit.note
+    forum = client.get_note(note.id)
+    volunteers = note.content.get('reviewing_volunteers', {}).get('value', [])
+    authorids = forum.content.get('authorids').get('value')
+
+    for v in volunteers:
+        if v not in authorids:
+            raise openreview.OpenReviewException(f'Volunteer {v} is not an author of this submission')
```
melisabok, line 10: what should we do if the exception is raised?

haroldrubio: The naming of the file is incorrect, it is named process but it's actually a preprocess - the error would just appear before the form gets submitted

PR #2387, opened 2024-10-21
Journal: use preferred emails invitation to get preferred email

celestemartinez:
The PCs can't see the preferred email when we do `profile.get_preferred_email()`. We should use the `Preferred_Emails` invitation instead so PCs can run this code themselves


carlosmondra (approved):

enrubio (approved):
Looks good

PR #2399, opened 2024-10-25
fix: change ithenticate group_id value and add index options

akshatchhabra:
Changed the group_id value and added index options while creating a submission.

### DIFF
--- a/openreview/venue/venue.py
+++ b/openreview/venue/venue.py
@@ -1137,7 +1137,7 @@ Total Errors: {len(errors)}
                     owner_first_name=first_name,
                     owner_last_name=last_name,
                     owner_email=owner_profile.get_preferred_email(),
-                    group_id=self.get_submission_id(),
+                    group_id=self.venue_id,
                     group_context={
                         "id": self.id,
                         "name": self.name,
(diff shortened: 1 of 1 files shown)


openreview/venue/venue.py
```diff
@@ -1137,7 +1137,7 @@ def ithenticate_create_and_upload_submission(self):
                     owner_first_name=first_name,
                     owner_last_name=last_name,
                     owner_email=owner_profile.get_preferred_email(),
-                    group_id=self.get_submission_id(),
+                    group_id=self.venue_id,
```
carlosmondra, line 1140: Just out of curiosity, do `self.get_submission_id()` and `submission.id` return the same value?
Also, lines 1144 to 1156 is commented code. Can we remove it?

akshatchhabra: `self.get_submission_id()` returns the submission invitation ID and `submission.id` returns the note ID of the submission. 

I've removed the commented code.

carlosmondra: I see. So confusing. OK, thanks!

carlosmondra (approved):

PR #2400, opened 2024-10-25
ARR: Fix stage rerunning

### LINKED ISSUE #2318: ARR: Blind Submission License Agreement gets overwritten
The current workflow for this is so that the agreement does not show up in the tasks tab of the Authors console when the Authors agree on submission. This is by deleting the duedate on the SubmissionX invitation, but when the ARR Configuration is run, this stage is getting re-run when it shouldn't because the relevant dates haven't changed - which is overwriting the duedate changes.

### LINKED ISSUE #2349: ARR: Submission Revision stage reruns with same date
We use a Submission Revision stage to create a task for authors to fill out the blind submission agreement if they did not agree to it while submitting.

We manually modify the child invitations to remove the duedate to prevent it from appearing in the tasks tab if they did agree to it. But, the ARR configuration will re-run the stage even if no changes are made and overwrites the manual changes

haroldrubio:
Resolves #2349 and resolves #2318

This PR fixes an issue where a stage keeps running because the `expdate` is mismatched because the expdate is set in the `InvitationBuilder` instead of in the note. If no expdate is provided for a stage on the venue request form, calculate it the same way its calculated in the `InvitationBuilder`

### DIFF
--- a/openreview/arr/helpers.py
+++ b/openreview/arr/helpers.py
@@ -2,7 +2,7 @@ import openreview
 from enum import Enum
 from datetime import datetime, timedelta
 from openreview.venue import matching
-
+from openreview.venue.invitation import SHORT_BUFFER_MIN
 
 from openreview.stages.arr_content import (
     arr_submission_content,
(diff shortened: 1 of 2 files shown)


celestemartinez:
I am confused how this fixes the issue, if this stage doesn't have a duedate to start with (so there is no expdate either?)

haroldrubio:
The stage has fields for the start date and due date and the InvitationBuilder populates the exp date as the invitation is posted. But the ARR code doesn't know that the exp date was set, only the start and due dates. So when it checks to see if the dates in the ARR Configuration note are different from the posted invitation:
https://github.com/openreview/openreview-py/blob/6afa3b9c6152aaab5993e276e619d71e445c58a2/openreview/arr/helpers.py#L1406-L1408
It expects that the expdate is None but finds that it's set to a value (duedate + 30mins), so it re-runs the stage even if nothing in the ARR Configuration was changed

celestemartinez:
I see, thanks for the clarification!

celestemartinez (approved):

PR #2401, opened 2024-10-25
ARR: Prevent meta-review ethics flagging

### LINKED ISSUE #2356: ARR: Prevent meta-review ethics flagging
Flagging for ethics review at the meta review stage is too late so we should remove the ability to do so, and remove the tests for this

haroldrubio:
Resolves #2356 

This PR disables ethics review flagging from meta-reviews. The tests were changed to:
1. Move the ethics review test from the meta-review test to the official review test
2. Asserts that no additional were sent during the meta-review test for ethics reviewing

### DIFF
--- a/openreview/arr/helpers.py
+++ b/openreview/arr/helpers.py
@@ -1547,16 +1547,8 @@ def flag_submission(
         lambda reply: any('Meta_Review' in inv for inv in reply['invitations']),
         forum.details['replies']
     ))
-    ethics_flag_from_metareviews = False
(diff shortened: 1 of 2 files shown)


celestemartinez:
Should the metareview be edited to not contain fields pertaining to ethics flagging, or will it stay the same?

haroldrubio:
We can keep the metareview ethics fields there - even if the ethics chairs of that cycle won't see it, the next cycle's AC/reviewers would still want that information

celestemartinez (approved):

PR #2404, opened 2024-10-28
Update matching with default max papers

emily-grabowski:


celestemartinez (approved):
Looks good! I was a bit confused at first because we usually do:

> june_venue.get_custom_max_papers_id(june_venue.get_reviewers_id())

to get the custom max papers id, by passing the whole reviewers id and not just the reviewers name. This way you already get the whole invitation id and you don't have to append the venue.id at the beginning. This is ok though.

openreview/venue/matching.py
```diff
@@ -1365,6 +1367,26 @@ def deploy(self, assignment_title, overwrite=False, enable_reviewer_reassignment
...
+            max_load_name = self.venue.get_custom_max_papers_id(self.match_group_name)
+            #update the default max papers in the custom max papers invitation
+            max_paper_invitation = self.client.get_invitation(id=f"{self.venue.id}/{max_load_name}")
+            max_paper_invitation.edit['weight']['param']['default']= default_max_papers
+            self.client.post_invitation_edit(
+                invitations=self.venue.get_meta_invitation_id(),
+                readers=[self.venue.id],
+                writers=[self.venue.id],
+                signatures=[self.venue.id],
+                invitation=max_paper_invitation
```
melisabok, line 1383: no need to post the whole invitation, you can post the default value only. 

```
invitation=openreview.api.Invitation(
 id = max_paper_invitation.id,
edit={
   "weight": {
       "param": {
             "default": default_max_papers
        }
   }
}

)
```

PR #2407, opened 2024-10-30
ARR: Fix assignment quotas

haroldrubio:
This PR resolves an issue where, if the 3rd edge (out of the Assignment + Invite_Assignment edges) posted to a paper is an Invite_Assignment edge, the process function fails because the pre-process saw it as posting a 4th edge.

The pre-process functions are changed to exclude the posted edge from the edge count if there is already an edge with that ID.

To allow the Assignment_Recruitment process function to complete, only non-accepted and non-declined Invite_Assignment edges are counted. The logic is still preserved because if an Invite_Assignment edge is accepted, then there must be an Assignment edge which are counted

### DIFF
--- a/openreview/venue/process/assignment_pre_process.js
+++ b/openreview/venue/process/assignment_pre_process.js
@@ -32,7 +32,7 @@ async function process(client, edge, invitation) {
     const filteredInviteAssignmentEdges = inviteAssignmentEdges.filter(e => !filteredLabels.includes(e?.label ?? ''))
 
     if (quota && filteredInviteAssignmentEdges.length + filteredAssignmentEdges.length > quota) {
-      return Promise.reject(new OpenReviewError({ name: 'Error', message: `Can not make assignment, total assignments and invitations must not exceed ${quota}` }))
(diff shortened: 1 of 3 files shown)


openreview/venue/process/assignment_pre_process.js
```diff
@@ -16,12 +16,20 @@ async function process(client, edge, invitation) {
...
+    const acceptLabel = invitations[0]?.content?.accepted_label?.value ?? '';
+
     const [{ edges: inviteAssignmentEdges }, { edges: assignmentEdges }] = await Promise.all([
         client.getEdges({ invitation: inviteAssignmentId, head: edge.head }),
         client.getEdges({ invitation: edge.invitation, head: edge.head })
     ])
 
-    if (quota && inviteAssignmentEdges.length + assignmentEdges.length >= quota) {
+    // Filter assignment edges to exclude the current edge.id
+    const filteredAssignmentEdges = assignmentEdges.filter(e => e.id !== edge.id)
```
celestemartinez, line 28: I thought that if this is a preprocess, then the edge is not posted yet?

celestemartinez: I see, this is the case were we update the edge in the `Invite_Assignment` process?

haroldrubio: Yes that's correct it's to handle updating an edge since updating an edge shouldn't count as a new edge

haroldrubio:
There's a request to be able to increase the quota for some papers where the reviews may have been AI generated and need extra reviewers - I'm thinking of a way to properly implement this

PR #2411, opened 2024-11-04
Fix/ Tests: check status before break in await_queue_edit()

celestemartinez:


openreview/arr/helpers.py
```diff
@@ -1446,6 +1448,10 @@ def set_stage(self, client_v1, client, venue, invitation_builder, request_form_n
             self._post_new_dates(client, venue, current_invitation)
         else:
             self._set_field_readers(venue)
+            expected_statuses = ['error', 'ok']
+            current_log_count = len(
+                [log for log in client.get_process_logs(invitation=self.super_invitation_id) if log['status'] in expected_statuses]
+            )
```
carlosmondra, line 1454: Would it be possible to have more than 1000 of these logs before you do the filtering by status?

haroldrubio: Yeah it shouldn't exceed 1000 - the issue actually only happens when the invitation is first made, any subsequent changes are only single edits to changed the dates of the invitation, or is handled by the `Venue` class code when a stage is run (equivalent to a PC posting a note on the request form)

carlosmondra (approved):
Looks good! Thanks!

PR #2413, opened 2024-11-05
ARR: Add PCs as impersonators on setup

### LINKED ISSUE #2283: ARR: Add some PCs as impersonators for every cycle
Let's add the call to add the impersonators in the setup function of the ARR class. 

I would suggest to do it for the PCs that manage all the cycles.

haroldrubio:
Resolves #2283


haroldrubio:
PCs that manage all cycles should be in the previous cycle and we take the PCs from the previous cycle unless we're given changes to the list.

We could add a new invitation that propagates adding a new email to all the previous cycles to make it easier for the tech team but I think that's a separate issue

PR #2414, opened 2024-11-05
ARR: Check for duplicate signatures before copying

### LINKED ISSUE #2368: ARR: Avoid posting notes twice when copying data
The `Share_Data` script does not check if a user has already posted a note before attempting to post one on their behalf -compare the signature of the note that's about to be posted to a set of all names for all users who have already posted a note

### LINKED ISSUE #2376: ARR: Withdrawn submissions failing previous URL validation
Papers that are withdrawn may still have reviews and reviewers that can be assigned in a later cycle. These paper should still pass the pre-process validation

haroldrubio:
Resolves #2368 and resolves #2376

Before the changes to the script, the additional registration form in the June cycle would break the tests since the registration tests have a maxReplies of 1


openreview/arr/management/setup_shared_data.py
```diff
@@ -91,6 +91,15 @@ def _is_not_available(month: str, year: int, current_date: datetime.datetime, po
             f"{previous_cycle_id}/{domain.content['ethics_reviewers_name']['value']}",
         ]
     ]
+    all_profiles = openreview.tools.get_profiles(client, client.get_group(domain.content['reviewers_id']['value']).members)
```
melisabok, line 94: a better option is to merge all the members first and call to tools.get_profiles only once.

tests/test_arr_venue_v2.py
```diff
@@ -833,6 +833,34 @@ def test_june_cycle(self, client, openreview_client, helpers, test_client):
                 }
             )
         )
+
+        # Post duplicate (messy data)
+        openreview_client.post_invitation_edit(
```
melisabok, line 838: what are you testing here?

haroldrubio: This is how I replicated the error I was getting when running this script several times

melisabok: what is the error that you were getting? an API error?

haroldrubio: Yes when using live data I was getting an error that it was violating max replies

melisabok: I don't see here you are replicating the error, you increasing the max replies and posting a new registration. 

we need an assertion to make sure we get what we expect. is there a chance to run the script to reproduce the error?

haroldrubio: Without the changes in the PR, the script will run and it will try to copy data from the June cycle into the August cycle. The API will try to post 2 replies from `~Reviewer_ARRTwo1` into the August cycle which correctly has `maxreplies=1`

melisabok (approved):

tests/test_arr_venue_v2.py
```diff
@@ -849,6 +851,32 @@ def test_june_cycle(self, client, openreview_client, helpers, test_client):
                 }
             )
         )
+
+        # Post duplicate and merge profiles
+        reviewer_two_merge_client.post_note_edit(
```
melisabok, line 856: the test is failing because the reviewer can not see the invitation, do you need to add it to the reviwers group?

haroldrubio:
I've been using the CircleCI tests to run my tests because there's some strange bug in my local that's preventing even master from passing

melisabok:
Fix master in your machine first before doing anything.

PR #2418, opened 2024-11-06
Feat/add to index option in request form

akshatchhabra:
These changes enable the venue to set the value for the add_to_index option in iThenticate. 

Also removed the default value for the EULA version.


carlosmondra:
It might be a good idea to add description to the other iThenticate fields.

akshatchhabra:
> The problem is that we don't know what they are going to call their groups. For example, AAAI uses Program Committee instead of Reviewers

Ah okay

carlosmondra (approved):

celestemartinez (approved):

PR #2419, opened 2024-11-07
Update request_remove_name_process.py

emily-grabowski:
Fix the reversed names that weren't being approved.


emily-grabowski:
The original PR wasn't accounting for the fact that when multiple accounts use the same name, numbers are used to differentiate between preferred names, so most names were not automatically passing. The tests were passing because there were no publications associated with the name, which also triggers automatic acceptance.

This PR updates the tests and the process function to account for that fact

celestemartinez (approved):

PR #2421, opened 2024-11-11
Feature/ Journal: allow volunteers to comment before they are approved

celestemartinez:


openreview/journal/journal.py
```diff
@@ -317,6 +317,12 @@ def get_solicit_review_approval_id(self, number=None, signature=None):
 
         return self.__get_invitation_id(name='Volunteer_to_Review_Approval', number=number)
 
+    def get_solicit_review_comment_id(self, number=None, reply_number=None):
+        if reply_number:
+            prefix = f'{self.venue_id}/{self.submission_group_name}{number}/Volunteer_to_Review{reply_number}'
```
melisabok, line 322: is the number shown in the button text?

celestemartinez: This is what the buttons look like:
Screenshot 2025-01-10 at 11 55 57 AM

The invitation ID is: `TMLR/Paper4/Volunteer_to_Review3/-/Volunteer_to_Review_Comment`. I do not like repeating `Volunteer_to_Review` but not sure how to best name the invitation. I didn't want to use id of the reviewer in the invitation id.

celestemartinez: Another thing I noticed is that the AEs can still see the official comment invitation, and I am worried they might try to use this to reply to the volunteer instead of the `Volunteer to Review Comment `button. Maybe there is a better way to allow the AE and the volunteer to communicate.

celestemartinez: @melisabok this is almost done. I think I want to add a preprocess that checks that AEs not use the Official_Comment invitatino to reply to volunteers.

celestemartinez: ok, let me remove the preprocess then, since it is not needed.

melisabok (approved):

PR #2426, opened 2024-11-16
ARR Dec - Adjusting label for response PDF

jkkummerfeld:
This PR changes one field name and description to better reflect the intent of the question.

Note - it would be good to also have a pre-process check that this is uploaded for any paper that is a resubmission (ie, if `previous_URL` contains something). I am not sure where that code lives though.


jkkummerfeld:
@haroldrubio Following up on this as we are approaching December 1st, when the new site will be set up.

jkkummerfeld:
Note that the latest updates include changes to the names of some key fields:

`explanation_of_revisions_PDF` -> `response_PDF`

`reassignment_request_action_editor` -> `reassignment_request_area_chair`

melisabok (approved):

PR #2429, opened 2024-11-20
Fix: Preserve transitive membership

carlosmondra:
This PR should fix the issue when two profiles are merged and then an email is removed, there is a chance that the transitive members will break.

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -769,7 +769,7 @@ class OpenReviewClient(object):
         return response.json()
 
 
-    def get_groups(self, id=None, prefix=None, member=None, signatory=None, web=None, limit=None, offset=None, after=None, stream=None, sort=None, with_count=False):
+    def get_groups(self, id=None, prefix=None, member=None, members=None, signatory=None, web=None, limit=None, offset=None, after=None, stream=None, sort=None, with_count=False):
         """
         Gets list of Group objects based on the filters provided. The Groups that will be returned match all the criteria passed in the parameters.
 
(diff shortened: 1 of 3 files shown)


celestemartinez:
Does this change the behavior of using `member` when doing get_group()? I see we use `member` in `get_groups()` and `get_all_groups()` throughout the venue and journal, should we change these to `members`?

carlosmondra:
> Does this change the behavior of using `member` when doing get_group()? I see we use `member` in `get_groups()` and `get_all_groups()` throughout the venue and journal, should we change these to `members`?

Nope. You can leave those. The difference between `GET /groups?members=<group_id_1>` and `GET /groups?member=<group_id_2>` is that the first one looks for groups where `group_id_1` appears as one of their members (inside the members array) and the second one lists the transitive members of `group_id_2`. Does that make sense?

openreview/api/client.py
```diff
@@ -769,16 +769,18 @@ def update_relation_readers(self, update):
...
+    def get_groups(self, id=None, prefix=None, member=None, members=None, signatory=None, web=None, limit=None, offset=None, after=None, stream=None, sort=None, with_count=False):
         """
         Gets list of Group objects based on the filters provided. The Groups that will be returned match all the criteria passed in the parameters.
 
         :param id: id of the Group
         :type id: str, optional
         :param prefix: Prefix that matches several Group ids
         :type prefix: str, optional
-        :param member: Groups that contain this member
+        :param member: Groups that that are trasitive members of the member value
```
celestemartinez, line 780:
```suggestion
        :param member: Groups that that are transitive members of the member value
```

carlosmondra:
> > > Does this change the behavior of using `member` when doing get_group()? I see we use `member` in `get_groups()` and `get_all_groups()` throughout the venue and journal, should we change these to `members`?
> > 
> > 
> > Nope. You can leave those as they are. The difference between `GET /groups?members=<group_id_1>` and `GET /groups?member=<group_id_2>` is that the first one looks for groups where `group_id_1` appears as one of their members (inside the members array) and the second one lists the transitive members of `group_id_2`. Does that make sense?
> 
> Ohh I see, so the behavior did not change. Looks good then!

Correct, I'm just using a different param to `GET /groups`

PR #2431, opened 2024-11-20
feat: update iThenticate EULA version using dateprocess functions

akshatchhabra:
1. Introduced a function to create submission invitations for iThenticate venues.
2. The submission invitation includes a date-process function that does the following:
    - Check the EULA version in the submission form.
    - Compare it against the current EULA version obtained from iThenticate.
3. If there is a version mismatch:
    - The date-process function posts an invitation edit.
    - This edit is made using the meta invitation.


openreview/venue/venue.py
```diff
@@ -496,7 +496,10 @@ def recruit_reviewers(self,
             allow_overlap_official_committee)
 
     def create_submission_stage(self):
-        self.invitation_builder.set_submission_invitation()
+        if self.iThenticate_plagiarism_check:
+             self.invitation_builder.set_iThenticate_submission_invitation()
+        else:
+            self.invitation_builder.set_submission_invitation()
```
carlosmondra, line 502: I would prefer to have something like this:
```suggestion
        self.invitation_builder.set_submission_invitation()
        if self.iThenticate_plagiarism_check:
             self.invitation_builder.set_iThenticate_fields()
```

That way if the method `invitation_builder.set_submission_invitation()` changes, we don't have to change it in two different places. It will also make the logic in `self.invitation_builder.set_iThenticate_fields()` simpler as we will only need to add the iThenticate fields.

melisabok (approved):

PR #2433, opened 2024-11-25
Update comment_process.py

racheljsmart:
This change should prevent messages from being sent to reviewers that are no longer in the reviewers group.


celestemartinez:
This is a temporary fix to the issue of unassigned reviewers getting emails. Carlos is working on a more permanent fix.

celestemartinez:
@carlosmondra we do not need this anymore, right? This was to fix the issue of unassigned users getting emails since they appear as readers of comments.

carlosmondra:
> @carlosmondra we do not need this anymore, right? This was to fix the issue of unassigned users getting emails since they appear as readers of comments.

Correct. We can close it. The fix is [here](https://github.com/openreview/openreview-api/pull/619)

PR #2438, opened 2024-11-28
Feat: Change comment readers when comment is deleted

carlosmondra:


openreview/venue/process/comment_process.py
```diff
@@ -16,6 +16,20 @@ def process(client, edit, invitation):
     comment = client.get_note(edit.note.id)
     paper_group_id=f'{venue_id}/{submission_name}{submission.number}'
 
+    # Change comment readers to be only the author if the comment is deleted
+    # Domain admins can still see the comment
+    if comment.ddate:
```
melisabok, line 21: why do you need to do this? could you please add a test for this?

carlosmondra: I updated the description of the PR. I don't have a complete solution for it, but I am willing to discuss alternatives.

melisabok: Thanks, what about making the edits not accesible if the note is deleted? I guess this can happen with other type of notes not only comments.

PR #2440, opened 2024-12-02
Chore: Reduce parallelism to save on resources

carlosmondra:


celestemartinez:
Looks good, thanks Carlos!

There is some commented out code in config.yml and some prints in the ICML tests. If these are not needed anymore, could we remove them?

celestemartinez (approved):

PR #2448, opened 2024-12-11
Feat: Update create profile method

carlosmondra:
This PR updates the method used to create profiles using the super user

Needs https://github.com/openreview/openreview-py/pull/2447 to be merged first

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -401,33 +401,37 @@ def create_profile(client, email, fullname, super_user='openreview.net'):
 
     return profile
 
-def create_authorid_profiles(client, note, print=print):
+def create_authorid_profiles(client, note):
     # for all submissions get authorids, if in form of email address, try to find associated profile
     # if profile doesn't exist, create one
     created_profiles = []
 
-    if 'authorids' in note.content and 'authors' in note.content:
-        author_names = [a.replace('*', '') for a in note.content['authors']]
-        author_emails = [e for e in note.content['authorids']]
-        if len(author_names) == len(author_emails):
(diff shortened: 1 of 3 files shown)


openreview/tools.py
```diff
@@ -343,7 +343,7 @@ def get_invitation(client, id):
         print('Can not retrieve invitation', e)
     return invitation
 
-def create_profile(client, email, fullname, url='http://no_url', allow_duplicates=False):
+def create_profile(client, email, fullname, super_user='openreview.net'):
```
celestemartinez, line 346: Will super_user always be `openreview.net`? Won't it be `OpenReview.net` on live?

carlosmondra: Yes, it will be. I wasn't sure how else to pass it because that information is in the config file of the API.

celestemartinez: I see. We could do something like this: https://github.com/openreview/openreview-py/blob/03d796c40e5bb950d3e716606f6bdaf05c308b8d/openreview/tools.py#L135 but if we are not using this in any process and we only do it manually, then it shouldn't be an issue

carlosmondra: Do you mean something like this?

```py
            super_user = 'openreview.net'
            if 'https://devapi' in client.baseurl or 'https://api' in client.baseurl:
                super_user =  'OpenReview.net'
```

carlosmondra: > but if we are not using this in any process and we only do it manually, then it shouldn't be an issue
Yeah, I don't see how we would use this in a process function

celestemartinez (approved):

PR #2451, opened 2024-12-12
Feature/ Update Tag class

celestemartinez:
Requires https://github.com/openreview/openreview-api/pull/547


openreview/api/client.py
```diff
@@ -3326,17 +3334,26 @@ def to_json(self):
...
 
-        if self.signatures:
-            body['signatures'] = self.signatures
+        if self.signature:
+            body['signature'] = self.signature
+
+        if self.profile:
+            body['profile'] = self.profile
+
+        if self.weight:
```
carlosmondra, line 3349: According to Sherry, this fails when self.weight == 0. This should fix it:
```suggestion
        if self.weight is not None:
```

tests/test_icml_conference.py
```diff
@@ -5722,14 +5722,30 @@ def test_forum_chat(self, openreview_client, helpers):
         ## Add tag emoji
         tag = sac_client.post_tag(openreview.api.Tag(
             invitation='ICML.cc/2023/Conference/Submission1/-/Chat_Reaction',
-            signatures=['ICML.cc/2023/Conference/Submission1/Senior_Area_Chairs'],
-            tag='😄',
-            replyto=note_edit['note']['id']
+            signature='ICML.cc/2023/Conference/Submission1/Senior_Area_Chairs',
```
melisabok, line 5725: does it mean we need to update the current tag invitations?

carlosmondra: Yes, https://github.com/openreview/openreview-api/pull/627.

I am open to suggestions, of course

.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ parameters:
     default: "master"
   openreview-api-v2-branch:
     type: string
-    default: "main"
+    default: "feature/tags"
```
melisabok, line 11: could you switch this to main?

celestemartinez: done

openreview/api/client.py
```diff
@@ -2885,6 +2894,12 @@ def to_json(self):
         if self.id:
             body['id'] = self.id
 
+        if self.invitations:
```
melisabok, line 2897: is there a reason to add "invitations" here? I don't think the API allows posting an invitation with the property invitations.

carlosmondra: However, when you get an invitation it had the `invitations` field. So I think this field is necessary in the Invitation class

melisabok: It can be in the class but I ignore them in then to_json function that is where we convert the object to a JSON to send to the API.

melisabok (approved):

PR #2459, opened 2025-01-06
Journal: update how assignments are archived for reviewers

celestemartinez:
~~TMLR wants to have the option to archive reviewer assingments if the reviewers has submitted their review.~~

Added an option to archive assignments once a paper is done, as long as the paper was submitted the previous year. This allows to free up quotas for the new year once a paper is finished. 

This PR also removes `inGroup` from the AE `Archived_Assignment` invitation, since once we archive assignments some of the AEs might not be in the AE group anymore.


openreview/journal/journal.py
```diff
@@ -1073,9 +1073,13 @@ def update_solicit_review(self, note_number, invitation):
         invitation.invitations = None
         self.invitation_builder.post_invitation_edit(invitation, replacement=True)
 
-    def archive_assignments(self):
+    def archive_assignments(self, check_reviews_done=False):
```
melisabok, line 1076: may be we should do this always. It doesn't make sense to archive an assignment if the review is not submitted, what do you think?

celestemartinez: Right now we are only archiving an assignment if the submission is "finished", whether the review was done or not. I think this still makes sense because those assignments shouldn't count anymore even if the review was not submitted. 

This PR is just adding that besides archiving assignments of "finished" submissions, assignments of still under review submissions can also be archived if the review was already submitted. This affects only reviewers. Not sure if we are going to need to archive assignments of finished submissions only again.

melisabok (approved):

PR #2464, opened 2025-01-08
Feature: add option to delay new assignments after review has been submitted

celestemartinez:
TMLR wants there to be a 2 week delay between when a reviewer submits their review and when they are once again available to be assigned to a new paper. 

In this case, I implemented a post process that will run 2 weeks after the review was submitted to decrease the pending reviews counter. If the journal does not want a delay, the counter is decreased as soon as the review is posted.


openreview/journal/invitation.py
```diff
@@ -3812,7 +3813,21 @@ def set_review_invitation(self):
 
         if self.journal.get_review_additional_fields():
             for key, value in self.journal.get_review_additional_fields().items():
-                invitation['edit']['note']['content'][key] = value if value else { "delete": True }         
+                invitation['edit']['note']['content'][key] = value if value else { "delete": True }
+
+        if self.journal.get_assignment_delay_after_submitted_review() > 0:
+            weeks = datetime.timedelta(weeks=self.journal.get_assignment_delay_after_submitted_review())
+            milliseconds = int(weeks.total_seconds() * 1000)
+            invitation['postprocesses'] = [
```
melisabok, line 3821: when a postprocesses run?

celestemartinez: The postprocess will run after a reply is posted to the invitation, so kind of like the process function. However, we can add a delay, so the postprocess will run after the specified delay (in ms). More info here: https://github.com/openreview/openreview-api/pull/630

melisabok: can we always add this post process and if the delay is 0 then we update the pending review edge right away?

then we need to remove the code that updates the edge in the review process

.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ parameters:
     default: "master"
   openreview-api-v2-branch:
     type: string
-    default: "main"
+    default: "feature/postprocesses"
```
melisabok, line 11: we need to wait until this is merged before merging this

melisabok (approved):

PR #2493, opened 2025-01-30
Fix: ithenticate process errors

akshatchhabra:
These changes fix errors in the date process function that adds the EULA agreement for venues using iThenticate.

Previously, this function was part of the submission invitation, but it failed because it was being removed while still executing.

After discussions with @carlosmondra, I moved the date process function to the invitation for iThenticate edges.

Now, the iThenticate edge invitation is automatically created for venues that use iThenticate.


openreview/venue/invitation.py
```diff
@@ -4471,6 +4459,7 @@ def set_iThenticate_plagiarism_check_invitation(self):
         
         paper_number = '${{2/head}/number}'
         edge_readers = [venue_id]
+        process_function = inspect.getsource(iThenticate_eula_process_function)
```
carlosmondra, line 4462: I am not against using this method. I'm curious why you didn't use `get_process_content` like we do for other invitations?

openreview/venue_request/process/ithenticate_eula_process.py
```diff
@@ -2,33 +2,36 @@
...
             iThenticate_agreement_object["value"]["param"]["enum"] = [f"Yes, I agree to iThenticate's EULA agreement version: {eula_version}"]
-            client.post_invitation_edit(invitation, invitation=venue.get_meta_invitation_id())
-
+            client.post_invitation_edit(
+                invitations=domain_group.content["meta_invitation_id"]["value"],
+                readers=[domain_group.domain],
+                writers=[domain_group.domain],
+                signatures=[domain_group.domain],
+                replacement=False,
+                invitation=submission_invitation,
```
carlosmondra, line 30: This is fine, but it's better if you only pass the fields that you are modifying.
So something like:
```py
invitation=Invitation(
    id=submission_invitation_id
    content={
        "iThenticate_agreement": ...
    }
)
```

openreview/venue_request/process/ithenticate_eula_process.py
```diff
@@ -43,7 +46,13 @@ def process(client, invitation):
...
-
+        
+        client.post_invitation_edit(
+            invitations=domain_group.content["meta_invitation_id"]["value"],
+            readers=[domain_group.domain],
+            writers=[domain_group.domain],
+            signatures=[domain_group.domain],
+            replacement=False,
+            invitation=submission_invitation,
+        )
```
carlosmondra, line 57: Perfectly fine, but I would use the same method as above :)

openreview/venue/invitation.py
akshatchhabra: I wanted to extract just the function and exclude the import statements. However, get_process_content reads the entire file, making it difficult to isolate only the function. I couldn't remove the import statements because doing so caused the test to fail.

carlosmondra: Ah got it

openreview/venue_request/process/ithenticate_eula_process.py
akshatchhabra: Thanks for the suggestion, I've added the change

openreview/venue_request/process/ithenticate_eula_process.py
akshatchhabra: Thanks for the suggestion, I've added the change

openreview/venue_request/process/ithenticate_eula_process.py
```diff
@@ -2,33 +2,39 @@
...
         if current_iThenticate_eula_version != eula_version:
-            iThenticate_agreement_object["description"] = f"The venue is using iThenticate for plagiarism detection. By submitting your paper, you agree to share your PDF with iThenticate and accept iThenticate's End User License Agreement. Read the full terms here: {eula_link}"
-            iThenticate_agreement_object["value"]["param"]["enum"] = [f"Yes, I agree to iThenticate's EULA agreement version: {eula_version}"]
-            client.post_invitation_edit(invitation, invitation=venue.get_meta_invitation_id())
-
+            modified_iThenticate_agreement_object = modified_invitation_content["iThenticate_agreement"]
+            modified_iThenticate_agreement_object["description"] = f"The venue is using iThenticate for plagiarism detection. By submitting your paper, you agree to share your PDF with iThenticate and accept iThenticate's End User License Agreement. Read the full terms here: {eula_link}"
+            modified_iThenticate_agreement_object["value"]["param"]["enum"] = [f"Yes, I agree to iThenticate's EULA agreement version: {eula_version}"]
```
carlosmondra, line 26: Where is the `modified_iThenticate_agreement_object` variable being used?

openreview/venue_request/process/ithenticate_eula_process.py
```diff
@@ -2,33 +2,39 @@
...
+            client.post_invitation_edit(
+                invitations=domain_group.content["meta_invitation_id"]["value"],
+                readers=[domain_group.domain],
+                writers=[domain_group.domain],
+                signatures=[domain_group.domain],
+                replacement=False,
+                invitation=modified_invitation,
```
carlosmondra, line 33: Does the `modified_invitation` include all the fields of the note content?

akshatchhabra: I've pushed changes that only send the modified fields of the note content

openreview/venue_request/process/ithenticate_eula_process.py
akshatchhabra: modified_iThenticate_agreement_object has been removed in the latest commit

carlosmondra (approved):

PR #2496, opened 2025-01-30
Update matching.py

### LINKED ISSUE #2472: Setting default Custom Max Papers weight
Emily made [this](https://github.com/openreview/openreview-py/pull/2404) change about setting the default max papers in the invitation from the deployed configuration. However, I now notice that we don't set the note as deployed until after `set_assignments()` finishes successfully, so by the time we try to get the deployed configuration, the status has not been changed yet. This means that the default value is not being set. 

I think instead of checking which configuration is deployed, we get the configuration with the title we pass to set_assignments(): `assignment_title`

emily-grabowski:
update to check for a matching assignment title instead of 'Deployed' state.

fixes #2472

### DIFF
--- a/openreview/venue/matching.py
+++ b/openreview/venue/matching.py
@@ -1368,9 +1368,9 @@ class Matching(object):
             self.setup_invite_assignment(hash_seed=hash_seed, invited_committee_name=f'''Emergency_{self.match_group_name}''')
 
         #get the default max papers from the assignment configuration if possible
-        deployed_matching_configurations = [x for x in self.client.get_all_notes(invitation=self.match_group.id+'/-/Assignment_Configuration') if x.content['status']['value']=='Deployed']
-        if deployed_matching_configurations:
-            default_max_papers = int(deployed_matching_configurations[0].content['max_papers']['value'])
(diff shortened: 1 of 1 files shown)


celestemartinez (approved):

celestemartinez (approved):

melisabok:
is there a test to verify this?

I see the previous fix was being tested but why we didn't catch this error?

celestemartinez:
> is there a test to verify this?
> 
> I see the previous fix was being tested but why we didn't catch this error?

Not sure how to test this, since the matcher is the one that calls `venue.set_assignments()`. This was working in the test because it looks like the ARR v2 test was setting the status of the config as 'Deployed'. However, this is done by the matcher after `set_assignments()` is called.

```
# impersonate user to get all the permissions to deploy the groups
venue.client.impersonate(self.config_note_interface.venue_id)
venue.set_assignments(
    assignment_title=self.config_note_interface.label,
    committee_id=self.config_note_interface.match_group,
    overwrite=True,
    enable_reviewer_reassignment=True,
)

self.config_note_interface.set_status(MatcherStatus.DEPLOYED)```

PR #2509, opened 2025-02-14
Update: Filters for messaging reviewers on ARR PC console

holylovenia:
This PR adds new/changes existing filter options so the final options will be as below:
1. Reviewers with unsubmitted official reviews
2. Reviewers with submitted official reviews (>0 submitted review)
3. Reviewers with 0 assignments
4. Reviewers with assignments
5. Reviewers with assignments who have at least one incomplete checklist
6. Reviewers with assignments who have submitted 0 reviews
7. Available reviewers with less than max cap assignments
8. Available reviewers with less than max cap assignments and signed up for emergencies
9. Unavailable reviewers (are not in the cycle and without assignments)


haroldrubio:
The tests are passing but the reviewer checklist reminder option needs some changing because it has the following bug:
If R1 submits a reviewer checklist but R2 and R3 haven't, then R2 and R3 won't get reminded (assuming these are their only assignments)

We need this PR: https://github.com/openreview/openreview-web/pull/2026
to add in the additional information so that we can check the signatures of each reviewer checklist and not make a decision using the presence of a single checklist

melisabok:
@haroldrubio can this one be merged today?

haroldrubio:
Yes! that would be great

haroldrubio (approved):

Issue #2523, opened 2025-03-04
Comment Stage: allow PCs to be notified when there is a direct message to them


melisabok:
Add a new checkbox that decides to send an email to the PCs if the comment has PCs as the only readers, besides the SAC and author of the comment.

This checkbox overrides the settings emails to pcs that we currently have.

Add description to the comment invitation using this feature: https://github.com/openreview/openreview-web/pull/2226

PR #2524, opened 2025-03-04
Comment Stage: add a new setting to notify PCs for direct comments only

melisabok:
Fixes: https://github.com/openreview/openreview-py/issues/2523

It depends on: https://github.com/openreview/openreview-web/pull/2226

- PCs can now decide if they want to get an email for every comment, for direct comments only or no emails:

Screenshot 2025-03-04 at 4 12 20 PM

- Email notification rules are displayed in the note editor so users know when the PCs will be notified:

Screenshot 2025-03-04 at 4 14 08 PM

@celestemartinez @racheljsmart @emily-grabowski I need some help with the text for the PC email options and instructions.


racheljsmart:
@melisabok About the "email program chairs about official comments" options, the options use the word "venue" but is it just replies to the submission forum? It wouldn't include the venue request form Official comment notifications?

A) Yes, email PCs for each official comment made in the venue - maybe "Yes, emails PCs for every Official Comment posted to a submission forum."
B) Yes, email PCs for each direct official comment made in the venue - maybe "Yes, emails PCs for each direct comment in a submission forum but not for nested replies to that first comment."
C) No, do not email PCs for each official comment made in a venue - "No, do not email PCs for each official comment made in a submission forum"

Is this the input you're asking for?

melisabok:
Hi @racheljsmart ,

The word  "venue" was already in the two previous options, I just added a third one and I wanted to get feedback about the wording "'Yes, email PCs for each direct official comment made in the venue',".

This options is similar to the one that we have for SACs, we always send an email if the comment is visible to the PCs or PCs/SACs only, that is the concept of "direct comment". 

I don't want to remove the "venue" word for all the options because I would need to keep backward compatibility with all the current official comment stages.

racheljsmart:
> This options is similar to the one that we have for SACs, we always send an email if the comment is visible to the PCs or PCs/SACs only, that is the concept of "direct comment".
> 


Sorry, I confused direct direct comment with "direct replies" that we use when getting the replies of a submission. So a direct comment is if only the PC/SACs are the readers?

melisabok:
> So a direct comment is if only the PC/SACs are the readers?

Yes, do you think we should use another word? I was trying to replicate the "Direct Message" tool that you have on Twitter 🫤

racheljsmart:
> > So a direct comment is if only the PC/SACs are the readers?
> 
> Yes, do you think we should use another word? I was trying to replicate the "Direct Message" tool that you have on Twitter 🫤

I'm not very familiar with Twitter. Maybe we should spell it out for them. "Yes, email PCs if only the PCs and SACs are readers of the comment." Although this could be explained in the description.

melisabok:
I added a text area to configure the instructions for submission, review, meta review and comment stages.

openreview/venue_request/venue_request.py
```diff
@@ -508,6 +522,7 @@ def setup_comment_stage(self):
                 'description': 'Should the PCs receive an email for each official comment made in the venue? Default is "No, do not email PCs for each official comment in the venue"',
                 'value-radio': [
                     'Yes, email PCs for each official comment made in the venue',
+                    'Yes, email PCs for each direct official comment made in the venue',
```
celestemartinez, line 525:
```suggestion
                    'Yes, email PCs only for private official comments made in the venue (comments visible only to Program Chairs and Senior Area Chairs, if applicable)',
```

celestemartinez: I know this is kind of long and I am open to other ideas, but I don't think just saying "private" or "direct" will be clear enough for PCs

melisabok: I like this option

openreview/stages/venue_stages.py
```diff
@@ -1155,6 +1163,39 @@ def get_mandatory_readers(self, conference, number):
...
             readers.append(conference.get_senior_area_chairs_id(number))
         return readers
+    
+    def get_description(self, conference):
+
+        if self.description:
+            return self.description
+        
+        instructions = '''
+- Please provide your official comments in the space provided below.
```
celestemartinez, line 1173: This is what I think the description should say (1 and 4 are always present, only one from 2 and 3):

1.Please select who should be able to see your comment under “Readers”. Readers marked as mandatory are required in order to post the comment. 

---------------------
2a. Program Chairs will be notified of all comments.
2b. Program Chairs will be notified of any comments that are visible only to the mandatory readers.
2c. Program Chairs will not be notified of any of your comments.

---------------------
3a. Senior Area Chairs will be notified of all comments.
3b. Senior Area Chairs will be notified of any comments that are visible only to the mandatory readers.

---------------------
4.All other readers will be notified of all comments.

openreview/stages/venue_stages.py
```diff
@@ -1155,6 +1163,39 @@ def get_mandatory_readers(self, conference, number):
...
+- Program Chairs will be notified of your direct comments, only mandatory readers.'''
+        else:
+            instructions += '''
+- Program Chairs will not be notified of your comments.'''
+
+        if conference.use_senior_area_chairs:
+            if self.email_sacs:
+                instructions += '''
+- Senior Area Chairs will be notified of your comments.'''
+            else:
```
celestemartinez, line 1191: It looks like if PCs select no to sending emails to SACs in the request form, this really means that they will be notified if the comment is visible only to mandatory readers. Maybe we should make this more clear in the venue request?

celestemartinez:
It looks like the tests are failing because the descriptions of the invitations are all null.

openreview/stages/venue_stages.py
melisabok: that's true, the current option is "No, do not email SACs for each official comment made in the venue". The new options can say "No, do not email SACs for each official comment. Only send email notifications for comments that are specifically visible to PCs and SACs."

What do you think?

celestemartinez: I would say something like: 

> No, do not email SACs for each official comment; instead send email notifications for comments that are only visible to PCs and SACs.

We could also word it similar to how we word it for the PCs:

'Yes, email SACs for each official comment.',
'Yes, email SACs only for private official comments made in the venue (comments visible only to Program Chairs and Senior Area Chairs)',

enrubio: Minor suggestion, but I think it's strange to ask PCs `"Should the SACs(if applicable) receive an email for each official comment made in the venue?"` and both options say Yes, even though it functions the same as before.

What do you think about removing "Yes" and just writing:
- Email SACs for each official comment made in the venue
- Email SACs only for private official comments made in the venue (comments visible only to Program Chairs and Senior Area Chairs)

Or we could change the description.

melisabok: I like that too but if I change the first option (the most used), I would need to keep backward compatibility with all the existing venues. I wouldn't want to do that now, I would prefer to use your suggestion in the new UI.

enrubio:
It seems like `direct_comment_email_pcs` in the venue group content isn't updated correctly. I tested by running the comment stage from the request form with different settings for **Email Program Chairs About Official Comments**.

celestemartinez (approved):

melisabok:
> It seems like direct_comment_email_pcs in the venue group content isn't updated correctly. I tested by running the comment stage from the request form with different settings for Email Program Chairs About Official Comments.

Can you describe the steps so I can try reproducing the bug with a test?

1. Configure Comment Stage and select notify PCs for private comments only
2. Post a comment and assert emails are being set to PCs for private comments only
3. Edit the Comment Stage and select not notify PCs.
4. Post a comment and asset emails are not being set to PCs

Is this the scenario?

openreview/venue_request/venue_request.py
```diff
@@ -533,6 +548,13 @@ def setup_comment_stage(self):
                 'required': False,
                 'default': 'Yes, enable chat between committee members',
                 'order': 33
+            },
+            'comment_description': {
+                'order': 34,
+                'value-regex': '[\\S\\s]{1,5000}',
+                'description': 'Specify a description for the comment stage. This will be shown in the comment form.',
```
enrubio, line 555: Can we add to the descriptions `Markdown is supported`?

Also, I like the default instructions that we build for the comment stage when the description is empty, but PCs don't know what it is or that we prepopulate it. Should we convey that to them? I'm not sure what the best way would be.

melisabok: Sure, I can add it to the description. We can ask Xukun to render the markdown icon in the text area when the field has markdown: true.

I don't understand the last statement, do you think we shouldn't prepopulate the description for them?

melisabok:
> I also noticed that when the SAC posts a comment to only PCs and SACs, the PC doesn't get notified. Is that expected? Other roles work fine.

This is a bug, the issue is the readers count validation, let me fix it

openreview/venue_request/venue_request.py
enrubio: I think the description we build is really good, but it's not clear in the request form that if `comment_description` is empty then we create it. The PCs can't see what the default is, does that make sense?

I think what we normally do in this situation is add it to the docs and link the docs in the request form. We're adding a new field to the stages so maybe it's good to update the docs, unless there a simpler way we can inform PCs of the default value? What do you think @celestemartinez, do you think it's necessary ?

celestemartinez: We can add this default value to the docs, either in the comment default form or [the comment stage](https://docs.openreview.net/reference/stages/comment-stage). I don't think it is particularly necessary but I think it could be nice to have it somewhere. 

It's not easy to show the default value in the request form since this is dynamically created depending on the options the PC choose at the time of creating/updating the comment stage.

melisabok: ok, I can mention in the description of the comment stage that a description will be autogenerated if the value is blank.

For other stages, I don't need to say anything, the value is optional

melisabok:
1. Add the markdown text to each description
2. Comment stage description says the instruction will be auto generated if no string is provided
3. Fixed bug about PCs not receiving email notification when there are 2 readers

Thanks for reviewing!

openreview/venue_request/venue_request.py
enrubio: That works too!

enrubio:
Everything looks good now! One last thing, how can they delete the instructions if they decide they no longer want it or decide they want to use our default instructions? I guess because it's markdown you can't pass `delete: true`, but maybe I'm missing something.

melisabok:
Good point, let me add a test case for that.

melisabok:
descriptions can be deleted now with the exception of the submission description (there is a pending issue in the API). A workaround could be to set a space in the description.

PR #2526, opened 2025-03-05
Custom Stage: let create a custom stage for any reply

melisabok:
Several changes have been made in this PR in order to customize the custom stage for different type of replies, list of changes below:

- Support a new `ReplyTo` type for `REBUTTALS`
- Support a new `ReplyTo` type for any invitation, use a string that indicate the name of the invitation (string after /-/)
- Define the name of the custom invitation based on the replyto invitation. For example, if we want to create a custom stage to reply to official reviews then the custom invitation will be names as "venue_id/Submisison1/Official_Review1/-/Custom_Stage_Name". This way we can have multiple levels of reply and avoid name collision.
- Support a new type of invitees `replytoReplytoSignatures` where we want the user to reply to their own thread. This is most needed for reviewers that use an anonreviewer id.
- Fix a bug in the custom stage process function to notify the reviewers/submitted group
- For rebuttal and custom stage include the code to create the reply invitation when the rebuttal/custom stage note is posted. Similar to review, meta review process functions.
- Add a test in the ICML conference to replicate the rebuttal use case for 2025


haroldrubio:
The changes make sense to me!

ARR has a rebuttal acknowledgement stage in API1 but it was difficult to bring that into API2 so I can ask if they'd want this again - but I'm not sure how to show the progress of this task (I don't know if the `customStageInvitations` property in the webfield can handle this logic)

I'm also curious about the child invitation propagation for large venues since it seems like a lot more data is getting fetched

melisabok:
ICML is trying to do the rebuttal ACK this year.

We haven't resolver the UI, I think they would need to use the API to check the status.

> I'm also curious about the child invitation propagation for large venues since it seems like a lot more data is getting fetched

True, there will be one invitation per reply but we don't get invitations in the PC console. Only in the forum pages and there should be only a few.

openreview/venue/process/invitation_edit_process.py
```diff
@@ -155,6 +162,12 @@ def post_invitation(note):
         if 'replyNumber' in invitation.edit['content']:
             content['replyNumber'] = { 'value': note.number }
 
+        if 'invitationPrefix' in invitation.edit['content']:
+            content['invitationPrefix'] = { 'value': note.invitations[0].replace('/-/', '/') + str(note.number) }
+
+        if 'replytoReplytoSignatures' in invitation.edit['content']:
+            content['replytoReplytoSignatures'] = { 'value': client.get_note(note.replyto).signatures[0] }
```
haroldrubio, line 169: Do you already have `client.get_note(note.replyto)` this note since you got all submissions and all replies? I'm just thinking maybe it's possible to avoid the API call or if you need to get a fresh note

haroldrubio (approved):

celestemartinez (approved):
I tested this with a couple of different settings and it seems to work well! Just a comment about something to keep in mind:
1. Some venues create multiple rebuttal invitations, like NeurIPS. Only one is created from the request form, so if we select `REBUTTALS` as the `replyto`, we just need to keep in mind it will use the rebuttal stage created from the request form. This is not an issue because we can always pass `Author_Rebuttal` as the `replyto` instead.

openreview/venue/process/invitation_edit_process.py
melisabok: Thanks for the catch! I fixed it.

melisabok:
@enrubio do you want to take a look or can we merge it?

enrubio (approved):

PR #2535, opened 2025-03-16
Fix: Invite and Assignment Pre-Process Quota

haroldrubio:
This PR checks for substring matching instead of strict equality for the current Invite_Assignment and Assignment edge in the quota.

This is to fix a bug where reviewers decline with a reason and the edge.label field contains `Decline: reason`. This edge is then incorrectly counted towards the quota to limit the number of active invitations for review + assignment edges.


melisabok:
@haroldrubio is this PR ready to review?

melisabok:
> @haroldrubio is this PR ready to review?

Can you answer this?

melisabok:
Instead of exclusion we should do inclusion:

quota <= num_assignment_edges + num_invite_assignment_edges(accept, invitation_sent, pending_sign_up)

openreview/venue/process/assignment_pre_process.js
```diff
@@ -28,8 +29,17 @@ async function process(client, edge, invitation) {
 
     // Filter assignment edges to exclude the current edge.id
     const filteredAssignmentEdges = assignmentEdges.filter(e => e.id !== edge.id)
-    // Filter invite assignment edges to exclude edges that are accepted
-    const filteredInviteAssignmentEdges = inviteAssignmentEdges.filter(e => !filteredLabels.includes(e?.label ?? ''))
+    // Convert includedLabels to lowercase for case-insensitive comparison
+    const lowerCaseIncludedLabels = includedLabels.map(label => label.toLowerCase());
+    
+    // Filter invite assignment edges to include only edges that contain any of the includedLabels as substrings (case-insensitive) and exclude the current edge.id
+    const filteredInviteAssignmentEdges = inviteAssignmentEdges.filter(e => {
```
melisabok, line 38: I'm just realizing that we are going to be counting twice an assignment that was created through an invitation, is that true?

haroldrubio: Fixed in recent commit by checking the assignment tails - should I make a test to make sure it doesn't do the double counting?

haroldrubio:
> Right now we list the edge IDs which could help us in debugging something, but I feel it's hard for ACs to read and understand

Adjusted the error to just show the edge count (by invite assignment/assignment) instead of the IDs

openreview/venue/process/assignment_pre_process.js
melisabok: yes, please. Make sure it doesn't count twice an assignment being done through an invite_assignment edge.

PR #2538, opened 2025-03-19
Feat: Add script to deploy openreview-py to the expertise instance

carlosmondra:
This script deploys openreview-py to the expertise instance. Currently when we deploy any version of the expertise to the instance, the master branch of openreview-py is also deployed.

@haroldrubio how is vertex AI installing openreview-py? What version is it using?

Depending on the answer, we may need to change how the deployment for openreview-py is done.

### DIFF
--- a/.github/workflows/deploy-expertise.yml
+++ b/.github/workflows/deploy-expertise.yml
@@ -1,0 +1,69 @@
+# This workflow deploys any branch, tag, or commit hash to the production environment.
+
+name: Production Deployment Expertise
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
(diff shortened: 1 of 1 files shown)


melisabok:
Thanks for doing this Carlos, I do think we need something similar the matching server too.

I'm adding Celeste as reviewer given she is in charge of making the releases of openreview-py

carlosmondra:
I believe there is no point in merging this though. I suppose that openreview-py is installed in a docker image now

carlosmondra:
Deploying openreview-py to the instance doesn't make sense anymore given that we are using vertex AI that requires a docker image

PR #2546, opened 2025-03-25
ARR: Switch SAC availability to a checkbox

haroldrubio:
This PR switches the SAC maximum load form to an agreement to share the load with the other SACs, rather than entering in a large value


melisabok (approved):

openreview/arr/management/setup_shared_data.py
```diff
@@ -242,6 +242,7 @@ def _is_not_available(month: str, year, current_date: datetime.datetime, posted_
                 )
 
     # Conditionally post unavailability notes
+    senior_area_chairs_id = reviewers_id = domain.content['senior_area_chairs_id']['value'].replace(venue_id, previous_cycle_id)
```
celestemartinez: Doesn't look like we use `reviewers_id` anywhere after this, but I think this is a typo? Should we remove `reviewers_id` here?

PR #2549, opened 2025-03-26
make sure the order of cdate <= duedate <= expdate

melisabok:
- Modify all the tests to make sure the cdate <= duedate <= expdate
- Delete invitations instead of expiring them when a submission is deleted, withdrawn or desk rejected. The `ddate` doesn't need to follow any order rule.
- Delete invitations instead of expiring them when we create the child invitations. 

This is needed before merging https://github.com/openreview/openreview-api/pull/681


carlosmondra (approved):

enrubio:
Changes look good. Just want to make sure I understand, this relates to the ordering of invitations in the venue configuration timeline right?

carlosmondra:
> Changes look good. Just want to make sure I understand, this relates to the ordering of invitations in the venue configuration timeline right?

This ensures that cdate <= duedate <= expdate is always consistent in an invitation. It should avoid mistakes when changing/adding dates

melisabok:
> Looks good! So now the _Expiration ('Withdraw_Expiration, Deletion_Expiration, etc) invitations will set a ddate instead of an expdate, correct?

Correct, I didn't change the invitation name but I could do it, I wasn't sure how to call Deletion_Expiration, Deletion_Deletion?

enrubio (approved):
Tested with the related api PR and it looks good!

PR #2554, opened 2025-04-01
Invitation description: let submission description be deleted

melisabok:
It depends on https://github.com/openreview/openreview-api/pull/676


celestemartinez (approved):

enrubio:
This looks good. Is there a difference between setting the description to `{ 'delete': True }` and `{ 'param': { 'const': { 'delete': True } } }`? I just noticed in the other invitations we set a param and const.

carlosmondra:
Yes, it should be the same

enrubio (approved):
Looks good

PR #2555, opened 2025-04-01
Import publications: allow importing papers from ORCID and arXiv

melisabok:
It requires: https://github.com/openreview/openreview-api/pull/837

1. Support externalIds field in the note schema
2. Deprecate existing DBLP import invitations
3. Support new OpenReview.net/Public_Article invitations to import from ORCID, arXiv and DBLP
4. Use a single invitation to claim authorship and remove an author for imported public articles
5. Implement public comments and tagging but this temporally disabled.


openreview/profile/management.py
```diff
@@ -271,8 +289,618 @@ def set_dblp_invitations(self):
...
+                                    }
+                                }
+                            },
+                            'pdf': {
+                                'order': 6,
+                                'description': 'Link to the PDF paper.',
+                                'value': {
+                                    'param': {
+                                        'type': 'string',
+                                        'regex': 'https://arxiv.org/pdf/.*',
```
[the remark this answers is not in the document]
melisabok: fixed

openreview/profile/process/author_coreference_pre_process.js
```diff
@@ -7,18 +7,22 @@ async function process(client, edit, invitation) {
...
   const userProfile = profiles[0];
   
   const usernames = userProfile.content.names.map(name => name.username);
   const names = userProfile.content.names.map(name => name.fullname);
 
   if (authorId === '') {
-    const authorName = publication.content.authorids.value[authorIndex];
+    const authorName = publication.content.authorids?.value[authorIndex];
     if (!usernames.some(username => username === authorName)) {
       return Promise.reject(new OpenReviewError({ name: 'Error', message: `The author name ${authorName} from index ${authorIndex} doesn't match with the names listed in your profile` }));
```
xkopenreview, line 27: currently ui allows the import when first name and last name are reversed
it will fail the validation of Authorship_Claim invitation

xkopenreview: current reverse name comparing logic in web is in
\lib\profiles.js
ln721~ln736

openreview/profile/process/orcid_record_process.js
```diff
@@ -0,0 +1,35 @@
...
+  // note.content.venueid = {
+  //   value: edit.domain
+  // }
+
+  note = {
+    id: edit.note.id,
+    content: {
+      authorids: {
+        value: edit.note.content.authors?.value.map((author) => {
+          return `https://orcid.org/orcid-search/search?searchQuery=${author}`;
```
xkopenreview: i'm not sure if it's always correct to construct search query as authorid

it's possible that the orcid record has the orcid of the author and the transform function is checking that (https://github.com/openreview/openreview-js/pull/78/files#diff-4b0765b9b068b0b3160b528dfbae64649f888db2de51164c8c22df7578a4876aR920)

melisabok: I think the orcid url is a better solution

tests/test_profile_management.py
```diff
@@ -355,9 +356,942 @@ def test_import_dblp_notes(self, client, openreview_client, test_client, helpers
...
+          "external-id-relationship" : "self"
+        } ]
+      },
+      "url" : {
+        "value" : "https://arxiv.org/abs/2407.15724"
+      },
+      "contributors" : {
+        "contributor" : [ {
+          "contributor-orcid" : {
+            "uri" : None,
```
melisabok: @xkopenreview uri is empty

openreview/profile/management.py
```diff
@@ -370,6 +778,540 @@ def set_dblp_invitations(self):
...
+                        'param': { 
+                            'items': [
+                                { 'prefix': '~.*', 'optional': True },
+                                { 'value': self.support_group_id, 'optional': True },
+                                { 'value': arxiv_uploader_group_id, 'optional': True } 
+                            ]
+                        } 
+                    },
+                    'writers':  [arxiv_uploader_group_id],
+                    'note': {
```
xkopenreview, line 1035: arxiv import invitation need to to allow note.id 
so that existing imports can be updated when arxiv has a new version

melisabok: how can I reproduce it?

xkopenreview: probably need to modify code to simulate there's a new version of imported note
unless there's a paper in arxiv that you can update the version

melisabok:
>when user B import using OrcidLinkB, paper1 would be shown as importable but would fail when posting the authorship claim 

I remember we discussed that user must add the name that matches with the orcid publication in order to be shown as "importable"

tests/test_profile_management.py
```diff
@@ -355,9 +356,1140 @@ def test_import_dblp_notes(self, client, openreview_client, test_client, helpers
...
+                    }
+                }
+            )
+        )
+
+        helpers.await_queue_edit(openreview_client, edit_id=edit['id'], process_index=0)
+        helpers.await_queue_edit(openreview_client, edit_id=edit['id'], process_index=1, error=True)
+
+        note = josiah_client.get_note(edit['note']['id'])
+        assert note.external_ids == ['doi:10.1103/PhysRevA.109.022426']
```
xkopenreview: doi will be all lowercase with https://github.com/openreview/openreview-js/pull/86

melisabok:
> in the case i described, user name match with the name shown in the import window, but does not match with paper imported by the other user

we need to show the publication in the window and ask the user to add the alternate name. I could change the claim authorship by checking that each word in the name must be present no matter the order?

tests/test_profile_management.py
melisabok: fixed

PR #2557, opened 2025-04-02
Venue Config Invitations

celestemartinez:


openreview/workflows/simple_dual_anonymous_workflow/process/edit_submission_content_process.py
```diff
@@ -0,0 +1,24 @@
...
+    domain = client.get_group(edit.domain)
+    venue_id = domain.id
+    meta_invitation_id = domain.get_content_value('meta_invitation_id')
+    
+    submission_content = edit.content['content']['value']
+
+    venue_invitations = client.get_all_invitations(prefix=venue_id + '/-/', type='invitation')
+
+    for invitation in venue_invitations:
+        if invitation.edit.get('invitation', {}).get('edit',{}).get('note',{}).get('id') == '${4/content/noteId/value}':
```
celestemartinez, line 12: @melisabok cheking for `'${4/content/noteId/value}'` in note.id worked well to find the revision invitations. 

Do you think I should use the `/Form_Fields` invitation or the meta invitation to update the content in the revision invitation? The problem with using `/Form_Fields` is it also expects the source (all_submissions, accepted_submissions), but I can read that from the existing invitation.

melisabok: let's use the meta invitation. 

I'm not sure if checking `${4/content/noteId/value}` is enough to detect a revision invitation. `noteId` can be whatever, I think we need to find a way to filter by submission invitation or Submitted venue id.

celestemartinez: Ok. I don't think there is anything else different in revision invitations that would help us differentiate them from other invitations, but I can keep thinking.

melisabok: what about the source parameter that tells the date process function how to get the notes for the `noteId` parameter?

celestemartinez: I thought about it, but we use `source` for other invitations as well, like `public_submissions` for public comment invitations and `flagged_for_ethics_review` for the ethics review stage. We don't exclusively use `source` for revision invitations.

melisabok: we need to make a combination of both conditions: source and invitation.edit.invitation.edit.note.id

melisabok: I think "reply_to" also should be part of the condition. But we can have edit invitation to edit a submission that don't need to be updated with the submission forum fields

celestemartinez: I don't think I quite understand, should I check the invitation does not have a replyto in edit.note.replyto?

melisabok: I think we need to check if reply_to is not part of the invitation.content where we store the source property, but I might be missing something, we can have a call.

melisabok (approved):

PR #2563, opened 2025-04-09
Fix/ Venue: Rebuttal Stage

celestemartinez:
1. Add an option to the request form and the rebuttal stage for PCs to choose whether ACs should get emails for rebuttals.
2. Save email_pcs and email_acs in the super invitation instead of the domain content, in case there are multiple rebuttals invitations
3. Fix the issue when multiple rebuttal stages are run from the request form. 
   - one fix is what Melisa suggested: deleting any invitations that were not re-posted
   - the other fix is passing `replacement=True` in case the id of the invitation is not changing

### DIFF
--- a/openreview/venue/process/invitation_edit_process.py
+++ b/openreview/venue/process/invitation_edit_process.py
@@ -24,7 +24,7 @@ def process(client, invitation):
         print('invitation is not yet active and no child invitations created', cdate)
         return
 
-    def delete_invitation(child_invitation):
+    def delete_invitation(child_invitation, ddate):
         client.post_invitation_edit(
             invitations=meta_invitation_id,
             readers=[venue_id],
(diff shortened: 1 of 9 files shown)


melisabok:
> To-do: fix the issue when multiple rebuttal stages are run from the request form.

could you try the following in the edit date process function?

```
current_child_invitations = client.get_all_invitations(invitation=invitation.id)

posted_invitations = openreview.tools.concurrent_requests(post_invitation, notes, desc=f'edit_invitation_process')
posted_invitations_by_id = { i.id: i for i in posted_invitations}

for current_invitation in current_child_invitations:
   if current_invitation.id not in posted_invitations_by_id:
      delete_invitation(current_invitation.id)
```

This should cover the case where delete_existing_invitations for accepted_submissions

openreview/venue/process/invitation_edit_process.py
```diff
@@ -24,21 +24,19 @@ def process(client, invitation):
...
+            readers=[venue_id],
+            writers=[venue_id],
+            signatures=[venue_id],
+            invitation=openreview.api.Invitation(
+                id=child_invitation.id,
+                ddate=now
             )
+        )
+
+    def delete_existing_invitations():
```
melisabok, line 39: I don't think you need to call this function anymore, right?

we are always going to delete the current invitations that were not created in this run

celestemartinez: Yes, you're right! Forgot to remove it

melisabok (approved):

PR #2564, opened 2025-04-11
Fix/ TMLR Fixes

celestemartinez:
1. Looks like Responsibility Acknowledgement reminders are being sent to archived reviewers. I have separated the reminder for this task from `reviewer_reminder_process.py` since there is no submission in this case and added a check to see if the reviewer is archived. 

This is what the email used to look like:
Screenshot 2025-04-11 at 3 06 40 PM

This is the new template (I removed the submission text)
Screenshot 2025-04-11 at 2 10 28 PM


melisabok (approved):
Thanks!

melisabok:
Is there a test for the reminder? I think we should add it.

PR #2565, opened 2025-04-14
ARR: Fix comment process keyerror

haroldrubio:
This PR fixes an issue where comments get deleted during the author response period, and causes the helper function that determines the thread ID (top-level parent comment) for all comments to fail.

If a comment gets posted to an orphaned thread, just set the thread to the forum

### DIFF
--- a/openreview/venue_request/process/request_form_pre_process.py
+++ b/openreview/venue_request/process/request_form_pre_process.py
@@ -32,3 +32,6 @@ def process(client, note, invitation):
 
         if abstract_deadline > (submission_deadline - datetime.timedelta(minutes=30)):
             raise openreview.OpenReviewException('The abstract registration deadline must be set at least 30 minutes before the submission deadline')
+
+    if note.content.get('api_version', '1') == '2' and len(note.content.get('venue_organizer_agreement', [])) != 6:
+        raise openreview.OpenReviewException('Please be sure to acknowledge and agree to all terms in the Venue Organizer Agreement.')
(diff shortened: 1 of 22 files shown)


openreview/arr/process/comment_process.py
```diff
@@ -2,9 +2,12 @@ def process(client, edit, invitation):
 
     def get_thread_id(tree, comment_id, forum):
         thread_id = comment_id
-        while tree[thread_id] != forum:
-            thread_id = tree[thread_id]
-        return thread_id
+        try:
+            while tree[thread_id] != forum:
+                thread_id = tree[thread_id]
```
melisabok, line 7: can you do `tree.get(thread_id, forum)` and avoid the try/except?

openreview/arr/process/comment_process.py
```diff
@@ -2,8 +2,8 @@ def process(client, edit, invitation):
 
     def get_thread_id(tree, comment_id, forum):
         thread_id = comment_id
-        while tree[thread_id] != forum:
-            thread_id = tree[thread_id]
+        while tree.get(thread_id, forum) != forum:
```
melisabok, line 5: you are doing tree.get twice per cycle in the loop? 

what about the following?

```
thread_id = tree.get(thread_id, forum)
where thread_id != forum:
   thread_id = tree.get(thread_id, forum)
```

I would prefer to avoid using "while" because it can lead to infinite loops, I would prefer use a for where we have a start and an end. 



```

celestemartinez (approved):

PR #2572, opened 2025-04-24
Add venue profile validation for invite assignment

enrubio:
This PR adds a new setting for the venue group content (`profile_minimum_requirements`) that allows venues to define profile requirements when inviting reviewers. Validation is added to the check_new_profiles cron job and to the invite assignment preprocess. Example of setting:

```
'profile_minimum_requirements': { 
    'value': {
        'relations': true,
        'history': true,
        'expertise': true,
        'publications': true,
        'active': true
    } 
}
```

If a user was invited, is in Pending Sign Up, and still has an incomplete profile, that user will stay in Pending Sign Up and receive email reminders each day until they complete their profile. Then their profile is checked for conflicts.

The email they receive will list what requirements they are missing.

- Relates to: https://github.com/openreview/openreview-js/pull/72

Changes need to match this PR in openreview-js, so before merging this we should make sure there's not more feedback on that PR that needs to be moved here.


openreview/venue/process/invite_assignment_pre_process.js
```diff
@@ -27,6 +28,53 @@ async function process(client, edge, invitation) {
   const profiles = await client.tools.getProfiles([edge.tail], true)
   const userProfile = profiles[0]
 
+  // Check for complete profile, if no profile then go to pending sign up
+  if (profileReqs && !userProfile.id.includes('@')) {
```
melisabok, line 32: I wonder if we can move this code to the tools.py so it can be used from the preprocess and the cron job.

enrubio: Ok! I can move it. To allow this JS preprocess to use it, should it get added here?: https://github.com/openreview/openreview-js/blob/8c7cd7f370459391a9cb4cc416223a386c287c6b/packages/client/src/tools.js

openreview/venue/process/invite_assignment_pre_process.js
```diff
@@ -27,6 +28,53 @@ async function process(client, edge, invitation) {
...
+        }
+  
+        if (actualValue === null || actualValue === undefined) {
+          break;
+        }
+      }
+
+      // If checking publications, filter notes for pdate
+      if (pathItems[pathItems.length - 1] === 'publications') {
+        actualValue = actualValue.filter(pub => pub.hasOwnProperty('pdate'));
```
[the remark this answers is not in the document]
melisabok: Yes!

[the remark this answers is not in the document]
enrubio: We can, like the expertise does. When we compute conflicts do/should we only look at public published publications? I was thinking we add a filter based on what info we use for conflicts.

openreview/venue/process/invite_assignment_pre_process.js
melisabok: you check by profile id instead of email address like `userProfile.id.startswith("~")`

melisabok: let's move this to the tools.py so the process function is not so large.

enrubio: But this is a JS preprocess so it would need to go in the JS version. What do you think?

melisabok: yes, let's do that

openreview/venue/process/invite_assignment_pre_process.js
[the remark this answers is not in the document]
enrubio: Talked with Melisa - decided to remove filter on publications and rely on client permissions.

openreview/venue/process/invite_assignment_pre_process.js
enrubio: This has been moved to tools.js: https://github.com/openreview/openreview-js/pull/72

openreview/venue/process/invite_assignment_pre_process.js
```diff
@@ -27,6 +28,11 @@ async function process(client, edge, invitation) {
   const profiles = await client.tools.getProfiles([edge.tail], true)
   const userProfile = profiles[0]
 
+  // Check for complete profile, if no profile then go to pending sign up
+  if (!client.tools.isProfileComplete(userProfile, profileReqs)) {
```
melisabok, line 32: what happens if profileReqs is empty?

openreview/venue/venue.py
```diff
@@ -1508,33 +1528,73 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+                                            invitation_edges = venue_client.get_edges(invitation=invite_assignment_invitation.id, label='Invitation Sent', head=submission.id, tail=user_profile.id)
                                             if invitation_edges:
                                                 invitation_edge = invitation_edges[0]
                                                 print(f'User invited twice, remove double invitation edge {invitation_edge.id}')
                                                 invitation_edge.ddate = openreview.tools.datetime_millis(datetime.datetime.now())
-                                                client.post_edge(invitation_edge)
+                                                venue_client.post_edge(invitation_edge)
+
+                                            ## Check venue profile requirements
+                                            min_requirements = venue_group.content.get('invited_reviewer_profile_minimum_requirements', {}).get('value')
```
melisabok, line 1548: can you move this code to tools.py?

enrubio:
I simplified the logic so that PCs should only require: history, relations, expertise, publications, and active state. Previously they could require any part of the profile, but that includes links which we should not allow.

Tests will fail because it relies on openreview-js changes.

openreview/venue/venue.py
```diff
@@ -1508,33 +1528,73 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+                                                    if type(expected_value) == int:
+                                                        if not actual_value or not isinstance(actual_value, list) or len(actual_value) < expected_value:
+                                                            is_incomplete = True
+                                                            break
+                                                    ## Check if field exists in profile (e.g. links)
+                                                    elif expected_value is True and not actual_value:
+                                                        is_incomplete = True
+                                                        break
+                                                    else:
+                                                        print(f'Invalid path: {profile_path}')
```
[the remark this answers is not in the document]
enrubio: This logic was removed and simplified.

openreview/venue/process/invite_assignment_pre_process.js
enrubio: I added a check for this. If there's no requirement defined by PCs then the profile check shouldn't run.

openreview/venue/venue.py
enrubio: The profile check was moved to tools.py and simplified (https://github.com/openreview/openreview-py/pull/2572#issuecomment-4031582592)

enrubio:
@melisabok @celestemartinez can you check this when you can?

If the logic in `check_profile_minimum_requirements` looks good, then the JS PR can be merged and the circle CI tests here can be re-run.

openreview/venue/process/invite_assignment_pre_process.js
```diff
@@ -85,6 +86,11 @@ async function process(client, edge, invitation) {
         return Promise.reject(new OpenReviewError({ name: 'Error', message: `Can not invite ${userProfile.id}, the user is an official reviewer` }))
       }
     }
+
+    // Check for complete profile, if no profile then go to pending sign up
+    if (profileReqs && !client.tools.checkProfileMinimumRequirements(userProfile, profileReqs)) {
```
melisabok, line 91: do these requirements apply to all the committee role or just reviewers?

openreview/conference/helpers.py
```diff
@@ -156,6 +156,7 @@ def get_conference(client, request_form_id, support_user='OpenReview.net/Support
             if venue.use_publication_chairs:
                 preferred_emails_groups.append(venue.get_publication_chairs_id())
             venue.preferred_emails_groups = preferred_emails_groups
+        venue.profile_minimum_requirements = venue_content.get('profile_minimum_requirements', {}).get('value', False)
```
melisabok, line 160: I would prefer a more descriptive name and instead of storing this in the venue domain, we can store it in the invite assignment invitation so it can be different per committee role. 

What about calling it "invitee_profile_minim_requirement"?

melisabok: I wish the new UI can have a subinvitation to edit the Invite_Assignment invitation to configure these requirements.

melisabok:
Closing this PR, no progres made.

PR #2574, opened 2025-04-24
Fix/ Venue request: validate withdrawal expdate

celestemartinez:
This fixes the issue where PCs set the withdrawal expiration date before the submission expdate, which means the withdrawal cdate > withdrawal expdate.

### DIFF
--- a/openreview/arr/process/submission_preprocess.py
+++ b/openreview/arr/process/submission_preprocess.py
@@ -35,6 +35,10 @@ def process(client, edit, invitation):
 
         if (arr_submission_v1 and 'aclweb.org/ACL/ARR' not in arr_submission_v1.invitation) or (arr_submission_v2 and not any('aclweb.org/ACL/ARR' in inv for inv in arr_submission_v2.invitations)):
             raise openreview.OpenReviewException('Provided paper link does not correspond to an ARR submission')
+        
+        # Check if the submission is from the current cycle
+        if (arr_submission_v2 and invitation.domain == arr_submission_v2.domain):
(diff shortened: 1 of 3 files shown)


celestemartinez:
@racheljsmart are there other dates in the request form that need to be validated?

openreview/venue_request/process/request_form_pre_process.py
```diff
@@ -15,4 +15,20 @@ def process(client, note, invitation):
...
+        except ValueError:
+            withdrawal_expiration = datetime.datetime.strptime(withdrawal_expiration, '%Y/%m/%d')
+
+        submission_deadline = note.content.get('Submission Deadline')
+        try:
+            submission_deadline = datetime.datetime.strptime(submission_deadline, '%Y/%m/%d %H:%M')
+        except ValueError:
+            submission_deadline = datetime.datetime.strptime(submission_deadline, '%Y/%m/%d')
+
+        if withdrawal_expiration < (submission_deadline + datetime.timedelta(minutes=30)):
```
melisabok, line 33: why is the withdrawal expiration date at the same time than the submission deadline? is it because they don't want authors to withdraw their submissions?

melisabok: I'm thinking that instead of validating this, we should adjust the withdrawal dates.

If withdrawal_expiration < submission_deadline + 30 min then don't event create the withdrawal invitation or set the cdate to be the same than the expiration

melisabok (approved):

melisabok:
Can I merge it?

celestemartinez:
> Can I merge it?

Not yet, I am fixing the issue with the Post Decision Stage as well. 

@melisabok you mentioned also validating that abstract deadline < submission deadline. How should I do this?

PR #2575, opened 2025-04-28
Feat/ TMLR: changes to decision form

celestemartinez:
TMLR asked for some changes to the AE decision form. 

This is what the form looks like:
Screenshot 2025-04-28 at 3 04 33 PM

and how the form is stored:
Screenshot 2025-04-28 at 3 05 35 PM

Also, added some code to the preprocess.

### DIFF
--- a/openreview/journal/process/submission_decision_pre_process.py
+++ b/openreview/journal/process/submission_decision_pre_process.py
@@ -3,3 +3,10 @@ def process(client, edit, invitation):
     if edit.note.content.get('recommendation', {}).get('value') == 'Reject' and edit.note.content.get('certifications', {}).get('value', []):
         raise openreview.OpenReviewException(f'Decision Reject can not have certifications')
 
+    if edit.note.content.get('claims_and_evidence', {}).get('value') == 'Yes' and edit.note.content.get('audience', {}).get('value') == 'Yes':
+        if edit.note.content.get('recommendation', {}).get('value') == 'Reject':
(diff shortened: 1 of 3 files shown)


celestemartinez:
I am open to suggestions for the field names of new fields. Also, after this has been checked, I still need to run it by TMLR and make sure they are ok with the changes.

celestemartinez:
I realize that changing the form the way I did will change it for all journals. Is this ok, or should I make the change only for TMLR?

melisabok:
> I realize that changing the form the way I did will change it for all journals. Is this ok, or should I make the change only for TMLR?

What about using the settings where you can add/remove fields to the decision form?

openreview/journal/process/submission_decision_pre_process.py
```diff
@@ -3,3 +3,10 @@ def process(client, edit, invitation):
     if edit.note.content.get('recommendation', {}).get('value') == 'Reject' and edit.note.content.get('certifications', {}).get('value', []):
         raise openreview.OpenReviewException(f'Decision Reject can not have certifications')
 
+    if edit.note.content.get('claims_and_evidence', {}).get('value') == 'Yes' and edit.note.content.get('audience', {}).get('value') == 'Yes':
```
melisabok, line 6: are these fields specific to TMLR?

I believe it doesn't affect other venues if these fields don't exist in the decision invitation.

celestemartinez: They are not specific to TMLR, they exist in the decision invitation for the journal but are textareas, so I thought this wouldn't affect them.

PR #2576, opened 2025-04-29
Feat/ Venue request: add venue organizer agreement

celestemartinez:
I have added the venue organizer agreement to the existing request form. This does not include the text about sharing emails, computing review counts and maming accepted papers public. 

~~I made the field optional but have a preprocess that checks all the checboxes have been checked. This way I won't have to change all api1 tests (but I can if we prefer it that way).~~

Screenshot 2025-04-29 at 3 13 24 PM


openreview/venue_request/venue_request.py
```diff
@@ -1616,6 +1616,19 @@ def setup_request_form(self):
                 'description': 'Please briefly describe how you heard about OpenReview.',
                 'order': 41
             },
+            'venue_organizer_agreement': {
+                'description': 'In order to use OpenReview, venue chairs must agree to the following:',
+                'values-checkbox': [
+                    'OpenReview natively supports a wide variety of reviewing workflow configurations. However, if we want significant reviewing process customizations or experiments, we will detail these requests to the OpenReview staff at least three months in advance.',
+                    'We will ask authors and reviewers to create an OpenReview Profile at least two weeks in advance of the paper submission deadlines.',
+                    'When assembling our group of reviewers and meta-reviewers, we will only include email addresses or OpenReview Profile IDs of people we know to have authored publications relevant to our venue.  (We will not solicit new reviewers using an open web form, because unfortunately some malicious actors sometimes try to create \'fake ids\' aiming to be assigned to review their own paper submissions.)',
```
celestemartinez, line 1624: @xkopenreview You can use this PR to check the issue I mentioned. I am using single-quotes here and escaping them, since using double-quotes does not work.

melisabok:
API 1 tests are disabled so I would prefer to make the field mandatory

melisabok:
@celestemartinez what do you think about removing the iThenticate fields in this PR?

celestemartinez:
> API 1 tests are disabled so I would prefer to make the field mandatory

ok, it is now required. We still need the preprocess anyway because we need to check all values are checked (and setting them as default makes them already checked, which we don't want). 

> @celestemartinez what do you think about removing the iThenticate fields in this PR?

Yes, I can remove them. Do we not need them at all anymore or am I moving them to another place?

melisabok (approved):

celestemartinez:
> @celestemartinez i think current change is using single quote around fake id. do you still plan to use double quote?

I saw that the fix you made is in the workflow PR and I don't know when we'll merge that, so I kept the single quotes here.

PR #2578, opened 2025-05-01
Comment Stage: use inGroup to select anon reviewers

melisabok:
Fixes https://github.com/openreview/openreview-web/issues/2273

When reviewers are unassigned, they won't showup as available readers because they are not part of the reviewers group anymore.


celestemartinez (approved):

enrubio (approved):
Changes look good and the comment readers are as expected. Thank you!

enrubio:
I noticed the readers of the anon groups are still changing whenever someone is unassigned/assigned. I know this is done in the API, will that change be reverted now that we have this change? Maybe a question for @carlosmondra

PR #2579, opened 2025-05-04
Feat/ Venue Config: refactor + AC workflow

celestemartinez:


openreview/workflows/workflows.py
```diff
@@ -0,0 +1,938 @@
...
+                signatories=[]
+            )
+        )
+
+
+        self.client.post_group_edit(
+            invitation=self.meta_invitation_id,
+            signatures=['~Super_User1'],
+            group=openreview.api.Group(
+                id=f'{support_group_id}/Venue_Request/Reviewers_Only',
```
melisabok, line 118: Do you need to create these groups?

openreview/workflows/workflows.py
```diff
@@ -0,0 +1,938 @@
...
+                }
+            }
+        )
+
+        self.post_invitation_edit(invitation)
+
+    def set_reviewers_only_comment(self):
+
+        support_group_id = self.support_group_id
+        comment_invitation_id = f'{support_group_id}/Venue_Request/Reviewers_Only/-/Comment'
```
melisabok, line 422: this invitation seems to be very similar to all the workflow requests, can we use one invitation for all requests? same as the deployment invitation.

melisabok:
- Should we have a folder for each workflow request? Only_Reviewers and Only_ACs_And_Reviewers. Each folder can have the deploy process function that extend the workflow templates.

openreview/workflows/workflow_process/area_chairs_group_template_process.py
```diff
@@ -0,0 +1,57 @@
+def process(client, edit, invitation):
```
melisabok, line 1: this is something that I would like to refactor to be the same for any group that we want to create that requires messaging and recruitment.

tests/test_acs_and_reviewers.py
```diff
@@ -0,0 +1,123 @@
...
+        assert openreview_client.get_invitation('openreview.net/Support/Venue_Request/ACs_and_Reviewers/-/Comment')
+
+        now = datetime.datetime.now()
+        due_date = now + datetime.timedelta(days=2)
+
+        request = pc_client.post_note_edit(invitation='openreview.net/Support/Venue_Request/-/ACs_and_Reviewers',
+            signatures=['~ProgramChair_EFGH1'],
+            note=openreview.api.Note(
+                content={
+                    'official_venue_name': { 'value': 'The EFGH Conference' },
```
melisabok, line 50: could we add two fields to specify the reviewers_name and area_chairs_name?

openreview/workflows/workflow_process/support_process.py
```diff
@@ -13,10 +13,14 @@ def process(client, edit, invitation):
     if edit.id != note_edits[0].id:
         print('not first edit, exiting...')
         return
+
```
melisabok, line 16: I would move this file outside of the workflows folder. Maybe another folder?

tests/test_reviewers_only.py
```diff
@@ -96,10 +102,10 @@ def test_setup(self, openreview_client, helpers):
             ))
         
         helpers.await_queue_edit(openreview_client, edit_id=edit['id'])
-        helpers.await_queue_edit(openreview_client, invitation='openreview.net/Support/Simple_Dual_Anonymous/Venue_Configuration_Request/-/Submission_Template')
```
melisabok, line 99: can you add to this test the parentInvitations for submissions and reviews?

For example:
```
assert submissions[0].parent_invitations == 'openreview.net/Support/-/Submission_Template'
```

openreview/workflows/workflows.py
celestemartinez: I am not sure about using the same deployment invitation, because each workflow will have its own deployment process. 
I can make the change for the comment invitation now.

melisabok: I suggest using the same code to create the invitation and parametrie the name of the process function?

tests/test_reviewers_only.py
melisabok: Because `ABCD.cc/2025/Conference/-/Submission` is already in the note.invitations

melisabok: can you try querying the review by parent_invitations = openreview.net/Support/-/Review_Template?

tests/test_acs_and_reviewers.py
celestemartinez: Done.

openreview/conference/helpers.py
```diff
@@ -31,6 +31,9 @@ def get_venue(client, venue_id, support_user='OpenReview.net/Support'):
     venue.allow_gurobi_solver = domain.content.get('allow_gurobi_solver', {}).get('value', False)
     venue.preferred_emails_groups = domain.content.get('preferred_emails_groups', [])
     
+    venue.submission_stage = openreview.stages.SubmissionStage(
```
celestemartinez, line 34: @melisabok Using the venue code to upload decisions works, but I had to add the submission_stage here. `submission_stage.name` is used when we call `get_paper_group_prefix()` to get the paper decision invitation. Is there a better way to do this?

melisabok: yes, it is great! then all the code will work, a lot of places call to venue.submission_stage so I think setting this up in the get_venue function seems correct. We should do the same for other stages if needed

openreview/venue/group.py
```diff
@@ -277,6 +286,9 @@ def create_venue_group(self):
         if self.venue.request_form_id:
             content['request_form_id'] = { 'value': self.venue.request_form_id }
 
+        if self.venue.request_form_invitation:
+            content['request_form_invitation'] = { 'value': self.venue.request_form_invitation }
```
melisabok, line 290: @xkopenreview is this keyword exists then the venue is running on the new workflow configuration

openreview/workflows/workflow_process/area_chairs_invited_group_template_process.py
```diff
@@ -0,0 +1,104 @@
+def process(client, edit, invitation):
```
melisabok, line 1: should we delete this file?

melisabok (approved):

PR #2586, opened 2025-05-14
Assignment Quotas: Fix race condition

haroldrubio:
This PR adds a fix to the assignment/invite assignment quota caused by a delay in inserting data into Mongo. This delay adds a race condition where two invites get sent out while receiving outdated information on the number of assignments to a paper. For example, 2 assignments are already made, and 2 invite assignment requests are made quickly. Only one of these should post but both of these will post causing errors in any subsequent assignments to that paper.

The fix:
1. Bypasses the quota for invite assignment edges already posted - any invite assignments will be able to be updated regardless of if they exceed the quota.
2. Add a small margin of error in the assignment pre-process that allows for cases that exceed the quota


openreview/venue/process/assignment_pre_process.js
```diff
@@ -31,7 +31,7 @@ async function process(client, edge, invitation) {
     // Filter invite assignment edges to exclude edges that are accepted
     const filteredInviteAssignmentEdges = inviteAssignmentEdges.filter(e => !filteredLabels.includes(e?.label ?? ''))
 
-    if (quota && filteredInviteAssignmentEdges.length + filteredAssignmentEdges.length >= quota) {
+    if (quota && filteredInviteAssignmentEdges.length + filteredAssignmentEdges.length >= quota + 1) {
```
melisabok, line 34: if you relax this validation, all the papers can have quota + 1 assignments, is this what we want?

melisabok: There are two options to solve this:

1. Reject the assignment for the invited reviewer that accepted the invitation. Send an email saying the paper already has enough reviewers and their review is not needed anymore.
2. Bypass the validation here by checking that if that assignment is coming from an invite assignment then let it post the edge. You can check that is the edge that we are trying to post already exists as an invite assignment (same head and tail) then we let the edge to be posted.

melisabok: @haroldrubio any updates on this?

melisabok (approved):

tests/test_arr_venue_v2.py
```diff
@@ -3804,17 +3869,43 @@ def test_sae_ae_assignments(self, client, openreview_client, helpers, test_clien
             ['~Reviewer_ARRThree1']
         )
 
+        ## Assert that recruitment process function works with > 3 reviewers (assignment can be broken by quota + 1)
```
celestemartinez, line 3872: Just double checking: this works not because we are allowing the quota to be broken by one, but because we are now accepting all assignments when they come from an Invite_Assignment?

haroldrubio: Yes that's correct. This way, if invites violate the quota, then that invitation can still work as normal. Eventually, the invite assignment quota will be reached once the database catches up and prevents further invite assignment edges

celestemartinez (approved):

PR #2595, opened 2025-06-02
Fix/ Matching: add Ethics Chairs as readers of ethics reviewers edges

celestemartinez:
This PR adds Ethics Chairs as readers of edges created for Ethics Reviewers. 

@melisabok should we remove the computation of conflicts for Ethics Reviewers? This change also makes conflict edges visible to Ethics Chairs.


melisabok (approved):

melisabok:
We can let the PCs decide if they want to compute conflict or not, right?

PR #2596, opened 2025-06-04
Post Decision Stage: do not change the venueid for submission that don't have a decision posted

melisabok:
Currently submissions that don't have a decision posted are being treated as rejected submission when the post decision stage runs.

This PR ignores these submissions and it doesn't change the venueid, readers not send any notification.

### DIFF
--- a/openreview/venue/venue.py
+++ b/openreview/venue/venue.py
@@ -429,8 +429,14 @@ class Venue(object):
 
         if venueid:
             return self.client.get_all_notes(content={ 'venueid': venueid}, sort=sort, details=details)
+        
+        venueids = [
+            self.get_submission_venue_id(),
+            self.venue_id,
+            self.get_rejected_submission_venue_id()
+        ]
 
-        return self.client.get_all_notes(invitation=self.get_submission_id(), sort=sort, details=details)
+        return self.client.get_all_notes(content={ 'venueid': ','.join(venueids)}, sort=sort, details=details)
 
     #use to expire revision invitations from request form
     def expire_invitation(self, invitation_id):
--- a/openreview/venue/venue.py
+++ b/openreview/venue/venue.py
@@ -415,7 +415,7 @@ class Venue(object):
 
     def get_submissions(self, venueid=None, accepted=False, sort='tmdate', details=None):
         if accepted:
-            accepted_notes = self.client.get_all_notes(content={ 'venueid': self.venue_id}, sort=sort)
+            accepted_notes = self.client.get_all_notes(content={ 'venueid': self.venue_id}, sort=sort, details=details)
             if len(accepted_notes) == 0:
                 accepted_notes = []
                 notes = self.client.get_all_notes(content={ 'venueid': f'{self.get_submission_venue_id()}'}, sort=sort, details='directReplies')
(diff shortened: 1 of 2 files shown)


openreview/venue/venue.py
```diff
@@ -426,14 +426,17 @@ def get_submissions(self, venueid=None, accepted=False, sort='tmdate', details=N
...
-                notes = self.client.get_all_notes(content={ 'venueid': self.venue_id}, sort=sort, details=details)
-                rejected = self.client.get_all_notes(content={ 'venueid': self.get_rejected_submission_venue_id()}, sort=sort, details=details)
-                if rejected:
-                    notes.extend(rejected)
-            return notes
+
+        if venueid:
+            return self.client.get_all_notes(content={ 'venueid': venueid}, sort=sort, details=details)
+        
+        venueids = [
```
melisabok, line 470: @celestemartinez this could be a breaking change

openreview/venue/venue.py
```diff
@@ -426,14 +426,17 @@ def get_submissions(self, venueid=None, accepted=False, sort='tmdate', details=N
...
+            self.get_submission_venue_id(),
+            self.venue_id,
+            self.get_rejected_submission_venue_id()
+        ]
+
+        return self.client.get_all_notes(content={ 'venueid': ','.join(venueids)}, sort=sort, details=details)
```
enrubio, line 476: I didn't know we could do this 😮

celestemartinez:
The change looks good! 

One thing I noticed is that we still run the process if the decision is deleted. This might be out of the scope of this PR, but I wonder if we should do this in the case an accept decision is changed to reject or deleted:
1. Any revision invitations for accepted submissions should be expired
2. Authors group should be removed from the Authors/Accepted group

melisabok:
I believe 2 is already happening.

1. I could add it to this PR so we can have it fixed, I know this is a very common case.

openreview/venue/process/decision_process.py
```diff
@@ -31,8 +51,7 @@ def process(client, edit, invitation):
         )
 
     if (authors_accepted_id):
-      accept_options = domain.get_content_value('accept_decision_options')
-      if openreview.tools.is_accept_decision(decision.content['decision']['value'], accept_options):
+      if is_accepted_decision:
```
melisabok: @celestemartinez this is add or remove based on the decision

celestemartinez: Yes, you are right, the authors are removed when the decision is changed. I believe they are not removed when the decision is deleted. Not sure how often this happens or if this is a big issue.

melisabok: ohhhh, good point, I will fix that.

melisabok: it should be fixed, in another PR we can update the venueid when a decision is deleted. I think we need to use invitations to edit the venueids instead of doing it with meta invitations

melisabok:
@enrubio good point. @celestemartinez should I fix this in this PR?

Can I remove the expiration of the post submission PR? what I can do is that the post submission should only be applied to submissions that are active AND do not have a decision, would that work?

melisabok:
ok, I won't make any change, please feel free to merge it 🙏

Issue #2594, opened 2025-06-02
CustomStage: Add support for source_submissions_query


enrubio:
Add support for `source_submissions_query` in the CustomStage so that we can create custom invitations for any paper `venueid`.

melisabok:
Can you explain the use case for AAAI? I would like to start working on this example.

It should be a task for the authors of desk-rejected papers, is that right?

enrubio:
Yes, they want to allow authors of desk-rejected papers to challenge the desk-rejection. So they want to collect some response from the authors stating their case, then handle them.

PR #2597, opened 2025-06-06
Feature/allow custom stage rejected submissions

melisabok:
Deprecate the source string: "all_submissions", "accepted_submissions", "public_submissions" and "flagged_for_ethics_review", "reply_to" and "source_submission_query". Instead use a source dictionary that allows several keys:

```
{
venueid: ['ICLR.cc/2025/Conference/Submission', 'ICLR.cc/2025/Conference/Submission/Desk_Rejected_Submission'],
readers: ['everyone'],
reply_to: 'Official_Review',
content: {
'flagged_for_ethics_review': True
},
with_decision_accept: True
}
```

each of these keywords are optionals, default value is { venueid: "venue_id/Submission"}

This is backward compatible and new invitations are now created using source as a dictionary. 

Tools module has a function to translate the source to expected format and two ways to create child invitations: for forums and replyto. This is reduce the duplicated code and have the source check in a single place.

CustomStage class can receive a source dictionary as parameter so we can specify more than one venue id.

Fixes: https://github.com/openreview/openreview-py/issues/2594 and https://github.com/openreview/openreview-py/issues/2599

Fixes the issue of changing the decision readers for submissions with accepted/rejected venueid and submission venueid (releasing decisions in batches)

it depends on https://github.com/openreview/openreview-api/pull/770


carlosmondra (approved):

openreview/stages/venue_stages.py
```diff
@@ -1690,18 +1690,18 @@ def get_signatures(self, conference, number):
...
 
         if self.source == self.Source.ACCEPTED_SUBMISSIONS:
-            source = 'accepted_submissions'
-        elif self.source == self.Source.PUBLIC_SUBMISSIONS:
-            source = 'public_submissions'
-        elif self.source == self.Source.ALL_SUBMISSIONS:
-            source = 'all_submissions'
-        elif self.source == self.Source.FLAGGED_SUBMISSIONS:
-            source = 'flagged_for_ethics_review'
+            return { 'venueid': [venue.venue_id, venue.get_submission_venue_id()], 'with_decision_accept': True }
```
celestemartinez, line 1696: For accepted submissions we pass `[venue.venue_id, venue.get_submission_venue_id()]` in case the venueid hasn't been changed, since we are also checking the paper has an accept decision?

melisabok: Correct

openreview/venue/venue.py
```diff
@@ -450,7 +450,10 @@ def get_rejected_submission_venue_id(self, submission_invitation_name=None):
         if self.submission_stage:
             return f'{self.venue_id}/Rejected_{self.submission_stage.name}'
         return f'{self.venue_id}/Rejected_Submission'
-
+    
+    def get_active_venue_ids(self, submission_invitation_name=None):
+        return [self.venue_id, self.get_submission_venue_id(submission_invitation_name), self.get_rejected_submission_venue_id(submission_invitation_name)]
```
enrubio, line 455: I'm missing why these venue IDs are considered active, I thought the submission venue ID is considered active? But we're including the accepted and rejected venue IDs for a paper (`self.venue_id` and `get_rejected_submission_venue_id`).

melisabok: There are some tasks that happen after the papers are marked as "accepted" or "rejected", like camera ready revisions,etc

openreview/venue/process/invitation_edit_process.py
```diff
@@ -37,47 +37,39 @@ def delete_invitation(child_invitation, ddate):
...
+
+            ## Keep backward compatibility with 'all_submissions' before and after running the post_decision_stage
             if not source_submissions:
                 source_submissions = client.get_all_notes(content={ 'venueid': ','.join([venue_id, rejected_venue_id]) }, sort='number:asc', details='replies')
+            
+            if 'with_decision_accept' in source:
+                source_submissions = [s for s in source_submissions 
+                                      if len([r for r in s.details['replies'] 
+                                        if f'{venue_id}/{submission_name}{s.number}/-/{decision_name}' in r['invitations'] 
+                                        and openreview.tools.is_accept_decision(r['content'][decision_field_name]['value'], accept_options) == source.get('with_decision_accept')]) > 0]
```
celestemartinez, line 56: @melisabok about the issue I found, I think the tests are passing because we are not using `should_match_invitation_source` in `invitation_edit_process.py` and here we are checking that the decision and `with_decision_accept` match.

openreview/venue/invitation.py
```diff
@@ -1989,6 +1990,11 @@ def set_decision_invitation(self):
                 },
                 'accept_decision_options': {
                     'value': ['Accept (Oral)', 'Accept (Poster)']
+                },
+                'source': {
+                    'value': {
+                        'venueid': [venue_id, self.venue.get_submission_venue_id(), self.venue.get_rejected_submission_venue_id()]
```
celestemartinez: this is the same as `self.venue.get_active_venue_ids()`, right?

openreview/venue/invitation.py
```diff
@@ -2739,7 +2745,7 @@ def set_submission_revision_invitation(self, submission_revision_stage=None):
                     'value': self.get_process_content('process/submission_revision_process.py')
                 },
                 'source': {
-                    'value': 'accepted_submissions' if only_accepted else 'all_submissions'
+                    'value': { 'venueid': [venue_id, self.venue.get_submission_venue_id()], 'with_decision_accept': True } if only_accepted else { 'venueid': self.venue.get_active_venue_ids() }
```
celestemartinez: I am worried that sometimes the PCs change the decision after running the post decision, so the venueid might not match the decision. Maybe we should update the venueid in the decision process if it has already been updated once, but we can do this in another PR.

openreview/venue/invitation.py
melisabok: yes, I forgot to replace it.

openreview/venue/process/invitation_edit_process.py
melisabok: ok, but it is called in the process function of decision process for example so I will add a test to reproduce it

openreview/venue/invitation.py
melisabok: yes, we should do it in the decision procoess.

Ideally, the venueid should be set using an invitation instead of code in the venue class like TMLR does. One invitation for Accepted_Submission and another one for Rejected_Submission

melisabok: let's make this change in another PR: move the post decision code to invitations

openreview/venue/process/invitation_edit_process.py
celestemartinez: I don't think it's being called there either? looks like we call it only in the submission, desk-rejection and withdrawal processes?

openreview/venue/process/withdrawal_reversion_submission_process.py
```diff
@@ -5,6 +5,7 @@ def process(client, edit, invitation):
     meta_invitation_id = domain.content['meta_invitation_id']['value']
     short_name = domain.content['subtitle']['value']
     contact = domain.content['contact']['value']
+    submission_venue_id = domain.content['submission_venue_id']['value']
```
enrubio: Is this supposed to be used somewhere? Looks like it's not currently.

openreview/venue/process/invitation_edit_process.py
```diff
@@ -37,47 +37,39 @@ def delete_invitation(child_invitation, ddate):
...
+
+            if 'readers' in source:
+                source_submissions = [s for s in source_submissions if set(source['readers']).issubset(set(s.readers))]
+
+            if 'content' in source:
+                for key, value in source.get('content', {}).items():
+                    source_submissions = [s for s in source_submissions if value == s.content.get(key, {}).get('value')]
+
+            if 'reply_to' in source:
+                source_submissions = [(openreview.api.Note.from_json(reply), s) for s in source_submissions for reply in s.details['replies'] if reply['invitations'][0].endswith(f'/-/{source.get("reply_to")}')]
```
enrubio, line 66: Why do we check only the first invitation in the reply? Can't there be multiple invitations? I see in the original code we checked the whole list like this: `if f'{venue_id}/{submission_name}{s.number}/-/{review_name}' in reply['invitations']` and in other cases checked only the first invitation, but I'm not sure why or if it would cause an issue.

melisabok: We assume that we check the invitation that was used to create the reply, like "Official_Review", this is code is the same as master, it hasn't changed in this PR

enrubio: Okay I understand, I saw it was different (in master in some cases we check the whole list) so just wanted to clarify that we are assuming it's created with the correct invitation.

enrubio (approved):
I looked through the code and ran the custom stage a few times with different configurations and didn't notice any problems. Very cool change, thank you!

celestemartinez (approved):

PR #2600, opened 2025-06-10
Journal: use stream true to load all the invitations, notes and groups

melisabok:


openreview/journal/webfield/editorsInChiefWebfield.js
```diff
@@ -175,10 +175,10 @@ var loadData = function() {
     Webfield2.api.getGroupsByNumber(VENUE_ID, ACTION_EDITOR_NAME),
     Webfield2.api.getGroupsByNumber(VENUE_ID, REVIEWERS_NAME, { withProfiles: true}),
     Webfield2.api.getAllSubmissions(SUBMISSION_ID, { domain: VENUE_ID }),
```
xkopenreview, line 177: how about getAllSubmissions
it calls getAll('/notes', query)

melisabok: getAllSubmissions adds the details param so we can not use stream=true here.

melisabok:
> is this change making the eic console to load faster?

I don't see any different, you are right but @carlosmondra found slow queries coming from this console.

openreview/journal/webfield/editorsInChiefWebfield.js
xkopenreview: the change will be in webfield2.js
from getall('/notes to get('/notes and add stream true

xkopenreview: discussed this in the daily meeting
when getAllSubmissions request for details then stream won't work
so leave it as it is now

PR #2610, opened 2025-06-20
search profiles with blocked status

melisabok:


melisabok:
> should all profile block tags be auto removed?
Yes, I think so but I'm not sure.

PR #2619, opened 2025-06-25
Fix/ Workflow fixes

celestemartinez:


tests/test_icml_conference_with_templates.py
```diff
@@ -884,36 +885,27 @@ def test_post_submission(self, client, openreview_client, test_client, helpers,
         assert desk_reject_inv.expdate == expdate
         assert pc_client.get_invitation('ICML.cc/2025/Conference/-/PC_Revision')
 
-        ## make submissions visible to SACs, ACs only
+        # ## make submissions visible to SACs, ACs only
         
-        ac_client = openreview.api.OpenReviewClient(username = 'ac1@icml.cc', password=helpers.strong_password)
```
celestemartinez, line 889: Now we are not letting PCs change the submission_readers in `Submission_Change_Before_Bidding`, so SACs, ACS and reviewers are always added.

melisabok: Sounds good for now, we might want to give them the option to remove them if they want to enable bidding for ACs only, for example. But this is not needed for the first release.

tests/test_icml_conference_with_templates.py
```diff
@@ -884,36 +885,27 @@ def test_post_submission(self, client, openreview_client, test_client, helpers,
...
-        'ICML.cc/2025/Conference/Submission1/Authors'] == submissions[0].readers
-        assert ['ICML.cc/2025/Conference',
-        'ICML.cc/2025/Conference/Submission1/Authors'] == submissions[0].writers
-        assert ['ICML.cc/2025/Conference/Submission1/Authors'] == submissions[0].signatures
-        assert 'authorids' not in submissions[0].content
-        assert 'authors' not in submissions[0].content
-        assert 'pdf' not in submissions[0].content
-        assert 'financial_aid' in submissions[0].content
-        assert not submissions[0].odate
+        # ac_client = openreview.api.OpenReviewClient(username = 'ac1@icml.cc', password=helpers.strong_password)
```
melisabok: can we keep this test and add the reviewers to the list in the assertion?

openreview/workflows/workflows.py
```diff
@@ -365,14 +372,8 @@ def set_conference_review_deployment(self):
...
-                    'ddate': {
-                        'param': {
-                            'range': [ 0, 9999999999999 ],
-                            'optional': True,
-                            'deletable': True                                 
-                        }
-                    },
-                    'signatures': ['${3/signatures}'],
+                    # 'signatures': ['${3/signatures}'],
+                    'signatures': ['${{3/note/id}/signatures}'],
```
celestemartinez: @melisabok the API won't let me not have `signatures` here, but I am not sure what the best way to keep the PC as signatures is

melisabok: instead of using a template, can you create the Deployment invitation per venue using Python code? I would create it in the setup_matching function if the venue uses the new workflow invitation.

celestemartinez: This is the venue deployment, not for assignments. I thought we wanted to have one invitation per request form type instead of creating one per venue request

melisabok: ohhh sorry about that, if you try removing "'optional': True" in note.id.param, I believe the signatures field is not manadatory.

I think the API requires the signatures field when the invitation is available to create new notes instead of editing them.

melisabok: This invitation should be used to update an existing note only.

celestemartinez: ohhh I see, that makes sense. Thanks!

openreview/venue/invitation.py
```diff
@@ -166,6 +172,9 @@ def set_submission_invitation(self):
             duedate = submission_duedate,
             expdate = tools.datetime_millis(submission_stage.exp_date) if submission_stage.exp_date else None,
             content = {
+                'users_to_notify': {
+                    'value': groups_to_notify
+                },
```
celestemartinez: I don't think we can remove `email_authors`, `email_pcs`, etc since the venue is using this. I am not sure what's the best way to not make this confusing for PCs:
1. Keep both for all invitations (PCs would be able to edit either field to edit notification settings)
2. Keep  booleans only for normal venues and use `users_to_notify` only for the new config invitations

melisabok: I choose number 2, deprecate boolean and use the array of options.

openreview/venue/process/review_process.py
```diff
@@ -8,20 +8,28 @@ def process(client, edit, invitation):
...
     submission_name = domain.get_content_value('submission_name')
     reviewers_name = domain.get_content_value('reviewers_name')
+    pretty_reviewers_name = reviewers_name.replace('_', ' ')
     area_chairs_name = domain.get_content_value('area_chairs_name')
     senior_area_chairs_name = domain.get_content_value('senior_area_chairs_name')
     reviewers_submitted_name = domain.get_content_value('reviewers_submitted_name')
     review_name = domain.get_content_value('review_name')
     sender = domain.get_content_value('message_sender')
 
+    parent_invitation = client.get_invitation(invitation.invitations[0])
```
celestemartinez, line 17: I've added more checks here. First I check if the parent invitation has the booleans. If yes, use those. Otherwise, search users_to_notify. Once we move fully to the new venue config, we can get rid of the booleans

openreview/venue/invitation.py
```diff
@@ -541,11 +556,14 @@ def set_review_invitation(self):
                 'script': self.invitation_edit_process
             }],
             content={
-                'email_program_chairs': {
-                    'value': review_stage.email_pcs
-                },
                 'source': {
                     'value': review_stage.get_submission_source(self.venue)
+                },
+                'review_rating': {
```
melisabok: 'rating_field_name'

celestemartinez:
I've made the changes we spoke about on Friday, so this should be ready to merge. Let me know if there is anything else that needs fixing!

melisabok (approved):
💯

PR #2625, opened 2025-07-07
Feat/TMLR: Beyond pdf

celestemartinez:
Changes:
- have a `pdf` (required) and `beyond_pdf` (optional) field in the submission form
- have a preprocess that checks the submission type
   - There is another option in `submission_type`: "Beyond PDF submission (pageless, webpage-style content)", so I added a preprocess for this but I am not sure if this is necessary

### DIFF
--- a/openreview/arr/helpers.py
+++ b/openreview/arr/helpers.py
@@ -1160,7 +1160,7 @@ class ARRWorkflow(object):
                     'writers': []
                 },
                 start_date=self.configuration_note.content.get('metadata_edit_start_date'),
-                due_date=self.configuration_note.content.get('metadata_edit_end_date')
+                exp_date=self.configuration_note.content.get('metadata_edit_end_date')
             )
         ]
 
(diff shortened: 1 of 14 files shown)


openreview/journal/journal.py
```diff
@@ -528,6 +528,9 @@ def has_expert_reviewers(self):
     def has_external_reviewers(self):
         return self.settings.get('external_reviewers', True)
 
+    def uses_beyond_pdf(self):
```
melisabok: do you think other journal would use this feature? I would not try to support it in the journal unless it is really needed.

In this case, it looks they only need to add a pre process validation so we can try to support the configuration of the preprocess instead of the setting "beyond_pdf"

celestemartinez: What would be the best way to support the configuration of the preprocess? We would still need a setting in the journal request form, right?

melisabok: yes, I would add a property in the settings that is the preprocess code. Or when we create the submission invitation, if the preprocess exists then do not overwrite it

openreview/journal/process/submission_pre_process.py
```diff
@@ -0,0 +1,7 @@
+def process(client, edit, invitation):
```
melisabok, line 1: do we want to store this code in the repo? it looks like it is related to TMLR only, we can use a different name if we decide to keep it: tmlr_submission_pre_process.py for example

PR #2634, opened 2025-07-16
Fix tests for readers/nonreaders validation

melisabok:


melisabok:
> the PCs will get this error if they try to remove them:

From which invitation they can remove the authors?

celestemartinez (approved):

PR #2636, opened 2025-07-17
Fix/message sign test

melisabok:


tests/test_reviewers_only.py
```diff
@@ -1395,6 +1395,9 @@ def test_email_reviews(self, openreview_client, helpers):
 
 To view this paper, please go to https://openreview.net/forum?id={submissions[0].id}
 
+This message was signed by ABCD 2025 Conference Automated Administrator
```
melisabok: @celestemartinez I wonder if the messages can be signed using the venueid so this message doesn't appear in the review notification email. Does the Automated Administrator have permission to sign as the venueid?

carlosmondra (approved):

PR #2639, opened 2025-07-23
Introduce a new comprehensive CoI policy

yoshitomo-matsubara:


melisabok:
could you try adding some tests here? https://github.com/yoshitomo-matsubara/openreview-py/blob/master/tests/test_tools.py#L560

I believe you can run `pytest tests/tests_tools.py`, if that doesn't work, CircleCI would run them for you.

yoshitomo-matsubara:
I couldn't run the test on my local machine, but Circle ran it for me

openreview/tools.py
```diff
@@ -1568,6 +1570,65 @@ def get_neurips_profile_info(profile, n_years=None):
         'publications': publications
     }
 
+def get_comprehensive_profile_info(profile, n_years=None):
```
melisabok, line 1573: I realized this function is very similar to get_neurips_profile_info but without filtering out intern positions and taking into account only relations ['Coauthor','Coworker']. 

I suggest get_neurips_profile_info calls get_comprehensive_profile_info and pass these two parameters:

`get_comprehensive_profile_info(profile, n_years=None, ignore_positions=None, only_relations=None, )`

and 
```
def get_neurips_profile_info(profile, n_years=None):
   get_comprehensive_profile_info(profile, n_years, ['intern'], ['Coauthor','Coworker'])
```

I can help to make this change.

melisabok: it looks like the difference between get_comprehensive_profile_info and get_profile_info (default) is that email domains are not taken into account for conflict detection, only institution domains. Having the option "Comprehensive" and "Default" might be confusing for the users, I would like to join them in a single option but I'm not sure if venues would want to still use email domains to detect conflicts, @carlosmondra @racheljsmart what do you think?

racheljsmart: @melisabok I think it would be fine in a single option as long as they can choose to include email domains or not, similar to setting a conflict year. It's tricky because users can set whatever in their institution domains but it's likely more detailed then whatever we get from the emails.

melisabok: Right, adding a new param list n_years will require passing a new parameter everywhere so I would prefer to add a new policy option, the question is that these options make sense:

- **Default**: takes into account the whole profile information for the last n years
- **Comprehensive**: takes into account the whole profile information except email domains for the last n years
- **NeurIPS**: takes into account the whole profile information for the last n years with the exception of:
1. email domains
2. "intern" position
3.  only includes co-author, co-worker relations

melisabok (approved):

PR #2647, opened 2025-08-01
Fix/ TMLR fixes

### LINKED ISSUE #2644: TMLR: review process
In the review process for the journal, we should check if we are processing the first edit, instead of checking if tcdate == tmdate. In some occasions, the review is edited soon after been posted and so the process function does not run.

celestemartinez:
- Run review process always for the first edit (fixes #2644)
- Run decision process always for first edit
- Limit the amount of times someone can volunteer to review a paper to 3 per month
- Create blocked group and add post process to check


openreview/journal/invitation.py
```diff
@@ -1046,6 +1046,10 @@ def set_submission_invitation(self):
             process=self.get_process_content('process/author_submission_process.py')
         )
 
+        existing_invitation = openreview.tools.get_invitation(self.client, submission_invitation_id)
+        if existing_invitation and existing_invitation.post_processes:
+            invitation.post_processes=existing_invitation.post_processes
```
melisabok, line 1058: why do you need this?

openreview/journal/process/review_process.py
```diff
@@ -19,7 +19,8 @@ def process(client, edit, invitation):
             client.post_edge(pending_review_edge)
 
     ## On update or delete return
-    if review_note.tcdate != review_note.tmdate:
+    review_edits = client.get_note_edits(note_id=review_note.id, sort='tcdate:asc')
```
melisabok: can you use count=1 to get the first one only?

openreview/journal/process/solicit_review_pre_process.py
```diff
@@ -1,2 +1,15 @@
...
+    solicit_note=edit.note
+    now = datetime.datetime.now()
+    cutoff = now - datetime.timedelta(days=30)
+
+    solicit_notes = client.get_notes(
+        parent_invitations=journal.get_solicit_review_id(),
+        signature=solicit_note.signatures[0],
+        mintcdate=openreview.tools.datetime_millis(cutoff))
+    
+    if len(solicit_notes) >= 3:
```
melisabok: I think this must be a setting of the journal. 

max_solicit_review: 3

if the setting doesn't exist then do not do the check

openreview/journal/process/submission_post_process.py
```diff
@@ -0,0 +1,36 @@
...
+    submission = client.get_note(edit.note.id)
+    submission_edits = client.get_note_edits(note_id=submission.id, sort='tcdate:asc')
+
+    # run post-process only if this is the first edit of the submission
+    if edit.id != submission_edits[0].id:
+        print('Submission edited, exit')
+        return
+    
+    # check if submission contains blocked author
+    blocked_authors_group_id = journal.get_blocked_authors_id()
```
melisabok: I think the name blocked authors can be confused with the profile blocked status that we have. What about calling the group: TMLR/Submission_Banned_Users that contains all the users that are banned to submit to TMLR.

openreview/journal/journal.py
```diff
@@ -395,6 +395,9 @@ def get_expert_reviewer_certification(self):
     def get_assignment_delay_after_submitted_review(self):
         return self.settings.get('assignment_delay_after_submitted_review', 0)
 
+    def get_blocked_authors_id(self):
+        return f'{self.venue_id}/Blocked_Authors'
```
melisabok: when is this group created? should we create it when the journal is deployed?

openreview/journal/process/submission_post_process.py
melisabok: if blocked_authors_group_id is a group that is empty or does not exist then we should skip the validation.

openreview/journal/journal.py
celestemartinez: I wasn't planning on adding it to the journal workflow itself, since I thought it was more of a TMLR only thing. I can add it as a setting and create the group when the journal is deployed if the setting exists.

openreview/journal/invitation.py
celestemartinez: I am adding a post process to the submission invitation, but I don't want it to get overwritten since I am only adding it for TMLR. Similar to [this comment](https://github.com/openreview/openreview-py/pull/2625#discussion_r2195752562)

melisabok (approved):

PR #2662, opened 2025-08-26
Add request_paper_similarity function

enrubio:
This PR adds the function `request_paper_similarity` which can be used to compute affinity scores between submissions.
- `venue_id` is the paper venue ID for venue A
- `alternate_venue_id` is the paper venue ID for venue B
- `invitation` is the invitation to retrieve papers for venue A
- `alternate_invitation` is the invitation to retrieve papers for venue B

If you are computing scores between papers of the same venue, you would pass the same value to both the entity A param and entity B param.

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -2639,16 +2639,20 @@ class OpenReviewClient(object):
 
         return response.json()
     
-    def request_paper_similarity(self, name, venue_id, alternate_venue_id, model=None, baseurl=None):
+    def request_paper_similarity(self, name, venue_id=None, alternate_venue_id=None, invitation=None, alternate_invitation=None, model=None, baseurl=None):
         """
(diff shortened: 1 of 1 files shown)


enrubio:
2 questions:
- How to add a test for this?
- If a PC were to use this, would they be able to compute scores against another venue?

melisabok:
> How to add a test for this?
call the expertise dev site or live site using this code (passing a base_url param)

> If a PC were to use this, would they be able to compute scores against another venue?

Only if they have reader permission on the other venue's submissions.

enrubio:
> call the expertise dev site or live site using this code (passing a base_url param)

Cool! Yeah I've been using it for the latest similarity score requests.

> Only if they have reader permission on the other venue's submissions.

That's good to know.

PR #2668, opened 2025-09-02
TMLR: J2C certification

celestemartinez:
Context: TMLR wants to add a J2C (Journal-to-Conference) Certification for papers with the following condition:
- AE score is 3 or more (from the decision field `recommendation_to_conference_track`)
- Average of AE + rev scores is 3 or more (from the recommendation and decision fields `recommendation_to_conference_track`)

This should also send an email about this certification to authors. 

Another change:
- TMLR wants the max number of active submissions for AE to be changes to 3. Since other journals also use this, I have added it as a setting that can be changed from the journal request form.

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -2557,6 +2557,7 @@ class OpenReviewClient(object):
                         venue_id, 
                         submission_content=None,
                         alternate_match_group = None, 
+                        alternate_expertise_selection_id = None,  
                         expertise_selection_id=None, 
                         model=None, 
                         baseurl=None, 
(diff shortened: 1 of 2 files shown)


openreview/journal/journal.py
```diff
@@ -403,6 +403,12 @@ def enable_blocked_authors(self):
 
     def get_blocked_authors_id(self):
         return f'{self.venue_id}/Submission_Banned_Users'
+    
+    def has_journal_to_conference_certification(self):
+        return self.settings.get('journal_to_conference_certification', False)
```
melisabok, line 408: what does it mean `journal_to_conference_certification`?

celestemartinez: this is just the name of this specific certification they want to add --> J2C Certification

openreview/journal/process/camera_ready_verification_process.py
```diff
@@ -3,6 +3,13 @@ def process(client, edit, invitation):
     journal = openreview.journal.Journal()
     venue_id = journal.venue_id
 
+    recommendation_mapping = {
```
celestemartinez: @melisabok is there a better place to keep this mapping?

melisabok (approved):

PR #2675, opened 2025-09-04
Feature/Timeline step to post LLM-generated review

celestemartinez:
To-dos:

- [x] create an overall AI_Reviewer group to sign the reviews
- [x] add subinvitations for model and prompt
- [x] add Carlos' code to call the LLM


openreview/workflows/templates.py
```diff
@@ -7355,3 +7356,213 @@ def setup_reviewers_review_days_late_template_invitation(self):
             signatures=['~Super_User1'],
             invitation=invitation
         )
+
+    def setup_llm_review_template_invitation(self):
```
melisabok: could you also change the name of the function and process file? let's use `LLM_PDF_Response`

PR #2685, opened 2025-09-11
LLM support: new module that let send a PDF as part of the context

melisabok:


openreview/llm/llm_gateway.py
```diff
@@ -0,0 +1,182 @@
...
+        Generate a completion using the Gemini model.
+        
+        Args:
+            messages: List of message dictionaries
+            **kwargs: Additional arguments to pass to litellm.completion
+            
+        Returns:
+            litellm completion response
+        """
+        import litellm  # Local import to avoid dependency if not used
```
melisabok: @carlosmondra this is the change to make the import run faster.

carlosmondra: Why did you place it here instead of the constructor? Just wondering.

melisabok: I thought about importing it when it is really needed it but I can move it to the constructor

PR #2686, opened 2025-09-12
ARR: Allow ACs/SACs to control comment participants

haroldrubio:
This PR adds a new invitation that allows ACs/SACs to modify the participants of the comment stage

### DIFF
--- a/tests/test_arr_venue_v2.py
+++ b/tests/test_arr_venue_v2.py
@@ -26,6 +26,8 @@ from openreview.stages.arr_content import (
     arr_ac_max_load_task,
     arr_sac_max_load_task
 )
+from openreview.workflows import EditInvitationsBuilder
+
 # API2 template from ICML
 class TestARRVenueV2():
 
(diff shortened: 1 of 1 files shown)


tests/test_arr_venue_v2.py
```diff
@@ -5588,6 +5590,111 @@ def test_emergency_reviewing_forms(self, client, openreview_client, helpers):
...
+        builder = EditInvitationsBuilder(
+            client,
+            'aclweb.org/ACL/ARR/2023/August'
+        )
+
+        builder.set_edit_reply_participants_invitation(
+            'aclweb.org/ACL/ARR/2023/August/Submission3/-/Official_Comment'
+        )
+
+        client.post_invitation_edit(
```
melisabok, line 5616: can you create this client using an AC or SAC? I don't think they have writer access to the comment invitation

openreview/workflows/edit_invitations.py
```diff
@@ -596,6 +596,236 @@ def set_edit_reply_readers_invitation(self, super_invitation_id, include_signatu
         
         self.save_invitation(invitation, replacement=False)
         return invitation
+    
+    def set_edit_reply_participants_invitation(self, super_invitation_id, include_signatures=True, due_date=None):
```
melisabok, line 600: I would rather add this to the ARR class to keep it simple. It looks like you need to call this for every submission so let's keep it as part of the ARR workflow. We do not have a clear way of doing this for any venue.

tests/test_arr_venue_v2.py
haroldrubio: Yeah I'll have to probably change how the comment invitation is built to include them as writers

openreview/workflows/edit_invitations.py
haroldrubio: Sure I can move it, the workflows were just a good template to start with

melisabok:
Thanks for the demo, I would prefer to show a link that takes them to the /invitation/edit page where it shows the edit button, we could automatically open the edit if the invitation is in the url, the same that we do for forum urls.

If you prefer to have it in the UI and this is a feature that ARR would use very frequent, I would add it to the ARR console component only (not the regular one) 

I also think the invitation should be simpler where they can add the reviewers or not and that value can be copied to the readers/signatures/invitees field.

haroldrubio:
> I would prefer to show a link that takes them to the /invitation/edit page

I'm confused on where the link would be because this link has to be built per submission. One option could be to have a list of /invitation/edit at the top of the consoles. The other would be to include the link in each of the paper rows. But in both cases it seems like there needs to be changes to openreview-web unless I'm missing something?

> I also think the invitation should be simpler where they can add the reviewers or not 

I would like a simpler interface as well, but is there a way to do this outside of a process function? The main blocker is that the edit signatures also needs to be set, and for reviewers/ACs we need to build a regex for the anon IDs.

melisabok:
1. We could set these invitations as pending tasks or show the link per submission in the PC console (yes, we would need to edit the UI)
2. I'm not asking to use a process function. The invitation can have a param in the edit.content then you can use the dollar sign substitution to complete the invitation invitees, signatures, readers, etc. Here is an example: https://openreview.net/invitation/edit?id=EurIPS.cc/2025/Workshop/DiffSys/-/Author_Rebuttal/Readers#invitationReply

melisabok:
You can keep the reviewers in the signatures field, it doesn't change anything. The only way to enable reviewers to make comments is adding them as invitees.

melisabok:
@haroldrubio can you update this issue as well?

Issue #2697, opened 2025-10-01
ARR: Remove Submission_Metadata_Revision from tasks


haroldrubio:
Remove this from the tasks page of the authors to prevent the authors from thinking this is required by preventing a duedate from being posted.

PR #2701, opened 2025-10-02
ARR: Remove metadata revision as a task

haroldrubio:
Resolves:
- #2697

- Adds an expiration date to the submission revision stage invitation
- Allows an empty string for deadline field
This is because the stage is read from the request form and not the reference, and previous submission revision stages may have a deadline
- Sets a pre-process to require at least the deadline or the expiration date

---
**Original Approach**
- Adds `get_active_venue_ids` function to the ARR class
- ARR class now calls the invitation builder instead of the `create_submission_revision_stage` function from the Venue class
- Uses the Venue class' invitation builder for all submission revision invitations except the `Submission_Metadata_Revision` invitation
- Duplicates logic from the Venue class invitation builder for the `Submission_Metadata_Revision` invitation but removes the duedate


openreview/arr/invitation.py
```diff
@@ -235,7 +235,147 @@ def set_desk_rejection_invitation(self):
         return self.venue_invitation_builder.set_desk_rejection_invitation()
 
     def set_submission_revision_invitation(self, submission_revision_stage=None):
-        return self.venue_invitation_builder.set_submission_revision_invitation(submission_revision_stage)
```
melisabok, line 238: can you just set a due_date = None in the SubmissionRevisionStage object instead of implementing a whole function?

haroldrubio: I don't think this works because they want to be able to expire it as well and only make it available for a small window after the submission deadline, if I set a None due_date, then it doesn't get an expdate:
https://github.com/openreview/openreview-py/blob/a492145532bd2cade21adc13f4af14ff4ccc9dd6/openreview/venue/invitation.py#L2777-L2778

melisabok: we can include an expdate in the stage class and put this logic there. I prefer to avoid the duplication of code.

openreview/venue_request/venue_request.py
```diff
@@ -746,10 +746,16 @@ def setup_submission_revision_stage(self):
             },
             'submission_revision_deadline': {
                 'description': 'By when should the authors finish revising submissions? Please enter a time and date in GMT using the following format: YYYY/MM/DD HH:MM (e.g. 2019/01/31 23:59) (Skip this if your venue does not have submission revisions)',
-                'value-regex': r'^[0-9]{4}\/([1-9]|0[1-9]|1[0-2])\/([1-9]|0[1-9]|[1-2][0-9]|3[0-1])(\s+)?((2[0-3]|[01][0-9]|[0-9]):[0-5][0-9])?(\s+)?$',
-                'required': True,
+                'value-regex': r'(^$)|(^[0-9]{4}\/([1-9]|0[1-9]|1[0-2])\/([1-9]|0[1-9]|[1-2][0-9]|3[0-1])(\s+)?((2[0-3]|[01][0-9]|[0-9]):[0-5][0-9])?(\s+)?$)',
```
melisabok, line 749: Let's keep duedate optional and expdate mandatory so no need to add a validation in the pre process.

All invitations should start having expiration dates otherwise they could be active forever leading to date process and process functions errors and keeping space in our queue.

PR #2702, opened 2025-10-02
Fix/new UI test

celestemartinez:
@melisabok it looks to me like the test was failing because the date process was trying to run twice, once at the invitation.edit.invitation.cdate and once at the invitation.mdate + 1000. I made the cdate in the past so now this should not happen again. Let me know if you think there might be something else and I can check again. 

This also increases the package version for when we are ready to deploy.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "openreview-py"
-version = "1.52.4"
+version = "1.52.5"
 description = "OpenReview API Python client library"
 authors = [{name = "OpenReview Team", email = "info@openreview.net"}]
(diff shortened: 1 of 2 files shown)


melisabok:
but this can happen when using the new UI too?

When you want to trigger the job, you usually set the cdate to now, right?

celestemartinez:
> but this can happen when using the new UI too?
> 
> When you want to trigger the job, you usually set the cdate to now, right?

Yes. I think we didn't use to have this issue because PCs had to input a date, but now they can choose now from the date picker and be more precise. I think not allowing dates in the past would fix this? Because the reason we have to add `#{4/mdate} + 1000` is in case the cdate is in the past, right?

PR #2706, opened 2025-10-03
ARR: Add custom withdrawal options

haroldrubio:
This PR allows ARR to input custom fields into the per-submission withdrawal invitations

### DIFF
--- a/openreview/stages/arr_content.py
+++ b/openreview/stages/arr_content.py
@@ -3327,3 +3327,47 @@ arr_submitted_author_content = {
         "order": 22
     }
 }
+
+arr_withdrawal_content = {
+    "withdrawal_confirmation": {
+        "order": 1,
+        "description": "Please confirm to withdraw.",
+        "value": {
+            "param": {
+                "type": "string",
+                "enum": [
+                    "I have read and agree with the venue's withdrawal policy on behalf of myself and my co-authors."
+                ],
+                "input": "checkbox"
+            }
+        }
+    },
+    "comment": {
+        "order": 2,
+        "description": "Add formatting using Markdown and formulas using LaTeX. For more information see https://openreview.net/faq.",
+        "value": {
+            "param": {
+                "type": "string",
+                "maxLength": 200000,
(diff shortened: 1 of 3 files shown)


openreview/arr/invitation.py
```diff
@@ -228,8 +228,8 @@ def set_public_comment_invitation(self):
     def set_decision_invitation(self):
         return self.venue_invitation_builder.set_decision_invitation()
 
-    def set_withdrawal_invitation(self):
-        return self.venue_invitation_builder.set_withdrawal_invitation()
+    def set_withdrawal_invitation(self, additional_fields=None):
+        return self.venue_invitation_builder.set_withdrawal_invitation(additional_fields=additional_fields)
```
melisabok: isn't there a chance to read `arr_withdrawal_content` here and no need to change `create_submission_stage`?

haroldrubio: The revision process function calls the `ARR.create_submission_stage` and from there it calls `Venue.create_submission_stage` and this call uses the Venue invitation builder. Should I set the venue's invitation builder to the ARR one? So it calls that function instead?

melisabok: I pushed a change where we set the additional field using the SubmisisonStage. Please take a look.

haroldrubio: Thanks! the changes make sense to me

openreview/stages/arr_content.py
```diff
@@ -3327,3 +3327,76 @@
...
+        },
+        "description": "If you withdraw more than 48 hours after submission deadline, you cannot resubmit this work in the next cycle.",
+        "order": 3
+    },
+    "confirm_restoring_rules": {
+        "value": {
+            "param": {
+                "input": "checkbox",
+                "enum": [
+                    "I understand that if I my withdrawn paper received even one review, this version has to be acknowledged."
```
celestemartinez:
```suggestion
                    "I understand that if my withdrawn paper received even one review, this version has to be acknowledged."
```

PR #2707, opened 2025-10-06
Venue: Compute dual submission metadata

enrubio:
This PR adds the function `compute_dual_submission_metadata` to the Venue class that creates a CSV file based on paper similarity scores. The CSV contains the paper IDs, scores, authors, titles, and abstracts. It takes:
- `alternate_venue`: The venue object of the venue you are computing against.
    - For a single venue, you would pass the same venue object.
    - For 2 different venues, you would pass the venue object of the other venue.
- `output_file_path`: The path to write the file to.
- `sparse_value`: The default is 5. This is a way to reduce the amount of data shared between venues and provide only meaningful scores. Also reduces the amount of data that the function is processing.
- `top_percent_cutoff`: The default is 1, meaning 1%. Same as above, this is to further filter the scores. This can be changed if we find that 1% isn't sufficient for venues.
- `job_id`: The job ID from the expertise. If none is passed, this function will compute similarity scores between active papers.
- `author_overlap_only`: Determines if the output file should contain only the cases where there is an overlap in the author list between papers. PCs may want this to help them sift through scores. However, cases where there is no overlap can help with finding fake profiles.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "openreview-py"
-version = "1.52.8"
+version = "1.52.9"
 description = "OpenReview API Python client library"
 authors = [{name = "OpenReview Team", email = "info@openreview.net"}]
 license = {text = "MIT"}


openreview/venue/venue.py
```diff
@@ -1906,3 +1906,130 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
                                 print(f'no profile active for {tail}')                                             
         
         return True
+
+def compute_similarity_scores_within_submissions(self, job_id, output_file_path, has_overlap=False):
+        short_name = self.short_name
+        venue_id = self.venue_id
+
+        print('Mapping paper IDs to author profile IDs...')
+        submissions = self.client.get_all_notes(invitation=f'{venue_id}/-/Submission')
```
melisabok: use `self.get_submission_id()` instead

openreview/venue/venue.py
```diff
@@ -1906,3 +1906,130 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+        
+        # Map note IDs to papers
+        paperid_map = {}
```
melisabok: you can do this in one line:

```
paperid_map = { s.id: s for s in submissions }
```

openreview/venue/venue.py
```diff
@@ -1906,3 +1906,130 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+        for note in submissions:
+            paperid_map[note.id] = note
+        
+        # Gather all author IDs
+        all_authors = set()
+
+        for note in submissions:
```
melisabok: can do two loops in a single one

openreview/venue/venue.py
```diff
@@ -1906,3 +1906,130 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+                if len(profiles) > 1:
+                    merge_profiles.append(flagged_id)
+
+        if merge_profiles:
+            raise openreview.OpenReviewException(f'Some author IDs link to multiple profiles. Please use the feedback form to contact us to investigate these profiles: {merge_profiles}')
+        
+        # Map papers to author profile IDs for that paper
+        paper_authorids_map = {}
+
+        for note in submissions:
```
melisabok: can you iterate paperid_map instead of submissions?

openreview/venue/venue.py
```diff
@@ -1906,3 +1908,160 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
                                 print(f'no profile active for {tail}')                                             
         
         return True
+
+    def compute_similarity_scores_within_submissions(self, output_file_path, sparse_value=5, job_id=None, has_overlap=False):
+        short_name = self.short_name
+
+        ## Getting paper/author data
+
+        submissions = self.client.get_all_notes(invitation=self.get_submission_id())
```
melisabok: let's use `self.get_submissions()` instead

openreview/venue/venue.py
```diff
@@ -1906,3 +1908,160 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+        all_author_profiles = openreview.tools.get_profiles(self.client, all_authors)
+        print(f'Retrieved {len(all_author_profiles)} author profiles')
+
+        # Map username to set of profile IDs (to detect duplicates)
+        username_to_id = {}
+
+        for profile in all_author_profiles:
+            names = [n.get('username') for n in profile.content.get('names', []) if n.get('username')]
+            for name in names:
+                username_to_id.setdefault(name, set()).add(profile.id)
```
melisabok: you can build the `flagged_profiles` her when adding the profile.id, you can check if the array is not empty

openreview/venue/venue.py
melisabok: it looks you need to get the submissions and profiles after you download the results from the expertise API so I would all this cade after getting the results in case the job failed or it has to wait hours to complete .

openreview/venue/venue.py
```diff
@@ -1906,3 +1908,160 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+        submissions = self.client.get_all_notes(invitation=self.get_submission_id())
+        print(f'Retrieved {len(submissions)} submissions')
+
+        paperid_map = { s.id: s for s in submissions }
+
+        all_authors = set()
+        for note in submissions:
+            all_authors.update(paperid_map[note.id].content['authorids']['value'])
+
+        all_author_profiles = openreview.tools.get_profiles(self.client, all_authors)
```
melisabok: you can use the param as_dict to get a dictionary where each key is a member of all_authors so no need to compute username_to_id

openreview/venue/venue.py
```diff
@@ -1906,3 +1908,160 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+                f'{short_name} abstract A',
+                f'{short_name} abstract B'
+            ])
+
+            for a_id, b_id, score in filtered_scores:
+
+                # Fetch metadata
+                a_title = paperid_map[a_id].content['title']['value']
+                a_abstract = paperid_map[a_id].content['abstract']['value'].replace("\n", "\\n")
+                a_authors_list = paper_authorids_map[a_id]
```
melisabok: having the paperid_map is enough. you can compute the author list as:

```
a_authors_list = [username_to_id[author_id] for author_id in paperid_map[a_id].content['authorids']['value']]
```

openreview/venue/venue.py
enrubio: It looks like using as_dict will map an id to only 1 profile, even if an id maps to multiple profiles

melisabok: an id shouldn't map to multiple profiles, if you find those cases, we need to fix them case by case. I don't think this script should be in charge of finding them.

openreview/venue/venue.py
```diff
@@ -1906,3 +1908,148 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+        print(f'Retrieved {len(all_author_profiles.keys())} author profiles')
+
+        ## Score filtering
+
+        # Keep top sparse_value scores per paper
+        print(f'Finding the top {sparse_value} scores per paper')
+        top_per_paper = {}
+        seen_pairs = set()
+
+        for r in results['results']:
```
melisabok, line 1942: what is this for loop for?

I understand the expertise results already returns the top 5 scores per submission. You only need to sort all the results and get the 1%

openreview/venue/venue.py
```diff
@@ -1906,3 +1908,148 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+        submissions = self.client.get_all_notes(invitation=self.get_submission_id())
+        print(f'Retrieved {len(submissions)} submissions')
+
+        papers_by_id = { s.id: s for s in submissions }
+
+        all_authors = set()
+        for note in submissions:
+            all_authors.update(papers_by_id[note.id].content['authorids']['value'])
+
+        all_author_profiles = openreview.tools.get_profiles(self.client, all_authors, as_dict=True)
```
melisabok: I recommend calling this variable `author_profile_by_id`

openreview/venue/venue.py
```diff
@@ -1906,3 +1908,148 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+        # Can be a lot of data. Stream to file so it's not stored in memory
+        results = self.client.get_expertise_results(job_id=job_id, wait_for_complete=True)
+        print('Sparse scores retrieved')
+
+        ## Getting paper/author data
+
+        submissions = self.client.get_all_notes(invitation=self.get_submission_id())
+        print(f'Retrieved {len(submissions)} submissions')
+
+        papers_by_id = { s.id: s for s in submissions }
```
melisabok: you iterating submissions twice, let's build papers_by_id and all_authors in the same loop

openreview/venue/venue.py
```diff
@@ -1906,3 +1908,148 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+                print('Filtering scores for overlapping author cases only')
+
+            for a_id, b_id, score in filtered_scores:
+
+                # Fetch metadata
+                a_title = papers_by_id[a_id].content['title']['value']
+                a_abstract = papers_by_id[a_id].content['abstract']['value'].replace("\n", "\\n")
+                # Use profile ID if available, otherwise use author ID in paper
+                a_authors_list = [
+                    all_author_profiles[a].id if all_author_profiles.get(a) else a
```
melisabok: you can do `all_author_profiles.get(a, Profile(id=a))`, return a if a is not in the dictionary instead of using an if,

openreview/venue/venue.py
enrubio: It looks like you can only pass `venueid` to `get_submissions()` and I want to get all papers in case the submission list changed between computing the scores and creating this CSV

openreview/venue/venue.py
enrubio: Discussed this and decided to keep calculating the top 5 scores per paper since the sparse scores returned a lot of data. Currently this uses heaps but the method can be changed.

openreview/venue/venue.py
enrubio: Cool, I didn't realize that

enrubio: It seems like when you call `get_profiles` with `as_dict=True`, if a profile is blocked/deleted the ID will show in the dict but will map to None. So I added the if back.

melisabok: in this case, I think we need to change the get_profiles function to always return a profile. This functions is being used for conflicts detection so if the profile is blocked/deleted, the conflict must be detected anyways, at least the profile.id must be there.

openreview/venue/venue.py
```diff
@@ -1906,3 +1908,177 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
...
+
+        papers_by_id_a = {s.id: s for s in submissions_a}
+        papers_by_id_b = papers_by_id_a if same_venue else {s.id: s for s in submissions_b}
+
+        all_authors = set()
+
+        for s in all_submissions:
+            all_authors.update(s.content['authorids']['value'])
+
+        author_profile_by_id = openreview.tools.get_profiles(self.client, all_authors, as_dict=True)
```
melisabok: You can delay the get_profiles call until you really needed it an it seems that you need it when you write the output file so I would move the call before that.

Also it means that you don't need to get the profiles for ALL the submissions, you just need the profiles for the submissions that are "detected" to be possible dual submissions.

melisabok: it looks like for get_all_notes is the same, you just need to get them when you need to output the file.

enrubio: That's true, this way it's not storing a giant dictionary of all the IDs. I pushed changes.

PR #2708, opened 2025-10-06
create tag invitations using templates for all the venues

melisabok:


openreview/workflows/edit_invitations.py
```diff
@@ -15,6 +15,10 @@ def __init__(self, client, venue_id, update_wait_time=5000):
         self.update_date_string = "#{4/mdate} + " + str(self.update_wait_time)
 
     def save_invitation(self, invitation, replacement=None):
+        if self.get_content_value('request_form_invitation') is None:
```
melisabok, line 18: @celestemartinez I'm adding this validation here to not create any edit invitations for venues that use the current venue request form. They won't have a UI to use these invitations so it is better to not create them.

openreview/venue/invitation.py
```diff
@@ -5052,4 +5052,127 @@ def set_edit_venue_group_invitations(self):
                     }
                 }
             )
-        )        
+        )
+
+    def set_venue_template_invitations(self):
+
+        super_id = 'openreview.net'
```
carlosmondra: Should this be 'OpenReview.net' for prod?

melisabok: thanks for the catch, let me fix it.

openreview/venue/invitation.py
```diff
@@ -5052,4 +5052,127 @@ def set_edit_venue_group_invitations(self):
...
                 }
             )
-        )        
+        )
+
+    def set_venue_template_invitations(self):
+
+        super_id = self.venue.support_user.split('/')[0] 
+        template_domain = f'{super_id}/Template'
+        submission_deadline = tools.datetime_millis(self.venue.submission_stage.exp_date if self.venue.submission_stage.exp_date else datetime.datetime.now(datetime.timezone.utc))
```
carlosmondra, line 5061: I don't understand why the submission deadline is set to now if there is no expiration date. What's the reasoning?

openreview/venue/invitation.py
```diff
@@ -5052,4 +5052,127 @@ def set_edit_venue_group_invitations(self):
...
+
+        self.client.post_invitation_edit(
+            invitations=f'{super_id}/-/Reviewers_Review_Count',
+            signatures=[template_domain],
+            content={
+                'venue_id': {'value': self.venue_id},
+                'reviewers_id': {'value': self.venue.get_reviewers_id() },
+                'activation_date': { 'value': submission_deadline + (60*60*1000*24*7*8) },
+            },
+            await_process=True
```
carlosmondra, line 5072: Why do we need to wait for the process?

openreview/venue/invitation.py
```diff
@@ -5052,4 +5052,127 @@ def set_edit_venue_group_invitations(self):
...
+            )
+
+        if self.venue.use_publication_chairs:
+            self.client.post_invitation_edit(
+                invitations=f'{super_id}/-/Publication_Chair_Role',
+                signatures=[template_domain],
+                content={
+                    'venue_id': {'value': self.venue_id},
+                    'committee_name': {'value': tools.singularize(self.venue.publication_chairs_name) },
+                    'activation_date': { 'value': submission_deadline + (60*60*1000*24*7*8) },
```
carlosmondra: Nitpick: You could extract submission_deadline + (60*60*1000*24*7*8) since it the same everywhere. I wonder if we should parametrize this in the function call, but maybe it doesn't make sense.

openreview/workflows/process/ethics_chair_role_process.py
```diff
@@ -0,0 +1,28 @@
+def process(client, invitation):
+
+    print('Compute roles for ethic chairs', invitation.id)
+    domain = client.get_group(invitation.domain)
+    ethics_chairs_id = domain.content.get('ethics_chairs_id', {}).get('value')
+
+    ethics_chairs_group = client.get_group(ethics_chairs_id)
```
carlosmondra: Should we return in case `ethics_chairs_id` is `None`? Otherwise, I think the process function will fail with `client.get_group`

openreview/workflows/process/ethics_reviewer_role_process.py
```diff
@@ -0,0 +1,41 @@
...
+        if profile.id.startswith('~') and profile.id not in tags_by_profile:
+            tags_by_profile[profile.id] = openreview.api.Tag(
+                invitation=invitation.id,
+                signature=domain.id,
+                profile=profile.id
+            )
+    
+    print('Post profile tags', len(tags_by_profile))
+    
+    client.delete_tags(invitation=invitation.id, wait_to_finish=True, soft_delete=True)
```
carlosmondra: I am not sure if we should soft_delete them. I would just delete them completely.

openreview/venue/invitation.py
melisabok: I got it from the new configuration UI where we set the deadlines to show in the timeline.

We should assume the submission has an expiration date but if not we do now + 8 weeks. 

Which date would you use? these tags need to be created when the conference decisions are posted, PCs could run them at any time by using the new configuration UI

openreview/venue/invitation.py
melisabok: because this edit runs a process and this post edit is running inside of a process function (venue deployment). So if the process function fails when I wan the deployment process function to also fail.

openreview/workflows/process/ethics_chair_role_process.py
melisabok: Good point, I can add it but I'm assuming that this invitation is only created when the ethics_chairs_id exists.

openreview/workflows/process/reviewer_role_process.py
```diff
@@ -0,0 +1,41 @@
...
+
+    for profile_id, profile in all_profiles.items():
+        if profile.id.startswith('~') and profile.id not in tags_by_profile:
+            tags_by_profile[profile.id] = openreview.api.Tag(
+                invitation=invitation.id,
+                signature=domain.id,
+                profile=profile.id
+            )
+    
+    print('Post profile tags', len(tags_by_profile))
```
carlosmondra, line 68: My only concern with this approach is the amount of memory we need to run this process function. Hopefully we won't face any issues. Just something to keep in mind.

melisabok: the concern is loading ALL reviewers profiles?

We already do this in the setup matching to compute conflicts. 

This invitation is meant to run only once BTW.

I'm open to other suggestions.

openreview/venue/invitation.py
carlosmondra: I see, but I don't see the +8 weeks there. Am I missing something?

openreview/workflows/process/reviewer_role_process.py
carlosmondra: > the concern is loading ALL reviewers profiles?

Yes

> I'm open to other suggestions.

I think it's fine for now. If we see memory issues we could post the tags in batches.

openreview/venue/invitation.py
carlosmondra: 😮 Interesting and a bit scary haha. OK, got it.

openreview/venue/invitation.py
melisabok: line 5069: `'activation_date': { 'value': submission_deadline + (60*60*1000*24*7*8) },`

this is the date when the tags will be created.

openreview/venue/invitation.py
melisabok: the new configuration UI is creating invitations with process functions that create other invitations. I want to root process function to fail is some of their children fail. 

Here is an example:

1. deployment process
1.1. create submission invitation
1.1.1. create submission/dates invitation to edit the dates
1.2 create review invitation
1.2.1 create review/dates invitation to edit the dates
...

openreview/venue/invitation.py
melisabok: done

openreview/workflows/process/ethics_chair_role_process.py
melisabok: done

openreview/workflows/process/ethics_reviewer_role_process.py
melisabok: Done.

openreview/venue/invitation.py
carlosmondra: Oh I understand now. I thought we were assigning now to the submission deadline.

openreview/workflows/templates.py
```diff
@@ -7357,6 +7358,111 @@ def setup_reviewers_review_days_late_template_invitation(self):
             invitation=invitation
         )
 
+    def setup_committee_roles_invitations(self):
+
+        def create_role_invitation(role):
+            invitation = Invitation(id=f'{self.super_id}/-/{role}_Role',
```
melisabok, line 7388: I'm thinking to create a super invitation here so all the role invitations are created using the same invitation so we can query all the roles at once.

melisabok: I prefer to keep the code as it is. This is ready.

openreview/tools.py
```diff
@@ -2139,4 +2139,13 @@ def create_forum_invitations(client, submission):
                         invitation=openreview.api.Invitation(id=forum_invitation.id,
                             ddate=openreview.tools.datetime_millis(datetime.datetime.now())
                         )            
-                    )
\ No newline at end of file
+                    )
+
+def singularize(word):
+    if word.endswith('ies'):
```
xkopenreview, line 2145: which role ends with ies?

melisabok: PCs can decide the name of the roles so I added it just in case.

xkopenreview (approved):

PR #2711, opened 2025-10-07
Feat/ Venue UI: Support ACs

celestemartinez:


openreview/workflows/workflow_process/deploy_reviewers_only_process.py
```diff
@@ -6,6 +6,10 @@ def process(client, edit, invitation):
     note = client.get_note(edit.note.id)
     venue_id = edit.note.content['venue_id']['value']
     reviewers_name = note.content['reviewers_name']['value']
+    area_chairs_name = note.content.get('area_chairs_name', {}).get('value')
+    area_chairs_id = f'{venue_id}/{area_chairs_name}' if area_chairs_name else None
+    paper_area_chairs_id_number = venue_id + '/Submission${{2/id}/number}/' + area_chairs_name if area_chairs_name else None
```
melisabok: This is not used anywhere

openreview/workflows/workflow_process/deploy_reviewers_only_process.py
```diff
@@ -6,6 +6,10 @@ def process(client, edit, invitation):
...
+    paper_area_chairs_id_note_number = venue_id + '/Submission${5/content/noteNumber/value}/' + area_chairs_name if area_chairs_name else None
```
melisabok: I wish you can compute this value using the venue class, `venue.get_area_chairs_id('${5/content/noteNumber/value}')`

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -39,18 +52,24 @@ def process(client, edit, invitation):
         end_date=submission_duedate + datetime.timedelta(weeks=6),
         reader_selection=True,
         check_mandatory_readers=True,
-        readers=[openreview.stages.CommentStage.Readers.REVIEWERS_ASSIGNED, openreview.stages.CommentStage.Readers.AUTHORS],
-        invitees=[openreview.stages.CommentStage.Readers.REVIEWERS_ASSIGNED, openreview.stages.CommentStage.Readers.AUTHORS]
+        readers=[openreview.stages.CommentStage.Readers.AREA_CHAIRS_ASSIGNED, openreview.stages.CommentStage.Readers.REVIEWERS_ASSIGNED, openreview.stages.CommentStage.Readers.AUTHORS],
```
melisabok, line 53: can this be default values based on the venue class when calling to `stage.get_readers`?

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -81,6 +100,37 @@ def process(client, edit, invitation):
 
     venue.create_submission_change_invitation(name='Submission_Change_Before_Bidding', activation_date=note.content['submission_deadline']['value'] + (30*60*1000))
 
+    # AC conflict and affinity score invitations
+    if area_chairs_name:
+        client.post_invitation_edit(
+            invitations=f'{invitation_prefix}/-/Reviewer_Conflict',
```
melisabok: Is this the name of the template? could it be called `Committee_Conflict`?

What is we call to the matching module to create the conflict invitation instead?

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -81,6 +100,37 @@ def process(client, edit, invitation):
...
+                'activation_date': { 'value': note.content['submission_deadline']['value'] + (60*60*1000*24*3) },
+                'submission_name': { 'value': 'Submission' },
+                'committee_name': { 'value': area_chairs_name },
+                'committee_role': { 'value': 'area_chairs' },
+                'committee_pretty_name': { 'value': venue.get_committee_name(area_chairs_name, pretty=True) }
+            },
+            await_process=True
+        )
+
+        client.post_invitation_edit(
```
melisabok: Same here, I would do:

```
venue.setup_matching_invitation()
```

the implementation would call to the matching module to create either the affinity score or conflict invitations

melisabok:
deploy_reviewers_only_process.py should be renamed to a different name. Something like "conference_review_workflow_deployment.py"

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
celestemartinez: I think we had spoken about doing this for the comment stage, and we ended up deciding not to do it but I can't remember why. I know there was an issue with the ethics stage: https://github.com/openreview/openreview-py/pull/2740

We can discuss this more if needed.

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
celestemartinez: I think we had done it this way because the Conflict and Affinity_Score invitations now have a date process that computes the edges. So if I create it in the venue class, I would then add code here too to add the date process?

melisabok: I think we can add a new function in the matching module to create these invitations that add the date process. It doesn't need to be the same code that we use for setup_matching

melisabok:
I would like to merge this and deploy soon. We should change the support page saying that we now support ACs.

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -21,6 +21,10 @@ def process(client, edit, invitation):
         double_blind=True
     )
 
+    area_chairs_name = note.content.get('area_chairs_name', {}).get('value')
```
melisabok: I prefer reading the settings from the venue class than reading the configuration note. The venue instance should have the property `use_area_chairs` set already.

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -29,6 +33,15 @@ def process(client, edit, invitation):
         )
     ]
 
+    if area_chairs_name:
```
melisabok:
```suggestion
    if venue.user_area_chairs:
```

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -29,6 +33,15 @@ def process(client, edit, invitation):
...
+        venue.bid_stages.append(
+            openreview.stages.BidStage(
+                area_chairs_id,
```
melisabok:
```suggestion
                venue.get_area_chairs_id(),
```

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -21,6 +21,10 @@ def process(client, edit, invitation):
         double_blind=True
     )
 
+    area_chairs_name = note.content.get('area_chairs_name', {}).get('value')
+    area_chairs_id = venue.get_area_chairs_id() if area_chairs_name else None
+    paper_area_chairs_id_note_number = venue.get_area_chairs_id(number='${5/content/noteNumber/value}') if area_chairs_name else None
```
melisabok: this variable can be set when it is needed

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -81,32 +100,11 @@ def process(client, edit, invitation):
...
-            'name': { 'value': 'Conflict' },
-            'activation_date': { 'value': note.content['submission_deadline']['value'] + (60*60*1000*24*3) },
-            'submission_name': { 'value': 'Submission' },
-            'reviewers_name': { 'value': reviewers_name }
-        },
-        await_process=True
-    )
+    # AC conflict and affinity score invitations
+    if area_chairs_name:
+        venue.setup_matching_invitations(area_chairs_id)
```
melisabok: maybe setup_matching_invitations can create the invitations for all the roles and do not call it for each role individually

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -139,6 +137,7 @@ def process(client, edit, invitation):
             'stage_name': { 'value': 'Official_Review' },
             'reviewers_name': { 'value': reviewers_name },
             'authors_name': { 'value': authors_name },
+            'additional_readers': { 'value': [paper_area_chairs_id_note_number] if area_chairs_name else [] },
```
melisabok:
```suggestion
            'additional_readers': { 'value': [venue.get_area_chairs_id(number='${5/content/noteNumber/value}') if venue.use_area_chairs else [] },
```

tests/test_reviewers_only.py
```diff
@@ -210,7 +215,7 @@ def test_setup(self, openreview_client, helpers):
         assert openreview_client.get_invitation('ABCD.cc/2025/Conference/-/Author_Decision_Notification')
 
         invitation = openreview_client.get_invitation('ABCD.cc/2025/Conference/Program_Committee/-/Submission_Group')
-        assert invitation and invitation.edit['group']['deanonymizers'] == ['ABCD.cc/2025/Conference', 'ABCD.cc/2025/Conference/Submission${3/content/noteNumber/value}/Program_Committee']
+        assert invitation and invitation.edit['group']['deanonymizers'] == ['ABCD.cc/2025/Conference']
```
melisabok, line 218: why is this assertion different?

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
celestemartinez: I am not sure we should do this, since this will return the default AC group id even when the venue has no ACs. This is why I check `area_chairs_id = venue.get_area_chairs_id() if venue.use_area_chairs else None`

celestemartinez: oh never mind, these are inside `if venue.use_area_chairs:`, so it is fine.

tests/test_reviewers_only.py
celestemartinez: I am setting the reviewer deanonymizers to be only the PCs by default (this is the default in the old request form), so I removed this line:
`self.reviewer_identity_readers = [openreview.stages.IdentityReaders.REVIEWERS_ASSIGNED]`

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
celestemartinez: Fixed.

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
celestemartinez: Fixed.

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
celestemartinez: Fixed.

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
celestemartinez: Fixed.

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
celestemartinez: Fixed.

Issue #2693, opened 2025-09-30
Revision History: manage visibility


racheljsmart:
A PC has requested:
`Could we please make sure that the Revision History is closed off to reviewers and area chairs? We only want PC chairs to be able to see the history.`

Currently the edits are determined by the readers of the note, and changing the readers of the edits can be challenging.

The desired feature would either be to hide the revisions from the committee completely (except PCs and authors) or be able to manage the readers of the edits more easily.

PR #2713, opened 2025-10-08
Fix/hide submission revision history

melisabok:
Fixes https://github.com/openreview/openreview-py/issues/2693

Make sure the submissions edits are only visible to the PCs + authors unless the PCs specifies a different setting.

This can fix several issues happened in the past where edits were released to ACs + reviewers and they fields were not correctly hidden.

### DIFF
--- a/openreview/conference/helpers.py
+++ b/openreview/conference/helpers.py
@@ -1026,10 +1026,7 @@ def get_submission_revision_stage(request_forum):
     if request_forum.content.get('api_version', '1') == '1':
         allow_author_reorder = request_forum.content.get('submission_author_edition', '') == 'Allow reorder of existing authors only'
 
-    submission_revision_history_readers = request_forum.content.get('submission_revision_history_readers')
-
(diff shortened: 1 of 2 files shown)


openreview/conference/helpers.py
```diff
@@ -1026,6 +1026,11 @@ def get_submission_revision_stage(request_forum):
     if request_forum.content.get('api_version', '1') == '1':
         allow_author_reorder = request_forum.content.get('submission_author_edition', '') == 'Allow reorder of existing authors only'
 
+    submission_revision_history_readers = request_forum.content.get('submission_revision_history_readers')
+
+    if submission_revision_history_readers and submission_revision_history_readers == 'Submission revision history should be visible to all the current submission readers':
```
celestemartinez: There's an issue here if you select `Submission revision history should be visible to venue organizers and submission authors only` in the Revision Stage from the request form, since we are only checking if they selected the same readers of the note and set `submission_revision_history_readers` accordingly. 

We should either set `submission_revision_history_readers` to None so we use the default value or directly set it to be venue_id and authors

melisabok: I see, let me reproduce it with a test.

PR #2727, opened 2025-10-16
use function to create random hash seed

melisabok:


carlosmondra:
> > I looked for other 1234 values in the repo and was able to find some for the hash_seed. Is that OK?
> 
> tests and conference related scripts should be ignored. Is there another file that you see 1234?

I found one more in builder.py, in the `recruit_reviewers` function

carlosmondra (approved):

PR #2734, opened 2025-10-23
Tests: Add ethics chair author comment

enrubio:
This PR adds tests for a comment invitation between Ethics Chairs and Authors. It allows Ethics Chairs to comment to each author individually and authors sign with their tilde ID. There are checks in the preprocess to make sure PCs or Ethics Chairs don't sign with their own tilde ID.

**The UI bug**:
Currently, if the readers of a reply does not contain a subset of the readers in the parent note, then the button will not show. We see this most often with the rebuttal when it's in reply to a review.

In this custom comment invitation, the `replyto` is set using `withForum` so the button appears for each reply. However, the readers are specified using `items` where PCs and Ethics Chairs are mandatory readers. 

What I expect is that if PCs or Ethics Chairs are not readers of the note, the button should not show. But it does - see below:
comment

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "openreview-py"
-version = "1.53.3"
+version = "1.54.0"
 description = "OpenReview API Python client library"
 authors = [{name = "OpenReview Team", email = "info@openreview.net"}]
 license = {text = "MIT"}


melisabok:
I see, why Emilia the readers invitation is a param instead of a const value?

If this is an invitation to talk between Authors and Ethics_Chairs, then the readers must be fixed.

xkopenreview:
> Xukun, we can still validate this and check if that every element of items must be in the parent note readers.

if it's value and optional false it's ok
i think it's not straightforward if invitation specified inGroup or prefix

enrubio:
> I see, why Emilia the readers invitation is a param instead of a const value?

This was because the Ethics Chairs wanted to select which author to message. They wanted to message each author individually

PR #2740, opened 2025-10-29
Feat/ Venue Stages: add default invitees/readers

celestemartinez:


openreview/venue/venue.py
```diff
@@ -685,6 +685,12 @@ def create_ethics_review_stage(self):
         self.invitation_builder.set_ethics_review_invitation()
         if self.ethics_review_stage.enable_comments:
             print('Setting up ethics review comments invitation')
+            if not self.comment_stage.invitees: # this means the comment stage has not been created from the request form
```
melisabok, line 688: is this the hack that I need to take a look?

celestemartinez: yes. Since we now add SACs and ACs as default when invitees is empty, they were being added in the case that the comment stage was being created for the first time with the ethics review stage. According to the tests, only ethics chairs and ethics reviewers should be invitees here since the Comment Stage hasn't been run from the request form.

melisabok: now I'm realizing that we might not want to enable comments by default in the new configuration. 

How is the default config? if the venue has reviewers and ACs then comment stage is enabled to PCs and ACs? but not reviewers?

maybe the test for the ethics comments should use another invitation name than Official_Comment.

celestemartinez: How it works now, if no readers are selected then we only add PCs as invitees and readers. This was changed in this PR to add assigned SACs, assigned ACs and assigned reviewers as default if no readers/invitees are passed. 

I believe in the new UI, we are automatically adding reviewers and authors as invitees/readers.

PR #2748, opened 2025-11-04
Fix/performance fixes

melisabok:
- Create tag role and endorsement invitations only when the venue is deployed (to not interfere on running venues)
- Activate tag role invitation after 20 weeks instead of 8 to make sure review process is finished
- Remove maxReplies from source invitations (DBLP, arXiv, ORCID)
- Add responseArchiveDate to preferred email invitations so they are being archived too
- Fix `profile is None` on stats process functions
- Move all the date process functions to parent invitations so the process code is in a single place
- Delete edges for archived papers in journals

### DIFF
--- a/openreview/arr/invitation.py
+++ b/openreview/arr/invitation.py
@@ -272,4 +272,7 @@ class InvitationBuilder(object):
 
     def set_reviewer_recommendation_invitation(self, start_date, due_date, total_recommendations):
         return self.venue_invitation_builder.set_reviewer_recommendation_invitation(start_date,  due_date,  total_recommendations)
+    
+    def set_venue_template_invitations(self):
+        return self.venue_invitation_builder.set_venue_template_invitations()
 


openreview/venue_request/process/deployProcess.py
```diff
@@ -5,6 +5,7 @@ def process(client, note, invitation):
     SUPPORT_GROUP = GROUP_PREFIX + '/Support'
     conference = openreview.helpers.get_conference(client, note.forum, SUPPORT_GROUP, setup=True)
     conference.create_submission_stage()
+    conference.invitation_builder.set_venue_template_invitations()
```
xkopenreview, line 8: where does invitation_builder has set_venue_template_invitations method

i saw self.invitation_builder = invitation.InvitationBuilder(client) in builder.py
but in invitation.py there's no set_venue_template_invitations

melisabok: it is here: https://github.com/openreview/openreview-py/blob/master/openreview/venue/invitation.py#L5129

PR #2750, opened 2025-11-05
ARR: Allow ARR PCs to Post Registration Forms

haroldrubio:
This PR is to allow the ARR organizers to post registration data on behalf of their users. This means the following changes need to be made:

1. The `profile_id` field needs to be added to all forms ARR expects to be able to post on behalf of the user (tentatively hidden + optional). The user should not be asked to manually input their profile ID again since they expect us to already have this given their signature (which is true but difficult if the signature is the venue/PC ID)
2. The `profile_id` must be validated in a pre-process, if present (maybe this can fixed by making type `profile`)
3. A second edit must be posted to notes posted by the organizers to allow the user to view and edit this form
4. (Not Implemented Yet) Whenever building edges from registration notes, read from either the signature or the `profile_id` field if the `signatures[0]` is not a tilde ID

### DIFF
--- a/openreview/arr/helpers.py
+++ b/openreview/arr/helpers.py
@@ -36,7 +36,11 @@ from openreview.stages.arr_content import (
     arr_metareview_rating_content,
     hide_fields_from_public,
     arr_submitted_author_forum,
-    arr_submitted_author_content
+    arr_submitted_author_content,
+    arr_delay_notification_content,
+    arr_emergency_declaration_content,
+    arr_great_or_irresponsible_ac_content,
+    arr_great_or_irresponsible_reviewer_content
 )
 
 from openreview.stages.default_content import comment_v2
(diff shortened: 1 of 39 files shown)


melisabok:
The issue with this approach is that the user won't see this task a complete because the signature is the PC and not the tilde id.

haroldrubio:
I could try the approach of using a single registration invitation for each - though if I recall from a while back, they were hesitant to do so for some reason.

I think the main things from the ARR perspective are:
1) Authors register to participate in ARR reviewing (reviewer or AC) but to be accepted as a reviewer or AC, ARR has their own scripts to verify whether an author is qualified for those roles (looking at their publication history in the ACL anthology, etc. - sources outside of OpenReview)

2) From a user experience perspective, they submit all of the information necessary to register them as a reviewer/AC in the author form. It would be redundant to ask them to fill out the same information again. I guess they could reduce the size of the author registration form and make them fill out the other forms themselves but that's more work on everyone else's part - the authors have a second deadline to keep in mind and the EiCs will need to send out notifications all while trying to turn around the matching quick enough to start reviewing

haroldrubio:
Here is what I'm thinking of sharing with the tech team/EiCS for the workflow changes:

1. Reviewer/Area Chair forms will be shared per cycle - registration, maximum load, license agreement, emergency reviewing
  - At any time, users can modify the values
  - To check historical data, you will need to read the edit history and parse the creation dates of certain edits to see if it lines up with the cycle
  - If a reviewer wants to indicate that they are unavailable this cycle, either (1) they have to explicitly set their form to 0 or (2) we zero out all maximum load requests per cycle to ask for their availability again
3. We can make this change transparent to the users - links to these forms will still be in their consoles and the UI will be the same
4. After filtering eligible authors from the available authors, you can add them to both the aclweb.org/ACL/ARR/Reviewers group and the aclweb.org/ACL/ARR/Year/Month/Reviewers group so they get access to that cycle’s console, as well as the shared form responses
5. For the cycle that they are added in, we can build their matching data directly from the submitted author form - but if they want their data to carry on to the next cycle, they have to fill out the forms themselves

PR #2770, opened 2025-11-20
ARR: Hide Explanation of Revisions PDF from Reviewers

haroldrubio:
This PR hides the `explanation_of_revisions_PDF` from all submission reviewers, and only reveals it to:
1) Submitted Reviewers
2) Previous Reviewers

- Adjusts the readership of this field in the preprint release edit
- Populates and builds the previous reviewer group during the setup of the matching data

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -1778,7 +1778,7 @@ class OpenReviewClient(object):
 
         return tools.concurrent_get(self, self.get_tags, **params)
 
-    def get_edges(self, id = None, invitation = None, head = None, tail = None, label = None, limit = None, offset = None, with_count=None, trash=None):
+    def get_edges(self, id = None, invitation = None, head = None, tail = None, label = None, limit = None, offset = None, with_count=None, trash=None, select=None):
         """
         Returns a list of Edge objects based on the filters provided.
 
@@ -1798,6 +1798,8 @@ class OpenReviewClient(object):
         params['limit'] = limit
         params['offset'] = offset
         params['trash'] = trash
+        if select is not None:
+            params['select'] = select
         if with_count is not None:
             params['count'] = with_count
 


openreview/arr/helpers.py
```diff
@@ -415,7 +415,20 @@ def _build_preprint_release_edit(client, venue, builder, request_form):
 
         hidden_field_names = hide_fields_from_public
         committee_members = venue.get_committee(number='${{4/id}/number}', with_authors=True)
+        explanation_field = 'explanation_of_revisions_PDF'
         note_content = { f: { 'readers': committee_members } for f in hidden_field_names }
+        
+        # Set custom readers for explanation_of_revisions_PDF to include Reviewers/Previous and Reviewers/Submitted
+        if explanation_field in note_content:
+            explanation_readers = [
+                venue_id + '/Program_Chairs',
```
melisabok: use `venue.get_program_chairs_id()`

openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -233,6 +234,31 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
                 members=[]
             )
         )
+        
+        # Create Reviewers/Previous group for resubmissions
+        if submission.id in resubmission_ids:
+            reviewers_previous_id = venue.get_reviewers_id(number=submission.number) + '/Previous'
```
melisabok: instead of using a new group, can you use the group of the previous submission from a previous cycle?

haroldrubio: https://github.com/openreview/openreview-py/blob/0e95bb100ebe97fd9706c625591e3df561727374/openreview/arr/helpers.py#L417-L431

I wouldn't be able to do this for the anonymous pre-print edit if I use the previous reviewers submitted directly, so I would need to handle this some other way

melisabok: what is an anonymous preprint edit?

haroldrubio: When authors opt-in for an anonymous pre-print, so to make their submission public, they set `preprint` to `True`. There's an invitation where the dateprocess posts edits using the `Preprint_Release_Submission` to submissions with `preprint: True` that sets the readers of the note to `everyone` and sets readers for all the fields that need to be hidden from the public

openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -302,6 +303,9 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
             for g in current_client.get_grouped_edges(invitation=rev_affinity_inv, head=submission.id, select='tail,id,weight', groupby='tail')
         }
         
+        # Collect previous reviewers' tilde IDs for Reviewers/Previous group
+        previous_reviewer_tilde_ids = []
```
melisabok: I don't think you need to compute this anymore.

openreview/arr/helpers.py
```diff
@@ -415,7 +415,8 @@ def _build_preprint_release_edit(client, venue, builder, request_form):
 
         hidden_field_names = hide_fields_from_public
         committee_members = venue.get_committee(number='${{4/id}/number}', with_authors=True)
-        note_content = { f: { 'readers': committee_members } for f in hidden_field_names }
+        explanation_field = 'explanation_of_revisions_PDF'
```
melisabok: it looks like this change is not needed.

openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -158,6 +158,7 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
     submissions = venue.get_submissions()
 
     resubmissions = get_resubmissions(submissions, previous_url_field)
+    resubmission_ids = {s.id for s in resubmissions}
```
melisabok: this change is not needed

openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -371,6 +372,33 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
...
+        if 'explanation_of_revisions_PDF' in submission.content:
+            explanation_readers = [
+                venue_id + '/Program_Chairs',
+                venue.get_senior_area_chairs_id(number=submission.number),
+                venue.get_area_chairs_id(number=submission.number),
+                venue.get_reviewers_id(number=submission.number, submitted=True),
+                venue.get_authors_id(number=submission.number)
+            ]
+            if not wants_new_reviewers:
+                explanation_readers.append(previous_reviewers.id)
```
melisabok, line 385: I don't understand this, if the submission doesn't ask for new reviewers, aren't the current cycle reviewers the same as the previous reviewers? why do you need to add them as readers?

openreview/arr/process/preprint_release_submission_process.py
```diff
@@ -37,6 +51,46 @@ def post_submission_edit(submission):
...
+                    ]
+                }
+            }
+            paper_forum = paper_link.split('?id=')[-1]
+            arr_submission_v1 = openreview.tools.get_note(client_v1, paper_forum)
+            arr_submission_v2 = openreview.tools.get_note(client, paper_forum)
+            
+            if arr_submission_v1:
+                v1_domain = arr_submission_v1.invitation.split('/-/')[0]
+                content['explanation_of_revisions_PDF']['readers'].append(
```
melisabok: don't you need to check if the submission asked for new reviewers or not?

openreview/arr/process/preprint_release_submission_process.py
```diff
@@ -37,6 +51,46 @@ def post_submission_edit(submission):
             note=updated_note,
             signatures=[venue_id]
         )
+
+        paper_link = submission.content.get('previous_URL', {}).get('value')
+        # If previous submission, change reader set to include previous reviewers submitted group
+        if paper_link:
+            content = {
+                'explanation_of_revisions_PDF': {
+                    'readers': [
```
melisabok: instead of building the whole readers array, you can use the `append` keyword in the edit

openreview/arr/management/setup_reviewer_matching.py
haroldrubio: The reasoning is that if the previous reviewers are the current cycle reviewers, then they shouldn't need to wait to submit a review to see this field.

Otherwise, only the submitted reviewers are allowed to see this field

melisabok: instead of adding the previous reviewers, why don't you decide between reviewers and reviewers/submitted as readers?

haroldrubio: It can be the case that not all previous reviewers are available for the current cycle. For examples, 2/3 could be previous reviewers and 1/3 would be a new reviewer - but only those 2/3 should be able to see this field

openreview/arr/process/preprint_release_submission_process.py
haroldrubio: Added check in recent commit

openreview/arr/process/preprint_release_submission_process.py
haroldrubio: Using append and delete to replace the reviewers with reviewers submitted group

melisabok (approved):

openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -371,6 +371,33 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
                 profile_id=reviewer_id,
                 edge_readers=[venue_id, senior_area_chairs_id, area_chairs_id, reviewer_id]
             )
+        
+        # Add previous reviewers to explanation_of_revisions_PDF readers
+        if 'explanation_of_revisions_PDF' in submission.content:
+            explanation_readers = [
+                venue_id + '/Program_Chairs',
```
celestemartinez: can you use `venue.get_program_chairs_id()` here?

celestemartinez (approved):

PR #2771, opened 2025-11-22
Update `emergency_load` value to be a radio input instead of a text field in arr_content.py.

holylovenia:
Update `emergency_load` value to be a radio input instead of a text field in arr_content.py.


melisabok:
@haroldrubio there are several tests that use a value higher than 4 for emergency_load, can you help Holy to update them?

PR #2773, opened 2025-11-24
Reviewer Recommendation: let SACs of assigned submissions to recommend reviewers

melisabok:
This PR fixes the Recommendation invitation to include the SACs as readers, invitees, signatures, edge readers and writers. 

The invitation configuration is the same than the Assignment invitation where SACs can change the assigned reviewers of a submission.

There is one thing missing that could be manually adjusted for CVPR: SACs need to have access to the edge browser link to recommend reviewers. 

Option 1: Remove the `start` parameter of the button link so the first column in the edge browser show the submission visible to the logged in users instead of their assigned papers because this doesn't apply to SACs.
Screenshot 2025-11-24 at 1 44 30 PM

Option 2: Modify the edge browser url in the "Area Chair Status" tab to go to the recommendation browser instead of assignment browser. This is very easy to change in the webfield config
Screenshot 2025-11-24 at 1 45 57 PM

Option 3: Show the edge browser url in the SAC console header, this as option 1 will rely on the readers permission of the submission to load the first column causing that SACs can see their own submissions.

I think the best approach is option 2.

### DIFF
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -4198,9 +4198,10 @@ To view your submission, click here: https://openreview.net/forum?id={{{{note_fo
             cdate=tools.datetime_millis(start_date) if start_date else None,
             duedate=tools.datetime_millis(due_date) if due_date else None,
             expdate=tools.datetime_millis(due_date + datetime.timedelta(minutes = SHORT_BUFFER_MIN)) if due_date else None,
-            invitees=[venue.get_area_chairs_id()],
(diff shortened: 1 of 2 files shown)


melisabok:
The SAC console "Modify Reviewers Assignments" link includes the `start` parameter that has the AC, each AC has its own link: https://github.com/openreview/openreview-py/blob/master/openreview/venue/webfield/seniorAreaChairsWebfield.js#L72
showing only the assigned submissions for that AC.

PR #2785, opened 2025-12-04
ARR: Migrate Registration Forms to Root Domain

### LINKED ISSUE #2768: ARR: Use top-level registration forms
**Issue**
ARR currently has several processes that copies the contents from one form to another, particularly the cases of:
1) Copying data from the previous cycle to the current cycle
2) Copying author data into the reviewer/area chair forms

We want to move away from this because:
1) We want to avoid the duplication of note content
2) We want to avoid the pattern of the venue being able to post data with a tilde ID in the signatures

**Proposed Solution**
- Build top-level ARR role groups (`aclweb.org/ACL/ARR/<Role>`) plus a root meta invitation at `aclweb.org/ACL/ARR/-/Edit`
- Build all registration forms from the root domain with the same schemas
- Read top-level notes while still writing per-cycle edges
- Update consoles/webfields to read from the top-level domain with `registrationNoteDomain`

**Changes To `@openreview-py`**  
- Add to `openreview/arr/arr.py` to build the top-level invitations, groups, the meta-invitation, and give permissions to these top-level EiCs to use these
- Extend `openreview/arr/helpers.py` with constructors for the shared groups, root meta invitation, and dual-membership sync helpers; expose IDs/config via request-form hooks.  
- Create top-level invitation builder for registration forms
- Create a new type of ARRStage to represent a top-level registration form
- Refactor matching/setup scripts (`management/setup_reviewer_matching.py`, `setup_sae_matching.py`, `setup_ae_matching.py`, `setup_authors_to_reviewers.py`, etc.) so they read root-domain notes and produce per-cycle `Custom_Max_Papers`, `Registered_Load`, `Emergency_*`, and `Reviewing_Resubmissions` edges.  
- Update process functions (`process/max_load_process.py`, `process/emergency_load_process.py`, others) to read the top-level notes
- Things to keep in mind: role overlap checks

**Changes To `@openreview-web`**  
- Add a `registrationNoteDomain` config variable for ARR consoles/webfields
- Update PC/reviewer/AC console configs in openreview-py to show top-level shared-form links
- Ensure Editors-in-Chief gain UI affordances for cross-cycle membership edits and recruiter flows once the shared groups exist.

**Potential Risks**  
- First-time introducing top-level invitations: syncing data between top-level and cycle groups
- Broad setup matching refactor (should be caught by the tests)
- Keep in mind the possibility of overlapping due-date windows

haroldrubio:
- Resolves #2768 

This PR changes the super invitation of the registration invitations for ARR to a new root at `aclweb.org/ACL/ARR` in order to (1) prevent automatic posting of data with a tilde ID signatures and (2) prevent duplication of note content (copying between cycle, copying between roles)

1. Copies members from the root to the current cycle on venue setup
2. Modify setup scripts to read data from the root domain

I need to decide on how to handle some forms that should be bound to the cycle (requests for letters of recognition) - initially I was planning pointing all registration forms from the openreview-web end to point to a new domain but some should still point to the current cycle

TODOs:
- [ ] Complete integration with openreview-web
- [x] Remove un-needed code (max load process and preprocess functions, etc.)
- [x] Running a date process from the current cycle's Custom_Max_Papers invitation to synchronize data from root registration forms


openreview/arr/arr.py
```diff
@@ -363,6 +366,25 @@ def expire_invitation(self, invitation_id):
...
 
+        # Initialize root-level ARR groups if not populated
+        setup_arr_root_groups(self.client, self.support_user)
+
+        # Add missing program chairs missing from the EIC group
+        members = self.client.get_group(f'{ROOT_DOMAIN}/Editors_In_Chief').members
+        for program_chair_id in program_chair_ids:
+            if program_chair_id not in members:
+                members.append(program_chair_id)
+        self.client.add_members_to_group(f'{ROOT_DOMAIN}/Editors_In_Chief', program_chair_ids)
```
melisabok: does it mean all the EICs from all the cycles will be accumulated in the root EIC group? why?

openreview/arr/helpers.py
```diff
@@ -2101,3 +2111,133 @@ def get_resubmissions(submissions, previous_url_field):
...
+            )
+    
+    # 2. Create meta invitation (aclweb.org/ACL/ARR/-/Edit)
+    meta_invitation_id = f'{domain}/-/Edit'
+    meta_invitation = openreview.tools.get_invitation(client, meta_invitation_id)
+    
+    if meta_invitation is None:
+        # Get process content from venue module
+        process_dir = os.path.join(os.path.dirname(os.path.dirname(__file__)), 'venue', 'process')
+        invitation_edit_process_path = os.path.join(process_dir, 'invitation_edit_process.py')
```
melisabok: is there a reason to add these process functions?

openreview/arr/management/setup_ae_matching.py
```diff
@@ -107,7 +108,8 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
     # Build load map
     print(f"num profiles {len(all_profiles)}")
     id_to_load_note = {}
-    for role_id in [area_chairs_id]:
+    root_role_id = f"{ROOT_DOMAIN}/{area_chairs_name}"
```
melisabok: can you store `main_area_chairs_id` in the domain.content?

openreview/arr/management/setup_ae_matching.py
```diff
@@ -163,6 +162,44 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
         )
         print(f"posting {len(cmp_to_post)} custom max papers for {role_id}")
         openreview.tools.post_bulk_edges(client=client, edges=cmp_to_post)
+
+    # Create Reviewing_Resubmissions edges from root domain load notes
+    for role_id in [area_chairs_id]:
+        resubmissions_to_post = []
+        role_resubmissions_inv = f"{role_id}/-/Reviewing_Resubmissions"
```
melisabok: what is this invitation for?

openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -181,7 +182,8 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
 
     # Build load map
     id_to_load_note = {}
-    for role_id in [reviewers_id]:
+    root_role_id = f"{ROOT_DOMAIN}/{reviewers_name}"
```
melisabok: can you store this in the domain or in the ARR class?

openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -260,6 +259,97 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
             wait_to_finish=True
         )
         openreview.tools.post_bulk_edges(client=client, edges=cmp_to_post)
+
+    # Create Reviewing_Resubmissions edges from root domain load notes
+    for role_id in [reviewers_id]:
+        resubmissions_to_post = []
+        role_resubmissions_inv = f"{role_id}/-/Reviewing_Resubmissions"
```
melisabok: I thought we were going to store the registration notes in the root group, not edges, why do you need this?

openreview/arr/management/setup_sae_matching.py
```diff
@@ -61,7 +63,8 @@ def process(client, invitation):
 
     # Build load map
     id_to_load_note = {}
-    for role_id in [senior_area_chairs_id]:
+    root_role_id = f"{ROOT_DOMAIN}/{senior_area_chairs_name}"
```
melisabok: same here, use domain group

openreview/arr/root_invitation.py
```diff
@@ -0,0 +1,184 @@
...
+
+            if len(process_logs) == 0:
+                raise openreview.OpenReviewException('Time out waiting for invitation to complete: ' + invitation.id)
+
+            if process_logs[0]['status'] == 'error':
+                raise openreview.OpenReviewException('Error saving invitation: ' + invitation.id)
+
+        return invitation
+
+    def set_registration_invitation(self, registration_stage):
```
melisabok, line 50: the registration invitations should be the same for a regular venues or root venue, I would like to try reusing the code for both cases.

It seems that function needs: venue_id, registration_stage. Can we support this and call the same function from both?

Another option is to add a new state to the Venue class that says (support_root_venue), if this is true the regisration notes are created on the root venue instead the current venue.

openreview/arr/arr.py
haroldrubio: Oh that true, I probably need to fix this since some of them will only be conference PCs and not part of the ARR team

I was thinking about using the root EiCs to give access to all previous ARR cycles without going to each request form, do you think that's doable/a good idea?

openreview/arr/helpers.py
haroldrubio: No I can remove this, I think it was there when I was testing out making different invitations but none of the root invitations should need to do the propagation of changes

openreview/arr/management/setup_ae_matching.py
haroldrubio: This is because there is a field in the maximum load note that allows them to indicate if they're willing to review a submission they previously reviewed - reviewers who select 0 load can select 'Yes' so they only get assigned resubmissions. The SACs/ACs don't have access to the maximum load note so that information is shared in the edges. Also, they can opt out of resubmissions so the SACs/ACs need to know this information

openreview/arr/root_invitation.py
haroldrubio: I was looking into changes for the Venue class and I'm running into a few issues:

The root venue doesn't have a venue request form (as I have implemented it right now) and doesn't get deployed, so there isn't a Venue object for the root venue. If there's a line that calls to the Venue object, I'd need to replace it.

Maybe it's worth it to deploy a full root venue instead of manually creating the meta invitation and only populating what we need, so I can better re-use code?

openreview/arr/management/setup_reviewer_matching.py
haroldrubio: The refactor removes the nesting inside the for loop - hopefully it's more clear that these are edges to the cycle and not the root group

openreview/arr/webfield/reviewersWebfield.js
```diff
@@ -0,0 +1,29 @@
...
+    officialReviewName: domain.content.review_name?.value,
+    reviewRatingName: domain.content.review_rating?.value,
+    areaChairName: domain.content.area_chairs_name?.value,
+    submissionName: domain.content.submission_name?.value,
+    submissionInvitationId: domain.content.submission_id?.value,
+    customMaxPapersInvitationId: domain.content.reviewers_custom_max_papers_id?.value,
+    recruitmentInvitationId: domain.content.reviewers_recruitment_id?.value,
+    reviewLoad: '',
+    hasPaperRanking: false,
+    registrationFormDomainMap: {
```
melisabok: I'm not sure about this change, can we pass the list of registration invitations to the consoles as a parameter?

melisabok:
Closing this PR, no progress made.

PR #2795, opened 2025-12-16
Improve documentation for ConferenceBuilder

nikita-zmanovskiy:
This PR adds missing API documentation for the ConferenceBuilder
class.

Changes:
- Added module and class references to api.rst
- Added a class-level docstring to ConferenceBuilder

This is a documentation-only change and does not affect
any API behavior.


melisabok:
ConferenceBuilder is deprecated and it will be removed soon. Please use Venue class or Journal class instead.

PR #2796, opened 2025-12-16
Fix/Venue: set default email groups

celestemartinez:
Since request forms are submitted from the UI, all request forms have the key `preferred_emails_groups`. So `note.content.get('preferred_emails_groups', [venue.get_authors_id()])` was always returning `[]`. 

This is the reason why none of the newly deployed venues have the preferred emails group and invitation created. I am now checking if `venue.preferred_emails_groups` is set from the request form, if not set it as `[venue.get_authors_id()]`


enrubio (approved):
Looks good!

PR #2798, opened 2025-12-17
Fix/ New UI Recruitment

celestemartinez:
There is an error in the recruitment process when check valid_invitees is empty. I am checking if any users where invited when posting the recruitment status note and when sending an email to PCs. The recruitment status note instead says "No users invited"

Screenshot 2025-12-17 at 12 09 10 PM

### DIFF
--- a/openreview/workflows/process/committee_recruitment_request_process.py
+++ b/openreview/workflows/process/committee_recruitment_request_process.py
@@ -82,16 +82,16 @@ def process(client, edit, invitation):
             if 'profile_not_found' not in recruitment_status['errors']:
                 recruitment_status['errors']['profile_not_found'] = []
             recruitment_status['errors']['profile_not_found'].append(email)
-        elif invited_group_ids:
-            invited_group_id=invited_group_ids[0]
-            if invited_group_id not in recruitment_status['already_invited']:
-                recruitment_status['already_invited'][invited_group_id] = [] 
-            recruitment_status['already_invited'][invited_group_id].append(email)
         elif member_group_ids:
             member_group_id = member_group_ids[0]
             if member_group_id not in recruitment_status['already_member']:
                 recruitment_status['already_member'][member_group_id] = []
             recruitment_status['already_member'][member_group_id].append(email)
+        elif invited_group_ids:
+            invited_group_id=invited_group_ids[0]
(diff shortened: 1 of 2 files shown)


openreview/workflows/process/committee_recruitment_request_process.py
```diff
@@ -82,16 +82,16 @@ def process(client, edit, invitation):
...
-            invited_group_id=invited_group_ids[0]
-            if invited_group_id not in recruitment_status['already_invited']:
-                recruitment_status['already_invited'][invited_group_id] = [] 
-            recruitment_status['already_invited'][invited_group_id].append(email)
         elif member_group_ids:
             member_group_id = member_group_ids[0]
             if member_group_id not in recruitment_status['already_member']:
                 recruitment_status['already_member'][member_group_id] = []
             recruitment_status['already_member'][member_group_id].append(email)
+        elif invited_group_ids:
```
celestemartinez, line 90: I am checking if the user is in the Reviewers group before checking the Reviewers/Invited group. Since this is an else if, we were never getting to the Reviewers group check

openreview/workflows/process/committee_recruitment_request_process.py
```diff
@@ -156,23 +156,27 @@ def recruit_user(invitee):
...
-        recipients=[f'{venue_id}/Program_Chairs'],
-        message=f'''The recruitment request process for the {committee_role.capitalize()} Committee has been completed.
+    invited_list = f'- [invited list](https://openreview.net/group/revisions?id={invited_group.id}&editId={added_edit["id"]})' if valid_invitees else ''
+
+    message = f'''The recruitment request process for the {committee_role.capitalize()} Committee has been completed.
 
 Invited: {recruitment_status["invited"]}
-Already invited: {len(recruitment_status["already_invited"])}
-Already member: {len(recruitment_status["already_member"])}
+Already invited: {len(recruitment_status["already_invited"].get(invited_group.id, []))}
```
celestemartinez, line 164: The already invited key is always in the recruitment status, we need to check if the invited group has been added and how many users are in it. For example, this is what the status looks like:

```
recruitment_status: {
  already_invited: {
      ABCD.cc/2025/Conference/Program_Committee/Invited: [
        reviewer_two@abcd.cc,
        reviewer@mail.com
      ]
    }
}

```

PR #2800, opened 2025-12-19
ARR: Update tracks

haroldrubio:
This PR updates the tracks for the January 2026 cycle (ACL conference)


openreview/stages/arr_content.py
```diff
@@ -2,32 +2,39 @@
 from datetime import datetime
 
 arr_tracks = [
-    "Computational Social Science and Cultural Analytics",
```
melisabok, line 5: is this a change just for this cycle or for all of them in the future?

haroldrubio: All cycles

melisabok (approved):

PR #2808, opened 2026-01-07
Fix: ARR Preprint Always Hides Authors

haroldrubio:
This PR always adds the appropriate readers in the preprint release invitation and adds documentation on appropriate timing of the preprint release

### DIFF
--- a/tests/test_edges.py
+++ b/tests/test_edges.py
@@ -69,9 +69,8 @@ class TestEdges:
     
     def test_rename_edges(self, client, openreview_client, helpers):
         guest = openreview.api.OpenReviewClient()
-        to_profile = guest.register_user(email = 'nadia@mail.com', fullname = 'Nadia L', password = helpers.strong_password)
-        assert to_profile
-        assert to_profile['id'] == '~Nadia_L1'
+        registration = guest.register_user(email = 'nadia@mail.com', fullname = 'Nadia L', password = helpers.strong_password)
+        assert registration['status'] == 'ok'
         super_user_edges = list(openreview.tools.iterget_edges(openreview_client, tail="~Super_User1"))
         openreview_client.rename_edges('~Super_User1', '~Nadia_L1')
         nadias_edges = list(openreview.tools.iterget_edges(openreview_client, tail="~Nadia_L1"))
--- a/tests/test_profile_management.py
+++ b/tests/test_profile_management.py
@@ -3724,7 +3724,7 @@ The OpenReview Team.
 
         profile = openreview_client.get_profile(email_or_id='~Lionel_Messi1')
         assert len(profile.content['emails']) == 2
-        assert len(profile.content['emailsConfirmed']) == 2
+        assert len(profile.content['emailsConfirmed']) == 1
 
         profile_content={
             'names': [


tests/test_arr_venue_v2.py
```diff
@@ -2720,6 +2720,30 @@ def test_post_submission(self, client, openreview_client, helpers, test_client,
         assert 'readers' not in submissions[0].content['reassignment_request_reviewers']
         assert 'readers' not in submissions[0].content['justification_for_not_keeping_action_editor_or_reviewers']
 
+        ## Papers without expected readership on authors are not revealed
+        ## Post Paper 6 to have no readers
+        openreview_client.post_note_edit(
```
melisabok: instead of deleting the readers, I would prefer to execute the Preprint Release before the Post Submission process function to reproduce the issue that happened yesterday. 

it is true that the Post Submission overwrites the preprint changes so you can run the preprint release again after the post submission to go back to the normal state.

tests/test_arr_venue_v2.py
```diff
@@ -2608,8 +2608,44 @@ def test_submitted_author_form(self, client, openreview_client, helpers, test_cl
     def test_post_submission(self, client, openreview_client, helpers, test_client, request_page, selenium):
 
         pc_client=openreview.Client(username='pc@aclrollingreview.org', password=helpers.strong_password)
+        pc_client_v2=openreview.api.OpenReviewClient(username='pc@aclrollingreview.org', password=helpers.strong_password)
         request_form=pc_client.get_notes(invitation='openreview.net/Support/-/Request_Form')[1]
 
+        ## release preprint submissions
```
melisabok, line 2620: before triggering the preprint release step, can you make sure the submissions don't have readers in the authors and authorids field? it is not clear to me if the Post_Submission already run or not.

tests/test_arr_venue_v2.py
```diff
@@ -2608,8 +2608,50 @@ def test_submitted_author_form(self, client, openreview_client, helpers, test_cl
...
+            )
+        )
+
+        helpers.await_queue()
+        helpers.await_queue_edit(openreview_client, 'aclweb.org/ACL/ARR/2023/August/-/Preprint_Release_Submission-0-1', count=1)
+        ## Assert released preprints have hidden authors and authorids
+
+        submissions = pc_client_v2.get_notes(invitation='aclweb.org/ACL/ARR/2023/August/-/Submission', sort='number:asc')
+        for submission in submissions:
+            if submission.content['preprint']['value'] == 'yes':
```
melisabok: instead of trying to find the submissions that are preprint, I would assert the submissions that we know that are preprints. 

You get the submissions sorted by number and you know that submission number 3 and 4 are preprints then you assert those submissions:

```
assert 'everyone' in submissions[2].readers
assert 'readers' in submissions[2].content['authors']
assert 'readers' in submission[2].content['authorids']

assert 'everyone' in submissions[3].readers
assert 'readers' in submissions[3].content['authors']
assert 'readers' in submission[3].content['authorids']
```

The if sentence could be wrong and the test would be passing without really checking if the readers were added.

celestemartinez:
I see that different fields are hidden by the Post Submission and by the Preprint process. For example, `consent_to_share_data` gets hidden by the Post Submission process, but not by the preprint, is this ok? This means there are some fields that are not hidden if the preprint process were to run first. 

Also, the preprint process releases `justification_for_author_changes` to the assigned committee, but the preprocess releases it only to venueid and authors. It looks like it matters in this case which order the processes are run in, not sure who should be the final readers in this case.

haroldrubio:
Thanks for catching that! I'll make a change to also hide these fields.

Also moving `justification_for_author_changes` to only the `hide_fields` list

PR #2810, opened 2026-01-12
Feature/ Journal: skip reviewer official recommendation

celestemartinez:
This PR adds a new setting to the journal request, `skip_official_recommendation`, which:
- skips the official recommendation, allowing AEs to directly move to the review rating stage after reviews are submitted
- sends an email to AEs after reviews are submitted that does not mention official recommendation

Note that in this workflow, the discussion period doesn't really happen, so the setting `discussion_period` is ignored, since the `decision_period` setting is used for the review rating duedate.

### DIFF
--- a/openreview/workflows/process/committee_recruitment_request_edit_reminder_process.py
+++ b/openreview/workflows/process/committee_recruitment_request_edit_reminder_process.py
@@ -22,7 +22,7 @@ def process(client, edit, invitation):
     for l in edit.content['invitee_details']['value'].strip().split('\n'):
         email_or_profile_id = l.split(',')[0].strip()
         if email_or_profile_id:
-            invitees.append(email_or_profile_id.lower() if not email_or_profile_id.startswith('~') else email_or_profile_id)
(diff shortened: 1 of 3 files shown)


openreview/journal/process/review_process.py
```diff
@@ -44,89 +44,118 @@ def process(client, edit, invitation):
...
-        client.post_message(
-            invitation=journal.get_meta_invitation_id(),
-            recipients=[journal.get_action_editors_id(number=submission.number)],
-            subject=f'''[{journal.short_name}] Start of author discussion for {journal.short_name} submission {submission.number}: {submission.content['title']['value']}''',
-            message=message,
-            replyTo=journal.contact_info,
-            signature=journal.venue_id,
-            sender=journal.get_message_sender()
-        )
+        if not journal.should_skip_official_recommendation:
```
melisabok:
```suggestion
        if not journal.should_skip_official_recommendation():
```

melisabok (approved):

PR #2818, opened 2026-01-21
Feat/ New UI: allow abstract/full submission deadline

celestemartinez:


openreview/venue/invitation.py
```diff
@@ -2929,8 +2935,10 @@ def set_submission_revision_invitation(self, submission_revision_stage=None):
                 }
             }
             edit_invitations_builder = openreview.workflows.EditInvitationsBuilder(self.client, self.venue_id)
-            edit_invitations_builder.set_edit_content_invitation(revision_invitation_id, content)
-            edit_invitations_builder.set_edit_dates_invitation(revision_invitation_id)
+            edit_invitations_builder.set_edit_content_invitation(revision_invitation_id, content if revision_stage.name != 'Full_Submission' else {}, allow_license_edition=revision_stage.allow_license_edition)
```
melisabok: why is the content empty when there is no Full_Submission invitation?

openreview/venue/invitation.py
```diff
@@ -2929,8 +2935,10 @@ def set_submission_revision_invitation(self, submission_revision_stage=None):
...
+            allow_cdate_edit = False if revision_stage.name == 'Full_Submission' else True
```
melisabok: I'm not sure about relying on the name of the invitation to indicate if the venue has an abstract/full deadline workflow, can we use another variable?

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -13,21 +13,26 @@ def process(client, edit, invitation):
...
     submission_duedate = datetime.datetime.fromtimestamp(note.content['submission_deadline']['value']/1000)
 
+    full_submission_deadline = note.content['full_submission_deadline']['value'] if 'full_submission_deadline' in note.content else None
+    full_submission_duedate = datetime.datetime.fromtimestamp(full_submission_deadline/1000) if full_submission_deadline else None        
+
     venue.submission_stage =  openreview.stages.SubmissionStage(
         start_date=submission_cdate,
         due_date=submission_duedate,
-        double_blind=True,
-        force_profiles=True
```
melisabok, line 20: Why is `force_profiles` removed here?

openreview/venue/invitation.py
celestemartinez: This is because for normal revision invitations, we give the option to PCs to enable the stage for accepted submissions or all submissions. We don't want to give this option for the Full_Submission stage.

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
celestemartinez: sorry, that must have happened when I fixed conflicts between the branch and master. Fixed.

openreview/venue/invitation.py
melisabok: so the PC will need to enter the whole content from scratch?

openreview/venue/invitation.py
celestemartinez: What about adding a parameter to SubmissionRevisionStage, `is_full_submission_revision`?

melisabok: maybe we should store it in the domain?

openreview/venue/invitation.py
celestemartinez: All submission revision invitations get created with the default submission content (we don't change the content when the submission content changes)

This content I am setting to {} here is related to the fields that PCs can edit:

For Full_Submission, PCs can only edit license and content
Image
For Camera_Ready_Revision, PCs can edit the source (which submissions to create revision invitations for)
Image

We don't need to pass source in the content to the Full_Submission invitation creating because these invitations get created for all papers.

openreview/venue/invitation.py
celestemartinez: but if we save it in the domain, when we are creating the different revision stages, we won't know which one is the Full_Submission one, right? That's why I thought this setting would need to be attached to the stage itself

melisabok: we can save in the domain: full_submission_invitation_id and if that matches with the name of the SubmissionRevisionStage then we don't let them edit the cdate.

openreview/venue/group.py
```diff
@@ -199,9 +199,12 @@ def create_venue_group(self):
             'decision_heading_map': { 'value': self.venue.decision_heading_map },
             'reviewers_message_submission_id': { 'value': self.venue.get_message_id(number='{number}') },
             'reviewers_message_id': { 'value': self.venue.get_message_id(committee_id=self.venue.get_reviewers_id()) },
-            'article_endorsement_id': { 'value': self.venue.get_article_endorsement_id() },
+            'article_endorsement_id': { 'value': self.venue.get_article_endorsement_id() }
         }
 
+        if self.venue.submission_stage.second_due_date:
+            content['full_submission_invitation_id'] = { 'value': f'{venue_id}/-/Full_Submission' }
```
melisabok: could we use the name of the invitation to build the Full_Submission?

`f'{venue_id}/-/Full_{submission_name}'`

celestemartinez: Done

PR #2821, opened 2026-01-21
Let recruitment notes be signed using the invitee email

melisabok:
- Sign recruitment notes using the invitee email or tilde id even if the user is not logged in
- Do not store the key and user in the note content, it is not needed anymore
- Remove preprocess validation given that API can validate if the user is invited and the key is correct.

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -1818,8 +1818,9 @@ def pretty_id(group_id):
     for token in tokens:
         transformed_token=re.sub(r'\..+', '', token).replace('-', '').replace('_', ' ')
         letters_only=re.sub(r'\d|\W', '', transformed_token)
+        has_no_ascii=not re.search(r'[a-zA-Z0-9]', transformed_token)
 
-        if letters_only != transformed_token.lower():
+        if letters_only != transformed_token.lower() or (has_no_ascii and transformed_token):
             transformed_tokens.append(transformed_token)
 
 
--- a/openreview/venue/group.py
+++ b/openreview/venue/group.py
@@ -346,17 +346,18 @@ class GroupBuilder(object):
             content['comment_notification_threshold'] = { 'value': self.venue.comment_notification_threshold }
 
         if self.venue.is_template_related_workflow():
+            submission_name = self.venue.submission_stage.name
             content['exclusion_workflow_invitations']  = {'value': [
                 f'{venue_id}/-/Edit',
-                f'/{venue_id}/Submission[0-9]+/',
+                f'/{venue_id}/{submission_name}[0-9]+/',
                 f'/{venue_id}/-/Venue.*/',
                 f'{venue_id}/{reviewers_name}/-/Message', # TODO: parametrize group names and invitation names
-                f'/{venue_id}/{reviewers_name}/-/(?!Submission_Group$|Bid|Conflict|Affinity_Score|Review_Count|Review_Assignment_Count|Review_Days_Late|Recruitment|Assignment).*/', # matching invitations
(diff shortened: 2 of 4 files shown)


openreview/workflows/templates.py
```diff
@@ -1031,14 +1031,14 @@ def setup_committee_group_recruitment_template_invitation(self):
                     'id': '${2/content/committee_id/value}/-/Recruitment_Response',
                     'duedate': '${2/content/due_date/value}',
                     'expdate': '${2/content/due_date/value}',
-                    'invitees': ['everyone'],
+                    'invitees': ['${3/content/committee_id/value}/Invited'],
                     'signatures': ['${3/content/venue_id/value}'], 
                     'readers': ['everyone'],
                     'writers': ['${3/content/venue_id/value}'],
                     'description': 'Configure the timeframe users invitees can accept or decline ${2/content/committee_pretty_name/value} recruitment invitations, whether or not they can accept with a reduced load, and customize the email responses when they accept or decline the invitation to serve as a ${2/content/committee_pretty_name/value}. Go to the **[${2/content/committee_pretty_name/value} group](/group/edit?id=${2/content/committee_id/value})** to recruit reviewers.',
-                    'preprocess': self.get_process_content('process/committee_recruitment_response_pre_process.js'),
```
celestemartinez, line 1040: It looks like the preprocess also checks if the user is already assigned to a submission before allowing them to decline the invitation. is it ok to not check this anymore?

openreview/workflows/templates.py
```diff
@@ -1031,14 +1031,14 @@ def setup_committee_group_recruitment_template_invitation(self):
...
                     'process': self.get_process_content('process/committee_recruitment_response_process.py'),
                     'web': self.get_webfield_content('webfield/committeeRecruitmentResponseWebfield.js'),
+                    'guestPosting': True,
                     'content': {
                         'hash_seed': {
```
celestemartinez: We don't need to store this anymore, right?

openreview/workflows/templates.py
melisabok: Good point, I can add it back.

openreview/workflows/templates.py
melisabok: this field is needed to tell the API that a secret field must be created for the invitation

celestemartinez: I thought the `guestPosting` did this, I was referring to the `hash_seed` stored in the content.

melisabok: ohhhh, yes, thanks for the catch, I will remove it

openreview/workflows/templates.py
melisabok: Preprocess was added back, I found a small bug setting the venueId, I added a test to verify that it works.

openreview/workflows/templates.py
melisabok: fixed

.circleci/config.yml
```diff
@@ -85,6 +85,7 @@ jobs:
           name: Clone openreview-web
           command: |
             git clone https://$OPENREVIEW_GITHUB@github.com/openreview/openreview-web.git ~/openreview-web
+            cd ~/openreview-web && git checkout feature/sign-notes-guest-email
```
melisabok: revert this after web PR is merged

openreview/tools.py
```diff
@@ -1361,7 +1362,13 @@ def recruit_user(client, user,
...
     client.post_message(recruitment_message_subject, [user], personalized_message, parentGroup=comittee_invited_id, replyTo=contact_email, invitation=message_invitation, signature=message_signature)
 
-def get_user_hash_key(user, hash_seed):
+def get_user_hash_key(user, hash_seed, invitation=None):
+    if invitation is not None:
+        jwt_payload = {
+            "group": user,
+            "invitation": invitation,
+        }
+        return jwt.encode(jwt_payload, hash_seed, algorithm="HS256")
```
xkopenreview, line 1371: i guess it will rely on invitation's exp date to invalidate the token

PR #2825, opened 2026-01-23
Fix/ Matching: revert edge readers to include assigned committee only

celestemartinez:
This PR adds only the assigned committee as readers of all edges, instead of the whole committee. This change has been done for all edges except for the following invitations:
- Custom_Max_Papers 
- Contraint_Label 

because these edges have a group as the head, not a note. In this case, the entire committee is still reader of the edges. 

No tests needed to be updated except for `test_venue_with_tracks.py` since this was the reason why we had made this change in the first place.

### DIFF
--- a/openreview/arr/management/setup_reviewer_matching.py
+++ b/openreview/arr/management/setup_reviewer_matching.py
@@ -209,6 +209,10 @@ def process(client, invitation):
         reviewers_id: [venue_id, senior_area_chairs_id, area_chairs_id]
     }
 
+    track_submission_edge_readers = {
+        reviewers_id: [venue_id, venue.get_senior_area_chairs_id(number='{number}'), venue.get_area_chairs_id(number='{number}')]
+    }
+
     # Create reviewers submitted groups 
     for submission in submissions:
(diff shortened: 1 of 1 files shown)


openreview/venue/matching.py
```diff
@@ -93,6 +93,8 @@ def _create_edge_invitation(self, edge_id, any_tail=False, default_label=None):
         
         is_assignment_invitation=edge_id.endswith('Assignment') or edge_id.endswith('Aggregate_Score')
         paper_number = '${{2/head}/number}'
+        if edge_id in [venue.get_custom_max_papers_id(self.match_group.id), venue.get_constraint_label_id(self.match_group.id)]:
```
melisabok, line 96: instead of checking the invitation id, could we check if the invitation.head is not a submission id then set paper_number = None?

I can imagine that will have more different invitations with the same pattern in the future.

celestemartinez: yes, I thought about doing this but the head does not get set until later in the code (and to set the head we check the invitation id anyway). I can look into this.

melisabok: I see that, ok, let's keep it as it is now.

PR #2829, opened 2026-01-28
Fix/ Add field to new UI request form for ACs

celestemartinez:


openreview/workflows/workflows.py
```diff
@@ -227,20 +227,32 @@ def set_conference_review_request(self):
...
                         },
-                        'area_chairs_name': {
+                        'area_chairs_support': {
                             'order': 12,
-                            'description': 'Please provide the designated name to be used for area chairs, for example Area_Chairs. Leave empty if your venue does not have area chairs',
+                            'description': "Does your venue have area chairs? Leave unchecked if your venue does not have area chairs.",
+                            'value': {
+                                'param': {
+                                    'type': "string",
+                                    'enum': ['Yes, my venue does have Area Chairs.'],
```
melisabok: can this be : `{ value: True, description: 'Yes, my venue does have Area Chairs. '}`?

It easier to use a booelan than checking a string value

openreview/workflows/workflows.py
```diff
@@ -227,20 +227,32 @@ def set_conference_review_request(self):
...
+                                    'input': 'checkbox',
+                                    'optional': True,
+                                    'deletable': True
+                                }
+                            }
+                        },
+                        'area_chairs_name': {
+                            'order': 13,
+                            'description': 'Please provide the designated name to be used for area chairs. Use underscores for spaces and capitalize as needed. Default is "Area_Chairs".',
```
melisabok: should we add "Ignore if area chair is not supported..."?

openreview/workflows/workflows.py
celestemartinez: Done

openreview/workflows/workflows.py
celestemartinez: Done

melisabok (approved):

PR #2839, opened 2026-02-05
Journal: add preprocess to official recommendation

celestemartinez:
TMLR wants to add a preprocess to the official recommendation, similar to the preprocess in the decision invitation.


openreview/journal/process/official_recommendation_pre_process.py
```diff
@@ -8,4 +8,14 @@ def process(client, edit, invitation):
     reviews=client.get_notes(invitation=journal.get_review_id(number=submission.number))
 
     if not any(review.signatures[0]==signatory_group_id for review in reviews):
-        raise openreview.OpenReviewException(f'You must submit your official review before submitting your recommendation')
\ No newline at end of file
+        raise openreview.OpenReviewException(f'You must submit your official review before submitting your recommendation')
+
+    if journal.venue_id == 'TMLR':
```
celestemartinez: @melisabok what is the best way to have these checks run only for TMLR? There is already a preprocess for the official recommendation so I had to add the TMLR specific code here. 

Would it be better to have a separate TMLR only preprocess file?

melisabok: I could think this way:

- store a value in the TMLR object: "official_recommendation_additional_validation" 
- get the value if exists and run it after the general validation using the `exec`. 

I'm thinking that we need to start thinking how to migrate the journal to the new UI where the EICs can add these validations by themselves.

melisabok (approved):

PR #2847, opened 2026-02-09
ARR: Optimize Matching Setup Scripts

haroldrubio:
This PR makes the following optimizations to the setup matching scripts:

1. Uses `concurrent_requests` to process resubmissions in parallel
2. Avoids updating affinity scores to 3 and instead posts a new edge to the `Resubmission_Score` invitation
3. Avoids posting in each iteration, appends resubmission updates to thread-safe data structures and posts in bulk after `concurrent_requests`

This PR is left in draft as these changes are not yet applied to the AC matching setup script

### DIFF
--- a/openreview/arr/management/setup_reviewer_matching.py
+++ b/openreview/arr/management/setup_reviewer_matching.py
@@ -292,6 +292,23 @@ def process(client, invitation):
             'label': label
         }
 
+    def process_reviewer_exception(reviewer_id, current_client, result):
+        if id_to_load_note.get(reviewer_id) and \
+            int(id_to_load_note[reviewer_id].content['maximum_load_this_cycle']['value']) == 0 and \
+                'Yes' in id_to_load_note[reviewer_id].content['maximum_load_this_cycle_for_resubmissions']['value']:
+            result['only_resubmissions'].append({
+                'role': reviewers_id,
+                'name': reviewer_id
+            })
+            rev_cmp = {
+                g['id']['tail'] : g['values'][0]
+                for g in current_client.get_grouped_edges(invitation=rev_cmp_inv, select='id,weight', groupby='tail')
+            }
+            result['reviewer_exception_updates'].append({
+                'reviewer_id': reviewer_id,
+                'existing_edge': rev_cmp[reviewer_id]
+            })
+
     def process_resubmission(submission):
         result = {
             'submission': submission,
@@ -300,23 +317,6 @@ def process(client, invitation):
             'only_resubmissions': [],
             'reviewer_exception_updates': []
         }
-
-        def process_reviewer_exception(reviewer_id, current_client):
-            if id_to_load_note.get(reviewer_id) and \
-                int(id_to_load_note[reviewer_id].content['maximum_load_this_cycle']['value']) == 0 and \
-                    'Yes' in id_to_load_note[reviewer_id].content['maximum_load_this_cycle_for_resubmissions']['value']:
-                result['only_resubmissions'].append({
-                    'role': reviewers_id,
-                    'name': reviewer_id
-                })
-                rev_cmp = {
-                    g['id']['tail'] : g['values'][0]
-                    for g in current_client.get_grouped_edges(invitation=rev_cmp_inv, select='id,weight', groupby='tail')
-                }
-                result['reviewer_exception_updates'].append({
-                    'reviewer_id': reviewer_id,
-                    'existing_edge': rev_cmp[reviewer_id]
-                })
 
         print(f"rewriting {submission.id}")
         # 1) Find all reassignments and reassignment requests -> 0 out or set to 3
@@ -365,7 +365,7 @@ def process(client, invitation):
             else:
                 updated_weight = 3
                 result['reassignment_status'].append(build_reassignment_edge(submission.id, reviewer_id, 'Requested'))
-                process_reviewer_exception(reviewer_id, current_client)
+                process_reviewer_exception(reviewer_id, current_client, result)
 
             replace_edge(
                 existing_edge=reviewer_edge,


openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -238,168 +202,201 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
...
+            if 'explanation_of_revisions_PDF' in submission.content:
+                local_explanation_readers = [
+                    venue.get_program_chairs_id(),
+                    venue.get_senior_area_chairs_id(number=submission.number),
+                    venue.get_area_chairs_id(number=submission.number),
+                    venue.get_reviewers_id(number=submission.number, submitted=wants_new_reviewers),
+                    venue.get_authors_id(number=submission.number)
+                ]
+
+            with shared_state_lock:
```
melisabok: what does it mean this?

openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -238,168 +202,201 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
...
-            if role_id not in load_invitation:
-                continue
+    status_inv = f"{reviewers_id}/-/{status_name}"
+    resubmission_score_inv = f"{reviewers_id}/-/{resubmission_score_inv_name}"
+    status_edges_to_post = []
+    resubmission_score_edges_to_post = []
+    explanation_reader_updates = []
+    shared_state_lock = threading.Lock()
+
+    def process_resubmission(submission):
```
melisabok: can you give us a summary of the resubmission function? this function is too large and it is difficult to follow.

tests/test_arr_venue_v2.py
```diff
@@ -3712,31 +3709,31 @@ def test_resubmission_and_track_matching_data(self, client, openreview_client, h
...
+        # For 1, assert that the AE affinity score is 3 and reviewer resubmission score exists
         ac_scores = {
             g['id']['tail'] : g['values'][0]
             for g in pc_client_v2.get_grouped_edges(invitation='aclweb.org/ACL/ARR/2023/August/Area_Chairs/-/Affinity_Score', head=submissions[1].id, select='tail,id,weight', groupby='tail')
         }
-        rev_scores = {
+        rev_resubmission_scores = {
             g['id']['tail'] : g['values'][0]
-            for g in pc_client_v2.get_grouped_edges(invitation='aclweb.org/ACL/ARR/2023/August/Reviewers/-/Affinity_Score', head=submissions[1].id, select='tail,id,weight', groupby='tail')
+            for g in pc_client_v2.get_grouped_edges(invitation='aclweb.org/ACL/ARR/2023/August/Reviewers/-/Resubmission_Score', head=submissions[1].id, select='tail,id,weight', groupby='tail')
```
melisabok: why are these test different?

haroldrubio: They're only different for now in draft, I only made changes to the reviewer script and not the AC script to make sure I have the logic correct - this resubmission loop is mostly shared logic between both scripts.

Once I change the AC script, the tests would be the same. But as we discussed, I'll be reverting the changes that require changes in the tests

haroldrubio:
(2) was because I wanted to avoid doing individual POST requests to update the affinity score edges to the previous reviewers to `weight: 3`

openreview/arr/management/setup_reviewer_matching.py
haroldrubio: Removed shared state, all state is returned now

openreview/arr/management/setup_reviewer_matching.py
haroldrubio: 1. Retrieves information about the submission from a previous cycle, returns early if the current cycle indicates that it is not a resubmission
2. Adds the current reviewer’s group to the previous reviewers. Add the reviewers/submitted group if the authors indicate that they want new reviewers
3. Resolve the previously submitted reviewers to tilde IDs and updates their affinity score to 0 or 3 depending on if the authors want new reviewers
4. If they want new reviewers, handle a special case for reviewers where they can opt-in to review resubmissions but not new submissions (max load = 0 in their note). Note the exception and update their custom max papers edge
5. Add readers to the `explanation_of_revisions_PDF` submission content field, the reviewers that get added depend on if the authors want new reviewers

melisabok:
The changes look good to me, I suggest adding the explanation of each step in the openreview-scripts documentation so ARR knows what the script is doing. Thanks Harold!

openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -289,6 +282,51 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
...
+                    'role': reviewers_id,
+                    'name': reviewer_id
+                })
+                rev_cmp = {
+                    g['id']['tail'] : g['values'][0]
+                    for g in current_client.get_grouped_edges(invitation=rev_cmp_inv, select='id,weight', groupby='tail')
+                }
+                result['reviewer_exception_updates'].append({
+                    'reviewer_id': reviewer_id,
+                    'existing_edge': rev_cmp[reviewer_id]
```
celestemartinez: Are these custom max paper edges? Will every reviewer always have an existing edge?

celestemartinez: I see later on the following comment, so it looks like we can be sure that this edge exists?
> #note implies cmp edge

openreview/arr/management/setup_reviewer_matching.py
```diff
@@ -289,6 +282,51 @@ def replace_edge(existing_edge=None, edge_inv=None, new_weight=None, submission_
...
+    def process_resubmission(submission):
+        result = {
+            'submission': submission,
+            'skip_scores': [],
+            'reassignment_status': [],
+            'only_resubmissions': [],
+            'reviewer_exception_updates': []
+        }
+
+        def process_reviewer_exception(reviewer_id, current_client):
```
carlosmondra: Why do we need this function inside another function? I think it would be better if it was out side of the process_resubmission function, if possible.

PR #2848, opened 2026-02-09
Fix/ New Venue UI: do not update group content on redeployment

celestemartinez:


tests/test_reviewers_only.py
```diff
@@ -44,6 +44,31 @@ def test_setup(self, openreview_client, helpers):
...
+        # edit invitation to allow redeployment for testing purposes
+        openreview_client.post_invitation_edit(
+            invitations='openreview.net/Support/-/Edit',
+            signatures=['~Super_User1'],
+            invitation=openreview.api.Invitation(
+                id = 'openreview.net/Support/Venue_Request/Conference_Review_Workflow/-/Deployment',
+                edit = {
+                    'note': {
+                        'content': {
+                            'redeployment': {
```
melisabok, line 56: Why do you need to add this new field?

celestemartinez: This was added just for the test so I could redeploy the venue and test this. I was able to reproduce this in my local but only by going to the queue and re-running the deployment process. If I try to redeploy without any change in the note edit, then the process does not re-run.

Issue #2831, opened 2026-02-02
New Configuration UI: post a message to the PCs when a scheduled process function fails because of missing data


melisabok:
Process function errors like: "Select a valid date to upload paper decisions", shouldn't be failing the process function.

I suggest to catch these errors in the process function and post a comment in the venue request form saying which data is missing.

The same happens for the deploy assignments.

celestemartinez:
Fixed here: #2853

PR #2853, opened 2026-02-12
New Venue UI: more fixes

celestemartinez:
- Avoid sending emails for comments posted by Support to the venue request
- Set the withdrawal expiration date one year after the submission deadline
- `setup_matching_invitations()` creates invitations for all roles (instead of calling it once per role)
- Create a `Status` invitation used by the different date processes to post a comment to the request form when an invitation runs with missing values
- Add the venueid to the comment sent to support to more easily group and find emails

### DIFF
--- a/openreview/arr/management/setup_shared_data.py
+++ b/openreview/arr/management/setup_shared_data.py
@@ -105,7 +105,7 @@ def process(client, invitation):
             domain.content['area_chairs_id']['value'].replace(venue_id, previous_cycle_id),
             domain.content['senior_area_chairs_id']['value'].replace(venue_id, previous_cycle_id),
             domain.content['ethics_chairs_id']['value'].replace(venue_id, previous_cycle_id),
-            f"{previous_cycle_id}/{domain.content['ethics_reviewers_name']['value']}",
+            domain.content['ethics_reviewers_id']['value'].replace(venue_id, previous_cycle_id),
         ]
     ]
 
--- a/openreview/arr/process/max_load_process.py
+++ b/openreview/arr/process/max_load_process.py
@@ -11,10 +11,7 @@ def process(client, edit, invitation):
     SAC_ID = domain.content['senior_area_chairs_id']['value']
     AC_ID = domain.content['area_chairs_id']['value']
     REV_ID = domain.content['reviewers_id']['value']
-    ETHICS_REV_ID = domain.content['ethics_chairs_id']['value'].replace(
-      domain.content['ethics_chairs_name']['value'],
-      domain.content['ethics_reviewers_name']['value']
-    )
+    ETHICS_REV_ID = domain.content['ethics_reviewers_id']['value']
     user = client.get_profile(edit.signatures[0]).id
 
     edge_readers = [CONFERENCE_ID]
--- a/openreview/conference/helpers.py
+++ b/openreview/conference/helpers.py
@@ -19,7 +19,7 @@ def get_venue(client, venue_id, support_user='OpenReview.net/Support'):
     venue.use_senior_area_chairs = 'senior_area_chairs_id' in domain.content
     venue.use_secondary_area_chairs = 'secondary_area_chairs_name' in domain.content
     venue.use_ethics_chairs = 'ethics_chairs_id' in domain.content
-    venue.use_ethics_reviewers = 'ethics_reviewers_name' in domain.content
+    venue.use_ethics_reviewers = 'ethics_reviewers_id' in domain.content or 'ethics_reviewers_name' in domain.content
     venue.use_publication_chairs = 'publication_chairs_id' in domain.content
     venue.automatic_reviewer_assignment = domain.content.get('automatic_reviewer_assignment', {}).get('value')
     venue.senior_area_chair_roles = domain.content.get('senior_area_chair_roles', {}).get('value', ['Senior_Area_Chairs'])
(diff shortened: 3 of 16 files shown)


openreview/workflows/process/compute_affinity_scores_process.py
```diff
@@ -12,11 +12,25 @@ def process(client, invitation):
...
+        if status_invitation_id:
+            # post status to request form
+            client.post_note_edit(
+                invitation=status_invitation_id,
+                signatures=[venue_id],
+                note=openreview.api.Note(
+                    signatures=[venue_id],
+                    content={
+                        'title': { 'value': f'{committee_name.replace("_", " ").title()} Affinity Scores Computation Failed' },
+                        'comment': { 'value': f'The process "{invitation.id.split("/")[-1].replace("_", " ").title()}" was scheduled to run, but we found no valid affinity score model to use. Please select a valid model and re-schedule this process to run at a later time.' }
```
melisabok: could you give more instruction about how to select the model?

Go to the workflow timeline UI, select Affinity Score Computation, click on edit, and select model, etc

openreview/workflows/process/deploy_assignments_process.py
```diff
@@ -8,24 +8,34 @@ def process(client, invitation):
...
+        if status_invitation_id:
+            # post status to request form
+            client.post_note_edit(
+                invitation=status_invitation_id,
+                signatures=[venue_id],
+                note=openreview.api.Note(
+                    signatures=[venue_id],
+                    content={
+                        'title': { 'value': f'{committee_name.replace("_", " ").title()} Assignment Deployment Failed' },
+                        'comment': { 'value': f'The process "{invitation.id.split("/")[-1].replace("_", " ").title()}" was scheduled to run, but we found no valid match configuration to deploy. Please select a valid match and re-schedule this process to run at a later time.' }
```
melisabok: add instructions in the email

openreview/workflows/workflow_process/venue_comment_process.py
```diff
@@ -44,10 +44,14 @@ def process(client, edit, invitation):
...
-        message=f'''A comment was posted to a service request.{comment_content}'''
-    )
\ No newline at end of file
+    #send email to support if commment comes from PCs
+    if comment.signatures[0].startswith('~'):
+        print('Sending email to support')
+        client.post_message(
+            invitation=f'{support_user}/-/Edit',
+            recipients=[support_user],
+            subject=f'''Comment posted to a request for service: {forum_note.content['title']['value']}''',
```
melisabok: now that we have as specific email notification to support, can we have the venue id in the subject so it is easier to find in the help scout view?

Something like "ICLR.cc/2026/Conference posted a comment [title]

put the link to the venue workflow ui in the comment text so it is easier to go to the config page

cc @racheljsmart

openreview/workflows/workflows.py
```diff
@@ -534,4 +535,121 @@ def set_conference_review_comment(self):
...
+        self.post_invitation_edit(invitation)
+
+    def set_conference_review_status_comment(self):
+
+        support_group_id = self.support_group_id
+        comment_invitation_id = f'{support_group_id}/Venue_Request/Conference_Review_Workflow/-/Status'
+
+        invitation = Invitation(id=comment_invitation_id,
+            invitees=[support_group_id],
+            readers=['everyone'],
```
melisabok: no need to be everyone here

openreview/workflows/workflows.py
```diff
@@ -534,4 +535,121 @@ def set_conference_review_comment(self):
...
+                            'replyto': '${4/content/noteId/value}',
+                            'ddate': {
+                                'param': {
+                                    'range': [ 0, 9999999999999 ],
+                                    'optional': True,
+                                    'deletable': True
+                                }
+                            },
+                            'signatures': ['${3/signatures}'],
+                            'readers': [support_group_id, '${5/content/venueId/value}/Program_Chairs'],
```
melisabok: I thought the status comment was going to be visible to us only, right?

openreview/workflows/workflows.py
```diff
@@ -534,4 +535,121 @@ def set_conference_review_comment(self):
...
+                                'type': 'string'
+                            }
+                        }
+                    }
+                },
+                'replacement': True,
+                'invitation': {
+                    'id': f'{support_group_id}/Venue_Request/Conference_Review_Workflow' + '${2/content/noteNumber/value}' + '/-/Status',
+                    'signatures': [self.super_id],
+                    'readers': ['everyone'],
```
melisabok: no need to be everyone

openreview/workflows/workflows.py
celestemartinez: This status invitation is the invitation to post a comment to the request form when a date process runs with missing fields (like the deployment). This is related to #2831

openreview/workflows/process/compute_affinity_scores_process.py
```diff
@@ -12,11 +12,26 @@ def process(client, invitation):
     venue_id = domain.id
     committee_name = invitation.get_content_value('committee_name')
     committee_id = f'{venue_id}/{committee_name}'
+    status_invitation_id = domain.get_content_value('status_invitation_id')
 
     support_user = domain.content['request_form_invitation']['value'].split('/Venue_Request')[0]
 
     affinity_scores_model = invitation.get_content_value('affinity_score_model')
     if not affinity_scores_model:
+        if status_invitation_id:
```
melisabok: can we create the status invitation id for all the venues so we avoid this if?

melisabok: can the status invitation be one for all the venues?

If we are going to use from the process functions, I believe it is possible.

openreview/workflows/workflows.py
```diff
@@ -534,4 +535,84 @@ def set_conference_review_comment(self):
...
+
+        support_group_id = self.support_group_id
+        status_invitation_id = f'{support_group_id}/Venue_Request/Conference_Review_Workflow/-/Status'
+
+        invitation = Invitation(id=status_invitation_id,
+            invitees=['active_venues'],
+            readers=['active_venues'],
+            writers=[support_group_id],
+            signatures=[support_group_id],
+            edit = {
```
celestemartinez, line 551: Using only one invitation for all request forms seems to work. @melisabok could you check the signatures and readers and correctly defined here? I am unsure if this is the right way to do it.

openreview/workflows/workflows.py
```diff
@@ -534,4 +535,84 @@ def set_conference_review_comment(self):
...
+                'signatures': {
+                    'param': {
```
melisabok, line 553: is this a param because it needs to be signed by the venue id?

celestemartinez: correct

PR #2856, opened 2026-02-13
ARR: Add pre-process to Submission Metadata Revision

haroldrubio:
This PR adds a pre-process to the Submission Metadata Revision invitation by adding a new function that creates the invitation in the ARR invitation builder. The ARR stage for this invitation specifically points to this function for creation.

### DIFF
--- a/openreview/arr/helpers.py
+++ b/openreview/arr/helpers.py
@@ -1878,10 +1878,7 @@ class ARRStage(object):
                         client, venue, invitation_builder, request_form_note
                     )
                 invitation_builder.set_process_invitation(self)
-            elif self.type in [
-                ARRStage.Type.SUBMISSION_REVISION_STAGE,
-                ARRStage.Type.SUBMISSION_METADATA_REVISION_STAGE
-            ]:
+            elif self.type == ARRStage.Type.SUBMISSION_METADATA_REVISION_STAGE:
                 invitation_builder.set_submission_metadata_revision_invitation(self)
 
             if self.extend:


openreview/arr/invitation.py
```diff
@@ -237,6 +237,156 @@ def set_desk_rejection_invitation(self):
...
+    script = meta_invitation.content['revision_preprocess_script']['value']
+    funcs = {
+        'openreview': openreview
+    }
+    exec(script, funcs)
+    funcs['process'](client, edit, invitation)
+'''
+            invitation.content['revision_preprocess_script'] = {'value': self.get_process_content(revision_stage.preprocess_path)}
+
+        if submission_license and revision_stage.allow_license_edition:
```
melisabok: do you need to ask for a license?

openreview/arr/helpers.py
```diff
@@ -1877,6 +1867,8 @@ def set_stage(self, client_v1, client, venue, invitation_builder, request_form_n
                         client, venue, invitation_builder, request_form_note
                     )
                 invitation_builder.set_process_invitation(self)
+            elif self.type == ARRStage.Type.SUBMISSION_REVISION_STAGE:
```
melisabok: can the type be more specific to the metadata invitation? this type can be used for other submission revision stages in the future

openreview/arr/invitation.py
```diff
@@ -237,6 +237,156 @@ def set_desk_rejection_invitation(self):
...
+        venue_id = self.venue_id
+        submission_license = self.venue.submission_license
+
+        revision_stage = openreview.stages.SubmissionRevisionStage(
+            name='Submission_Metadata_Revision',
+            source={'venueid': self.venue.get_submission_venue_id()},
+            start_date=arr_stage.start_date,
+            due_date=arr_stage.due_date,
+            exp_date=arr_stage.exp_date,
+            remove_fields=arr_stage.stage_arguments.get('remove_fields', ['authors', 'authorids', 'pdf']),
```
melisabok: it looks like you already removed the fields in the arr_stage, why are you doing it again?

openreview/arr/invitation.py
```diff
@@ -237,6 +237,156 @@ def set_desk_rejection_invitation(self):
...
+            only_accepted=arr_stage.stage_arguments.get('only_accepted', False),
+            allow_author_reorder=arr_stage.stage_arguments.get('allow_author_reorder', openreview.stages.AuthorReorder.DISALLOW_EDIT),
+            preprocess_path='process/submission_preprocess.py'
```
melisabok, line 253: submission_preprocess should be a JS function in the future please

openreview/arr/invitation.py
haroldrubio: At this stage I don't think I need to, but I'll do some testing about this

Edit: Seems like I was able to omit this block

tests/test_arr_venue_v2.py
```diff
@@ -3052,6 +3057,28 @@ def test_metadata_edit(self, client, openreview_client, helpers, test_client, re
...
+            'reassignment_request_area_chair': {'value': 'No, I want the same area chair from our previous submission (subject to their availability).'},
+            'reassignment_request_reviewers': {'value': 'Yes, I want a different set of reviewers'},
+            'justification_for_not_keeping_action_editor_or_reviewers': {'value': 'metadata edit justification'}
+        })
+        with pytest.raises(openreview.OpenReviewException, match=r'The provided URL points to a submission in the current cycle. Please provide a link to a previous ARR submission.'):
+            test_client.post_note_edit(
+                invitation=f"aclweb.org/ACL/ARR/2023/August/Submission1/-/Submission_Metadata_Revision",
+                signatures=['aclweb.org/ACL/ARR/2023/August/Submission1/Authors'],
+                note=openreview.api.Note(
+                    content=current_cycle_link_content
```
melisabok, line 3110: I would prefer to create the metadata content from scratch to visualize the fields that are being edited in this test.

openreview/arr/invitation.py
haroldrubio: Created an issue to do this in the future

openreview/arr/helpers.py
```diff
@@ -1877,6 +1878,11 @@ def set_stage(self, client_v1, client, venue, invitation_builder, request_form_n
                         client, venue, invitation_builder, request_form_note
                     )
                 invitation_builder.set_process_invitation(self)
+            elif self.type in [
+                ARRStage.Type.SUBMISSION_REVISION_STAGE,
```
melisabok: why do you need the type SUBMISSION_REVISION_STAGE in this if?

tests/test_arr_venue_v2.py
melisabok, line 3062: how do you get two different error messages without passing a different content?

openreview/arr/helpers.py
haroldrubio: Added a specific Submission Metadata Revision type

openreview/arr/invitation.py
haroldrubio: Removed the redundant hardcoded remove fields array and only relying on the ARRStage

tests/test_arr_venue_v2.py
haroldrubio: Wrote the metadata content to a variable to visualize the values in the test

openreview/arr/helpers.py
haroldrubio: I don't need it there - I've removed it

tests/test_arr_venue_v2.py
haroldrubio: The error is a multi-error for all of those additional fields. Should I just have a single assert with the full multi-error string?

melisabok: ohhh I see, makes sense. I think you can keep it, maybe the match should be a regex that matches both errors so you don't need to post_note_edit twice.

melisabok (approved):

openreview/arr/helpers.py
```diff
@@ -1483,12 +1473,14 @@ class Type(Enum):
             STAGE_NOTE (2): Built-in OpenReview stage that's available on the request form
             PROCESS_INVITATION (3): An invitation that stores an ARR script in the form of a process function
             SUBMISSION_REVISION_STAGE (4): An invitation that allows revisions to the submission
+            SUBMISSION_METADATA_REVISION_STAGE (5): An invitation that allows metadata-only revisions to the submission
```
celestemartinez, line 1476: What exactly is the difference between the `SUBMISSION_REVISION_STAGE` and `SUBMISSION_METADATA_REVISION_STAGE` stages?

haroldrubio: I need to treat the Submission Metadata Revision separately because it calls a specific ARR invitation builder function that sets Super User as signatures and writers of the child invitations so that it can get notes across cycles.

There's a specific Submission Revision Stage here because they need special handling since the stage can be used for many Submission Revision invitations

celestemartinez: I see, thanks! That's what I thought but wanted to make sure I was understanding this properly.

celestemartinez (approved):

PR #2858, opened 2026-02-13
ignore email notifications to venue organizers if they are inviters o…

melisabok:
…f assignment invitations

### DIFF
--- a/openreview/venue/process/paper_recruitment_process.py
+++ b/openreview/venue/process/paper_recruitment_process.py
@@ -92,7 +92,7 @@ Confirmation of the assignment is pending until the invited reviewer creates a p
 OpenReview Team'''
 
             ## - Send email
-            response = client.post_message(subject, edge.signatures, message, invitation=meta_invitation_id, signature=venue_id, replyTo=contact, sender=sender)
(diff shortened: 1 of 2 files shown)


openreview/venue/process/paper_recruitment_process.py
```diff
@@ -92,7 +92,7 @@ def process(client, edit, invitation):
 OpenReview Team'''
 
             ## - Send email
-            response = client.post_message(subject, edge.signatures, message, invitation=meta_invitation_id, signature=venue_id, replyTo=contact, sender=sender)
+            response = client.post_message(subject, edge.signatures, message, invitation=meta_invitation_id, signature=venue_id, replyTo=contact, sender=sender, ignoreRecipients=[venue_id])
```
melisabok, line 95: @celestemartinez I'm not sure if this is the correct fix. Could it be that a PC is also an AC and would like to receive an email with the invited response?

If that is the case then I should do an if instead:

```
if venue_id not in edge.signatures or pc_group_id not in edge.signature:
     send the email
```

melisabok:
closing this PR, discussed with Celeste and Harold that PCs don't need it.

PR #2862, opened 2026-02-17
ARR: Avoid posting SAC loads in setup script

haroldrubio:
This PR removes the step in the SAC matching setup script that computes Custom_Max_Papers edges for each SAC based on tracks. The SAC matching edges are posted independently of the matcher so this step is no longer needed.

### DIFF
--- a/openreview/arr/management/setup_sae_matching.py
+++ b/openreview/arr/management/setup_sae_matching.py
@@ -19,7 +19,6 @@ def process(client, invitation):
     request_form_id = domain.content['request_form_id']['value']
     previous_url_field = 'previous_URL'
     reviewers_id = domain.content['reviewers_id']['value']
-    meta_invitation_id = domain.content['meta_invitation_id']['value']
     senior_area_chairs_id = domain.content['senior_area_chairs_id']['value']
     tracks_field_name = 'research_area'
 


openreview/arr/management/setup_sae_matching.py
```diff
@@ -95,52 +96,19 @@ def process(client, invitation):
...
-                )
-            )
-        client.delete_edges(
-            invitation=role_cmp_inv,
-            soft_delete=True,
-            wait_to_finish=True
-        )
-        openreview.tools.post_bulk_edges(client=client, edges=cmp_to_post)
+    # SAC maximum loads handled by ARR matching scripts
+    custom_max_papers_invitation = client.get_invitation(f"{senior_area_chairs_id}/-/Custom_Max_Papers")
```
melisabok: please use `venue.get_custom_max_papers_id(committee_id)`  to build the invitation id

openreview/arr/management/setup_sae_matching.py
```diff
@@ -95,52 +96,19 @@ def process(client, invitation):
...
+    weight_param = custom_max_papers_invitation.edit['weight']['param']
+    if 'enum' in weight_param:
```
melisabok: the reason why we need an enum is because the edge browser renders a dropdown with the numbers from 1 to 100. If you remove this, the PCs won't be able to edit the quota using the UI.

I would user higher number if needed or ask Xukun to implement a textbox when enum is present

haroldrubio: Oh yes that's right. Well the change to minimum isn't necessary, I can remove it

openreview/arr/management/setup_sae_matching.py
```diff
@@ -19,6 +19,7 @@ def process(client, invitation):
     request_form_id = domain.content['request_form_id']['value']
     previous_url_field = 'previous_URL'
     reviewers_id = domain.content['reviewers_id']['value']
+    meta_invitation_id = domain.content['meta_invitation_id']['value']
```
melisabok: this PR is removing code only, why do you need to create this variable?

haroldrubio: Missed this when I was changing the PR to avoid modifying the Custom Max Papers, I removed the extra variable

melisabok (approved):

PR #2863, opened 2026-02-18
ARR: Handle no previous Reviewers/Area_Chairs group

haroldrubio:
This PR uses `tools.get_group` to fetch `<previous_cycle>/SubmissionX/Reviewers` and `<previous_cycle>/SubmissionX/Area_Chairs` to return `None` instead of erroring.

Replicating this issue in the tests when a submission is deleted before the submission deadline passes

### DIFF
--- a/tests/test_arr_venue_v2.py
+++ b/tests/test_arr_venue_v2.py
@@ -4883,18 +4883,12 @@ reviewerextra2@aclrollingreview.com, Reviewer ARRExtraTwo
         venue = openreview.helpers.get_conference(client, request_form.id, 'openreview.net/Support')
         submissions = pc_client_v2.get_notes(invitation='aclweb.org/ACL/ARR/2023/August/-/Submission', sort='number:asc')
         ethics_client = openreview.api.OpenReviewClient(username = 'reviewerethics@aclrollingreview.com', password=helpers.strong_password)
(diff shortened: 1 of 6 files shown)


melisabok (approved):

tests/test_arr_venue_v2.py
```diff
@@ -1637,6 +1637,76 @@ def test_submission_preprocess(self, client, openreview_client, test_client, hel
...
+            signatures=['~SomeFirstName_User1'],
+            note=openreview.api.Note(content=deleted_source_content)
+        )
+        delete_edit = openreview_client.post_note_edit(
+            invitation='aclweb.org/ACL/ARR/2023/June/-/Edit',
+            readers=['aclweb.org/ACL/ARR/2023/June/Program_Chairs'],
+            writers=['aclweb.org/ACL/ARR/2023/June/Program_Chairs'],
+            signatures=['aclweb.org/ACL/ARR/2023/June/Program_Chairs'],
+            note=openreview.api.Note(
+                id=desk_reject_submission_edit['note']['id'],
```
melisabok: it looks like this variable `desk_reject_submission_edit` doesn't exist. How did you make this test pass?

haroldrubio: I ran the tests but removed some logic and renamed variables from a different attempt to reproduce the issue before pushing, I missed this variable rename but it should be fixed now

PR #2864, opened 2026-02-18
New venue configuration UI: recruitment request fixes

melisabok:
- Restrict the invitee list to 50k characters
- Validate that emails have the correct format
- Remind invitees that are in the invited group, ignore the rest


openreview/workflows/templates.py
```diff
@@ -708,6 +708,7 @@ def setup_committee_group_recruitment_template_invitation(self):
     funcs = {
         'openreview': openreview,
         'datetime': datetime,
+        're': re
```
xkopenreview, line 711: it throws the following process log error
"NameError: name 're' is not defined. Did you forget to import 're'?"

melisabok: I fixed it by updating the invitations

PR #2868, opened 2026-02-19
Client: Use stream not concurrent for edges

haroldrubio:


haroldrubio:
`get_all_tags` is also doing a `concurrent_get` call - should this be changed to use `stream=true` like edges? I'm not sure if this is supported for the tags endpoint

PR #2873, opened 2026-02-20
Fix/ Venue request: add a preprocess to check if venue id has already been used

### LINKED ISSUE #2807: Deployment preprocess: Check if venue ID already exists
This is to prevent us from accidentally overwriting an existing venue. Maybe we can check for membership in the venues or active_venues groups. We should also consider what happens if there's a deployment error and we need to redeploy.

celestemartinez:
Fixes #2807

This PR adds two preprocesses, one to each deployment (old request form and new UI request form). Before deployment happens, this preprocess checks if the venue group exists. If it does, it checks that the `request_form_id` in the venue_group.content matches the id of the request form being deployed. If it doesn't match, an error is thrown.


openreview/venue_request/process/deploy_pre_process.py
```diff
@@ -0,0 +1,18 @@
+def process(client, note, invitation):
+
+    venue_id = note.content.get('venue_id', '')
+
+    baseurl_v2 = 'http://localhost:3001'
```
celestemartinez: @melisabok is there a better way to do this? For the old request form, I need to use the api2 client to get the domain group otherwise I get the group but not the content. 
Or could I make get_group() using api1 return the content if it exists?

melisabok: you need to always use api 2 to get group. 

please use the tools functions to get the urls: https://github.com/openreview/openreview-py/blob/master/openreview/tools.py#L1912

celestemartinez: Thanks! Fixed

PR #2874, opened 2026-02-20
make sure submission name is parametrized

melisabok:
Remove the hardcoded 'Submission' string and use the value of submission_name available in the domain group. Make sure chinese characters are supported


xkopenreview:
I'm not sure if this is correct, i think we need to know the rational of pc making the change
because it's not just 'Submission' string. some logic relies on the 's' plural form in group which is meaningless in Chinese so the logic would likely break.

if the pc has difficulty understand the interface it might be better for pc to use a translation extension
or for openreview to support internationalization

PR #2898, opened 2026-02-27
Feat/Journal: skip review

celestemartinez:
I am adding a new button to each forum, `Official Recommendation Enabling`, which:
- is enabled if the journal does not skip the official recommendation stage
- allows EICs to move to the discussion and official recommendation period with fewer than 3 reviews
- ~is enabled on the duedate of the review invitation (so I added a dateprocess that runs at the review duedate)~
- is enabled at the same time as the review stage and EICs can decide when to move to the next stage

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,42 +13,11 @@ parameters:
     type: string
     default: "3.13.1"
 
-jobs:
-  # test job
-  install-package:
-    parameters:
-      python-version:
-        type: string
-    docker:
-      - image: cimg/python:<<parameters.python-version>>-node
+# Reusable commands to reduce duplication between jobs
+commands:
+  setup-mongodb:
+    description: "Initialize MongoDB replica set"
     steps:
-      - checkout
-      - run:
-          name: Install Dependencies
-          command: |
-            pip install -e .
-      - run:
-          name: Verify Installation
-          command: python -c "import openreview"
-  build:
-    # The resource_class feature allows configuring CPU and RAM resources for each job. Different resource classes are available for different executors. https://circleci.com/docs/2.0/configuration-reference/#resourceclass
-    resource_class: large
-    parallelism: 10
-    working_directory: ~/openreview-py-repo
-    parameters:
-      python-version:
-        type: string
-    docker:
-      - image: cimg/python:<<parameters.python-version>>-node
-      - image: cimg/redis:7.2
-      - image: mongo:8.0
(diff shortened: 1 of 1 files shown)


openreview/journal/process/official_recommendation_enabling_process.py
```diff
@@ -0,0 +1,121 @@
+def process(client, edit, invitation):
+
+    journal = openreview.journal.Journal()
+
+    submission = client.get_note(edit.note.forum)
```
melisabok, line 5: is this the same code than the review_process? can you put it in a single place?

openreview/journal/invitation.py
```diff
@@ -3904,6 +3906,18 @@ def set_review_invitation(self):
...
 
+        if not self.journal.should_skip_official_recommendation():
+            # allow EICs to enable official recommendation
+            invitation_content['duedate_script'] = {
+                'value': self.get_process_content('process/review_duedate_process.py')
+            }
+            invitation['dateprocesses'].append(
+                {
+                    'dates': [ "#{4/duedate} + 1000" ],
+                    'script': self.get_super_dateprocess_content('duedate_script')
```
melisabok: I'm not sure about creating the invitation when the review duedate is reached, the duedate changes for every new reviewer assignment made. 
What about creating it when the review invitation is created? it is up to the EIC when to enable it.

celestemartinez: I see, I didn't realize the review deadline changed. Yes, I can create it when the review invitation is created.

openreview/journal/process/official_recommendation_enabling_process.py
celestemartinez: it is similar, but the review process has more code that does things like notifying the readers and updating Pending_Review edges, so I don't think I can easily have this in one process function. Do you mean adding the repeated code to a function in the journal and calling it from both process functions?

melisabok: yes, we can put the repeated code in a journal method or as a process function stored in an object.content

openreview/journal/journal.py
celestemartinez: @melisabok what is this method for? Is this to manually update all existing invitations? I don't think we are calling this from anywhere in the code.

I am trying to understand if I need to add anything here.

melisabok: I used it to migrate some data in the past when we renamed the Solicit invitations, you can remove the code.

PR #2905, opened 2026-03-02
ARR: Automate Author Response Extension

### LINKED ISSUE #2784: ARR: Automate reviewer response extension
There are some steps that were automated for the October cycle that the workflow chairs should be able to do programmatically:

1. Enable/Disable Author Response (remove Authors invitee and signature from Official_Comment)
2. Enable/Disable Reviewer Response (remove Authors from readers from Official_Comment)
3. Extend Review_Issue_Report (or arbitrarily any submission invitation)

I was thinking of creating some kind of super invitation where the edit content contains the note ID, note number, author response date, reviewer response date and a JSON that maps from invitation name to the extended date. There would also be a date process to check that the invitations weren't prematurely closed by the stages

haroldrubio:
- Fixes #2784

This PR adds an invitation with a cronjob that automates the following logic for papers who receive their reviews after the official author response period ends:

- Extend the author response deadline to 3 days after the third review is received.
- Extend the reviewer response deadline to 4 days after the third review is received.
- Extend the review issue report deadline to 5 days after the third review is received.


openreview/arr/management/setup_author_response_extension.py
```diff
@@ -0,0 +1,337 @@
+def process(client, invitation):
```
melisabok, line 1: this code seems very large, what about moving it to the ARR class and call the class from here?

It can also be used by the ARR tech team if they need to run it on demand

openreview/arr/helpers.py
```diff
@@ -404,6 +404,24 @@ class ARRWorkflow(object):
             "order": 59,
             "required": False
         },
+        "author_response_extension_start_date": {
```
melisabok: what is the relation of these dates with the official comment stage?

This stage should only be running when the comment stage starts, right?

What if we create the invitation to become active as soon as the comment stage becomes active?

openreview/arr/management/setup_author_response_extension.py
haroldrubio: I'll work on moving the logic to the ARR class that's a good idea

openreview/arr/helpers.py
haroldrubio: This is different from the comment stage start and end dates - the official comment outside of the rebuttal/author response is used for comments between reviewers and area chairs/senior area chairs, so I don't think I can tie this to any existing dates

But yes that is true, that this only runs whenever the comment stage is active

openreview/arr/arr.py
```diff
@@ -167,6 +167,259 @@ def set_arr_stages(self, configuration_note):
...
+                )
+                print(f'  - Updated Official_Comment invitation for paper {paper_number}')
+            else:
+                print(f'  - Official_Comment already open for paper {paper_number}')
+
+        except Exception as e:
+            print(f'  - Error updating Official_Comment for paper {paper_number}: {str(e)}')
+
+        # Open Review_Issue_Report invitations for each review (1-10)
+        for review_number in range(1, 11):
```
melisabok: is it safe to use a range of numbers? some reviews can be posted and then deleted. Why you do not iterate through all the submission reviews?

openreview/arr/management/setup_author_response_extension.py
```diff
@@ -0,0 +1,35 @@
...
+    request_form_id = domain.content['request_form_id']['value']
+
+    client_v1 = openreview.Client(
+        baseurl=openreview.tools.get_base_urls(client)[0],
+        token=client.token
+    )
+
+    request_form = client_v1.get_note(request_form_id)
+    support_group = request_form.invitation.split('/-/')[0]
+    venue = openreview.helpers.get_conference(client_v1, request_form_id, support_group)
```
melisabok: no need to pass the support_group, there is already a default value for this.

openreview/arr/management/setup_author_response_extension.py
```diff
@@ -0,0 +1,33 @@
...
+    domain = client.get_group(invitation.domain)
+    request_form_id = domain.content['request_form_id']['value']
+
+    client_v1 = openreview.Client(
+        baseurl=openreview.tools.get_base_urls(client)[0],
+        token=client.token
+    )
+
+    venue = openreview.helpers.get_conference(client_v1, request_form_id)
+
```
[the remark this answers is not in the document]
haroldrubio: Resolved in a previous commit to use the default value from the function

tests/test_arr_venue_v2.py
```diff
@@ -6141,6 +6141,242 @@ def test_review_issue_forms(self, client, openreview_client, helpers, test_clien
...
+        assert process_inv.content['review_issue_report_delay_ms']['value'] == 432000000
+
+        ## Step 2: Test < 3 reviews case — process should re-open Official_Comment for papers with < 3 reviews
+        ## Pick a submission with < 3 reviews
+        few_reviews_submission = submissions[1]
+
+        authors_group = 'aclweb.org/ACL/ARR/2023/August/Submission2/Authors'
+        comment_inv = openreview_client.get_invitation('aclweb.org/ACL/ARR/2023/August/Submission2/-/Official_Comment')
+        assert authors_group in comment_inv.invitees
+        assert authors_group in comment_inv.readers
```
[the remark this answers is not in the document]
haroldrubio: Added an additional assertion after getting the submission

openreview/arr/arr.py
```diff
@@ -174,6 +174,335 @@ def set_arr_stages(self, configuration_note):
...
+                    ]
+                    if 'edit' not in edit_params:
+                        edit_params['edit'] = {}
+                    if 'note' not in edit_params['edit']:
+                        edit_params['edit']['note'] = {}
+                    edit_params['edit']['note']['readers'] = {
+                        'param': {'enum': filtered_readers}
+                    }
+                    needs_update = True
+                    print(f'  - Removing Authors from Official_Comment inner readers')
```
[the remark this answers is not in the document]
haroldrubio: Added a block of code to re-open the official comment, and added a test to re-produce this behavior

openreview/arr/helpers.py
```diff
@@ -684,6 +684,22 @@ def __init__(self, client_v2, venue, configuration_note, request_form_id, suppor
...
+                    'content': {
+                        'author_response_delay_ms': {'value': 259200000},      # 3 days
+                        'reviewer_response_delay_ms': {'value': 345600000},    # 4 days
+                        'review_issue_report_delay_ms': {'value': 432000000},  # 5 days
+                    }
+                },
+                start_date=self.configuration_note.content.get('author_response_extension_start_date'),
+                exp_date=self.configuration_note.content.get('author_response_extension_end_date'),
+                process='management/setup_author_response_extension.py',
+                cron=self.configuration_note.content.get('author_response_extension_cron')
```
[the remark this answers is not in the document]
haroldrubio: Added `author_response_extension_cron` as a required field to avoid building the invitation without a valid cron

melisabok (approved):

openreview/arr/helpers.py
```diff
@@ -321,6 +321,24 @@ class ARRWorkflow(object):
...
+            "order": 58,
+            "required": False
+        },
+        "author_response_extension_end_date": {
+            "description": "When should the author response extension manager stop running?",
+            "value-regex": "^[0-9]{4}\\/([1-9]|0[1-9]|1[0-2])\\/([1-9]|0[1-9]|[1-2][0-9]|3[0-1])(\\s+)?((2[0-3]|[01][0-9]|[0-9]):[0-5][0-9])?(\\s+)?$",
+            "order": 59,
+            "required": False
+        },
+        "author_response_extension_cron": {
```
celestemartinez, line 337: Could we have a default expression here? I am worried about PCs knowing how to define this.

openreview/arr/arr.py
```diff
@@ -175,6 +175,379 @@ def set_arr_stages(self, configuration_note):
...
+            except Exception as e:
+                print(f'Error processing paper {paper_number}: {str(e)}')
+                continue
+
+        print(f'Completed author response extension management for {self.id}')
+
+    def _get_review_issue_report_invitations(self, reviews):
+        for review in reviews:
+            review_number = review.number
+            if review_number is None or not review.invitations:
```
celestemartinez: Why do we need this check here?

openreview/arr/arr.py
```diff
@@ -175,6 +175,379 @@ def set_arr_stages(self, configuration_note):
...
+
+            # Check inner readers (note readers)
+            current_readers = invitation.edit['note']['readers']['param']['enum']
+            if authors_group not in current_readers:
+                if 'edit' not in edit_params:
+                    edit_params['edit'] = {}
+                edit_params['edit']['note'] = {
+                    'readers': {
+                        'param': {
+                            'enum': current_readers + [authors_group]
```
celestemartinez, line 311: It looks like we use enum here for readers but only after the ethics stage is run? Before that we use items [for example](https://openreview.net/invitation/edit?id=aclweb.org/ACL/ARR/2026/March/Submission994/-/Official_Comment)

Will this always run afer the ethics review stage?

openreview/arr/arr.py
```diff
@@ -175,6 +175,379 @@ def set_arr_stages(self, configuration_note):
...
+        try:
+            invitation = self.client.get_invitation(
+                self.get_invitation_id(name=comment_name, number=paper_number)
+            )
+
+            needs_update = False
+            edit_params = {}
+            authors_group = self.get_authors_id(paper_number)
+
+            if now_millis <= author_response_close:
```
celestemartinez, line 416: do authors need to be added to the note readers here, or is that not necessary?

openreview/arr/arr.py
haroldrubio: Yes it will always run after the ethics review stage is run - should it stay as enum?

openreview/arr/arr.py
haroldrubio: Yes I think you're right, I added an `or now_millis <= author_response_close` to where I added authors to the note readers

openreview/arr/helpers.py
haroldrubio: Added a default to run twice per day

openreview/arr/arr.py
haroldrubio: I don't think we need it anymore so I removed it

openreview/arr/arr.py
celestemartinez: I think it can stay as enum. Perhaps I can try working on using items even for after running the ethics stage but that can be another PR

openreview/arr/arr.py
```diff
@@ -175,6 +175,376 @@ def set_arr_stages(self, configuration_note):
...
+        cdate = invitation.cdate
+
+        # Check if invitation is active
+        if cdate > now:
+            print(f'Author response extension process not yet active, cdate: {cdate}')
+            return
+
+        # Fetch all submissions with replies
+        submissions = self.client.get_all_notes(
+            invitation=self.get_submission_id(),
```
celestemartinez: Looks like this will happen for withdrawn/desk-rejected papers as well, is that an issue?

celestemartinez (approved):

PR #2908, opened 2026-03-03
ARR: Merge Post-Submission and Preprint Release

### LINKED ISSUE #2809: ARR: Merge Preprint Release into Post Submission
Currently, the Preprint Release step happens separately from Post Submission. Since the Preprint Release step is configurable, there can be a race condition since these two processes are reading and updating the submissions.

These should be merged into 1 step to guarantee execution order while still providing guardrails for the preprints

Initial approach to mirror existing approaches to modify/extend Venue class behavior for ARR:
``` python
def create_submission_stage(self):
    stage_value = self.venue.create_submission_stage()

    ...

    # New behavior to Post_Submission date process to ARR script
    self._modify_post_submission_invitation()

    return stage_value

def create_post_submission_stage(self):
    self.venue.create_post_submission_stage()
    # New behavior to Post_Submission date process to ARR script
    self._patch_post_submission_invitation()

def _modify_post_submission_invitation(self):
    post_submission_id = self.venue.get_post_submission_id()
    inv = self.client.get_invitation(post_submission_id)

    arr_script = self.invitation_builder.get_process_content(
        'process/post_submission_process.py'
    )

    # inv.date_processes = same dates, but script swapped to arr_script
    # self.client.post_invitation_edit(... invitation=inv ...)
```

haroldrubio:
- Resolves #2809 
This PR merges the post submission and preprint release steps in the ARR workflow into an ARR-specific post-submission dateprocess


openreview/arr/arr.py
```diff
@@ -489,10 +489,35 @@ def create_submission_stage(self):
             replacement=False,
             invitation=invitation
         )
+        self._patch_post_submission_invitation()
         return stage_value
 
     def create_post_submission_stage(self):
-        return self.venue.create_post_submission_stage()
+        self.venue.create_post_submission_stage()
```
melisabok: what if you create the Post_Submission invitation here? or pass the process function path as a param?

openreview/arr/process/post_submission_process.py
```diff
@@ -0,0 +1,165 @@
...
+
+        client.post_note_edit(
+            invitation=meta_invitation_id,
+            readers=[venue_id],
+            writers=[venue_id],
+            signatures=[venue_id],
+            note=updated_note
+        )
+
+    openreview.tools.concurrent_requests(
```
melisabok: This seems to be two different types of post_submission invitations so I would try doing that:

1. Post_Submission: post edits for all the submissions where preprint == No. This can be specified as a source parameter in the same invitation. 
2. Preprint_Post_Submission: post edits for all the submissions where preprint == Yes. This can be specified as the source parameter in the same invitation

@celestemartinez this is something that we discussed yesterday and this is a good case to support it. 

1. specify the source in the invitation.content
2. process function get the submissions from the source
3 apply the edits

openreview/arr/invitation.py
```diff
@@ -198,18 +198,51 @@ def set_preferred_emails_invitation(self):
...
+    def set_preprint_post_submission_invitation(self):
+        return self.set_post_submission_invitation(
+            invitation_id=self.venue.get_preprint_post_submission_id(),
+            source={
+                'venueid': self.venue.get_submission_venue_id(),
+                'content':{
+                    'preprint': 'yes'
+                }
+            },
+            process_path='../arr/process/post_submission_process.py'
```
melisabok: do you really need a different process function here?

I thought that preprint is just making the readers to everyone, what else is missing?

haroldrubio: Yes you're right we don't need it there - I thought I would need it to set the `explanation_of_revisions_PDF` for resubmissions but that is already happening in the script to setup the reviewer matching data. It only needs to be set again for `preprint: yes`

openreview/arr/process/post_submission_process.py
```diff
@@ -0,0 +1,120 @@
+def process(client, invitation):
```
melisabok: do you need this file?

haroldrubio: We still need this process function to release the submissions to the public - but maybe I should change the name of the file since it's not for the regular post submission stage?

melisabok: but the Preprint Post Submission should be created with readers: everyone so no need to use a different process function, right?

haroldrubio: Hmm we could try using hide fields but some of fields have only the authors as readers and other fields have the committee as readers

melisabok: the goal of having a post submission invitation is to define the readers in the invitation itself more than doing it on the fly in process functions. What is not working about hiding the fields inside the preprint post submission invitation?

openreview/arr/invitation.py
```diff
@@ -213,17 +213,112 @@ def set_post_submission_invitation(self, invitation_id=None, source=None):
...
+                        'readers': {
+                            'param': {
+                                'const': {
+                                    'delete': True
+                                }
+                            }
+                        }
+                    }
+
+        for field in hide_fields_from_public:
```
melisabok, line 270: where is hide_fields_from_public being defined?

haroldrubio: It's being imported: https://github.com/openreview/openreview-py/blob/347724d16cca8f48b0412e428fc251ce12d44b5a/openreview/arr/invitation.py#L10

tests/test_arr_venue_v2.py
```diff
@@ -683,7 +689,7 @@ def test_june_cycle(self, client, openreview_client, helpers, test_client):
                 'reassignment_request_reviewers': { 'value': 'This is not a resubmission' },
                 'software': {'value': '/pdf/' + 'p' * 40 +'.zip' },
                 'data': {'value': '/pdf/' + 'p' * 40 +'.zip' },
-                'preprint': { 'value': 'yes'},
+                'preprint': { 'value': 'no'},
```
melisabok, line 720: why did you have to change this test?

tests/test_arr_venue_v2.py
```diff
@@ -2661,81 +2670,6 @@ def test_post_submission(self, client, openreview_client, helpers, test_client,
...
-            )
-        )
-
-        helpers.await_queue()
-        helpers.await_queue_edit(openreview_client, 'aclweb.org/ACL/ARR/2023/August/-/Preprint_Release_Submission-0-1', count=1)
-        ## Assert released preprints have hidden authors and authorids
-        ## check index 3 and 5 (even paper numbers) specifically
-        ## check index 4 has no author readers since they were not released
-
-        submissions = pc_client_v2.get_notes(invitation='aclweb.org/ACL/ARR/2023/August/-/Submission', sort='number:asc')
```
melisabok, line 2719: can you keep these assertions?

tests/test_arr_venue_v2.py
haroldrubio: The June ones have changed `preprint` to `no` because there are some tests that rely on the readers of the group to still be the assigned committee. On L692, it affects this assertion in the commitment tests:
https://github.com/openreview/openreview-py/blob/347724d16cca8f48b0412e428fc251ce12d44b5a/tests/test_arr_venue_v2.py#L6936

which fails because the readers are everyone. This is failing in this PR and not before, because the preprint release did not run for June. Now it does, because it runs with the post submission stage

melisabok (approved):

tests/test_arr_venue_v2.py
haroldrubio: Added assertions back after post submission is called

openreview/arr/process/post_submission_process.py
haroldrubio: Modified the invitation schema to hide the fields from the edit

openreview/venue/invitation.py
```diff
@@ -395,7 +395,7 @@ def set_submission_deletion_invitation(self, submission_revision_stage):
         self.save_invitation(expire_invitation, replacement=True)
         return invitation
 
-    def set_post_submission_invitation(self):
+    def set_post_submission_invitation(self, invitation_id=None, source=None, process_path='process/post_submission_process.py'):
```
celestemartinez: Why do you need to make these changes to the venue invitation? ARR is not using this `set_post_submission_invitation` anymore, right?

haroldrubio: Something I still need is the ability to change the source, so only the non-preprints go through the usual post submission invitation - but I removed the rest

celestemartinez (approved):

PR #2913, opened 2026-03-04
Fix/ New Venue UI: allow public submissions after submission deadline

celestemartinez:
Main changes:
1. allow PCs to select 'everyone' as readers in `Submission_Change_Before_Bidding/Reviewing`. Add a preprocess that checks the readers are selected correctly:
a. if `everyone ` is selected, no other reader options can be selected
b. if `everyone` is not selected, then PCs and paper suthors must be selected
2. Add odate and bibtex to submission as soon as it becomes public
3. Submission release will update a rejected paper's bibtex if the paper is public
4. Submission Release will no longer run automatically, PCs need to select which submissions to make public
a. a comment is posted to the request form if it runs without PC input
5. Similar to # 1, allow PCs to select everyone as readers of replies. 
6. Remove authors and everyone as readers option from stages like review, metareview, decision. The PCs should use the Official_Review_Release invitation to release to authors or the public (since this invitation removes the authors as nonreaders)
7. Create the Meta_Review_Release invitation (related to # 5)


openreview/workflows/process/submission_release.py
```diff
@@ -36,7 +57,6 @@ def get_source_submission_tuples(all_submissions):
         source_submissions = [(submission, openreview.api.Note.from_json(reply)) for submission in all_submissions for reply in submission.details['directReplies'] if f'{venue_id}/{submission_name}{submission.number}/-/{decision_name}' in reply['invitations']]
```
celestemartinez, line 57: @melisabok I realized the way I am getting submissions here already excludes submissions with no decision, so I just added a test to make sure this is working as expected.

openreview/workflows/process/edit_note_readers_preprocess.py
```diff
@@ -0,0 +1,17 @@
...
+    authors_name = domain.get_content_value('authors_name')
+    readers = edit.content['readers']['value']
+
+    is_submission_change_invitation = 'Submission_Change_Before' in invitation.id.split('/Readers')[0]
+
+    if 'everyone' in readers and len(readers) > 1:
+        raise openreview.OpenReviewException('The "everyone" reader option cannot be included with other reader options.')
+    
+    if 'everyone' not in readers:
+        if venue_id not in readers and not any(r for r in readers if r.endswith('/Program_Chairs')):
```
melisabok, line 14: if venue_id and venue_id/PCs two different options?

I think venue_id should be internally added always.

celestemartinez: I think this comes from the venue class stages, sometimes the PCs are added and sometimes the venue id is added, so the edit invitations work the same. I can make it more uniform across and always add the venue id. 

The option in the edit invitation depends on the super invitation, because if the venue is added in one but the PCs apears as option in the other, then the option does not appear as already selected and that is confusing.

tests/test_acs_and_reviewers.py
```diff
@@ -1148,19 +1270,26 @@ def test_release_submissions(self, openreview_client, helpers):
...
+        valid_bibtex = '''@misc{
+anonymous'''+str(year)+'''paper,
+title={Paper title 3},
+author={Anonymous},
+booktitle={Submitted to The EFGH Conference},
+year={'''+str(year)+'''},
+url={https://openreview.net/forum?id='''     
+
+        valid_bibtex = valid_bibtex +   submissions[0].forum + '''}
+}'''
```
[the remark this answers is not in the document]
celestemartinez: Fixed

openreview/workflows/process/submission_release.py
```diff
@@ -86,19 +106,31 @@ def edit_submission(submission_tuple):
...
         elif not note_accepted:
+            note_content = {
+                'venueid': {
+                    'value': rejected_venue_id
+                },
+                'venue': {
+                    'value': venue
+                }
+            }
+            if submission.content.get('bibtex', {}).get('value'):
```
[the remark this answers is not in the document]
celestemartinez: Fixed.

openreview/workflows/process/submission_release.py
```diff
@@ -86,19 +106,31 @@ def edit_submission(submission_tuple):
...
+                note_content['_bibtex'] = {
+                    'value': openreview.tools.generate_bibtex(
+                        note=submission,
+                        venue_fullname=title,
+                        year=str(datetime.datetime.now().year),
+                        url_forum=submission.forum,
+                        paper_status = 'rejected',
+                        anonymous=True
+                    )
```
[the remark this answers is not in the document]
celestemartinez: fixed.

openreview/workflows/process/edit_note_readers_preprocess.py
celestemartinez: @melisabok It looks like submissions are created with the venue as readers, but replies are created by default with the PC as readers. Should I always set the venue id as readers for all notes?

melisabok: I think we need to keep consistency, using the venue as reader is more general IMO but I do not want to break any existing behavior. I know that the comment stage has selectable readers and seeing the PC group is more clear than the venue.

I also think that if we impersonate the venue, then we can see everything so venue doesn't need to be reader of anything because it acts as a super user.

celestemartinez: I can see if this change would break too many things. For the comment invitation, we can use description for the venue id so users keep seeing the PCs as readers in the note editor.

melisabok: I remember the API is validating this, is the PC is not selected but readers is everyone then no need to add the optional: False PC member.

melisabok (approved):

PR #2919, opened 2026-03-10
Workflow Timeline: run setup_matching on deployment for all groups

### LINKED ISSUE #2917: Reviewer Assignment Deployment: Can run before Assignment invitations are created
There was a case where Reviewers_Assignment_Deployment ran and called set_assignments() but the Proposed Assignment invitation wasn't created yet, causing a process function error. We should check if the assignment invitation exists first.

celestemartinez:
This PR:
- fixes and tests the AC assignments deployment
- runs setup matching on deployment (fixes #2917) and skips submission and reviewer checks for new UI venues
- removes the `Assignment_Configuration` template, since this invitation is created by the setup matching
- adds a process to the `Assignment_Configuration` invitation if the venue uses the new UI
- creates a new method `set_assignment_invitations` which is called during deployment and creates the necessary invitations for matching and deployment

### DIFF
--- a/openreview/arr/invitation.py
+++ b/openreview/arr/invitation.py
@@ -186,6 +186,9 @@ class InvitationBuilder(object):
     def expire_invitation(self, invitation_id):
         return self.venue_invitation_builder.expire_invitation(invitation_id)
 
+    def unexpire_invitation(self, invitation_id):
+        return self.venue_invitation_builder.unexpire_invitation(invitation_id)
+
     def set_meta_invitation(self):
         return self.venue_invitation_builder.set_meta_invitation()
     
@@ -195,8 +198,14 @@ class InvitationBuilder(object):
     def set_submission_invitation(self):
         return self.venue_invitation_builder.set_submission_invitation()
 
+    def set_submission_deletion_invitation(self, submission_revision_stage):
+        return self.venue_invitation_builder.set_submission_deletion_invitation(submission_revision_stage)
+
     def set_post_submission_invitation(self):
         return self.venue_invitation_builder.set_post_submission_invitation()
+
+    def set_submission_change_invitation(self, name, activation_date):
+        return self.venue_invitation_builder.set_submission_change_invitation(name, activation_date)
 
     def set_pc_submission_revision_invitation(self):
         return self.venue_invitation_builder.set_pc_submission_revision_invitation()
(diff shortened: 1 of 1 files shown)


openreview/workflows/process/compute_affinity_scores_process.py
```diff
@@ -38,6 +38,17 @@ def process(client, invitation):
 
     venue = openreview.helpers.get_venue(client, venue_id, support_user)
 
+    match_group = client.get_group(committee_id)
+
+    if not match_group.members:
+        print(f'No members found in the {committee_name} group. No affinnity scores were computed')
```
melisabok, line 44: is this string being shown in the workflow configuration UI as part of the log details?

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
```diff
@@ -224,20 +214,19 @@ def process(client, edit, invitation):
...
-        signatures=[invitation_prefix],
-        content={
-            'venue_id': { 'value': venue_id },
-            'name': { 'value': 'Assignment_Configuration' },
-            'submission_name': { 'value': 'Submission' },
-            'reviewers_name': { 'value': reviewers_name }
-        },
-        await_process=True
-    )
+    # client.post_invitation_edit(
```
melisabok: remove this?

openreview/workflows/process/compute_affinity_scores_process.py
celestemartinez: yes, I added this recently since the setup matching is no longer returning an error.

openreview/workflows/workflow_process/conference_review_workflow_deployment.py
celestemartinez: done

openreview/workflows/process/assignment_configuration_process.py
```diff
@@ -3,11 +3,11 @@ def process(client, edit, invitation):
     domain = client.get_group(edit.domain)
     venue_id = domain.id
     meta_invitation_id = domain.get_content_value('meta_invitation_id')
-    reviewers_name = domain.get_content_value('reviewers_name')
+    committee_name = invitation.get_content_value('committee_name')
 
     assignment_title = edit.note.content['title']['value']
 
-    match_name_invitation = client.get_invitation(f'{venue_id}/-/{reviewers_name}_Assignment_Deployment/Match')
+    match_name_invitation = client.get_invitation(f'{venue_id}/-/{committee_name}_Assignment_Deployment/Match')
```
melisabok, line 10: why the reviewers_id is not the prefix of the invitation? like we do with Assignment_Configuration?

PR #2931, opened 2026-03-13
fixing entity names for expertise results

cagostino:
This PR addresses changes to output entity processing of expertise results based on the changes made to the expertise repo in this [PR](https://github.com/openreview/openreview-expertise/pull/333)


openreview/conference/matching.py
```diff
@@ -645,10 +645,10 @@ def _compute_scores(self, score_invitation_id, submissions, model='specter+mfr')
                 matching_status['no_publications'] = result['metadata']['no_publications']
 
                 if self.alternate_matching_group:
-                    scores = [[entry['submission_member'], entry['match_member'], entry['score']] for entry in result['results']]
+                    scores = [[entry.get('entityB', entry.get('submission_member')), entry.get('entityA', entry.get('match_member')), entry['score']] for entry in result['results']]
```
melisabok: @enrubio can you confirm if this is the right order of the columns?

enrubio: Yes this is correct. For scores between groups, entity B is the submission_member and it should be the first item in the list. For example, when we compute SAC-AC scores, column 1 is the AC (which becomes the `head`) and column 2 is the SAC (`tail`).

openreview/conference/matching.py
```diff
@@ -645,10 +645,10 @@ def _compute_scores(self, score_invitation_id, submissions, model='specter+mfr')
...
                     return self._build_profile_scores(score_invitation_id, scores=scores), matching_status
 
-                scores = [[entry['submission'], entry['user'], entry['score']] for entry in result['results']]
+                scores = [[entry.get('entityA', entry.get('submission')), entry.get('entityB', entry.get('user')), entry['score']] for entry in result['results']]
```
enrubio: I just want to make sure I'm understanding the columns correctly since it's easy to mix up. Group to Group scores looks good to me. When we request scores for a Group to Papers (in request_expertise), it looks like entity A is always the group and entity B are the papers (type Note). So I'm confused why the naming gets switched and "submission" is replaced with entity A here? I could be missing something.

enrubio: Talked about it in the daily - the entity A that is sent in the request can be different from the entity A returned in the results.

celestemartinez: Spoke about this again, Chris will make a change to the entities from the request match the entities from the results.

enrubio (approved):
Changes look good! Glad it's consistent now between or-py and the expertise.

PR #2933, opened 2026-03-13
ARR: Post webfields once

### LINKED ISSUE #2558: ARR: Set the Area Chairs webfield only once
Set the content of the area chair webfield when the ARR venue is deployed and don't re set it every time the configuration changes.

This way organizers can make manual modifications without their changes being reverted.

haroldrubio:
- Resolves #2558
This PR sets the ARR webfields only once in the setup function by checking for a specific ARR comment added to each of the webfields


openreview/arr/arr.py
```diff
@@ -366,58 +367,68 @@ def setup(self, program_chair_ids=[], publication_chairs_ids=[]):
...
         with open(os.path.join(os.path.dirname(__file__), 'webfield/homepageWebfield.js')) as f:
             content = f.read()
-            self.client.post_group_edit(
-                invitation=self.get_meta_invitation_id(),
-                signatures=[self.venue_id],
-                group=openreview.api.Group(
-                    id=self.venue_id,
-                    web=content
+            venue_group = openreview.tools.get_group(self.client, self.venue_id)
+            if WEBFIELD_MARKER not in venue_group.web:
```
melisabok, line 371: webfield are posted once in regular venues, why is this not the case for ARR?

haroldrubio: In the Venue class, the webfields are only posted if the group has not yet been created. In ARR, the webfields are posted anytime the `setup` function is run. When the group edits were added to the ARR `setup` function, I thought this function was only run once so they are getting posted unconditionally

melisabok: Use this PR instead: https://github.com/openreview/openreview-py/pull/2939

PR #2934, opened 2026-03-13
Fix/ Submission Change: add a source and filter by source in process function

celestemartinez:
This PR adds a source to the `Submission_Change_Before_Bidding/Reviewing` and makes the process function read and filter by this source.

This is being added so we can check if this invitation is active in the submission process and post the respective edit.

### DIFF
--- a/tests/test_reviewers_only.py
+++ b/tests/test_reviewers_only.py
@@ -1899,8 +1899,8 @@ For more details, please check the following links:
         helpers.await_queue_edit(openreview_client, edit_id='ABCD.cc/2025/Conference/-/Official_Comment-0-1', count=3)
 
         invitations = openreview_client.get_invitations(invitation='ABCD.cc/2025/Conference/-/Official_Comment')
-        # these get created only for active papers
-        assert len(invitations) == 9
+        # these get created only for all papers since we added source now
+        assert len(invitations) == 10
         inv = openreview_client.get_invitation('ABCD.cc/2025/Conference/Submission1/-/Official_Comment')
 
         assert inv


openreview/tools.py
```diff
@@ -2055,6 +2059,7 @@ def should_match_invitation_source(client, invitation, submission, note=None):
     content_keys = invitation.edit.get('content', {}).keys()
     
     if not content_keys:
```
celestemartinez: @melisabok I am trying to understand why we are checking this here. The `Submission_Change_Before_Bidding` invitation does not have content keys, so no submissions match. Maybe I am not supposed to be using `should_match_invitation_source` here?

melisabok: this function was intended to catch the invitations that create child invitations. I didn't think about invitations editing existing notes. I guess this is done here so when the submission process functions find all the invitations to create the child invitations ignores other the ones that do not correspond  but I do no remember any case. Try removing this if and see if the tests pass.

openreview/workflows/process/submission_change_process.py
```diff
@@ -4,6 +4,7 @@ def process(client, invitation):
     venue_id = domain.id
```
melisabok: I wish this could be the same process function than the post_submission that we have in the old UI, is it possible?

celestemartinez: fixed here: https://github.com/openreview/openreview-py/pull/2937

openreview/tools.py
```diff
@@ -2062,12 +2062,12 @@ def should_match_invitation_source(client, invitation, submission, note=None):
         if is_accept_decision(decision_value, accept_options) != with_decision_accept:
             return False
 
+    note_invitation = invitation.edit.get('note') or invitation.edit.get('invitation', {}).get('edit', {}).get('note')
```
celestemartinez: @melisabok I had to add this here. Since we removed the `if not content_keys:` check, the submission process was trying to create children invitation for Desk_Rejection_Expiration, for example, which has no note inside the edit. I think this check should catch all other invitations we should be ignoring.

openreview/venue/process/post_submission_process.py
```diff
@@ -15,33 +15,40 @@ def process(client, invitation):
...
+                note=updated_note,
+                signatures=[venue_id]
+            )
     
     ## Release the submissions to specified readers if venueid is still submission
-    submissions = client.get_all_notes(content= { 'venueid': submission_venue_id }, domain=venue_id)
+    source = openreview.tools.get_invitation_source(invitation, domain)
+    print('source', source)
+    venue_ids = source.get('venueid', [submission_venue_id])
+    submissions = client.get_all_notes(content= { 'venueid': ','.join([venue_ids] if isinstance(venue_ids, str) else venue_ids)}, sort='number:asc', domain=venue_id)
```
melisabok: I think it would be less expensive to get all notes from invitation and then rely on `should_match_invitation_source` to match with the source.venueid.

I would also like to filter the submissions and then call to post_submission_edit. 

1. Get all submissions
2. Filter submissions by `should_match_invitation_source`
3. Call concurrent request using filtered submissions

openreview/tools.py
melisabok: I'm realizing that we are restricting this to note invitations but we may have group or tag invitations.

Isn't there another way to avoid the Desk_Rejection_Expiration?

openreview/venue/process/post_submission_process.py
celestemartinez: done

openreview/tools.py
```diff
@@ -1975,7 +1975,7 @@ def get_invitation_source(invitation, domain):
     meta_review_name = domain.content.get('meta_review_name', {}).get('value', None)
     rebuttal_name = domain.content.get('rebuttal_name', {}).get('value', None)
 
-    source = invitation.content.get('source', { 'value': { 'venueid': submission_venue_id } }).get('value', { 'venueid': submission_venue_id }) if invitation.content else { 'venueid': submission_venue_id }
+    source = invitation.content.get('source', { 'value': { 'venueid': submission_venue_id } }).get('value', { 'venueid': submission_venue_id }) if invitation.content else {}
```
celestemartinez, line 1978: returning no source if invitation does not have it

openreview/tools.py
```diff
@@ -2024,6 +2024,9 @@ def should_match_invitation_source(client, invitation, submission, note=None):
 
     source = get_invitation_source(invitation, domain)
 
+    if not source:
```
celestemartinez, line 2027: @melisabok I know we spoke about having
```
if not source and not content_keys:
    return False
```
Can you remind me why checking content_keys is also necessary? This seems to work as it is but wondering if I am mising something.

tests/test_reviewers_only.py
```diff
@@ -1855,7 +1874,8 @@ def test_comment_stage(self, openreview_client, helpers):
         helpers.await_queue_edit(openreview_client, edit_id='ABCD.cc/2025/Conference/-/Official_Comment-0-1', count=3)
 
         invitations = openreview_client.get_invitations(invitation='ABCD.cc/2025/Conference/-/Official_Comment')
-        assert len(invitations) == 10
+        # these get created only for active papers, should we add a source to the official comment?
```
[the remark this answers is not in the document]
celestemartinez: fixed.

tests/test_reviewers_only.py
```diff
@@ -1497,6 +1498,20 @@ def test_reviewers_deployment(self, openreview_client, helpers):
...
+            invitation='ABCD.cc/2025/Conference/-/Edit',
+            signatures=['ABCD.cc/2025/Conference'],
+            note=openreview.api.Note(
+                id=submission.id,
+                content={
+                    'venueid': { 'value': 'ABCD.cc/2025/Conference/Rejected_Submission' },
+                    'venue': { 'value': 'Submitted to ABCD 2025' }
+                }
+            )
+        )
```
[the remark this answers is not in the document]
celestemartinez: ignore, this edit is made with the meta invitation

openreview/tools.py
```diff
@@ -2077,10 +2077,10 @@ def should_match_invitation_source(client, invitation, submission, note=None):
     if 'noteReaders' in content_keys:
         return False
     
-    if 'noteId' not in content_keys:
+    if content_keys and 'noteId' not in content_keys:
         return False
     
-    if 'noteNumber' not in content_keys:
+    if content_keys and 'noteNumber' not in content_keys:
         return False
```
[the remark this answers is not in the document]
celestemartinez: not sure I understand this feedback

tests/test_reviewers_only.py
```diff
@@ -1518,6 +1533,10 @@ def test_review_stage(self, openreview_client, helpers):
         assert not 'readers' in submissions[0].content['pdf']
         assert submissions[0].content['data_release']['readers'] == ['ABCD.cc/2025/Conference', 'ABCD.cc/2025/Conference/Submission1/Authors']
 
+        # submission #1 still has previous readers
```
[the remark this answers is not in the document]
celestemartinez: fixed.

openreview/venue/process/post_submission_process.py
```diff
@@ -41,13 +42,16 @@ def post_submission_edit(submission):
         )
     
     ## Release the submissions to specified readers if venueid is still submission
-    submissions = client.get_all_notes(content= { 'venueid': submission_venue_id }, domain=venue_id)
+    source = openreview.tools.get_invitation_source(invitation, domain)
+    all_submissions = client.get_all_notes(invitation=f'{venue_id}/-/{submission_name}', sort='number:asc', domain=venue_id)
```
melisabok: can you use `submission_id` instead?

celestemartinez: fixed

PR #2965, opened 2026-03-23
Fix/cookie

carlosmondra:


openreview/openreview.py
```diff
@@ -149,7 +149,7 @@ def __handle_token(self, response):
         self.token = str(response['token'])
         self.profile = Profile( id = response['user']['profile']['id'] )
         self.headers['Authorization'] ='Bearer ' + self.token
-        self.user = jwt.decode(self.token, options={"verify_signature": False})
+        self.user = response['user']
```
melisabok, line 152: So the test that was failing was using a client of API 1

carlosmondra: I am not sure. What I did was to tell the AI to check the difference between my changes and master when everything was passing and identify the difference. That's how it founnd it.

melisabok: Thanks, I will merge it now.

Issue #2744, opened 2025-10-31
Assignment: When users accept invite assignment they receive 2 emails


enrubio:
It seems like when a user accepts an invitation to review a paper we send them 2 emails, one from the Assignment invitation and one from Assignment Recruitment. Both say almost the same thing, they confirm the assignment and link to the console. I think we should just send 1 email to not spam them and reduce the amount of emails we send.

There may be other places where we send them an extra email, for example when a user accepts but is pending sign up.

PR #2970, opened 2026-03-24
testing duplicate email behaviors

cagostino:
Fixes https://github.com/openreview/openreview-py/issues/2744

### DIFF
--- a/openreview/venue/group.py
+++ b/openreview/venue/group.py
@@ -389,7 +389,6 @@ class GroupBuilder(object):
                 signatures=[self.openreview_template],
                 content={
                     'venue_id': { 'value': venue_id },
-                    'message_reply_to': { 'value': self.venue.contact },
                     'venue_short_name': { 'value': self.venue.short_name },
                     'venue_from_email': { 'value': self.venue.get_message_sender()['fromEmail'] }
                 },
@@ -476,7 +475,6 @@ class GroupBuilder(object):
                 content={
                     'venue_id': { 'value': venue_id },
                     'group_id': { 'value': authors_id },
-                    'message_reply_to': { 'value': self.venue.contact  },
                     'venue_short_name': { 'value': self.venue.short_name },
                     'venue_from_email': { 'value': self.venue.get_message_sender()['fromEmail'] }
                 },
@@ -490,7 +488,6 @@ class GroupBuilder(object):
                 content={
                     'venue_id': { 'value': venue_id },
                     'group_id': { 'value': authors_accepted_id },
-                    'message_reply_to': { 'value': self.venue.contact  },
                     'venue_short_name': { 'value': self.venue.short_name },
                     'venue_from_email': { 'value': self.venue.get_message_sender()['fromEmail'] }
                 },
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -3587,6 +3587,8 @@ To view your submission, click here: https://openreview.net/forum?id={{{{note_fo
 
             registration_content = registration_stage.get_content(api_version='2', conference=self.venue)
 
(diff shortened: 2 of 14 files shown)


tests/test_icml_conference.py
```diff
@@ -1938,6 +1938,9 @@ def test_assignment(self, client, openreview_client, helpers, request_page, sele
 Please note that responding to this email will direct your reply to pc@icml.cc.
 '''
 
+        messages = openreview_client.get_messages(to='melisa@icml.cc', subject='[ICML 2023] You have been assigned as a Reviewer for paper number 1')
```
melisabok: before this code, add:

```
assignment_edges=pc_client.get_edges(invitation='ICML.cc/2023/Conference/Reviewers/-/Assignment', head=submissions[0].id, tail='~Melisa_ICML1')

assert len(assignment_edges) == 1

helpers.await_queue_edit(openreview_client, assignment_edges[0].id)
```

If  len(assignment_edges)  == 0 then the invite assignment process function hasn't finished so you need to wait for that process function to be finished as well.

tests/test_icml_conference.py
```diff
@@ -1938,6 +1938,13 @@ def test_assignment(self, client, openreview_client, helpers, request_page, sele
 Please note that responding to this email will direct your reply to pc@icml.cc.
 '''
 
+        assignment_edges=pc_client.get_edges(invitation='ICML.cc/2023/Conference/Reviewers/-/Assignment', head=submissions[0].id, tail='~Melisa_ICML1')
+        assert len(assignment_edges) == 1
+        helpers.await_queue_edit(openreview_client, assignment_edges[0].id)
+
+        messages = openreview_client.get_messages(to='melisa@icml.cc', subject='[ICML 2023] You have been assigned as a Reviewer for paper number 1')
+        assert messages and len(messages) == 1
```
celestemartinez: I just realized that this recruitment is being done before the assignments are deployed, so if I am not mistaken, a second email won't be sent now. 

The good news is this is already being tested for the `check_new_profiles` function in lines:
https://github.com/openreview/openreview-py/blob/e7c08c5c8b6f3599282f8344136a1f43f69cb0a2/tests/test_icml_conference.py#L2216
https://github.com/openreview/openreview-py/blob/e7c08c5c8b6f3599282f8344136a1f43f69cb0a2/tests/test_icml_conference.py#L2243

I would just add another test later on to test the assignment being done immediately after the acceptance (without having to wait for the user to sign up)
here is a good place: https://github.com/openreview/openreview-py/blob/e7c08c5c8b6f3599282f8344136a1f43f69cb0a2/tests/test_icml_conference.py#L2357

openreview/venue/venue.py
```diff
@@ -1903,26 +1903,6 @@ def mark_as_accepted(venue_group, edge, submission, user_profile, invite_assignm
                     external_paper_committee_id=paper_reviewer_invited_id.replace('/Invited', '').replace('{number}', str(submission.number))
                     client.add_members_to_group(external_paper_committee_id, edge.tail)
 
-                if assigment_label:
```
melisabok, line 1906: It looks you are missing a change, invite_assignment_process.py. This change is related to reviewers that were invited to review a paper, they accepted but they didn't have a profile at that time. 

There should a test for both cases, to test this change, the test must call Venue.check_new_profiles manually because this is expected to be run as a cron job.

As a summary, there are two ways to confirm an assignment:
1. The user accepts the invitation, the user has a profile, the invitation edge becomes "Accepted", the assignment edge is created.
2. The user accepts the invitation, the user does not have a profile, the invitation edge becomes "Pending Sign Up", the user signs up, the assignment edge is created.


You fixed case 2, case 1 is missing.

PR #3000, opened 2026-04-03
Add Docker Compose setup for running integration tests

carlosmondra:
## Summary
Adds a Docker Compose configuration that runs the full integration test stack (MongoDB, Redis, Elasticsearch, API v1, API v2, web frontend, and pytest) in containers, eliminating the need to install and manage services locally.

## Changes
- **`docker/docker-compose.yml`**: Defines 8 services sharing a single network namespace via a pod container, so `localhost` works everywhere and existing `circleci.json` configs are reused as-is. Uses healthchecks and `depends_on` conditions to enforce startup order.
- **`docker/run-tests.sh`**: Wrapper script that keeps infrastructure running between test runs and only restarts the API servers (via `cleanStart`) to get a fresh database each time.
- **`docker/Dockerfile.{api,web,test}`**: Minimal images — Node 22 + Python for APIs, Node 22 for web, Python 3.13 + Firefox ESR + geckodriver for the test runner.
- **`docker/scripts/{api-v1,api-v2,web,test}-entrypoint.sh`**: Entrypoint scripts that rsync source from read-only mounts, cache npm and pip installs in named volumes (hash-based skip), and use editable pip installs so Python code changes are picked up without reinstalling.
- **`openreview/tools.py`**: Make `LOCAL_API_V1`, `LOCAL_API_V2`, and `LOCAL_SITE` configurable via environment variables with localhost defaults.
- **`README.md`**: Add Docker Compose as the recommended test method with usage examples; reorganize existing manual setup as a secondary option.
- **`.vscode/launch.json`**: Use `"module": "pytest"` instead of hardcoded homebrew path.


cagostino (approved):
was able to run the pr's run-tests.sh end to end, build and stack bring-up worked, shared network namespace via pod container allows reuse of circleci.json

melisabok (approved):

docker/run.py
```diff
@@ -0,0 +1,428 @@
...
+    print("  API v2: http://localhost:3001")
+    print()
+    print("View logs per service in separate terminals:")
+    print("  docker compose logs -f api-v1")
+    print("  docker compose logs -f api-v2")
+    print("  docker compose logs -f web")
+    print("  docker compose logs -f mongo")
+    print()
+    print("Stop with: docker compose down")
+    print("  (or Ctrl+C here)")
```
celestemartinez: Currently, `docker compose down` and `Ctrl+C` perform different functions in serve mode:
- `docker compose down` will shutdown all the services
- `Ctrl+C` only closes the current shell but leaves the services running

This might be a desired feature but I think we should clarify in the prints/READ.md

docker/run.py
```diff
@@ -0,0 +1,428 @@
...
+    finally:
+        if keep_infra:
+            teardown_apis(serve_mode=True)
+            print("Infrastructure still running. Tear down with: docker compose down")
+        else:
+            print("\nShutting down...")
+            teardown(serve_mode=True)
+
+
+def mode_shell(target, no_clean=False):
```
celestemartinez: This mode is not working for me. I tried both `./run.py --shell` and `./run.py --shell test`. In both cases I get an error: `ERROR: file or directory not found: bash`

README.md
```diff
@@ -77,22 +77,146 @@ The openreview-py library can be used to easily access and modify any data store
...
+### Serve Mode (Manual Browser Testing)
+
+Start all services with ports exposed to your host machine for manual browser testing:
+
+```bash
+# Start services only (browse http://localhost:3030)
+./run.py --serve
+
+# Populate the database by running a test, then keep services running
+./run.py --serve tests/test_icml_conference.py
```
celestemartinez, line 120: What is the difference between this and 
`./run.py --keep-infra tests/test_client.py`? 

Do we need both?

carlosmondra: --serve is for manual browser testing. --keep-infra is for fast iteration on test runs by keeping mongo, redis, ES and web up.

docker/run.py
carlosmondra: Good catch! Thanks, fixed.

docker/run.py
carlosmondra: Good catch :). Let me check. I think they should behave the same way. I'll push the changes.

celestemartinez:
These are my thoughts on which commands are the most useful:
- test mode: very useful.
- serve mode: very useful, my favorite since I can do manual browser testing. 
- shell mode: very useful as well, especially if I need to run multiple tests. 

Note: it would be great if we could instead have a combination between serve and shell mode, where I could do manual browser testing while having shell access to run scripts/tests. I would use this mode more than the two separate ones individually. 
- --setup-only: not sure about the use case for this since we can't really interact with any of the components. Is it only to test the setup launches fine?
- --no-clean: I like the idea but not clear on how to use it between runs. This is how I used it and what I expect:
```
1. execute `./run.py --serve tests/test_registration_step.py`. Once it finished, I can see the WXYZ conference in my browser.
2. Ctrl+C to stop all services
3. execute `./run.py --serve --no-clean`. Once it launches, I expected the WXYZ conference to still be visible in the browser but it wasn't. 
```
Is this the right way to do this/think about this?

carlosmondra:
Thanks Celeste for the very thorough feedback.

> Note: it would be great if we could instead have a combination between serve and shell mode, where I could do manual browser testing while having shell access to run scripts/tests. I would use this mode more than the two separate ones individually.

Done! You can now run `./run.py --serve --shell` together

> * --setup-only: not sure about the use case for this since we can't really interact with any of the components. Is it only to test the setup launches fine?

Removed

> * --no-clean: I like the idea but not clear on how to use it between runs. This is how I used it and what I expect:
> 
> ```
> 1. execute `./run.py --serve tests/test_registration_step.py`. Once it finished, I can see the WXYZ conference in my browser.
> 2. Ctrl+C to stop all services
> 3. execute `./run.py --serve --no-clean`. Once it launches, I expected the WXYZ conference to still be visible in the browser but it wasn't. 
> ```
> 
> Is this the right way to do this/think about this?

Yes, I think this process makes sense, so I made sure that it works

celestemartinez:
Hopefully last comment:

When trying to run this:
`./run.py --serve --shell tests/test_icml_conference.py`
I get the following error: 
`run.py: error: argument --shell: invalid choice: 'tests/test_icml_conference.py' (choose from 'api-v1', 'api-v2', 'test', 'web')`

Changing the order of --shell and --serve does work, but it looks like --shell is expecting an argument.

README.md
```diff
@@ -77,22 +77,160 @@ The openreview-py library can be used to easily access and modify any data store
...
+
+# Second run: reuse infrastructure and database (skips cleanStart)
+./run.py --serve --no-clean
+
+# Or combine: populate, then browse with shell access
+./run.py --keep-infra tests/test_registration_step.py
+./run.py --serve --shell --no-clean
+```
+
+> Note: `--no-clean` requires `keep_infra` to be useful. Without it, the database is lost when services are torn down.
```
celestemartinez:
```suggestion
> Note: `--no-clean` requires `keep_infra` to be used. Without it, the database is lost when services are torn down.
```

carlosmondra:
> Hopefully last comment:
> 
> When trying to run this: `./run.py --serve --shell tests/test_icml_conference.py` I get the following error: `run.py: error: argument --shell: invalid choice: 'tests/test_icml_conference.py' (choose from 'api-v1', 'api-v2', 'test', 'web')`
> 
> Changing the order of --shell and --serve does work, but it looks like --shell is expecting an argument.

It was supposed to accept an argument to use different shells, but maybe this is a feature we don't need at the moment, so I removed it. The command should work now.

celestemartinez (approved):

PR #3009, opened 2026-04-09
 Support multiple reviewer/area chair roles per submission

melisabok:
## Summary

Lets a venue define more than one reviewer role (and more than one area chair role), and say which
per-submission group each role is assigned into — one shared group, or one group per role. Single-role
venues behave identically to before.

## Motivation

Some venues need to run distinct reviewer committees for different kinds of submissions or different
evaluation criteria — for example `Expert_Reviewers` vs `Technical_Reviewers`, or a main AC track
alongside a technical AC track. The workflow already accepted alternate names in `reviewer_roles` /
`area_chair_roles` at the top-level group, but the per-submission plumbing collapsed everything back to a
single primary role: one `Submission1/Reviewers` group, one Official_Review invitation, one assignment
deployment, one `Submitted` subgroup. That made it impossible to keep the committees separate once
submissions came in — assignments, review forms, and "submitted" tracking all landed on the wrong group,
and notifications fanned out to the umbrella instead of the reviewer's own role.

This branch treats the multi-role case as a first-class configuration. The request form asks for the same
two values the `Venue` class already holds, rather than an abstraction it has to translate:

- `reviewer_groups_names` — the recruiting roles. Optional, defaults to `['Reviewers']`. One main
  `venue_id/<role>` group is created per entry at deployment, each with its own recruitment, webfield,
  matching and assignment invitations.
- `submission_reviewer_group_names` — the groups created per submission. Optional, defaults to the first
  entry of `reviewer_groups_names`. One `Submission_Group` invitation is created per entry.

`area_chair_groups_names` / `submission_area_chair_group_names` mirror this for area chairs.

Roles are paired with per-submission groups **by position**. Give one name and every role is assigned
into it; give one name per role, in the same order, and each role gets its own group. Any other length is
rejected by the request form preprocess, since a wrong-length list would otherwise deploy reviewers into
the wrong group silently.

The per-submission group name does not have to be one of the roles — a venue with
`['Expert_Reviewers', 'Technical_Reviewers']` merging into a single group can name it `Reviewers`, and
gets `Submission1/Reviewers`. No top-level `venue_id/Reviewers` group is needed; the `Submission_Group`
invitation lives under that prefix and creates the per-submission groups when assignments are deployed.

The role name is then threaded through group creation, invitation setup, matching, assignment deployment,
and the review process function, so per-role groups, per-role review forms, and per-role notifications
all work end-to-end.

## Key changes

- New Venue / ReviewStage attributes `submission_reviewer_roles` and `submission_area_chair_roles` drive
  how per-submission groups and review invitations are fanned out.
- New `Venue.get_submission_committee_name(role)` is the single place that maps a role to its
  per-submission group; both `invitation.py` and `matching.py` call it instead of repeating the pairing
  logic.
- `get_reviewers_id` / `get_area_chairs_id` / `get_anon_reviewer_id` take an optional `name=` so callers
  can target a specific role.
- Group builder, invitation builder, matching, and assignment deployment iterate over the configured
  roles instead of using a single primary name.
- Assignment invitations carry a `submission_committee_name`; matching deployment reads it instead of
  inferring the target group name from the committee id, and falls back to the positional mapping.
- The review process function resolves the reviewer's role from the anon signature prefix and scopes the
  `Submitted` subgroup and reviewer-email recipients to that role.
- New `reviewer_groups_names` / `submission_reviewer_group_names` / `area_chair_groups_names` /
  `submission_area_chair_group_names` / `senior_area_chair_groups_names` fields on the
  `Conference_Review_Workflow` request form, with length validation in the request form preprocess.
- The AC `Reviewer_Reassignment` invitation offers all reviewer roles in a `reviewers_name` enum,
  defaulting to the role paired with that AC role by position. The proposed assignment title is typed in
  rather than synchronized from the assignment configurations.

## Companion change

`openreview-web` branch `feature/multiple-roles` updates `components/group/WorkflowInvitations.js` to
expand `reviewer_roles` / `area_chair_roles` / `senior_area_chair_roles` into one workflow group per role
in the PC console, instead of a single group per committee kind.

## Test plan

New end-to-end coverage, one file per configuration:

- [x] `pytest tests/test_merged_committee_roles.py` — two reviewer roles and two AC roles sharing a
      single per-submission group. `submission_reviewer_group_names` is `['Reviewers']`, a name that is
      not one of the roles; `submission_area_chair_group_names` reuses the `Area_Chairs` role.
- [x] `pytest tests/test_two_submission_committee_roles.py` — two reviewer roles and two AC roles, each
      with its own per-submission group, including the per-role review forms and AC console
      reassignment links.
- [x] `pytest tests/test_webconf_conference_with_templates.py` — many track roles with the default
      single per-submission group.
- [x] `pytest tests/test_venue_deployment.py` — request form validation and deployment regression.

Single-role and mixed regressions:

- [ ] `pytest tests/test_reviewers_only.py`
- [ ] `pytest tests/test_acs_and_reviewers.py`
- [ ] `pytest tests/test_matching_v2.py tests/test_icml_conference_with_templates.py tests/test_tasks.py tests/test_change_venue_email.py`


openreview/venue/process/review_process.py
```diff
@@ -60,7 +64,13 @@ def create_group(group_id, members=[]):
...
         )
 
-    create_group(paper_reviewers_submitted_id, [review.signatures[0]])
+    # Match the anon signature to its role by prefix: e.g.
+    # 'venue/Submission1/Expert_Reviewer_XYCV' belongs to role 'Expert_Reviewers'.
+    for role in reviewer_roles:
+        anon_prefix = f'{paper_group_id}/{role[:-1] if role.endswith("s") else role}_'
+        if review.signatures[0].startswith(anon_prefix):
+            create_group(f'{paper_group_id}/{role}/{reviewers_submitted_name}', [review.signatures[0]])
+            break
```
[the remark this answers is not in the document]
melisabok: Submitted groups are always created, ignore.

openreview/workflows/process/deploy_assignments_process.py
```diff
@@ -50,6 +50,15 @@ def process(client, invitation):
...
 
+    venue = openreview.helpers.get_venue(client, venue_id, support_user)
+
+    venue.set_assignments(
+        assignment_title=match_name,
+        committee_id=committee_id,
+        overwrite=True,
+        enable_reviewer_reassignment=True
+    )
+
```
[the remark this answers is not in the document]
melisabok: Fixed, thanks!

melisabok:
These changes can be seen in the UI using this branch (still in progress): https://github.com/openreview/openreview-web/pull/2753

enrubio:
There are 2 issues I flagged in Celeste's PR that enables reviewer reassignment that I think should be fixed in this PR because it relates to multiple groups:
1. [assignment_configuration_process.py](https://github.com/openreview/openreview-py/pull/3078#discussion_r3639273872):

> Here we always post an edit to f'{venue_id}/{area_chairs_name}/-/Reviewer_Reassignment' but area_chairs_name will always be the 1st AC role in the list of AC roles. In group.py we loop through each role in area_chair_roles to create the invitation. Do we need to do the same here?

2. [matching.py](https://github.com/openreview/openreview-py/pull/3078#discussion_r3639628571):

> Since each AC group has its own reviewers_proposed_assignment_title, maybe we need to find which AC group to delete the title from? Otherwise this only deletes it for the first role in the AC role list.
Not sure if this is something to consider, but I know we want to support multiple roles soon.

racheljsmart:
In the PC console, there is an Edit Reviewer Assignment link provided for each submission in the Submission Status under the Official Review status. If there are multiple roles and assignments, what link will be shared here?

melisabok:
> In the PC console, there is an Edit Reviewer Assignment link provided for each submission in the Submission Status under the Official Review status. If there are multiple roles and assignments, what link will be shared here?

now it shows only the first one (first role).

The web draft PR tries to show all of them but it is not ready, I need to work with Xukun to cover all the cases.

openreview/workflows/edit_invitations.py
```diff
@@ -2045,12 +2089,37 @@ def set_edit_reviewer_reassignment_invitation(self, group_id):
...
+                        'value': {
+                            'param': {
+                                'type': 'string',
+                                'enum': reviewer_roles if reviewer_roles else [reviewers_name],
+                                'default': reviewers_name,
+                                'optional': True
+                            }
+                        }
+                    },
+                  'reviewers_proposed_assignment_title': {
```
celestemartinez, line 2150: So PCs will need to type the matching title here, it is no longer a dropdown?

melisabok: yes, sorry about this but it is easy to manage if there are many roles. 

in the case of having many roles, it is hard to map them with each proposed assignment title, I think typing the title is fine.

celestemartinez (approved):

openreview/stages/venue_stages.py
```diff
@@ -720,18 +721,23 @@ def __init__(self,
...
+        self.submission_reviewer_roles = submission_reviewer_roles
+
+    def _get_reviewer_roles(self, conference):
+        return self.submission_reviewer_roles or [conference.reviewers_name]
 
     def _get_reviewer_readers(self, conference, number, review_signature=None):
+        reviewer_roles = self._get_reviewer_roles(conference)
         if self.release_to_reviewers is ReviewStage.Readers.REVIEWERS:
-            return conference.get_reviewers_id()
+            return [conference.get_reviewers_id(name=name) for name in reviewer_roles]
```
enrubio: I think we need to make these same changes for the other stages as well? Other stages call `get_reviewers_id()` or `get_anon_reviewer_id()` for example but we don't pass a `name` there, e.g.: Submission stage, Comment, MetaReview, etc. so it would still resolve to the first name in the role list.

What do you think? @melisabok

melisabok: I agree, let me check what is missing.

melisabok: Done.

enrubio: I still see some places where we didn't pass a name to `get_reviewers_id` or didn't replace it with the new functions you created. For example, `get_readers` in SubmissionStage, `get_signatures` in CustomStage, etc.

melisabok: ok, I tried to cover all the stages but I didn't add many tests for each stage when there is more than one submission group. I would like to wait until we have a venue using this to make sure we are doing the right thing, I don't want to break any existing behavior. Please review again.

enrubio: Ah okay, got it. That makes sense, thank you

enrubio:
Changes to the stages look good to me but looks like ARR tests fail now. Maybe because ARR is missing the new venue fields? `submission_reviewer_roles` and `submission_area_chair_roles`?

enrubio (approved):
Changes look good

openreview/workflows/workflow_process/committee_group_template_process.py
celestemartinez: Looks like this content dictionary does not exist anymore and it's causing errors.

PR #3010, opened 2026-04-09
Add description support to CustomStage

### LINKED ISSUE #2724: Custom Stage: Add support for invitation description
It would be great if we could specify instructions for custom invitations (`description` inside the `invitation.edit.invitation`). This is so that PCs don't have to run the custom stage then go to the invitation to directly add the description.

cagostino:
Fixes #2724

CustomStage had no description parameter, requiring PCs to run the stage and then manually edit the invitation to set one. ReviewStage and MetaReviewStage already supported this — this PR adds the same support to CustomStage.

- Added description=None param to CustomStage.__init__
- Wired it into set_custom_stage_invitation() following the same pattern as set_review_stage_invitation() and set_meta_review_stage_invitation()

### DIFF
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -3193,6 +3193,11 @@ To view your submission, click here: https://openreview.net/forum?id={{{{note_fo
         if custom_stage.preprocess_path:
             invitation.edit['invitation']['preprocess'] = self.get_process_content(custom_stage.preprocess_path)
 
+        if custom_stage.description:
+            invitation.edit['invitation']['description'] = custom_stage.description
+        else:
+            invitation.edit['invitation']['description'] = { 'param': { 'const': { 'delete': True } } }
+
         self.save_invitation(invitation, replacement=False)
         return invitation
 
--- a/openreview/stages/venue_stages.py
+++ b/openreview/stages/venue_stages.py
@@ -1629,6 +1629,7 @@ class CustomStage(object):
                  notify_readers=False, 
                  email_template=None, 
                  allow_de_anonymization=False,
+                 description=None,
                  child_invitations_name=None):
         self.name = name
         self.child_invitations_name = child_invitations_name if child_invitations_name else self.name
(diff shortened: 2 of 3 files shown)


melisabok:
Please add a test to this change. Here is a place where we use CustomStage: https://github.com/openreview/openreview-py/blob/master/tests/test_icml_conference.py#L4716

melisabok (approved):

openreview/venue/invitation.py
```diff
@@ -3193,6 +3193,11 @@ def set_custom_stage_invitation(self):
         if custom_stage.preprocess_path:
             invitation.edit['invitation']['preprocess'] = self.get_process_content(custom_stage.preprocess_path)
 
+        if custom_stage.description:
+            invitation.edit['invitation']['description'] = custom_stage.description
+        else:
+            invitation.edit['invitation']['description'] = { 'param': { 'const': { 'delete': True } } }
+
```
enrubio, line 3274: Since we use `replacement=False` for the custom stage, if a field isn't included in the edit then no change is made to it. In this case for the description, if they don't include one with each edit, then it gets deleted.

I think to keep the behavior consistent we don't delete the description and they should explicitly pass `delete: true`.

celestemartinez: I think it's fine to leave like this, this is consistent with how we delete fields in the content if they are not in the edit (and consistent with  how we add/remove description in the review stage, for example)

celestemartinez (approved):

openreview/venue/invitation.py
enrubio: > this is consistent with how we delete fields in the content if they are not in the edit 

I thought if it's not in the edit then the field doesn't get deleted unless you specify `replacement=True` or they write `delete: true` for the field?

I'm fine with leaving it though, just wanted to point it out.

celestemartinez: yes, you are right. What I meant is we are passing delete: true if the field is in the invitation but not in the edit, so this is consistent with how we are already doing it: https://github.com/openreview/openreview-py/blob/cbc7a6091c6537ebc133e35f5d36529901c735ac/openreview/stages/venue_stages.py#L1830

enrubio: I understand now thank you!

enrubio (approved):
This looks good to me

PR #3024, opened 2026-04-16
Fix/ save_invitation should check if the log finished with ok or error status

celestemartinez:
This PR changes the retrieval logic in `save_invitation` to consider only logs with status of `ok` or `error`. Previously, it was getting the log with status `running` and returning while the date process was still running. 

I wasn't able to reproduce the error in a test, I think the date process runs too quickly, but I was able to print the log that was being returned and it had status `running`.

### DIFF
--- a/openreview/workflows/edit_invitations.py
+++ b/openreview/workflows/edit_invitations.py
@@ -29,12 +29,15 @@ class EditInvitationsBuilder(object):
         invitation = self.client.get_invitation(invitation.id)
 
         if invitation.date_processes and len(invitation.date_processes[0]['dates']) > 1 and self.update_date_string == invitation.date_processes[0]['dates'][1]:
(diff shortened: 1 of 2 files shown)


openreview/workflows/edit_invitations.py
```diff
@@ -29,12 +29,15 @@ def save_invitation(self, invitation, replacement=None):
         invitation = self.client.get_invitation(invitation.id)
 
         if invitation.date_processes and len(invitation.date_processes[0]['dates']) > 1 and self.update_date_string == invitation.date_processes[0]['dates'][1]:
-            process_logs = self.client.get_process_logs(id=invitation.id + '-0-1', min_sdate = invitation.tmdate + self.update_wait_time - 1000)
+            expected_statuses = {'error', 'ok'}
```
celestemartinez, line 32: edit invitations do not have a date process, but changed it here too to be consistent.

melisabok (approved):

PR #3025, opened 2026-04-17
ARR: Remove all but latest 2 cycles

haroldrubio:
This PR adds a step in the `setup` function in the ARR class to keep only the most recent 2 cycles in `active_venues`


openreview/arr/arr.py
```diff
@@ -735,8 +735,29 @@ def get_submissions(self, venueid=None, accepted=False, sort=None, details=None)
...
+
+            venue_group = tools.get_group(self.client, venue_id)
+            venue_cdate = venue_group.cdate
+            arr_venues.append((venue_cdate, venue_id == self.venue_id, venue_id))
+
+        if len(arr_venues) <= keep_count:
+            return
+
+        arr_venues.sort(reverse=True)
+        stale_venue_ids = [venue_id for _, _, venue_id in arr_venues[keep_count:]]
```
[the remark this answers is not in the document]
haroldrubio: Added a guard for a None return on get_group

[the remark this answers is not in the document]
haroldrubio: Ignore - avoid the complicated semantics of parsing by venue ID, cdate is always more accurate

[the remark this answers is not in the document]
haroldrubio: Ignore since we should be using cdates, the alternative I suppose is to get all groups to retrieve information from every active_venue but I also think this should be avoided and pay the latency cost

openreview/arr/arr.py
```diff
@@ -735,8 +735,31 @@ def get_submissions(self, venueid=None, accepted=False, sort=None, details=None)
...
+
+        for venue_id in active_venues.members or []:
+            if not venue_id.startswith('aclweb.org/ACL/ARR'):
+                continue
+
+            venue_group = tools.get_group(self.client, venue_id)
+            if venue_group is None:
+                continue
+            venue_cdate = venue_group.cdate
+            arr_venues.append((venue_cdate, venue_id == self.venue_id, venue_id))
```
celestemartinez: What is the purpose of saving `venue_id == self.venue_id`?

melisabok: to know if that venue is the current cycle?

melisabok (approved):

openreview/arr/arr.py
celestemartinez: yes, I thought so, but I guess I don't see it being used anywhere. Maybe I am missing something

openreview/arr/arr.py
```diff
@@ -735,8 +735,31 @@ def get_submissions(self, venueid=None, accepted=False, sort=None, details=None)
...
+
+        if len(arr_venues) <= keep_count:
+            return
+
+        arr_venues.sort(reverse=True)
+        stale_venue_ids = [venue_id for _, _, venue_id in arr_venues[keep_count:]]
```
melisabok: @celestemartinez here we should filter out the current cycle.

haroldrubio: I can make the change to do it more explicitly

celestemartinez (approved):

PR #3032, opened 2026-04-23
Fix timeout errors on venue deployment

melisabok:
## Summary

Raises the client-side `await_process` timeout from 50 seconds to 10 minutes, and cuts one layer of nested process-function work out of the committee-group setup chain. Addresses "Process timed out" errors reported by the support team when deploying new venues.

## Motivation

Support reported sporadic `OpenReviewException("Process timed out")` errors when posting the `Conference_Review_Workflow/-/Deployment` edit. The root cause was in the V2 client's internal `__await_process` helper:

- It only polled for up to **100 iterations × 0.5s = 50 seconds** before giving up.
- `deployProcess.py` runs a large synchronous chain of group and invitation creation (including the `Committee_Group` process, which itself synchronously creates ~8 groups/invitations per committee role). Under load, 50s is easily exceeded.
- For comparison, the server-side wait in [openreview/venue/invitation.py `save_invitation`](openreview/venue/invitation.py) already tolerates up to 30 minutes for its `-0-1` date-process log to complete. The client's 50s budget was ~36× shorter than what the rest of the stack accepts.

While investigating, I also noticed the `Committee_Group` process fans out through `await_process=True` on seven sub-calls, but only one of those sub-calls (`Committee_Invited_Group`) actually had a server-side process function — so that one process was the only genuinely nested work. Inlining it into the parent process removes the last real nested process firing for committee setup.

### Client timeout bump

- [openreview/api/client.py](openreview/api/client.py) — `__await_process` now polls up to **1200 iterations × 0.5s = 10 minutes** before raising `"Process timed out"`. Also removes the per-iteration `print('Check logs for process function', …)` debug line that was noisy in ops output.

### Inline `Committee_Invited_Group` setup into the parent `Committee_Group` process

- [openreview/workflows/workflow_process/committee_group_template_process.py](openreview/workflows/workflow_process/committee_group_template_process.py):
  - After posting the `Committee_Invited_Group` group edit, the parent process now directly posts the `Group_Message` invitation for the `/Invited` group and captures `{committee_role}_invited_message_id` into the single final meta-invitation group-edit batch. Previously that work lived in a separate `committee_invited_group_template_process.py` that ran as a nested process via `await_process=True`.
  - `venue_short_name` and `venue_from_email` are computed once as locals and reused by both `Group_Message` posts (committee group + invited group).
  - Removes all seven `await_process=True` flags from the child `post_group_edit` / `post_invitation_edit` calls. Most of those templates never had a top-level process attached, so the flag was a no-op on the server; the one that did (`Committee_Invited_Group`) no longer needs one because its work is inlined.
- [openreview/workflows/templates.py](openreview/workflows/templates.py) — remove `process=self.get_process_content('workflow_process/committee_invited_group_template_process.py')` from the `Committee_Invited_Group` template invitation so posting to it no longer fires a nested process.
- **Deleted** [openreview/workflows/workflow_process/committee_invited_group_template_process.py](openreview/workflows/workflow_process/committee_invited_group_template_process.py) — its logic is now in the parent `committee_group_template_process.py`.

## Impact

- Deploy no longer fails with a 50s client-side timeout; the 10-minute budget matches the tolerance the server already has for comparable operations.
- One fewer nested process firing per committee role during deploy (e.g. 3 fewer for a venue with Reviewers + Area_Chairs + Senior_Area_Chairs). No behavior change — the same groups, invitations, and domain content fields are produced.

## Test plan

- [ ] `pytest tests/test_reviewers_only.py` passes (primary regression).
- [x] `pytest tests/test_acs_and_reviewers.py` passes (covers the full 3-role committee chain).
- [ ] `pytest tests/test_group_recruitment.py` passes (exercises `Committee_Invited_Group`, `Committee_Declined_Group`, and the recruitment invitations).
- [ ] Manually deploy a venue via the Support workflow on a staging instance and confirm no `"Process timed out"` is raised and the venue reaches the same end state (reviewer / AC / SAC groups, `/Invited`, `/Declined`, recruitment request / reminder / response invitations, domain content keys populated).


celestemartinez (approved):

enrubio (approved):
These changes look good, I like that one layer of nested process functions were removed.

PR #3041, opened 2026-04-27
Add `unified_authors` mode to `SubmissionStage`

melisabok:
## Summary

Adds a `unified_authors` flag on `SubmissionStage` that switches submission
content from the legacy two-field schema (`authors`: `string[]` +
`authorids`: `profile{}`) to a single `authors` field of type `author{}`, with
one structured entry per author (`fullname`, `username`, optional
`institutions`). The template deployment workflow opts in; default
`SubmissionStage()` keeps the legacy schema, so existing venues are unaffected.

## Changes

- **Schema** (`stages/venue_stages.py`, `stages/default_content.py`): new
  `unified_authors` flag and `submission_v2_unified_authors` template
  (`username` constrained to a profile ID, `^~\S+$`). When enabled, the
  `authorids` field is dropped and `force_profiles` is skipped.
- **Invitations** (`venue/invitation.py`): readers/writers resolve to a flat
  profile-ID list in both modes — legacy `${...authorids/value}`, unified
  `${...authors/value/*/username}`. Withdrawal / desk-reject /
  `Submission_Change_Before_*` only configure `authorids` for legacy venues.
- **Deployment** (`workflows/.../conference_review_workflow_deployment.py`):
  passes `unified_authors=True`.
- **`Note` API**: `note.authors` (fullname strings) and `note.authorids`
  (profile IDs / emails) now work for both schemas on V1 (`openreview.py`) and
  V2 (`api/client.py`). Process scripts, `tools.generate_bibtex`, `venue.py`,
  and `matching.py` read author IDs through these properties.
- **Process scripts**: read IDs via `note.authorids` and email the per-paper
  Authors group instead of individual recipients. Revision processes update
  group membership *before* notifying (so new co-authors get the email).
- **Profile name removal**: `request_remove_name_process.py` detects
  publications via `note.authorids`; `request_remove_name_decision_process.py`
  rewrites the deleted username/fullname inside the `author{}` object in both
  **notes** and **invitations** (literal author-object list), with the legacy
  `authorids` path preserved.

## Tests

- Migrated all 9 `Conference_Review_Workflow` / template-deployment tests to
  the unified submission format (posting `authors` as objects; assertions use
  `submission.authorids` and expect no `authorids` field).
- **`tests/test_submission_authors_email.py`** (new): author-by-email under the
  unified schema — default rejects emails (`must be a valid profile ID`); a
  venue can opt in by relaxing the `username` regex via `Submission/Form_Fields`.
- **`test_profile_management.py::test_rename_unified_authors_submission`** (new):
  name removal rewrites the username/fullname across the note's `author{}`
  object, the per-paper Authors group, and a literal-embedding invitation.

## Backward compatibility

Default `SubmissionStage()` produces the unchanged legacy schema; the flag must
be opted into explicitly. Downstream code reads author IDs through
`note.authorids`, which handles both schemas transparently.

## Test plan

- [x] 9 migrated template-deployment tests pass on the V2 stack.
- [x] Legacy `SubmissionStage()` unchanged; `unified_authors=True` emits a
      single `author{}` field with no `authorids`.
- [x] `Note.authors` / `Note.authorids` verified on both schemas (V1 + V2).
- [x] `test_submission_authors_email.py` passes.
- [x] `test_profile_management.py` full suite passes (28 passed, 1 pre-existing
      skip), including the new and legacy name-removal tests.


openreview/stages/default_content.py
```diff
@@ -540,7 +540,38 @@
...
+        'param': {
+            'type': 'author{}',
+            'properties': {
+                'fullname': { 'param': { 'type': 'string' } },
+                'username': {
+                    'param': {
+                        'type': 'string',
+                        'regex': r"^~\S+$|^[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$",
+                        'mismatchError': 'must be a valid email or profile ID'
+                    }
```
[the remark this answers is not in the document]
melisabok: fixed

openreview/venue/process/invite_assignment_pre_process.py
```diff
@@ -61,7 +61,7 @@ def process(client, edge, invitation):
 
         print(f'Check conflicts for {user_profile.id}')
         ## - Check conflicts
-        authorids = submission.content['authorids']['value']
```
enrubio, line 64: This file isn't used, we use the JS version

openreview/venue/venue.py
enrubio: Minor, but if a new UI venue used iThenticate and it reached this condition, I think it would fail since they're not using `authorids` anymore.

enrubio:
I was looking at the tests in my local and each institution gets its own row at the top of the submission, but if there are many authors from different institutions I feel it takes up too much space and pushes the submission content further down. Maybe they can be inline and we just show the domain? If we want to show the full name, maybe they hover to display it? @xkopenreview @melisabok 

paper-header

openreview/venue/process/invite_assignment_pre_process.py
melisabok: fixed

openreview/venue/venue.py
melisabok: fixed

openreview/venue/process/invite_assignment_pre_process.py
enrubio: The fix still needs to be made, but in the JS version, we access the `submission.content.authorids?.value`

melisabok: Thanks for the catch, I added a test for unified authors and fixed the preprocess function.

openreview/stages/venue_stages.py
```diff
@@ -588,20 +599,26 @@ def get_content(self, api_version='2', conference=None):
...
-            }
-            content['authorids'] = {
-                'value': ['${{4/id}/content/authorids/value}'],
-                'order':4
-            }
+            if conference and conference.submission_stage.unified_authors:
+                content['authors'] = {
+                    'value': ['${{4/id}/content/authors/value}'],
+                    'order': 3
+                }
```
enrubio, line 606: Previously we had `hidden: true` for the authors field, we don't need it here anymore?

melisabok: the new venues will have only one field to represent the author list ("authors"). This field is not hidden, it has to be shown in the submission form to search for profiles.

enrubio: Ah sorry got it, I confused hidden with readers. That makes sense.

enrubio (approved):
Changes look good to me. I still feel that the institution list at the top of the submission page can be improved, but that can be a separate discussion, just wanted to flag it.

tests/test_submission_authors_email.py
```diff
@@ -0,0 +1,242 @@
...
+                            'We acknowledge that, if our venue\'s reviewing workflow is non-standard, or if our venue is expecting more than a few hundred submissions for any one deadline, we should designate our own Workflow Chair, who will read the OpenReview documentation and manage our workflow configurations throughout the reviewing process.',
+                            'We acknowledge that OpenReview staff work Monday-Friday during standard business hours US Eastern time, and we cannot expect support responses outside those times.  For this reason, we recommend setting submission and reviewing deadlines Monday through Thursday.',
+                            'We will treat the OpenReview staff with kindness and consideration.',
+                            'We acknowledge that authors and reviewers will be required to share their preferred email.',
+                            'We acknowledge that review counts will be collected for all the reviewers and publicly available in OpenReview.',
```
celestemartinez:
```suggestion
                            'We acknowledge that role participation will be collected for all participants—reviewers, area chairs, and senior area chairs—and made publicly available in the OpenReview profile of each participant.',
```

celestemartinez: I suspect this test didn't run? It is failing in my local

celestemartinez (approved):

carlosmondra (approved):

PR #3043, opened 2026-04-28
Desk Rejection Reversion: Re-create forum invitations

### LINKED ISSUE #3038: Official review of desk rejected submission doesn't get propagated changes
In the case of:

1. A submission is desk rejected, the official review invitation gets deleted. The expiration date also passes
2. The review stage is called with an updated expiration date
3. The submission's desk rejection is reverted and the official review invitation is restored

The official review invitation is restored with the original expiration date, so the reviewers are unable to submit a review

haroldrubio:
- Resolves #3038 
This PR calls `create_forum_invitations` in the desk rejection reversion process function to re-create all forum invitation in order to propagate changes to the parent invitation while the submission was desk-rejected


melisabok (approved):
Nice solution!

celestemartinez (commented):
Do we need to do this for withdrawal reversions as well?

enrubio:
> Do we need to do this for withdrawal reversions as well? 

I think yes

haroldrubio:
There was an issue that I don't think I got around to addressing which is that `create_forum_invitations` doesn't handle invitations with `noteReaders` like the invitation edit process function - so I think if you add a test for this in the ARR tests looking at the checklists, it might fail

melisabok:
>There was an issue that I don't think I got around to addressing which is that create_forum_invitations doesn't handle invitations with noteReaders like the invitation edit process function - so I think if you add a test for this in the ARR tests looking at the checklists, it might fail

This has been fixed.

Another issue is the Metadata Revision invitation is being signed by super user so we can not edit it when the desk reject is reverted. We are skipping it. 

In the future we should stop using the super user to run ARR process function. Use an ARR group to give access to all ARR cycles.

melisabok:
@celestemartinez could you review this again?

celestemartinez (approved):

haroldrubio:
> In the future we should stop using the super user to run ARR process function. Use an ARR group to give access to all ARR cycles.

Yes I agree, thanks for the fixes!

PR #3053, opened 2026-05-12
Fix ORCID Authorship_Claim apostrophe mismatch

### LINKED ISSUE #415: ORCID import is unreachable for users with apostrophes in their names — name validator rejects the only string that would match
OpenReview's ORCID importer compares profile names byte-for-byte against the credit-name field in ORCID work records. For works deposited via Crossref (the most common path), the apostrophe is U+2019 (RIGHT SINGLE QUOTATION MARK), not U+0027. However, the profile-name input validator rejects U+2019 with the error "Only letters, single hyphens, single dots at the end of a name, and single spaces are allowed." The straight ASCII apostrophe U+0027 is also rejected. There is therefore no name a user with an apostrophe in their surname can save that will pass the importer's match check.
Repro:

Profile name "Luke O'Connor" (straight ', U+0027) — saves fine, ORCID importer disables all checkboxes.
Try to add variant "Luke O'Connor" (curly ', U+2019) — rejected by validator (error 2026-05-07-14862806).
Crossref/ORCID side: curl -s 'https://api.crossref.org/works/10.1038/s41588-025-02169-3' | jq -r '.message.author[] | select(.family | test("Connor")) | .family' | xxd returns 4f e2 80 99 43 6f 6e 6e 6f 72 — i.e., O + U+2019 + Connor.

Suggested fix: Unicode-normalize both sides of the comparison (NFKC, plus a canonical apostrophe folding U+2019/U+02BC/U+02B9 → U+0027) before matching, in both the importer and the validator. Either accept the curly form in the validator, or fold it on the importer side. Related: #99, #116, #212.

carlosmondra:
Fixes openreview/openreview#415

## Summary
Crossref deposits ORCID `credit-name` values with U+2019 (right single quote), but OpenReview profile names use U+0027 (straight apostrophe). The `Authorship_Claim` preprocess did a strict byte-for-byte `indexOf`, so legitimate authors of papers like `O'Connor` were unable to claim Crossref-sourced publications — the UI surfaced this as disabled "claim authorship" checkboxes.

The fix normalizes both sides of the comparison (NFKC + fold U+2019 / U+02BC / U+02B9 → U+0027) before `indexOf`. The un-normalized publication name is preserved in the error message so users see what was actually compared.

## Changes
- **`openreview/profile/process/author_coreference_pre_process.js`**: added a `normalizeName` helper; applied it to the user's profile names and to the publication's author name before the `indexOf` lookup.
- **`openreview/profile/process/deprecated_dblp_author_coreference_pre_process.js`**: same fix — this file is still wired up via `ProfileManagement.set_deprecated_dblp_ivitations()` and attached to `DBLP.org/-/Author_Coreference`.

## Tests
- **`tests/test_profile_management.py::test_orcid_authorship_claim_unicode_apostrophe`**: verifies that a profile whose `fullname` uses U+0027 can claim an ORCID-imported publication whose author name uses U+2019; and that a genuinely different name still rejects with the un-normalized publication name preserved in the error message.

## Deployment
The preprocess JS is stored on the `Authorship_Claim` invitation (and the deprecated `DBLP.org/-/Author_Coreference` invitation). This change only takes effect after `ProfileManagement.setup()` is re-run so the invitation's `preprocess` field is refreshed with the new code.

## Out of scope (deliberate)
- \`openreview-api\`'s profile-name validator — the U+0027-only canonical-apostrophe policy stays as-is.
- Profile-name storage canonicalization — no migration needed; existing U+0027 profile names will match U+2019 Crossref records after this fix.

### DIFF
--- a/openreview/profile/process/author_coreference_pre_process.js
+++ b/openreview/profile/process/author_coreference_pre_process.js
@@ -1,6 +1,9 @@
 async function process(client, edit, invitation) {
   client.throwErrors = true;
-  
+
+  const normalizeName = (s) =>
+    typeof s === 'string' ? s.normalize('NFKC').replace(/[’ʼʹ]/g, "'") : s;
+
   const authorIndex = edit.content.author_index.value;
   const authorId = edit.content.author_id.value;
   const authorName = edit.content.author_name.value;
@@ -8,7 +11,7 @@ async function process(client, edit, invitation) {
   if (Tools.prettyId(authorId) !== authorName) {
     return Promise.reject(new OpenReviewError({ name: 'Error', message: `The author id ${authorId} doesn't match with the author name ${authorName}` }));
   }
-  
+
   const { notes } = await client.getNotes({ id: edit.note.id });
   const publication = notes[0];
 
@@ -24,7 +27,7 @@ async function process(client, edit, invitation) {
   const userProfile = profiles[0];
 
   const usernames = userProfile.content.names.map(name => name.username).filter(username => !!username);
-  const names = userProfile.content.names.map(name => name.fullname);
+  const names = userProfile.content.names.map(name => normalizeName(name.fullname));
 
   const usernameIndex = usernames.indexOf(authorId);
 
@@ -33,7 +36,7 @@ async function process(client, edit, invitation) {
   }
 
   const publicationAuthorName = publication.content.authors.value[authorIndex]?.fullname;
-  const nameIndex = names.indexOf(publicationAuthorName);
+  const nameIndex = names.indexOf(normalizeName(publicationAuthorName));
   if (nameIndex === -1) {
     return Promise.reject(new OpenReviewError({ name: 'Error', message: `The author name ${publicationAuthorName} from index ${authorIndex} doesn't match with the names listed in your profile` }));
   }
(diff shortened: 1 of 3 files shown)


melisabok (approved):

carlosmondra:
@xkopenreview checked the PR and the problem is not here. The transformation function is converting the tick correctly to `'`, the fix is going to be in openreview-web. The same idea will be applied

xkopenreview:
the fix should probably be placed in the orcid import modal check logic

the issue is:
orcid record has alternate tick, user name has normal tick and user can't select the papers in orcid because the name does not match
user wanted to add alternate tick to name but it's not allowed in names

the whole flow is
web post json with alternate tick
process function call transformation and received alternate tick and name with alternate tick is posted. at this time, forum note has alternate tick
if the user is claiming a paper with alternate tick, the user will input name and tilde id. the name user input is normal tick which match with profile name and the process validation in Authorship_Claim invitation is to compare the name and id user entered and they match. Then name user entered (normal tick) is not compared with the note name (alternate) tick.
after Authorship_Claim edit is posted, the forum note author name is replaced with user entered name (normal tick)

PR #3062, opened 2026-05-28
Edit submission release: control field visibility by editing the release invitation

melisabok:
## Summary

Lets program chairs control **which submission fields are visible, and to whom, when submissions are released** — including releasing author identities and PDFs — by editing the release invitation's content schema, instead of relying on a single `reveal_author_identities` boolean and a process function that computes readers per note.

Field visibility is now expressed as invitation content: a field's `readers` is either a constant list (restrict to those readers) or the escaped delete `{ 'const': { 'delete': True } }` (remove the field-level readers so the field inherits the note's readers). The API stamps that schema onto every released note, so the release becomes declarative and synchronous — no async process step that can fail silently.

This depends on the companion **openreview-api** change that adds the `{ const: { delete: true } }` escape (stored verbatim in the invitation, applied as `{ delete: true }` to note edits).

## Motivation

Previously, releasing author identities was a single boolean on the Readers step, and `submission_release.py` computed and posted `authors`/`authorids` `readers` itself. That approach:

- couldn't express "release this field to whoever can read the note" without naming readers explicitly, and couldn't remove a field's readers at all (a bare `{ delete: true }` in an invitation edit deletes the key from the invitation, not from the notes);
- pushed field-visibility logic into an async process function whose failures the PC never sees;
- was hard-coded to authors/authorids, with no general way to hide/reveal other fields such as the PDF.

### Release template wiring — `workflow_process/submission_release_template_process.py`
- Wire in `set_edit_submission_release_source_invitation` (the **Which_Submissions** source editor) and `set_edit_content_invitation` (the **Form_Fields** content-schema editor) for the release invitation. Both builder methods already exist; this PR connects them to the release step.
- Drop the `reveal_author_identities` boolean from the **Readers** step — the Readers step now edits only note readers; field visibility is edited through **Form_Fields**.

### Release process — `workflows/process/submission_release.py`
- Stop posting `authors`/`authorids` `readers` from the process. Those are now defined in the invitation content schema and stamped onto every note edit by the API.
- Derive bibtex anonymity from the stored `authors.readers` schema (`{ const: { delete: true } }` ⇒ authors revealed).
- Remove the `reveal_author_identities is None` guard and the "Submission Release Failed" status-note flow it drove (no longer reachable — visibility is always defined by the schema).
- The process now posts only what cannot be an invitation constant: `venue`, `_bibtex`, and `odate`/`pdate`.

### Client — `api/client.py`
- Add `OpenReviewClient.delete_invitation(invitation_id)` helper.

### Tests — `tests/test_reviewers_only.py`, `tests/test_acs_and_reviewers.py`
- Release author identities / PDFs by editing the release invitation's `Form_Fields` content schema: post a restricted `readers` list to hide, or `{ 'const': { 'delete': True } }` to release the field to the note's readers.
- Assert both the stored invitation schema and the resulting note content (field `readers` removed on release; note readers unchanged; bibtex regenerated accordingly).
- Replace the old `reveal_author_identities` posts with the schema-driven flow.

## Behavior

| PC action (via Form_Fields) | Stored in invitation | Effect on released notes |
|---|---|---|
| Set field `readers` to a list | `readers: [ ... ]` | Field restricted to those readers |
| Set field `readers` to escaped delete | `readers: { const: { delete: true } }` | Field-level readers removed → field inherits the note's readers |
| Leave field unedited | unchanged | Unchanged |

Re-editing a field after the release activation date has passed re-runs the release with the updated schema, so hide/reveal round-trips without re-scheduling.

## Dependency

Requires the openreview-api PR that introduces `{ const: { delete: true } }` (schema acceptance + unwrap-onto-entity-edits). Without it, posting the escaped delete through Form_Fields is rejected.

## Testing

- `tests/test_reviewers_only.py` — full release flow: accepted release with author identities revealed via escaped delete, rejected reveal-then-hide, PDF hide/unhide, all through Form_Fields. **21 passed.**
- `tests/test_acs_and_reviewers.py` — accepted reveal + rejected hide of author identities via Form_Fields. **14 passed.**

Both files must be run whole (shared server state), per `CLAUDE.md`.


openreview/workflows/process/submission_release.py
```diff
@@ -17,36 +17,20 @@ def process(client, invitation):
...
     meta_invitation_id = domain.content['meta_invitation_id']['value']
     decision_option = invitation.get_content_value('decision_option')
     release_accepted = True if decision_option == 'Accepted' else False
-    status_invitation_id = domain.get_content_value('status_invitation_id')
-    request_form_id = domain.get_content_value('request_form_id')
 
-    support_user = invitation.invitations[0].split('Template')[0] + 'Support'
-
-    reveal_authors = invitation.get_content_value('reveal_author_identities')
-    if reveal_authors is None:
```
celestemartinez, line 34: I see we removed this. Does it mean that this step will always run automatically without input from the PCs?

melisabok: can we? the default could be: same submission readers and do not anonymize authors. 

if they want to make them public and reveal authors then they would need to edit the invitation using Form_Fields.

celestemartinez: Yes, this is fine, especially so that the venue/venueid of submissions always update after decisions.

openreview/workflows/workflow_process/submission_release_template_process.py
```diff
@@ -25,16 +25,8 @@ def process(client, edit, invitation):
     )
 
     edit_invitations_builder = openreview.workflows.EditInvitationsBuilder(client, domain.id)
+    edit_invitations_builder.set_edit_submission_release_source_invitation(invitation_id)
```
celestemartinez: I think we don't need to show this subinvitation. PCs shoudn't change the source (I am working on creating one invitation per decision option). Or if we allow them to edit it, we show it as a json so they can edit the source directly

Image

tests/test_acs_and_reviewers.py
```diff
@@ -1600,33 +1600,60 @@ def test_release_submissions(self, openreview_client, helpers):
...
-                    'value': True
                 }
             }
         )
         helpers.await_queue_edit(openreview_client, edit_id='EFGH.cc/2025/Conference/-/Accepted_Submission_Release-0-1', count=2)
 
+        # release the author identities of accepted submissions through the content schema
+        pc_client.post_invitation_edit(
+            invitations='EFGH.cc/2025/Conference/-/Accepted_Submission_Release/Form_Fields',
+            content={
```
celestemartinez, line 1624: This should be added by default to the invitation right? If we want authorids to be released by default (I know this is what we spoke about, I can add it to my PR instead)

melisabok: yes, I will leave it as it is, you can change it in your PR.

celestemartinez:
In `templates.py`, we are still creating the Submission Release invitation with `authors` and `authorids`: https://github.com/openreview/openreview-py/blob/d48e8123ecea9c58728126bd08653febd56b966c/openreview/workflows/templates.py#L1829

We should remove `authorids` since all new venues use the new authors schema
image

melisabok:
> In templates.py, we are still creating the Submission Release invitation with authors and authorids:

Fixed

celestemartinez (approved):

tests/test_acs_and_reviewers.py
enrubio: To clarify because I'm a bit confused - currently with these changes, if they did not edit the form fields of the Accepted Submission Release, the author identities would still be hidden right? That's what it looks like to me. So this is fine to release for now while Celeste makes the change in another PR?

melisabok: yes, Celeste will do it in her PR.

openreview/workflows/process/submission_release.py
```diff
@@ -17,36 +17,20 @@ def process(client, invitation):
...
-                }
-            )
-        )
-        return
+    # The authors/authorids readers are defined in the invitation content schema: a readers
+    # constant, or the escaped delete { 'const': { 'delete': True } } that the API stamps as
+    # { 'delete': True } on every note edit. The process only uses the schema to build the
+    # bibtex accordingly.
+    authors_readers_schema = invitation.edit.get('note', {}).get('content', {}).get('authors', {}).get('readers')
+    reveal_authors = authors_readers_schema == { 'const': { 'delete': True } }
```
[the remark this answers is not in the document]
enrubio: @melisabok is this true? I do see we have multiple ways to write `{delete: true}` but I'm unsure if we need to account for them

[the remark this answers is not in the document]
melisabok: the invitation needs to have only the value `{ 'const': { 'delete': True } }` to remove the author readers when the invitation is being used. 

that is the new feature. 

using param/const or directly delete:true is removing the field from the invitation.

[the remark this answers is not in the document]
enrubio: oohhh I understand why it's correct now, thank you

enrubio (approved):
Changes look good to me and it looks like some issues are being fixed in another PR

PR #3066, opened 2026-06-01
Add `reviewer_to_reviewer_anonymity` journal setting

melisabok:
Request from SLADS journal.

## Summary

Adds an optional journal setting that hides reviewer identities from other
assigned reviewers. By default the setting is `False`, preserving the current
TMLR behavior where assigned reviewers can see one another's identities.

When enabled, the per-paper `Reviewers` group no longer lists itself as a
deanonymizer, so reviewers see co-reviewers only as anonymized IDs
(e.g. `Reviewer_xYz`). Author–reviewer double-blind anonymity is unchanged.

## Changes

- `journal.py`: new `is_reviewer_to_reviewer_anonymous()` reader (defaults to `False`).
- `group.py`: per-paper reviewers group excludes the reviewers group from
  `deanonymizers` when the setting is enabled.

## Tests

- `test_slads_journal.py`: enables `reviewer_to_reviewer_anonymity` and asserts
  reviewers are excluded from the reviewers group `deanonymizers`.
- `test_journal.py`: asserts TMLR keeps reviewers mutually visible by default.

### DIFF
--- a/openreview/journal/group.py
+++ b/openreview/journal/group.py
@@ -496,9 +496,12 @@ Visit [this page](https://openreview.net/group?id={self.journal.get_expert_revie
 
         reviewers_group=openreview.tools.get_group(self.client, reviewers_group_id)
         if not reviewers_group:
+            reviewer_deanonymizers = [venue_id, action_editors_group_id]
+            if not self.journal.is_reviewer_to_reviewer_anonymous():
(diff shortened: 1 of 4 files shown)


melisabok:
merging because it is needed for SLADS, please post any comments after merge.

PR #3069, opened 2026-06-01
Add self-service profile vouching

melisabok:
## Summary
Lets an established OpenReview user vouch for a profile that the moderation team has **rejected**, activating it without further manual steps. The user posts a `openreview.net/Support/-/Vouch` tag (from the profile UI); a `preprocess` validates the voucher and a `process` activates the rejected profile.

## Changes
- **`profile/process/vouch_pre_process.py`** (new): validates the voucher before the tag is saved — must be `Active`/`Active Institutional`, not have been vouched themselves, have a confirmed institutional email (`get_institutions`), have a declared relation to the vouched user, and stay within their vouching quotas.
  - **Quotas**: a voucher may create at most **5 vouches per rolling 30-day window** and at most **20 vouches in their lifetime**. Both limits are read from the invitation content (`monthLimit` / `lifetimeLimit`) with the defaults as fallback, so the invitation is the single source of truth.
  - The voucher's existing vouches are fetched once (capped just above the lifetime limit) and counted against both quotas in memory. The monthly count uses each tag's **`tcdate` (true creation date)** rather than `tmdate`, so a vouch that is edited later — e.g. re-posted by a profile name-removal cascade — still counts toward the month it was originally created in, not the month of the edit.
- **`profile/process/vouch_process.py`** (new): if the vouched profile is in state `Rejected`, activates it via `moderate_profile('accept')`; otherwise skips.
- **`profile/management.py`**: wires the `Vouch` invitation — `readers=['everyone']`, `invitees=['~']`, the new `preprocess`/`process`, and `content` carrying the quotas the UI surfaces to users:
  ```python
  content={
      'lifetimeLimit': { 'value': 20 },
      'monthLimit': { 'value': 5 }
  }
  ```
- **`api/client.py`**: `activate_user` no longer raises `KeyError` when the activation response has no token (the pending-moderation case).

## Tests
- **`test_profile_management.py::test_vouch_for_profile`**: temporarily enables moderation (restored in `finally`), rejects each profile, then asserts a qualified voucher activates it and that each validation rejects appropriately (non-active voucher, already-vouched voucher, no institutional email, missing relation, >5/month). Also asserts the `Vouch` invitation exposes the quota content (`lifetimeLimit.value == 20`, `monthLimit.value == 5`) so the UI can display users their limits.

## Depends on
[A companion openreview-api change so tag invitations run their `preprocess`, and so a support-group-signed process can moderate profiles. Coordinated separately with the API team.](https://github.com/openreview/openreview-api/pull/1136)

UPDATE: added a test to verify tags are renamed when the name is deleted


openreview/profile/management.py
```diff
@@ -171,10 +171,12 @@ def set_profile_moderation_invitations(self):
...
-                readers=[self.support_group_id],
+                readers=['everyone'],
                 writers=[self.support_group_id],
                 signatures=[self.support_group_id],
-                invitees=[self.support_group_id],
+                invitees=['~'],
+                preprocess=self.get_process_content('process/vouch_pre_process.py'),
+                process=self.get_process_content('process/vouch_process.py'),
                 tag={
                     'id': {
```
xkopenreview, line 185: i think it's missing label here

melisabok: let me add it

melisabok: done

tests/test_profile_management.py
```diff
@@ -4083,4 +4083,168 @@ def test_post_tag_for_blocked_profile(self, openreview_client, support_client, h
...
+            signatures=['openreview.net/Support'],
+            note=openreview.api.Note(
+                id=config_note.id,
+                content={
+                    'moderate': { 'value': 'Yes' },
+                    'weekend_moderation': { 'value': 'Yes' }
+                }
+            )
+        )
+
```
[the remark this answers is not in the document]
melisabok: OpenReview_Config doesn't have a process function so no need to await here

tests/test_profile_management.py
```diff
@@ -4083,4 +4083,168 @@ def test_post_tag_for_blocked_profile(self, openreview_client, support_client, h
...
+                invitation='openreview.net/-/OpenReview_Config',
+                signatures=['openreview.net/Support'],
+                note=openreview.api.Note(
+                    id=config_note.id,
+                    content={
+                        'moderate': { 'value': 'No' },
+                        'weekend_moderation': { 'value': 'Yes' }
+                    }
+                )
+            )
```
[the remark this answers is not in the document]
melisabok: no need to await here

openreview/profile/process/vouch_pre_process.py
```diff
@@ -0,0 +1,37 @@
...
+
+    ## The voucher must have a confirmed institutional email
+    confirmed_emails = voucher_profile.content.get('emailsConfirmed', []) or []
+    has_institutional_email = False
+    for email in confirmed_emails:
+        domain = email.split('@')[-1]
+        institutions = client.get_institutions(domain=domain)
+        if institutions and institutions.get('count', 0) > 0:
+            has_institutional_email = True
+            break
```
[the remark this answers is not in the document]
melisabok: fixed

openreview/profile/process/vouch_pre_process.py
```diff
@@ -0,0 +1,37 @@
...
+
+    if not has_institutional_email:
+        raise openreview.OpenReviewException('You are not allowed to vouch for another user: you must have an institutional email in your profile.')
+
+    ## The voucher must have created fewer than 5 vouches in the last month
+    one_month_ago = openreview.tools.datetime_millis(datetime.datetime.now() - datetime.timedelta(days=30))
+    voucher_vouch_tags = client.get_tags(invitation=invitation.id, signature=voucher_id)
+    recent_vouches = [t for t in voucher_vouch_tags if (t.tcdate or t.cdate or 0) >= one_month_ago]
```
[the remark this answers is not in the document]
melisabok: fixed

celestemartinez (approved):
Looks good to me!

openreview/profile/process/vouch_pre_process.py
```diff
@@ -0,0 +1,71 @@
...
+        for relation in relations
+    )
+    if not has_relation:
+        raise openreview.OpenReviewException('You are not allowed to vouch for this user: you must have a relation to them in your profile.')
+
+    ## The voucher must have created fewer than 5 vouches in the last month. Push the
+    ## 30-day window into the query (mintmdate) and cap the result (limit) so the
+    ## preprocess stays fast even for voucers with a long vouching history.
+    one_month_ago = openreview.tools.datetime_millis(datetime.datetime.now() - datetime.timedelta(days=30))
+    recent_vouches = client.get_tags(invitation=invitation.id, signature=voucher_id, mintmdate=one_month_ago, limit=6)
```
enrubio: If we get vouches by tmdate would it cause an issue with the count since a vouch tag gets re-posted by the name removal process?

xkopenreview: i think the 5 vouches pers 30 days limit might be too permissive, it means the person can get 60 profiles in to OpenReview each year if the person does not care about the vouched profile staying in their profile.

melisabok: I can change the tmdate to tcdate, good point. 

Xukun, I would like a person to be able to vouch for their coauthors when submitting for the first time. Which number do you suggest?

xkopenreview:
right now profile edit page assume the lifetime limit is set in
content?.lifetimeLimit?.value
please let me know if it's set in a different field in the vouch invitation

openreview/profile/process/vouch_pre_process.py
melisabok: fixed

melisabok:
@xkopenreview  I added monthLimit and lifetimeLimit, please check the recent commit

xkopenreview:
the following steps will generate a invitee error which might be hard to interpret
1. voucher with institutional emailA added vouchee in relation
2. voucher added non-institutional emailB and saved profile
3. voucher requested deletion of emailA and su deleted it. voucher keep the same browsing session
4. voucher do vouch for vouchee
5. instead of the non-institutional email error, the non-invitee error is thrown

this is because the token still has the deleted email and transitive member starting from user.id (the deleted email) will not have ~

image

---
discussed in the daily meeting
decided not to do anything as it's rare

xkopenreview (approved):

PR #3076, opened 2026-06-09
 Add test verifying date processes are rescheduled on a venue domain rename

melisabok:
### Summary
The API now re-schedules date process functions when a venue is renamed to a new domain — it moves the venue's process logs and rewrites in-flight BullMQ jobs so nothing is left pointing at the old domain (and adds a `domain` query param to `GET /logs/process`). This PR adds a regression test on the `openreview-py` side that verifies that behavior end to end. No library changes — `rename_venue`/`rename_domain` remain thin pass-throughs.

### What the test does
`tests/test_venue_deployment.py::TestVenueDeployment::test_rename_reschedules_date_processes`:

1. Deploys a venue (`TVRN.cc/2026/Conference`).
2. Renames it to a new domain (`TVRN.org/2026/Conference`) via `rename_venue`.
3. Asserts the date processes are **rescheduled under the new domain**: `Withdrawal-0-1` runs there, and `Withdrawal`, `Withdrawal/Dates`, and `Preferred_Emails` (with its `date_processes`) exist under the new domain.
4. Asserts the **old domain is clean**: the invitation is gone and no process logs remain pointing at the old domain (they were moved to the new one).
5. Asserts a pending date process **actually executes** under the new domain: it schedules the `Withdrawal` `-0-0` (cdate) process a few minutes out via `Withdrawal/Dates` so it is an in-flight job at rename time, renames, then waits for that `-0-0` to fire under the new domain — and confirms it does **not** run under the old domain.

### Notes
- The execution check (step 5) makes the test run ~4 minutes, since it waits for the scheduled date process to fire after the rename.
- `openreview/api/client.py` and `openreview/tools.py` contain only trailing-whitespace cleanups.


celestemartinez (approved):

enrubio (approved):
Tests look good

PR #3083, opened 2026-06-11
 Ignore `independent-researcher.org` domain in conflict-of-interest computation

melisabok:
## Summary

We introduced the domain `independent-researcher.org` as the institution domain for users who declare themselves as independent researchers in their profile history. Because this is a shared placeholder domain rather than a real shared affiliation, it should **not** create conflicts of interest between users.

This PR ensures the conflict-of-interest computation ignores the `independent-researcher.org` domain, so two independent researchers are no longer flagged as conflicted with each other through their history domain.

## Changes

- **`openreview/tools.py`**: Added `independent-researcher.org` to the `common_domains` list in `info_function_builder`. This list is the single point that every conflict policy (`default`, `NeurIPS`, `Comprehensive`, and custom callables) flows through, so the domain is discarded from the conflict domain set after subdomain expansion — across all policies.

- **`tests/test_tools.py`**: Added `test_independent_researcher_conflicts` covering:
  - Two independent researchers (both with `independent-researcher.org` in history) produce **no conflict**, under the `default`, `NeurIPS`, and `Comprehensive` policies.
  - A genuinely shared institution domain (e.g. `cmu.edu`) is **still** detected as a conflict, even when one profile also declares `independent-researcher.org`.
  - The independent-researcher domain remains ignored even when present alongside another institution.

## Notes

`subdomains('independent-researcher.org')` returns just `['independent-researcher.org']` (since `.org` is a TLD), so discarding that single entry fully removes it from the conflict domain set without affecting any other domain.

## Testing

```
tests/test_tools.py::TestTools::test_independent_researcher_conflicts PASSED
============================== 1 passed in 0.10s ===============================
```

### DIFF
--- a/openreview/tools.py
+++ b/openreview/tools.py
@@ -1473,7 +1473,8 @@ def get_all_venues(client):
 def info_function_builder(policy_function):
     def inner(profile, n_years=None, submission_venueid=None):
         common_domains = ['gmail.com', 'qq.com', '126.com', '163.com',
-                    'outlook.com', 'hotmail.com', 'yahoo.com', 'foxmail.com', 'aol.com', 'msn.com', 'ymail.com', 'googlemail.com', 'live.com']
+                    'outlook.com', 'hotmail.com', 'yahoo.com', 'foxmail.com', 'aol.com', 'msn.com', 'ymail.com', 'googlemail.com', 'live.com',
+                    'independent-researcher.org']
         argspec = inspect.getfullargspec(policy_function)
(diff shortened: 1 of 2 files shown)


carlosmondra (approved):

celestemartinez (approved):

enrubio (approved):
Changes and tests look good

xkopenreview (approved):

xkopenreview:
i think the common domains can be those reflected in missing_institutions but we decide to skip

PR #3093, opened 2026-06-23
Fix `TypeError` in Profile Email Removal when profile has DBLP publications

melisabok:
## Problem

Removing an email via the `openreview.net/Support/-/Profile_Email_Removal` invitation crashed when the target profile was an author on a modern DBLP record (`openreview.net/Public_Article/DBLP.org/-/Record`):

```
File ".../request_remove_email_process.py", line 75, in process
    for index, author in enumerate(publication.content.get('authorids', {}).get('value')):
TypeError: 'NoneType' object is not iterable
```

These DBLP notes store authorship as `authors: [{fullname, username}, ...]` dicts (the tilde id lives in `username`) and have **no top-level `authorids` field**. The existing `.get('authorids', {})` default only guards against a *missing* key — not a present key whose `value` is `None` — so `enumerate(None)` raised and the whole email-removal process aborted.

## Fix

`openreview/profile/process/request_remove_email_process.py`:

```python
for index, author in enumerate(publication.content.get('authorids', {}).get('value') or []):
```

Notes without an `authorids` value are now skipped. This is correct for DBLP `Public_Article` notes: their author is already referenced by the tilde id (e.g. `~Harold_Last1`), not the raw email being removed, so there is nothing to rewrite in the note content — it just needs to be left untouched rather than crashed on.

## Tests

`tests/test_profile_management.py::test_remove_email_address` now adds a DBLP publication (with the profile claiming authorship) before the email removal, reproducing the original crash. Assertions were updated to:

- bump the publication count `3 → 4`,
- look notes up by title instead of by fragile positional index (the extra note shifts list ordering),
- verify the DBLP note is left intact (no `authorids` field; `authors` still maps `Harold Last → ~Harold_Last1`).

Full file passes: `28 passed, 1 skipped`.

### DIFF
--- a/openreview/profile/process/request_remove_email_process.py
+++ b/openreview/profile/process/request_remove_email_process.py
@@ -72,7 +72,7 @@ def process(client, edit, invitation):
         readers = None
         writers = None
         needs_change = False
-        for index, author in enumerate(publication.content.get('authorids', {}).get('value')):
+        for index, author in enumerate(publication.content.get('authorids', {}).get('value') or []):
             if email == author:
                 authors.append(preferred_name)
                 authorids.append(preferred_id)
--- a/tests/test_profile_management.py
+++ b/tests/test_profile_management.py
@@ -3436,6 +3436,120 @@ The OpenReview Team.
         assert len(profile.content['names']) == 1
         assert profile.content['names'][0]['username'] == '~Marie_Last1'
 
+    def test_replace_email_unified_authors_submission(self, openreview_client, helpers, support_client):
+        '''An author listed by email (no profile yet) in the unified ``author{}``
+        schema should have its username (and the derived author group membership)
+        replaced with the profile ID once a profile is created for that email.'''
+
+        ## 1. Configure the Submission invitation to allow emails in the username field of the authors object
+        openreview_client.post_invitation_edit(
+            invitations='UNAME.org/2026/Conference/-/Edit',
+            readers=['UNAME.org/2026/Conference'],
+            writers=['UNAME.org/2026/Conference'],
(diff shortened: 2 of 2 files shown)


tests/test_profile_management.py
```diff
@@ -3964,7 +3964,55 @@ def test_remove_email_address(self, openreview_client, helpers, support_client):
...
+            content = {
+                'xml': { 'value': dblp_xml }
+            },
+            note = openreview.api.Note(
+                external_id = 'dblp:journals/corr/abs-9999-12345',
+                content={
+                    'title': { 'value': 'A DBLP Paper' },
+                    'authors': {
+                        'value': [
+                            {'fullname': 'Harold Last', 'username': ''},
```
celestemartinez, line 4106: So for `OpenReview.net/Public_Article/DBLP.org/-/Record` notes, the username will always be either a profile ID or empty?

openreview/profile/process/request_remove_email_process.py
enrubio, line 45: Do API 1 notes have this issue? Would we need the same fix here?

tests/test_profile_management.py
melisabok: correct

openreview/profile/process/request_remove_email_process.py
melisabok: API1 notes do not have author objects

enrubio (approved):
Fix looks good to me

celestemartinez (approved):

melisabok:
do not merge this yet, I want to add a case where note.content.authors.username is an email address and it is replaced correctly.

PR #3102, opened 2026-06-29
Feat/decision emails

celestemartinez:
The main changes include supporting dynamic decision options (like "Accept (Oral)", "Accept (Poster)", etc.), generating separate notification invitations for each decision type, and ensuring that email templates and processes are updated automatically when decision options change. The test suite has also been updated to verify these new behaviors.

**Dynamic Decision Notification Management**

* The system now creates and manages separate notification invitations for each decision option (e.g., `Author_Accept_Oral_Decision_Notification`, `Author_Reject_Decision_Notification`). When decision options are changed, the system automatically deletes old invitations and creates new ones as needed. [[1]](diffhunk://#diff-1fa89c4d1fb091dec33cbb2b11119802c44e48a18a6c52d887111331c3b09423R22-R77) [[2]](diffhunk://#diff-74260f29e7fc95bead83e92cd455b248137d36e7c83c327786064179127996bcL238-R256) [[3]](diffhunk://#diff-30d9aec9183a045aa1dc63028d7bf001f4e17ed5ee9b684fd9dcb4fc05c1bd3cR1076-R1106)

* The `decision_option` is now passed through the workflow, and decision notification emails are sent only to submissions matching the relevant decision. This ensures that authors receive the correct notification for their decision outcome. [[1]](diffhunk://#diff-38907fc3d85563371197af87b7fe34413b945de8239cab0d22713e697ea1d6e0R24-R29) [[2]](diffhunk://#diff-38907fc3d85563371197af87b7fe34413b945de8239cab0d22713e697ea1d6e0R68-R71) [[3]](diffhunk://#diff-38907fc3d85563371197af87b7fe34413b945de8239cab0d22713e697ea1d6e0L81-R89)

**Test Suite Updates**

* Tests have been updated to check for the presence and correct state of the new, dynamically named decision notification invitations, and to verify that the system correctly handles changes to decision options. [[1]](diffhunk://#diff-30d9aec9183a045aa1dc63028d7bf001f4e17ed5ee9b684fd9dcb4fc05c1bd3cR1076-R1106) [[2]](diffhunk://#diff-30d9aec9183a045aa1dc63028d7bf001f4e17ed5ee9b684fd9dcb4fc05c1bd3cL1095-R1120) [[3]](diffhunk://#diff-30d9aec9183a045aa1dc63028d7bf001f4e17ed5ee9b684fd9dcb4fc05c1bd3cL1210-R1235) [[4]](diffhunk://#diff-a8f6f4a0736227326c4df8d2ce912f33d9dbefee1f980addaf13ef9c1da7e47dL137-R138)

### DIFF
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -2207,7 +2207,7 @@ To view your submission, click here: https://openreview.net/forum?id={{{{note_fo
                     'value': decision_stage.options
                 },
                 'accept_decision_options': {
-                    'value': ['Accept (Oral)', 'Accept (Poster)']
+                    'value': decision_stage.accept_options if decision_stage.accept_options else ['Accept (Oral)', 'Accept (Poster)']
                 },
                 'source': {
                     'value': {
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -2207,7 +2207,7 @@ To view your submission, click here: https://openreview.net/forum?id={{{{note_fo
                     'value': decision_stage.options
                 },
                 'accept_decision_options': {
-                    'value': decision_stage.accept_options if decision_stage.accept_options else ['Accept (Oral)', 'Accept (Poster)']
+                    'value': decision_stage.accept_options if decision_stage.accept_options else [option for option in decision_stage.options if 'accept' in option.lower()]
                 },
                 'source': {
                     'value': {
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -2215,7 +2215,7 @@ To view your submission, click here: https://openreview.net/forum?id={{{{note_fo
                     'value': decision_stage.options
                 },
                 'accept_decision_options': {
-                    'value': decision_stage.accept_options if decision_stage.accept_options else [option for option in decision_stage.options if 'accept' in option.lower()]
+                    'value': decision_stage.accept_options
                 },
                 'source': {
                     'value': {
--- a/openreview/workflows/edit_invitations.py
+++ b/openreview/workflows/edit_invitations.py
@@ -1460,7 +1460,7 @@ class EditInvitationsBuilder(object):
             }
 
         invitation_content = {
-            'email_subject': {
+            'subject': {
                 'value': '${4/content/email_subject/value}'
             }
         }
@@ -1473,7 +1473,7 @@ class EditInvitationsBuilder(object):
                 'value': '${4/content/reject_email_content/value}'
             }
         elif is_review_invitation:
-            invitation_content['email_content'] = {
+            invitation_content['message'] = {
                 'value': '${4/content/email_content/value}'
             }
 
--- a/openreview/workflows/edit_invitations.py
+++ b/openreview/workflows/edit_invitations.py
@@ -1409,10 +1409,73 @@ class EditInvitationsBuilder(object):
         self.save_invitation(invitation, replacement=True)
         return invitation
 
-    def set_edit_email_template_invitation(self, super_invitation_id):
+    def set_edit_email_template_invitation(self, super_invitation_id, is_review_invitation=False):
 
         venue_id = self.venue_id
-        invitation_id = super_invitation_id + '/Message'
+        invitation_id = super_invitation_id + '/Templates'
+
+        edit_content = {
+            'email_subject': {
+                'description': 'The subject of the email to be sent to authors.  Make sure not to remove the parenthesized tokens.',
+                'value': {
+                    'param': {
+                        'type': 'string',
+                        'regex': '.+',
+                    }
+                }
+            }
+        }
+
+        if not is_review_invitation:
+            edit_content['accept_email_content'] = {
+                'description': 'The content of the email to be sent to authors for accepted submissions.  Make sure not to remove the parenthesized tokens.',
+                'value': {
+                    'param': {
+                        'type': 'string',
+                        'maxLength': 100000,
+                        'input': 'textarea'
+                    }
+                }
+            }
+            edit_content['reject_email_content'] = {
+                'description': 'The content of the email to be sent to authors for rejected submissions.  Make sure not to remove the parenthesized tokens.',
+                'value': {
+                    'param': {
+                        'type': 'string',
+                        'maxLength': 100000,
+                        'input': 'textarea'
+                    }
+                }
+            }
+        elif is_review_invitation:
+            edit_content['email_content'] = {
+                'description': 'The content of the email to be sent to authors.  Make sure not to remove the parenthesized tokens.',
+                'value': {
+                    'param': {
+                        'type': 'string',
+                        'maxLength': 100000,
+                        'input': 'textarea'
+                    }
+                }
+            }
+
+        invitation_content = {
+            'email_subject': {
+                'value': '${4/content/email_subject/value}'
+            }
+        }
+
+        if not is_review_invitation:
+            invitation_content['accept_message'] = {
+                'value': '${4/content/accept_email_content/value}'
+            }
+            invitation_content['reject_message'] = {
+                'value': '${4/content/reject_email_content/value}'
+            }
+        elif is_review_invitation:
+            invitation_content['email_content'] = {
+                'value': '${4/content/email_content/value}'
+            }
 
         invitation = Invitation(
             id = invitation_id,
@@ -1421,41 +1484,14 @@ class EditInvitationsBuilder(object):
             readers = [venue_id],
             writers = [venue_id],
             edit = {
-                'content': {
-                    'email_subject': {
-                        'description': 'The subject of the email to be sent to authors.  Make sure not to remove the parenthesized tokens.',
-                        'value': {
-                            'param': {
-                                'type': 'string',
-                                'regex': '.+',
-                            }
-                        }
-                    },
-                    'email_content': {
-                        'description': 'The content of the email to be sent to authors.  Make sure not to remove the parenthesized tokens.',
-                        'value': {
-                            'param': {
-                                'type': 'string',
-                                'maxLength': 100000,
-                                'input': 'textarea'
-                            }
-                        }
-                    }
-                },
+                'content': edit_content,
                 'signatures': [self.get_content_value('program_chairs_id', f'{venue_id}/Program_Chairs')],
                 'readers': [venue_id],
                 'writers': [venue_id],
                 'invitation': {
                     'id': super_invitation_id,
                     'signatures': [venue_id],
-                    'content': {
-                        'subject': {
-                            'value': '${4/content/email_subject/value}'
-                        },
-                        'message': {
-                            'value': '${4/content/email_content/value}'
-                        }
-                    }
+                    'content': invitation_content
                 }
             }
         )
--- a/openreview/workflows/edit_invitations.py
+++ b/openreview/workflows/edit_invitations.py
@@ -1436,32 +1436,8 @@ class EditInvitationsBuilder(object):
                         'regex': '.+',
                     }
                 }
-            }
-        }
-
-        if not is_review_invitation:
-            edit_content['accept_email_content'] = {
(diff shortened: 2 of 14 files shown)


openreview/workflows/process/email_decisions_process.py
```diff
@@ -20,7 +20,13 @@ def process(client, invitation):
...
     email_subject = invitation.get_content_value('subject')
-    email_content = invitation.get_content_value('message')
+    accept_email_content = invitation.get_content_value('accept_message')
+    reject_email_content = invitation.get_content_value('reject_message')
+    decision_invitation = client.get_invitation(f'{venue_id}/-/{decision_name}')
+    accept_options = decision_invitation.content.get('accept_decision_options', {}).get('value')
+    print('accept_options', accept_options)
+    decision_field_name = domain.content.get('decision_field_name', {}).get('value', 'decision')
+    print('decision_field_name', decision_field_name)
     fields_to_include = invitation.get_content_value('fields_to_include')
```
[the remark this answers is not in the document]
celestemartinez: I think this can be ignored: older venues with the legacy `message` value will also have the older process function that expects the legacy `message`

openreview/workflows/process/email_decisions_process.py
```diff
@@ -61,12 +68,23 @@ def send_decision_email(submission):
...
-                submission_number=submission.number,
-                submission_title=submission.content['title']['value'],
-                formatted_decision=formatted_decision,
-                submission_forum=submission.id
-            )
+            decision_value = decision.content.get(decision_field_name, {}).get('value')
+            note_accepted = openreview.tools.is_accept_decision(decision_value, accept_options) if decision_value else False
+            print('decision_value:', decision_value, 'note_accepted:', note_accepted)
+            if note_accepted:
+                message = accept_email_content.format(
```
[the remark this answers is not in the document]
celestemartinez: I realize this is how it works in the request form. However, this is a bit more complicated in the new UI since I would have to edit both the invitation and the sub-invitation every time the decision options change (and I also worry this might automatically send emails to authors if the invitation is already active). 

Also, I am not checking if the email has already been sent based on the subject. Maybe I should do this and let the PCs know that if they don't change the subject and an email has already been sent, it will not be sent again. 

Thoughts? @enrubio @melisabok

celestemartinez:
Working on the following:
- creating one decision notification invitation per decision option and updating them as the decision options are updated
- check if an email has already been sent with that subject and skip if true

openreview/venue/invitation.py
```diff
@@ -2207,7 +2207,7 @@ def set_decision_invitation(self):
                     'value': decision_stage.options
                 },
                 'accept_decision_options': {
-                    'value': ['Accept (Oral)', 'Accept (Poster)']
+                    'value': decision_stage.accept_options if decision_stage.accept_options else ['Accept (Oral)', 'Accept (Poster)']
                 },
```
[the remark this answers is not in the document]
celestemartinez: Fixed.

tests/test_acs_and_reviewers.py
```diff
@@ -1219,6 +1244,142 @@ def test_decision_release_stage(self, openreview_client, helpers):
         ]
         assert decisions[0].nonreaders == []
 
+    def test_decion_notification_stage(self, openreview_client, helpers):
```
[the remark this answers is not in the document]
celestemartinez: Fixed.

openreview/workflows/process/email_decisions_process.py
```diff
@@ -21,8 +21,12 @@ def process(client, invitation):
 
     email_subject = invitation.get_content_value('subject')
     email_content = invitation.get_content_value('message')
+    decision_invitation = client.get_invitation(f'{venue_id}/-/{decision_name}')
+    accept_options = decision_invitation.content.get('accept_decision_options', {}).get('value')
+    decision_field_name = domain.content.get('decision_field_name', {}).get('value', 'decision')
```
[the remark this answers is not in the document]
celestemartinez: Fixed.

openreview/workflows/templates.py
```diff
@@ -1459,8 +1467,7 @@ def setup_email_decisions_template_invitation(self):
                             'param': {
                                 'type': 'string',
                                 'maxLength': 100,
-                                'regex': '^[a-zA-Z0-9_]*$',
-                                'default': 'Author_Decision_Notification'
+                                'regex': '^[a-zA-Z0-9_]*$'
```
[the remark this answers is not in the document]
celestemartinez: Fixed.

openreview/workflows/workflow_process/edit_decision_options_process.py
```diff
@@ -16,4 +18,60 @@ def process(client, edit, invitation):
...
+        client.post_invitation_edit(
+            invitations=meta_invitation_id,
+            readers=[venue_id],
+            writers=[venue_id],
+            signatures=[venue_id],
+            invitation=openreview.api.Invitation(
+                id=f'{venue_id}/-/Author_{formatted_decision_option}_Decision_Notification',
+                ddate=now
+            )
+        )
```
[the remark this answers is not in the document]
celestemartinez: Fixed.

melisabok:
it looks like an ICLR test is failing now.

openreview/venue/invitation.py
```diff
@@ -2215,7 +2215,7 @@ def set_decision_invitation(self):
                     'value': decision_stage.options
                 },
                 'accept_decision_options': {
-                    'value': ['Accept (Oral)', 'Accept (Poster)']
+                    'value': decision_stage.accept_options if decision_stage.accept_options else [option for option in decision_stage.options if 'accept' in option.lower()]
```
melisabok: why do we have an else here? can we have a default value for decision_stage.accept_options?

celestemartinez: I am not sure there should be a default. For the new UI it doesn't matter because we would always create the invitations with the default values, but if invitations are created manually this could happen:
1. The PCs choose `options` that are not the default ones, for example: Accept (Oral), Accept (Poster), Revision Needed, Reject
2. The PCs do not choose `accept_options`, so the default are used: Accept

I think it is safer to build them from the `options`. What do you think?

celestemartinez:
> it looks like an ICLR test is failing now.

Should be fixed!

openreview/venue/invitation.py
celestemartinez: Unless you mean setting the default from the options in the stage itself, so we don't have to do it in invitation.py

openreview/workflows/workflow_process/edit_decision_options_process.py
```diff
@@ -16,4 +18,60 @@ def process(client, edit, invitation):
...
+
+    # post new decision notification invitations for the added decision options
+    request_form_inv = domain.get_content_value('request_form_invitation')
+    invitation_prefix =request_form_inv.split('Support')[0] + 'Template'
+    short_name = domain.get_content_value('subtitle')
+    from_email = domain.content['message_sender']['value']['fromEmail']
+
+    for decision_option in added_decision_options:
+        formatted_decision_option = decision_option.replace(' ', '_').replace('(', '').replace(')', '')
+        client.post_invitation_edit(
```
enrubio, line 66: I think we need to update the `formatted_decision_option` replacement in both places because they could also enter a string like `Accept: Poster`. It looks like it’d cause a process function error since the `name` doesn’t pass the regex validation.

To be safer we could also create the new invitations before hard-deleting the old ones, but not necessary.

openreview/workflows/process/email_decisions_process.py
```diff
@@ -37,45 +38,60 @@ def process(client, invitation):
...
+    if not filtered_submissions:
+        print(f'No emails were sent since there are no {decision_option.lower()} submissions')
+        return
 
-    openreview.tools.concurrent_requests(send_decision_email, active_submissions)
+    openreview.tools.concurrent_requests(send_decision_email, filtered_submissions)
 
-    print('Decision emails sent to authors')
\ No newline at end of file
+    print(f'{len(filtered_submissions)} {decision_option} decision emails sent to authors')
```
enrubio: Minor but this may not be correct since we filter out duplicate messages. Might be confusing for us/PCs if we look at the logs.

openreview/workflows/edit_invitations.py
```diff
@@ -1422,10 +1422,32 @@ def set_edit_email_date_invitation(self, super_invitation_id, due_date=None):
         self.save_invitation(invitation, replacement=True)
         return invitation
 
-    def set_edit_email_template_invitation(self, super_invitation_id):
+    def set_edit_email_template_invitation(self, super_invitation_id, is_review_invitation=False):
```
enrubio: Just checking - looks like `is_review_invitation` isn't use in this invitation, should it be?

openreview/workflows/workflow_process/edit_decision_options_process.py
celestemartinez: What do you mean update in both places? Maybe we can have a regex for the decision options so they don't add `:`

enrubio: Oh I meant that we format the decision in both line 50 & 65 so both would need to be updated. But I like the idea of using regex instead

openreview/workflows/edit_invitations.py
celestemartinez: Fixed.

openreview/workflows/process/email_decisions_process.py
celestemartinez: Fixed.

openreview/workflows/workflow_process/edit_decision_options_process.py
celestemartinez: Regex added.

openreview/venue/invitation.py
celestemartinez: [Fixed](https://github.com/openreview/openreview-py/pull/3102/commits/5c34a98fcc5443fa6a78b3fc3cb63cf8a1e38d43)! Let me know if that was not what you meant

melisabok (approved):

openreview/workflows/workflow_process/edit_decision_options_process.py
enrubio: I think the regex we may need to use is something like this: `^[\w ()]+$`
Which should match alphanumeric characters, underscores, spaces, parentheses.
Just because they _could_ include other characters that would cause it to fail.

openreview/workflows/workflow_process/email_authors_template_process.py
```diff
@@ -2,12 +2,32 @@ def process(client, edit, invitation):
 
     domain = client.get_group(edit.domain)
     meta_invitation_id = domain.content.get('meta_invitation_id', {}).get('value')
+    short_name = domain.content['subtitle']['value']
```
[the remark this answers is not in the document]
celestemartinez: Ignore

openreview/workflows/workflow_process/edit_decision_options_process.py
```diff
@@ -16,4 +18,60 @@ def process(client, edit, invitation):
...
+    now = openreview.tools.datetime_millis(datetime.datetime.now())
+
+    if len(invitation_edits) > 1:
+        # delete the previous edit's decision notification invitations
+        previous_edit = invitation_edits[-2]
+        previous_decision_options = previous_edit.content['decision_options']['value']
+    else:
+        #delete the default decision notification invitations
+        previous_decision_options = ['Accept', 'Reject']
+
```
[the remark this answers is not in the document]
celestemartinez: Ignore

openreview/workflows/workflow_process/edit_decision_options_process.py
celestemartinez: I don't think we've had issues with the decision options in the past but I can change this

celestemartinez: Fixed.

enrubio (approved):
Changes look great!

PR #3107, opened 2026-06-30
Retry POST requests aborted by stale/broken connections

carlosmondra:
## Summary
A connection reset/abort on a POST (e.g. a stale keep-alive socket the server closed during a long idle phase) was fatal because `LogRetry` only retried on 5xx statuses and urllib3 never retries POST by default. This makes both API clients recover from broken-connection resets instead of raising `ConnectionError` mid-job.

## Changes
- **`openreview/api/client.py`** & **`openreview/openreview.py`**: Teach `LogRetry` to detect a broken-connection signature (connection aborted/reset, ECONNRESET, broken pipe) and mark the request method retryable for that case only, letting the existing backoff machinery reconnect and resend. Plain read timeouts are deliberately excluded so writes the server may have already processed are not replayed.

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -27,9 +27,31 @@ from .. import tools
 from .. import mfa
 
 class LogRetry(Retry):
-     
+
+    # Connection failures whose signature means the request almost certainly
+    # never reached the server (e.g. a stale keep-alive socket the server had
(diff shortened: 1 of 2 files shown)


carlosmondra:
It didn't fix the issue

PR #3135, opened 2026-07-23
Improve performance of `invitation_edit_process`

melisabok:
## Problem

The process that creates/updates per-paper child invitations (reviews, meta reviews, comments, etc.) was slow for large venues. For N matching submissions it made roughly **3N API calls**: one `post_invitation_edit` per paper plus two extra GET round-trips per paper that can be avoided entirely.

## Changes

1. **Removed the `get_invitation` call after each `post_invitation_edit`.** The edit response already contains the values of the saved invitation, so the child invitation is now built directly from `paper_invitation_edit['invitation']` instead of being re-fetched by id.

2. **Removed the `get_notes` call per child invitation in `update_note_readers`.** The submissions are fetched with `details='replies'`, so the notes under each child invitation are already in memory — they are now selected by filtering `forum_note.details['replies']` by invitation id (falling back to the forum note itself when the invitation edits the submission directly).

3. **Parallelized the deletion sweep.** Child invitations that no longer match the source (e.g. desk-rejected or withdrawn papers) were deleted one at a time in a serial loop; they now go through `openreview.tools.concurrent_requests` like the creation pass.

4. **Readability:** the tuple unpacking in `post_invitation` no longer shadows its parameter (`childNote, forumNote = note`), and `update_note_readers` receives the `(note, forum_note)` tuple and unpacks it internally, keeping the two-argument call site.

## Impact

Per process run, API calls go from ~3N + 2 down to **N + 2** (1 submissions fetch, 1 child-invitations fetch, N mandatory invitation-edit posts), plus only the note edits that actually change readers. No GET requests scale with the number of submissions anymore, and stale-invitation deletion is no longer serial.

## Testing

- `tests/test_venue_submission.py` — 17 passed on a clean database.

### DIFF
--- a/openreview/venue/process/invitation_edit_process.py
+++ b/openreview/venue/process/invitation_edit_process.py
@@ -73,9 +73,12 @@ def process(client, invitation):
 
         return filter_by_source(source)
     
-    def update_note_readers(submission, paper_invitation):
-        ## Update readers of current notes
-        notes = client.get_notes(invitation=paper_invitation.id)
+    def update_note_readers(note, paper_invitation):
+        submission, forum_note = note
+        ## Update readers of current notes, they are available in the forum replies
+        notes = [openreview.api.Note.from_json(reply) for reply in forum_note.details['replies'] if paper_invitation.id in reply['invitations']]
+        if paper_invitation.id in forum_note.invitations:
+            notes.append(forum_note)
         invitation_readers = paper_invitation.edit['note'].get('readers', [])
 
         ## if invitation has param in readers, we ignore the update
@@ -130,7 +133,7 @@ def process(client, invitation):
 
     def post_invitation(note):
 
-        note, forumNote = note
+        childNote, forumNote = note
 
         def find_note_from_details(note_id):
             if note_id == forumNote.id:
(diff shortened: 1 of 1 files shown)


openreview/venue/process/invitation_edit_process.py
```diff
@@ -73,9 +73,12 @@ def filter_by_source(source):
...
     
-    def update_note_readers(submission, paper_invitation):
-        ## Update readers of current notes
-        notes = client.get_notes(invitation=paper_invitation.id)
+    def update_note_readers(note, paper_invitation):
+        submission, forum_note = note
+        ## Update readers of current notes, they are available in the forum replies
+        notes = [openreview.api.Note.from_json(reply) for reply in forum_note.details['replies'] if paper_invitation.id in reply['invitations']]
+        if paper_invitation.id in forum_note.invitations:
+            notes.append(forum_note)
```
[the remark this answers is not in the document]
melisabok: ignore, they should always come, check how we get the source.

enrubio (approved):
Changes look good to me!

openreview/venue/process/invitation_edit_process.py
```diff
@@ -73,9 +73,12 @@ def filter_by_source(source):
 
         return filter_by_source(source)
     
-    def update_note_readers(submission, paper_invitation):
-        ## Update readers of current notes
-        notes = client.get_notes(invitation=paper_invitation.id)
+    def update_note_readers(note, paper_invitation):
+        submission, forum_note = note
+        ## Update readers of current notes, they are available in the forum replies
+        notes = [openreview.api.Note.from_json(reply) for reply in forum_note.details['replies'] if paper_invitation.id in reply['invitations']]
```
celestemartinez, line 79: Just trying to understand these two lines:
- line 79 gets all notes if the notes are replies of the submission
- lines 80-81 append the submission itself if the invitation matches

These two things are distinct sets, right?

melisabok: this should replace the get_notes by child invitation, because we do not know which type of child invitation is, we try to find all the notes that contain that match invitation, it could be the forum itself or one of their replies. 

May be line 80-81 because we asume this process function is only being used for forum replies and not forum edits.

carlosmondra (approved):

celestemartinez (approved):

PR #3137, opened 2026-07-24
Venue: support filtering submissions by decision options

celestemartinez:
This pull request enhances support for filtering submissions based on specific decision options, not just a general accept/reject status. It introduces logic to handle the new `decision_options` parameter throughout the codebase, ensuring that features like invitation filtering and content visibility work correctly when targeting submissions with particular decision decisions. Additionally, a new test verifies this behavior.

**Filtering and Matching Enhancements:**

* Added logic in `openreview/stages/venue_stages.py` to determine if only accepted decision options are targeted by comparing the source's `decision_options` to the venue's accept options, and updated content visibility accordingly. [[1]](diffhunk://#diff-5753bcb1ca462c4479bbbc4be41859a89bd080fce2d2a4cb2859d0fe31a7a722L636-R636) [[2]](diffhunk://#diff-5753bcb1ca462c4479bbbc4be41859a89bd080fce2d2a4cb2859d0fe31a7a722R650-R661)
* Updated `openreview/tools.py` so that both `with_decision_accept` and `decision_options` can trigger decision-based filtering, and added checks to ensure submissions match the specified decision options. [[1]](diffhunk://#diff-a0a408fd57af11c3a81cfd044e307e78e5b96b29d4ec2f55a62cd5667b12b45dL2149-R2149) [[2]](diffhunk://#diff-a0a408fd57af11c3a81cfd044e307e78e5b96b29d4ec2f55a62cd5667b12b45dL2162-R2173)
* Modified `openreview/venue/process/invitation_edit_process.py` to give precedence to `decision_options` over `with_decision_accept` when filtering source submissions.

**Testing:**

* Added a new test `test_filter_by_decision_option` in `tests/test_workshop_v2.py` to verify that filtering submissions by a specific decision option (e.g., "Invite to Venue") works as expected.

### DIFF
--- a/openreview/stages/venue_stages.py
+++ b/openreview/stages/venue_stages.py
@@ -633,7 +633,7 @@ class SubmissionRevisionStage():
                 if field not in content:
                     content[field] = { 'delete': True }
 
-            only_accepted = self.only_accepted
+            only_accepted = self.only_accepted or self.only_accepted_decision_options(conference)
 
             hidden_field_names = conference.submission_stage.get_hidden_field_names()
             
@@ -646,6 +646,18 @@ class SubmissionRevisionStage():
                     content[field]['readers'] = { 'delete': True }                        
 
         return content
+
+    def only_accepted_decision_options(self, venue):
+        # True when the source's decision_options exactly match the venue's accept
+        # options, so the stage targets only accepted papers and authors/authorids
+        # should be hidden like the only_accepted / with_decision_accept path.
+        decision_options = self.source.get('decision_options')
+        if not decision_options:
+            return False
+        accept_options = venue.client.get_group(venue.venue_id).content.get('accept_decision_options', {}).get('value')
+        if not accept_options:
+            return False
+        return set(decision_options) == set(accept_options)
 
     def get_source_submissions(self, venue):
 
--- a/openreview/venue/process/invitation_edit_process.py
+++ b/openreview/venue/process/invitation_edit_process.py
@@ -52,10 +52,18 @@ def process(client, invitation):
                 source_submissions = client.get_all_notes(content={ 'venueid': ','.join([venue_id, rejected_venue_id]) }, sort='number:asc', details='replies', domain=venue_id)
             
             if 'with_decision_accept' in source:
-                source_submissions = [s for s in source_submissions 
-                                      if len([r for r in s.details['replies'] 
-                                        if f'{venue_id}/{submission_name}{s.number}/-/{decision_name}' in r['invitations'] 
+                source_submissions = [s for s in source_submissions
+                                      if len([r for r in s.details['replies']
+                                        if f'{venue_id}/{submission_name}{s.number}/-/{decision_name}' in r['invitations']
                                         and openreview.tools.is_accept_decision(r['content'][decision_field_name]['value'], accept_options) == source.get('with_decision_accept')]) > 0]
+
+            # decision_options has precedence over with_decision_accept
+            if 'decision_options' in source:
+                decision_options = source.get('decision_options')
+                source_submissions = [s for s in source_submissions
+                                      if any(f'{venue_id}/{submission_name}{s.number}/-/{decision_name}' in r['invitations']
+                                        and r['content'][decision_field_name]['value'] in decision_options
+                                        for r in s.details['replies'])]
 
             if 'readers' in source:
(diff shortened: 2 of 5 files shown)


openreview/venue/process/invitation_edit_process.py
```diff
@@ -52,11 +52,19 @@ def filter_by_source(source):
...
-                source_submissions = [s for s in source_submissions 
-                                      if len([r for r in s.details['replies'] 
-                                        if f'{venue_id}/{submission_name}{s.number}/-/{decision_name}' in r['invitations'] 
+                source_submissions = [s for s in source_submissions
+                                      if len([r for r in s.details['replies']
+                                        if f'{venue_id}/{submission_name}{s.number}/-/{decision_name}' in r['invitations']
                                         and openreview.tools.is_accept_decision(r['content'][decision_field_name]['value'], accept_options) == source.get('with_decision_accept')]) > 0]
 
+            # decision_options has precedence over with_decision_accept
+            if 'decision_options' in source:
```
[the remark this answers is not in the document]
celestemartinez: The source_submissions returned when filtering by `with_decision_accept` get overwritten if `decision_options` is present. I think we can ignore

openreview/venue/process/invitation_edit_process.py
```diff
@@ -52,11 +52,19 @@ def filter_by_source(source):
                 source_submissions = client.get_all_notes(content={ 'venueid': ','.join([venue_id, rejected_venue_id]) }, sort='number:asc', details='replies', domain=venue_id)
             
             if 'with_decision_accept' in source:
```
melisabok: `with_decision_accept` should be deprecated and use `decision_options` instead. Can you confirm that the new workflow won't use with_decision_accept anymore?

celestemartinez: I am now thinking that having `with_decision_accept` was nice because I don't need to update this value every time PCs change the decision options. It looks like the submission release invitation is still using `with_decision_accept` so I need to update that

melisabok: ok, we can keep it. Do we ask the PCs to tell us which decisions represent an accept using the new UI?

celestemartinez: yes, there is a place for them to type in which decision options represent acceptance:

image

openreview/stages/venue_stages.py
```diff
@@ -647,6 +647,18 @@ def get_content(self, api_version='2', conference=None):
...
+        # True when the source's decision_options exactly match the venue's accept
+        # options, so the stage targets only accepted papers and authors/authorids
+        # should be hidden like the only_accepted / with_decision_accept path.
+        decision_options = self.source.get('decision_options')
+        if not decision_options:
+            return False
+        accept_options = venue.client.get_group(venue.venue_id).content.get('accept_decision_options', {}).get('value')
+        if not accept_options:
+            return False
+        return set(decision_options) == set(accept_options)
```
enrubio, line 660: `decision_options` could be a subset of `accept_options` too, right?

celestemartinez: Here I wanted `only_accepted` to work exactly like it used to before, which means it should be True only when the `decision_options` include all the `accept_options`. This is only used in the SubmissionRevision stage.

openreview/venue/process/invitation_edit_process.py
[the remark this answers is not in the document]
enrubio: It looks like in the case where `with_decision_accept = True` and `decision_options = ['Reject']` I think the submission list is empty? But it sounds like you'd expect it to be rejected submissions?

openreview/venue/process/invitation_edit_process.py
melisabok: are we using the "server as reviewer" validation type here? I think we should.

celestemartinez: I had created a new PR for that: https://github.com/openreview/openreview-py/pull/3141, we still need to figure out web

openreview/venue/process/invitation_edit_process.py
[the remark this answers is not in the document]
celestemartinez: This has been fixed. We should check first `decision_options` and if it exists in the `source` we should ignore `with_decision_accept`. We can eventually get rid of `with_decision_accept`

enrubio (approved):
Changes look good!

melisabok (approved):

PR #3139, opened 2026-07-27
Add human verification to all invitations with attachments

melisabok:
## Summary

Adds a default `humanVerificationRequired` rate limit to every invitation that
accepts a file/attachment upload. This throttles automated abuse of attachment
endpoints (e.g. a bot uploading files or posting submissions in a tight loop)
while leaving normal usage unaffected.

The default limit is:

```python
{ "limit": 15, "windowMs": 3600000 }   # 15 uploads per hour, per user
```

Previously this was only applied to the Submission invitation, and only when a
venue explicitly set a `submission_human_verification` value on its domain
group. That opt-in property is removed — the limit is now applied automatically
and uniformly wherever an invitation's content contains an attachment field.

### Core

- **`openreview/tools.py`** — new helper `content_has_attachments(content)` that
  detects whether an invitation's content definition contains any `file`-type
  field, plus a `DEFAULT_HUMAN_VERIFICATION` constant holding the default limit.
- **`openreview/venue/invitation.py`** — new builder helper
  `get_human_verification(content)` returns the default limit when the content
  has attachments, otherwise `None`. It is now applied across all venue stage
  invitations whose content is configurable: Submission, PC Submission Revision,
  Review, Rebuttal, Meta Review, Decision, Submission Revision / Full Submission,
  Custom Stage, Ethics Review, and Registration. The committee matching-setup
  invitation (affinity-score CSV upload) gets the default directly.
- **`openreview/workflows/edit_invitations.py`** — the file-upload PC
  invitations (`Upload_Scores`, `Decision_CSV`) get the default directly.
- **`openreview/profile/management.py`** — the Archive `Direct_Upload`,
  `Anonymous_Preprint/-/Submission`, and News `Article` invitations get the
  default directly.
- **`openreview/journal/invitation.py`** — the journal Author Submission,
  Revision, Camera-Ready Revision, and EIC Revision invitations get the default
  directly.
- **`openreview/arr/invitation.py`** — the ARR Submission Metadata Revision
  invitation applies the same attachment detection.

### Removal of the old opt-in property

The `submission_human_verification` domain/request-form property is removed, since
the limit is now applied automatically:

- `openreview/venue/venue.py` — removed the attribute initialization.
- `openreview/venue/helpers.py` — removed the domain-group read.
- `openreview/conference/helpers.py` — removed the request-form read.

## Tests

- **Assertions added** verifying the default limit is present on attachment
  invitations: `test_venue_submission.py` (Submission + a per-paper
  Camera-Ready Revision child invitation), `test_submission_limits.py`
  (template-workflow Submission), `test_journal.py` (TMLR Submission), and
  `test_profile_management.py` (Archive `Direct_Upload`).
- **`test_neurips_conference_v2.py`** — dropped the now-obsolete domain-group
  setup that set the old `submission_human_verification` property; its existing
  assertion now validates the built-in default instead.
- **Bulk-posting tests fixed** — five integration tests post 50–101 submissions
  as a single user in a loop, which would trip the new 15/hour limit. Each bulk
  loop now posts through the super user (`openreview_client`), which is exempt
  from the rate limit, keeping the same author signatures:
  `test_cvpr_conference_v2.py`, `test_icml_conference.py`,
  `test_icml_conference_with_templates.py`, `test_venue_with_tracks.py`,
  `test_arr_venue_v2.py`.

## Notes

- A completeness sweep confirmed every `type: 'file'` field in the V2 codebase
  now flows through one of these paths, and that no other test file bulk-posts
  more than 15 submissions as a single user.
- `humanVerificationRequired` is a V2-only invitation property; the legacy
  `conference/` and `venue_request/` (V1) builders are intentionally untouched.

### DIFF
--- a/openreview/venue/invitation.py
+++ b/openreview/venue/invitation.py
@@ -115,6 +115,11 @@ class InvitationBuilder(object):
             process = f.read()
             return process
 
+    def get_human_verification(self, content):
+        if tools.content_has_attachments(content):
+            return tools.DEFAULT_HUMAN_VERIFICATION
+        return None
+
     def set_meta_invitation(self):
         venue_id=self.venue_id
         meta_invitation = openreview.tools.get_invitation(self.client, self.venue.get_meta_invitation_id())
@@ -175,7 +180,7 @@ class InvitationBuilder(object):
             cdate = submission_cdate,
             duedate = submission_duedate,
             expdate = tools.datetime_millis(submission_stage.exp_date) if submission_stage.exp_date else None,
-            humanVerificationRequired = self.venue.submission_human_verification,
+            humanVerificationRequired = self.get_human_verification(content),
             content = {
                 'submission_email_template': {
                     'value': f'''Your submission to {self.venue.short_name} has been {{{{action}}}}.
@@ -513,6 +518,7 @@ To view your submission, click here: https://openreview.net/forum?id={{{{note_fo
             readers = [venue_id],
             writers = [venue_id],
             cdate = cdate,
+            humanVerificationRequired = self.get_human_verification(content),
             edit = {
                 'signatures': [self.venue.get_program_chairs_id()],
                 'readers': [self.venue.get_program_chairs_id(), self.venue.get_authors_id('${2/note/number}')],
@@ -688,6 +694,10 @@ To view your submission, click here: https://openreview.net/forum?id={{{{note_fo
     exec(script, funcs)
     funcs['process'](client, edit, invitation)
 '''
+
+        human_verification = self.get_human_verification(content)
+        if human_verification:
+            invitation.edit['invitation']['humanVerificationRequired'] = human_verification
 
         if review_duedate:
             invitation.edit['invitation']['duedate'] = review_duedate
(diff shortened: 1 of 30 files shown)


openreview/venue/invitation.py
```diff
@@ -115,6 +115,11 @@ def get_process_content(self, file_path):
             process = f.read()
             return process
 
+    def get_human_verification(self, content):
+        if tools.content_has_attachments(content):
+            return tools.DEFAULT_HUMAN_VERIFICATION
+        return None
```
carlosmondra: I think it should do human verification as long as it is a submission invitation. I wouldn't condition it on it having attachments

melisabok: so rebuttals with attachments should not have human verification.

melisabok: done

carlosmondra (approved):

openreview/venue/invitation.py
```diff
@@ -175,7 +175,7 @@ def set_submission_invitation(self):
             cdate = submission_cdate,
             duedate = submission_duedate,
             expdate = tools.datetime_millis(submission_stage.exp_date) if submission_stage.exp_date else None,
-            humanVerificationRequired = self.venue.submission_human_verification,
+            humanVerificationRequired = tools.DEFAULT_HUMAN_VERIFICATION,
```
celestemartinez, line 178: Will we ever want to change this value? I was thinking we might want to still read it from the domain if it exists and if not we use the default? I am not sure if we will ever want to overwrite this value

melisabok: with the new workflow configuration, all the invitations will be created on deployment time so I don't think that saving it in the domain group would help.

celestemartinez (approved):

enrubio (approved):
This looks good to me

PR #3140, opened 2026-07-27
adjusting endpoint in client to download data from signed url automatically when requested

cagostino:


openreview/api/client.py
```diff
@@ -3479,7 +3479,7 @@ def get_expertise_metadata(self, job_id, baseurl=None):
         print('get expertise metadata', response_json)
         return response_json
 
-    def get_expertise_results(self, job_id, baseurl=None, wait_for_complete=False, format='json'):
+    def get_expertise_results(self, job_id, baseurl=None, wait_for_complete=False, format='json', full=False, output_filename=None):
```
melisabok: instead of a new param, could we have a get_expertise_all_results?

melisabok (approved):

enrubio (approved):
This works, thanks!

PR #3142, opened 2026-07-30
fix empty list of recipients

melisabok:
Make sure the list of recipients is not empty before posting a message and do not send any email to PCs if the emails_pcs is False.

### DIFF
--- a/openreview/venue/process/comment_process.py
+++ b/openreview/venue/process/comment_process.py
@@ -134,15 +134,20 @@ To view the comment, click here: https://openreview.net/forum?id={submission.id}
                 )
 
     #send email to author of comment
-    client.post_message(
-        invitation=meta_invitation_id,
-        recipients=[edit.tauthor] if edit.tauthor != 'OpenReview.net' else [],
-        subject=f'''[{short_name}] Your comment was received on Paper Number: {submission.number}, Paper Title: "{submission.content['title']['value']}"''',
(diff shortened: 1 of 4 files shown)


celestemartinez (approved):

openreview/venue/process/comment_process.py
```diff
@@ -134,15 +134,20 @@ def process(client, edit, invitation):
...
-    client.post_message(
-        invitation=meta_invitation_id,
-        recipients=[edit.tauthor] if edit.tauthor != 'OpenReview.net' else [],
-        subject=f'''[{short_name}] Your comment was received on Paper Number: {submission.number}, Paper Title: "{submission.content['title']['value']}"''',
-        message=f'''Your comment was received on a submission to {short_name}.{content}''',
-        replyTo=contact,
-        signature=venue_id,
-        sender=sender
-    )
+    comment_author_recipients = [edit.tauthor] if edit.tauthor.lower() != 'openreview.net' else []
```
enrubio, line 137: Minor but maybe we want to update the same check in the rebuttal process and in arr's comment process. It also has:
`recipients=[edit.tauthor] if edit.tauthor != 'OpenReview.net' else []`

melisabok: done

enrubio (approved):
Looks good!

PR #3143, opened 2026-07-30
Bulk attachment download: `tools.get_all_attachments()`

melisabok:
## Problem

Downloading the files of a venue meant one HTTP request per submission:

```python
for submission in client.get_all_notes(content={'venueid': venue_id}):
    data = client.get_attachment(id=submission.id, field_name='latex_source_zip')
```

The API already supports fetching up to 50 attachments in a single request — `GET /attachment?ids=<comma separated>&name=<field>` archives them and returns one zip — but the client had no explicit way to use it, and nothing in the library turned "all submissions of a venue" into "all their files on disk".

### 1. `OpenReviewClient.get_attachments(ids, field_name)` — `openreview/api/client.py`

Returns the binary content of a zip file with one file per note.

```python
zip_bytes = client.get_attachments(ids=['abc123', 'def456'], field_name='pdf')
with open('attachments.zip', 'wb') as f:
    f.write(zip_bytes)
```

- No client-side validation of the list: the API already validates it (`anyRequired: ['id', 'ids']` and `maxUniqueItems` for `GET/attachment`), and the limit is the API's to change without needing a client release.
- Parameter order is `ids` first, then `field_name`.

### 2. `openreview.tools.get_all_attachments(client, venueid, field_name, output_dir=None)`

Downloads the attachments of every note with the given `venueid` and extracts them into a directory, returning the list of file paths.

```python
files = openreview.tools.get_all_attachments(
    client,
    venueid='auai.org/UAI/2026/Conference',
    field_name='latex_source_zip'
)
```

- Fetches submissions with `client.get_all_notes(content={'venueid': venueid}, sort='number:asc')`.
- **Skips notes without a value in `field_name`.** This matters: the server resolves the ids of a batch with `Promise.all`, so a single note missing that field makes the whole request 404. Filtering client-side keeps optional fields (supplementary material, latex sources) usable.
- Requests the rest in batches of 50, **sequentially**, and extracts each returned zip into `output_dir` (defaults to a directory named after the field). Concurrent batches were considered and deliberately left out to avoid tripping API rate limits.
- **Extracts each member individually instead of `extractall`**, so the returned list contains the path of every file that was actually written.
- **Never leaves a lone note in the last batch** (101 notes → 50 / 49 / 2). The API streams the raw file instead of a zip when a single id resolves, which would not open as a zip. A venue with exactly one note holding the field is not supported for the same reason.

## Verification

- Batching, extraction, filtering of notes without the field, and the request the tool issues were exercised with a stub client across 2 / 51 / 100 / 101 / 1000 notes.
- No integration tests added; the HTTP round trip itself is not covered by automated tests.

### DIFF
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -829,6 +829,43 @@ class OpenReviewClient(object):
         response = self.__handle_response(response)
         return response.content
 
+    def get_attachments(self, ids, field_name):
+        """
+        Gets a zip file with the attachments of the notes with the provided ids.
+
+        Note that when only one of the ids resolves to a note the server streams that
+        file directly instead of a zip.
+
+        :param ids: List of Note ids or Reference ids. The max number of ids is 50
+        :type ids: list[str]
+        :param field_name: name of the field associated with the attachment files
+        :type field_name: str
+
+        :return: The binary content of a zip file containing one file per note
+        :rtype: bytes
+
+        Example:
+
+        >>> f = client.get_attachments(ids=['Place Note-ID here', 'Place Note-ID here'], field_name='pdf')
+        >>> with open('attachments.zip','wb') as op: op.write(f)
+
+        """
+
+        if not ids:
+            raise OpenReviewException('Provide a non empty list of ids')
+
+        if len(set(ids)) > 50:
+            raise OpenReviewException('The max number of ids is 50')
+
+        params = {
+            'name': field_name,
+            'ids': ','.join(ids)
+        }
+
+        response = self.session.get(self.baseurl + '/attachment', params=tools.format_params(params), headers = self.headers)
+        response = self.__handle_response(response)
+        return response.content
+
     def get_venues(self, id=None, ids=None, invitations=None):
         """Get a list of Venue objects based on the filters provided.
 
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -829,37 +829,6 @@ class OpenReviewClient(object):
         response = self.__handle_response(response)
         return response.content
 
-    def get_attachments(self, ids, field_name):
-        """
-        Gets a zip file with the attachments of the notes with the provided ids.
-
-        Note that when only one of the ids resolves to a note the server streams that
-        file directly instead of a zip.
-
-        :param ids: List of Note ids or Reference ids. The max number of ids is limited by the API
-        :type ids: list[str]
-        :param field_name: name of the field associated with the attachment files
-        :type field_name: str
-
-        :return: The binary content of a zip file containing one file per note
-        :rtype: bytes
-
-        Example:
-
-        >>> f = client.get_attachments(ids=['Place Note-ID here', 'Place Note-ID here'], field_name='pdf')
-        >>> with open('attachments.zip','wb') as op: op.write(f)
-
-        """
-
-        params = {
-            'name': field_name,
-            'ids': ','.join(ids)
-        }
-
-        response = self.session.get(self.baseurl + '/attachment', params=tools.format_params(params), headers = self.headers)
-        response = self.__handle_response(response)
-        return response.content
-
     def get_venues(self, id=None, ids=None, invitations=None):
         """Get a list of Venue objects based on the filters provided.
 
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -836,7 +836,7 @@ class OpenReviewClient(object):
         Note that when only one of the ids resolves to a note the server streams that
         file directly instead of a zip.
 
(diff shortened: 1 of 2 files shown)


openreview/api/client.py
```diff
@@ -829,6 +829,43 @@ def get_attachment(self, field_name, id=None, ids=None, group_id=None, invitatio
...
+        >>> f = client.get_attachments(ids=['Place Note-ID here', 'Place Note-ID here'], field_name='pdf')
+        >>> with open('attachments.zip','wb') as op: op.write(f)
+
+        """
+
+        if not ids:
+            raise OpenReviewException('Provide a non empty list of ids')
+
+        if len(set(ids)) > 50:
+            raise OpenReviewException('The max number of ids is 50')
```
carlosmondra: Should we allow the API to validate this instead? I would say at least the limit of 50. Given that the API can change that value

openreview/tools.py
```diff
@@ -2271,6 +2273,88 @@ def singularize(word):
...
+    os.makedirs(output_dir, exist_ok=True)
+
+    file_paths = []
+    for batch in tqdm(batches, desc='Downloading attachments'):
+        if len(batch) == 1:
+            note = batch[0]
+            file_path = os.path.join(output_dir, get_attachment_file_name(note, field_name))
+            with open(file_path, 'wb') as f:
+                f.write(client.get_attachment(field_name, id=note.id))
+            file_paths.append(file_path)
```
carlosmondra: I would remove this part and the new method `get_attachment_file_name`, unless you have a good reason to keep it.

openreview/api/client.py
melisabok: ok, I will remove the validation

openreview/tools.py
```diff
@@ -2271,6 +2273,88 @@ def singularize(word):
...
+
+    ## The API returns the file itself instead of a zip file when only one id is requested,
+    ## so avoid leaving a single note in the last batch
+    if len(batches) > 1 and len(batches[-1]) == 1:
+        batches[-1].insert(0, batches[-2].pop())
+
+    os.makedirs(output_dir, exist_ok=True)
+
+    file_paths = []
+    for batch in tqdm(batches, desc='Downloading attachments'):
```
melisabok, line 2338: @carlosmondra do you think we should use concurrent requests here? to make the download faster?

carlosmondra (approved):

openreview/tools.py
melisabok: decided to not use it

openreview/api/client.py
melisabok: fixed

openreview/tools.py
melisabok: fixed

enrubio:
Just checking if this is duplicated or not -- the current `get_attachment` function does support `ids` and getting up to 50 attachments. I made the change ~1year ago, it looks similar to the `get_attachments` function here.

melisabok:
Thanks for the catch Emilia! I removed the function in the client and leaving the tools function to download ALL the attachments.

openreview/tools.py
```diff
@@ -3,8 +3,11 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 import inspect
 
+import io
 import json
 import os
+import shutil
+import zipfile
```
enrubio, line 10: Do we need to pin the versions?

openreview/tools.py
```diff
@@ -2277,6 +2280,65 @@ def singularize(word):
...
+    if len(batches) > 1 and len(batches[-1]) == 1:
+        batches[-1].insert(0, batches[-2].pop())
+
+    os.makedirs(output_dir, exist_ok=True)
+
+    file_paths = []
+    for batch in tqdm(batches, desc='Downloading attachments'):
+        archive_content = client.get_attachment(ids=[note.id for note in batch], field_name=field_name)
+        with zipfile.ZipFile(io.BytesIO(archive_content)) as archive:
+            for member in archive.infolist():
```
enrubio, line 2341: If there's only 1 note it fails here with `BadZipFile: File is not a zip file`

celestemartinez (approved):

openreview/tools.py
melisabok: what do you mean? I haven't added any new library to the repo.

openreview/tools.py
melisabok: fixed

openreview/tools.py
enrubio: oh sorry, I didn't realize zipfile was part of python

enrubio (approved):
Changes look good

PR #3147, opened 2026-08-03
Fix/remove stream=True when callong to get_all_notes without details

melisabok:
Do not make the exception to call the get notes with stream = True if there are no details requested. 

Using after param is faster than streaming when the venue is very large.


carlosmondra (approved):

enrubio (approved):
Looks good!

celestemartinez (approved):

PR #3148, opened 2026-08-03
Feature/profile document upload

melisabok:
Added test to verify document upload.


celestemartinez (approved):

openreview/api/client.py
```diff
@@ -1089,6 +1090,155 @@ def moderate_profile(self, profile_id, decision, reason=None):
...
+            json = {
+                'profileId': profile_id,
+                'type': document_type
+            },
+            headers = self.headers)
+
+        response = self.__handle_response(response)
+        return response.json()
+
+    def post_profile_document(self, file_path, token, document_type='identity'):
```
enrubio: Could it be possible we create a parental consent link but then it gets uploaded as an identity doc? I'm not sure what will call `post_profile_document` but I wonder if we should remove the default value for `document_type` and make it required to prevent that.

melisabok: done!

enrubio (approved):
Changes look good!

PR #3149, opened 2026-08-03
create one invitation per decision string

celestemartinez:
This PR creates one `Submission_Release` per decision option, similar to the `Author_Notification` invitations. 

By default, any accept submission release invitations release submissions to the public, along with `title`, `abstract` and `author` names. Readers are added to the `pdf` field. The `pdate` is set to the invitation's `cdate`, but can be changed by PCs using the `/Dates` invitation.

By default, any reject submission release invitations keep submissions hidden (visible to PCs, assigned committee and authors) and adds readers to the `pdf` and `authors` fields. 

The PCs can change the readers of submissions using the `/Readers` invitation, as well as the readers of individual fields using the `/Form_Fields` invitation. 

Everything is defined directly in the invitation. The process function currently only computes the `_bibtex` and `odate`.

I didn't add the following fields directly in the invitation:
- `odate`: this is set only if the readers are set to `everyone`, but the PCs can change the readers, so I think the process still needs to check and set this.


tests/test_reviewers_only.py
```diff
@@ -2530,7 +2530,7 @@ def test_release_submissions(self, openreview_client, helpers):
         assert submissions[0].content['venue']['value'] == 'ABCD 2025 Conference Submission'
         assert '_bibtex' not in submissions[0].content
 
-        inv = pc_client.get_invitation('ABCD.cc/2025/Conference/-/Accepted_Submission_Release')
+        inv = pc_client.get_invitation('ABCD.cc/2025/Conference/-/Accept_Submission_Release')
```
melisabok: should we call it `ABCD.cc/2025/Conference/-/Submission_Accept_Release`

openreview/workflows/templates.py
```diff
@@ -1750,7 +1760,7 @@ def set_paper_release_template_invitation(self):
                 },
                 'domain': '${1/content/venue_id/value}',
                 'invitation': {
-                    'id': '${2/content/venue_id/value}/-/${2/content/decision_option/value}_${2/content/submission_name/value}_Release',
+                    'id': '${2/content/venue_id/value}/-/${2/content/decision_option_id/value}_${2/content/submission_name/value}_Release',
```
melisabok: we need to make sure there are no spaces here or unsupported characters.

celestemartinez: This is being checked by the decision_option field: https://github.com/openreview/openreview-py/blob/97c0081ca42a9b73a40c05730d67dbedc3e3c124/openreview/workflows/edit_invitations.py#L1186 since we also use this in the notification invitation id

openreview/workflows/edit_invitations.py
```diff
@@ -939,6 +939,21 @@ def set_edit_dates_one_level_invitation(self, super_invitation_id, include_due_d
...
+                        'type': 'date',
+                        'range': [ 0, 9999999999999 ]
+                    }
+                }
+            }
+            invitation.edit['invitation']['edit'] = {
+                'note': {
+                    'pdate': '${4/content/publication_date/value}'
+                }
+            }
```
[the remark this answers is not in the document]
celestemartinez: `pdate` is not on the same level as the other dates, that's why the path is different

openreview/workflows/templates.py
```diff
@@ -1861,6 +1889,224 @@ def set_paper_release_template_invitation(self):
 
         self.post_invitation_edit(invitation)
 
+    def set_reject_paper_release_template_invitation(self):
+
+            invitation = Invitation(id=f'{self.template_domain}/-/Reject_Submission_Release',
+                invitees=['active_venues'],
+                readers=['everyone'],
```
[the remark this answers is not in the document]
celestemartinez: fixed.

openreview/workflows/workflow_process/edit_decision_options_process.py
```diff
@@ -47,12 +52,22 @@ def process(client, edit, invitation):
...
+        # notification invitations
+        notification_invitation_id = f'{venue_id}/-/Author_{formatted_decision_option}_Decision_Notification'
+        notification_invitations_to_delete = client.get_invitations(prefix=notification_invitation_id)
+
+        for inv in notification_invitations_to_delete:
+            print(f'Deleting Notification invitation: {inv.id}')
+            client.delete_invitation(inv.id)
+
+        # submission release invitaations
+        release_invitation_id = f'{venue_id}/-/{formatted_decision_option}_Submission_Release'
```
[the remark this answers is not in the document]
celestemartinez: fixed.

openreview/workflows/workflow_process/edit_decision_options_process.py
```diff
@@ -62,16 +77,61 @@ def process(client, edit, invitation):
     from_email = domain.content['message_sender']['value']['fromEmail']
```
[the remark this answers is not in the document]
celestemartinez: fixed.

tests/test_reviewers_only.py
```diff
@@ -1844,45 +1844,49 @@ def test_review_stage(self, openreview_client, helpers):
...
-        invitation = openreview_client.get_invitation('ABCD.cc/2025/Conference/-/LLM_PDF_Response')
-        assert invitation.content['prompt']['value'] == 'This is the prompt submitted by PCs'
-        assert invitation.content['model']['value'] == 'gemini/gemini-2.0-flash'
-        assert invitation.content['api_key']['value'] == '1234567abcdefg'
-        assert invitation.content['api_key']['readers'] == ['ABCD.cc/2025/Conference']
+    # def test_LLM_PDF_response_stage(self, openreview_client, helpers):
+
+    #     pc_client = openreview.api.OpenReviewClient(username='programchair@abcd.cc', password=helpers.strong_password)
+
+    #     review_invitation = openreview_client.get_invitation('ABCD.cc/2025/Conference/-/Official_Review')
```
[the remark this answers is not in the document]
celestemartinez: removed

celestemartinez:
Changed the invitation ID to be f'{decision_option}_{submission_name}_Change_Before_Decision'

image

tests/test_reviewers_only.py
celestemartinez: This has been updated!

melisabok:
The screenshot looks good to me.

melisabok (approved):

openreview/workflows/templates.py
```diff
@@ -1799,7 +1820,229 @@ def set_paper_release_template_invitation(self):
...
+                                        'const': {
+                                            'delete': True
+                                        }
+                                    }
+                                },
+                                'pdf': {
+                                    'readers': [
+                                        '${7/content/venue_id/value}',
+                                        '${7/content/venue_id/value}/${7/content/submission_name/value}${{4/id}/number}/${7/content/authors_name/value}'
+                                    ]
```
enrubio, line 1856: Here we always set readers to the venue + authors, but what if it was previously visible to the assigned committee? Is it intentional that we hide it from them as well?

openreview/workflows/process/submission_release.py
enrubio: Should we pass the accept options to `decision_to_venue`? We do that in the other places we call it.

openreview/workflows/workflow_process/edit_decision_options_process.py
```diff
@@ -47,31 +52,85 @@ def process(client, edit, invitation):
...
+                'decision': { 'value': decision_option }
+            }
+        )
+
+        is_accepted_option = openreview.tools.is_accept_decision(decision_option, accept_decision_options)
+        additional_readers = []
+        if area_chairs_name:
+            if senior_area_chairs_name:
+                additional_readers.append(f'{venue_id}/{submission_name}' + '${{2/id}/number}/' + senior_area_chairs_name)
+            additional_readers.append(f'{venue_id}/{submission_name}' + '${{2/id}/number}/' + area_chairs_name)
```
enrubio, line 111: Minor but maybe `senior_area_chairs_name` should not be nested if a venue has SACs but no ACs. Don't think it happens often, but _could_ happen.

celestemartinez: We do not allow the request to be created with SACs but no ACs, not sure this is something we will ever have.

openreview/workflows/templates.py
celestemartinez: I think this is fine, invitations are always created with a default and PCs can always change this. The readers can be changed at any time by the PCs and I don't think we should be trying to sync these invitations every time that happens. Also, since this will run automatically, I think being restrictive is best and PCs can always re-run it

openreview/workflows/process/submission_release.py
celestemartinez: Fixed.

openreview/workflows/workflow_process/edit_decision_options_process.py
enrubio: okay, I thought we did this for some venue but I couldn't remember exactly, this is fine then

openreview/workflows/workflow_process/edit_decision_options_process.py
enrubio, line 51: I think here we need to consider a change to the accept options as well? For example, if a venue had this:
`decision_options = Accept (Oral), Accept (Poster), Reject`
`accept_options = Accept (Oral)`
Then the Accept (Poster) inv will use the rejection template. 

But say on the next edit they fix the accept options, I think because there's no change in the `decision_options` between edits, the Accept Poster inv wouldn't get recreated using the accept template?

openreview/workflows/templates.py
enrubio: Those are good points, thank you

openreview/workflows/workflow_process/edit_decision_options_process.py
celestemartinez: I wouldn't really expect this to happen but I can make a fix. I think it will just be easier to delete all invitations and re-post all of them.

celestemartinez: Claude recommended to repost the invitation only the the decisions has been added/removed from `accept_options`, so I am not deleting/reposting all invitations. I don't expect this to happen very often.

openreview/workflows/workflow_process/edit_decision_options_process.py
```diff
@@ -43,35 +50,98 @@ def process(client, edit, invitation):
...
             client.delete_invitation(inv.id)
 
     # post new decision notification invitations for the added decision options
     request_form_inv = domain.get_content_value('request_form_invitation')
     invitation_prefix =request_form_inv.split('Support')[0] + 'Template'
-    short_name = domain.get_content_value('subtitle')
     from_email = domain.content['message_sender']['value']['fromEmail']
 
     for decision_option in added_decision_options:
-        formatted_decision_option = decision_option.replace(' ', '_').replace('(', '').replace(')', '')
```
enrubio, line 65: Do we also need to add: `| rebucketed_decision_options:` here so the related notification invitation gets updated?

openreview/workflows/workflow_process/edit_decision_options_process.py
enrubio: I feel I do see venues update their decisions and just want to make sure they use the correct template.

Small question - If a decision is removed from `accept_options` (so now it's considered Reject), I think the invitation for that decision will still contain a `pdate` since we're reposting them and not recreating. Is that fine?

openreview/workflows/workflow_process/edit_decision_options_process.py
celestemartinez: Fixed. Only the default template changes here.

openreview/workflows/workflow_process/edit_decision_options_process.py
celestemartinez: Oh I agree, I just didn't think decisions are re-bucketed very often. I added replacement true to make sure the invitation gets completely replaced.

enrubio (approved):
It looks great, thanks Celeste!

PR #3152, opened 2026-08-06
Bulk edit throughput and resilience: bigger pools, scoped POST retries, self-healing fan-outs

melisabok:
## Motivation

Creating per-submission child invitations for a large venue (AAAI 2027, ~48k submissions) took ~30 minutes and could be killed entirely by one transient server error. Together with the API-side cache and error-status fixes (already merged), this PR takes the same run to **~7 minutes with zero manual intervention**, surviving worker restarts mid-run (measured: 40,823 edits in 7:19, one envoy connection reset absorbed by retry, 0 failures).

### 1. Connection pool sized for concurrency (both clients)

`HTTPAdapter(pool_maxsize=128)` in `openreview.Client` and `openreview.api.OpenReviewClient`. The requests default of 10 pooled connections meant that at high `max_workers` every request beyond 10 discarded its connection and re-paid the TCP+TLS handshake (`Connection pool is full, discarding connection` warnings). Pool now matches the highest `max_workers` used by `tools.concurrent_requests`.

### 2. POST retries, scoped to the edits endpoints only (V2 client)

Two retry strategies sharing one policy (8 attempts, backoff, `status_forcelist` 429/5xx), differing **only** in `allowed_methods`:

- The three edits endpoints (`/notes/edits`, `/invitations/edits`, `/groups/edits`) get an adapter with POST added — mounted per-URL, using requests' longest-prefix routing. Edit posts are safe to retry: re-posting the same edit converges to the same entity.
- Every other POST (`/messages`, etc.) keeps urllib3's default methods, which exclude POST — non-idempotent requests are never auto-retried. The V1 client is unchanged (no edits endpoints).

This relies on the API's error-status contract (merged in openreview/openreview-api `fix/preprocess-error-code`): deliberate rejections — validation errors, process function `OpenReviewException`s — return **400** and are never retried; 5xx means a genuine server fault (including python processes killed mid-request, which the API intentionally returns as 500 *so that* clients retry them).

### 3. `tools.concurrent_requests`: failure containment and self-healing

Two new opt-in parameters, defaults preserve existing behavior for all 21 library call sites:

- `return_exceptions=True` — a failing execution contributes its exception object to the results instead of aborting the remaining ones (mirrors `asyncio.gather`).
- `retries=N` (+ `retry_max_workers`, default `max_workers // 4`) — after the main pass, re-runs only the failed values, up to N passes at reduced concurrency (a struggling server gets room to recover). Results stay position-aligned with `params`. Failures that survive all passes raise (or are returned under `return_exceptions`).

### 4. Venue process functions: full concurrency + retry

- `invitation_edit_process.py`: both fan-outs (child posts and deletes) run at `max_workers=128, retries=1`. Ordering is load-bearing and commented: if any post still fails after retry, the process raises **before** the delete phase — a child whose post failed would otherwise be missing from the posted set and be wrongly deleted.
- `group_edit_process.py`, `post_submission_process.py`: `max_workers=128` (previously capped at 16, or 10 on 2-core workers).

**Why 128:** the workers are I/O-bound threads (blocked on sockets, GIL released), so the value is "requests in flight", not CPU parallelism — it is machine-independent on the client side (128 threads ≈ 128 MB of mostly-virtual stack and 128 file descriptors, well under default limits on any worker VM; measured client CPU negligible even on 2 cores). The number is calibrated to the **server**: by Little's law, saturating it needs throughput × latency ≈ 93/s × 1.37s ≈ 127 in-flight (production measurement, 40k-edit run). Fewer workers under-saturate (64 → ~47/s ceiling); more cannot help while the server is the limiter (measured flat throughput from 16 to 126 workers) and only lengthen the server queue. The old `min(16, cpu_count()*5)` default was the actual machine-dependence bug — it silently dropped to 10 on 2-core cloud workers. The library default stays 16; only these three venue-scale bulk operations opt into 128, and their retry passes run at 32 as built-in backoff.

## Testing

- New `tests/test_client.py::test_preprocess_error_status`: a preprocess rejection surfaces as an immediate `OpenReviewException` with status **400** — if the API regressed to 500, the edits retry would turn it into a `RetryError` and this test fails loudly. Pins the client half of the error-status contract.
- `tests/test_journal.py` full run in Docker against the fixed API: 14/14 in 21:47 with **zero** retry lines and zero `RetryError` — the original CI failure (Beyond PDF preprocess rejection retried 8 times into an opaque error) verified fixed end to end.
- `tests/test_client.py`: 28 passed. Retry semantics (transient recovery, permanent raise after N passes, order preservation, unchanged defaults) covered by unit-level smoke checks during development.

## Production measurements (AAAI 2027 bulk invitation creation, 128 workers)

| Stage | Throughput | Note |
|---|---|---|
| Before | ~27/s (~30 min) | server serialized by cache SCAN (fixed API-side) |
| After API cache fix | ~45/s | |
| Final | **~93/s, 40,823 edits in 7:19** | 1 envoy reset auto-retried, 0 failures, flat throughput across the run |

## Dependencies

The API-side prerequisites are **already merged** in openreview-api: the cache invalidation fixes and the error-status contract (`fix/preprocess-error-code`). CI runs against api `main`, which contains both, so this branch requires no branch pinning or merge ordering.


carlosmondra (approved):

celestemartinez (approved):

openreview/api/client.py
```diff
@@ -133,10 +133,26 @@ def __init__(self, baseurl = None, username = None, password = None, token= None
...
         self.session = requests.Session()
-        adapter = HTTPAdapter(max_retries=retry_strategy)
+        adapter = HTTPAdapter(max_retries=retry_strategy, pool_maxsize=128)
+        edits_adapter = HTTPAdapter(max_retries=edits_retry_strategy, pool_maxsize=128)
         self.session.mount('https://', adapter)
         self.session.mount('http://', adapter)
+        # requests picks the longest matching prefix, so only the edits endpoints
+        # get the POST-retrying adapter
+        for edits_url in [self.note_edits_url, self.invitation_edits_url, self.group_edits_url]:
+            self.session.mount(edits_url, edits_adapter)
```
enrubio, line 156: It seems like if we are retrying note edits where there is no `note.id` then we can post duplicate notes. It was flagged by claude. I suppose it could happen if someone is running a script posting a bunch of new notes. Looks like this isn't the case for the invitation edits or group edits. The suggestion was to remove `note_edits_url` from this list, but not sure if we want to do that.

openreview/api/client.py
```diff
@@ -133,10 +133,26 @@ def __init__(self, baseurl = None, username = None, password = None, token= None
...
+        # Same retry policy, differing only in allowed_methods: POST is added so edit
+        # posts are retried on 429/5xx. Edit posts are safe to retry (re-posting the
+        # same edit converges to the same entity) and the API reserves 5xx for server
+        # faults — deliberate rejections such as process function validation return
+        # 400, which is never retried. Other POSTs (e.g. messages) are not idempotent
+        # and keep urllib3's default allowed_methods, which exclude POST.
+        edits_retry_strategy = LogRetry(
+            allowed_methods=frozenset(['HEAD', 'GET', 'PUT', 'DELETE', 'OPTIONS', 'TRACE', 'POST']),
+            **retry_params
+        )
```
enrubio, line 147: Looks like we need to set `raise_on_status=False` here too, otherwise for OpenReviewExceptions get returned as RetryErrors and the actual error message is lost.

openreview/api/client.py
melisabok: That could also happens if the user get the error, having duplicated notes is an issue that the API should fix.

Checking the API code, the notes/edits endpoint uses transactions, as soon as you haven an error the transaction is rolled back, can you confirm @carlosmondra ?

openreview/api/client.py
melisabok: Fixed!

openreview/api/client.py
enrubio: Confirmed this in daily meeting, thanks!

PR #3155, opened 2026-08-06
Fix/arr submission revision

celestemartinez:
**Problem**: The `Blind_Submission_License_Agreement` stage builds its form by removing all
submission fields except the license question. 

When the ARR Configuration is posted again, the stage reuses the saved list. If a Revision added or deleted a field from the submission form in between, the remove_options list in the invitation does not match the remove_options list in the Submission Revision note, causing a `NotMatchError`.

**Fix:** rebuild the list from the venue's current submission form instead of
reusing the saved one. Only applies to the license agreement stage.


enrubio (approved):
Looks great!

melisabok (approved):

PR #3160, opened 2026-08-11
 Add Journal Request to the support page with a support-controlled Deployment step

melisabok:
## Summary

Journals are currently deployed by hand: the support team posts a note to
`openreview.net/Support/-/Journal_Request` with the `venue_id` and `secret_key` typed
into the form, and the journal is created immediately by the note's process. There is
no public entry point on the support page and no separation between requesting a
journal and deploying it.

This PR turns the journal request into a proper support-page workflow with three
invitations, each with a single job:

| Invitation | Role |
|---|---|
| `Support/-/Journal_Request` | **Create-only** public request form, listed on the support page. Never edits notes and never touches journal data; its process creates the per-request Revision invitation and emails the listed editors a confirmation. |
| `Support/Journal_Request{number}/-/Revision` | The **only** way to edit a request, for its entire life. Created at submission for `[Support] + editors`; re-posted at deployment for `[venue, Support]`. Pre-deployment edits are a no-op process-wise; post-deployment edits re-run the journal setup so settings changes take effect. |
| `Support/-/Journal_Request_Deployment` | **New, support-only.** Assigns the `venue_id` (the single source of that value) and atomically sets the request note's readers/writers to the journal groups via `${...}` substitution. Its process creates the `Journal` object for the first time plus the per-request tooling (group, comment, recruitment, and Revision invitations). A preprocess rejects deploying a `venue_id` already bound to a different request. |

## Flow

1. **Submit**: any user posts the request form from the support page (new *Journal
   Request* button and tab in `supportRequestsWeb.js`). The form has no `venue_id` or
   `secret_key` fields; the `settings` field is prefilled with a default JSON of all
   static journal settings and its description links to the journal-settings
   documentation. The listed editors get read access to the pending request and a
   confirmation email, and can edit it through the Revision invitation.
2. **Deploy**: the support team posts the Deployment edit with the `venue_id`. The
   journal is fully set up (groups, invitations, recruitment tooling) and the request
   forum is released to `[Support, venue, venue/Action_Editors]`.
3. **Maintain**: EICs edit the request through the Revision invitation (no
   `secret_key`; the venue groups keep forum access — the invitation never touches
   note readers/writers), and each edit re-runs `get_journal(setup=True)`.

## Secret key is now system-managed

`secret_key` is no longer a user input anywhere. `Journal.setup()` generates it
(`tools.create_hash_seed()`) when the journal has none, and the group builder stores
it in the **venue group content** with the rest of the journal settings, with field
readers `[venue_id]` (same pattern as the iThenticate API key) — visible only to the
venue and never editable through any invitation.

`JournalRequest.get_journal` reads the key from the venue group, falling back to the
request note for journals deployed before this change (the next setup run migrates it
into the group). The `Reviewer_Recruitment`/`Action_Editor_Recruitment`/
`Reviewer_Recruitment_by_AE` processes no longer read the key off the request note —
they build the journal via `get_journal`, so recruitment links keep signing with the
same key regardless of where it is stored.

## Other library changes

- `Journal.update_affinity_scores` and the `check_new_profiles` reminders loop skip
  requests that have not been deployed (`get_journal` returns `None` for a note
  without `venue_id` — previously a pending request would crash these maintenance
  jobs with a `KeyError`).
- `JournalRequest.get_request_form_content()` is the shared field-definition source
  for the form and Revision invitations.
- The Revision invitation signs as `~Super_User1` (like the recruitment invitations)
  so its process can re-post the journal meta invitation.

## Tests

- New `tests/test_journal_request_workflow.py`: user-submitted request (no
  deployment, editor access, confirmation email sent exactly once, Revision available
  from submission), deployment (journal + tooling created, venue-scoped Revision,
  generated 16-char secret in the venue group only, form cannot edit notes),
  post-deployment Revision settings edit (setup re-runs, key preserved), and
  venue-id-reuse rejection.
- All 9 legacy journal test files migrated to the two-step flow (request post +
  Deployment edit, `secret_key` removed from request content, post-deployment edits
  via the per-request Revision invitation).
- `test_journal_reviewer_recruitment_by_ae` hardened: `get_messages` results are
  sorted by `cdate` before index-based assertions (the API does not guarantee return
  order).

Full journal suite (10 files, 54 tests) passes in a single clean-start run.

## Deployment notes

- Production scripts that post journal requests with `venue_id`/`secret_key` in the
  content must switch to the two-step flow (the form rejects those fields).
- The support group webfield needs a one-time update to show the new button/tab
  (`VenueRequest.__init__` only installs it when the group has no web).
- Existing journals get their per-request Revision invitation and group-stored secret
  key automatically the next time their request is edited/redeployed; a one-off
  migration script can front-load this if preferred.


openreview/journal/journal_request/process/request_form_process.py
```diff
@@ -0,0 +1,25 @@
...
+
+    note = client.get_note(edit.note.id)
+
+    ## create the Revision invitation so the editors can edit their pending request
+    journal_request = openreview.journal.JournalRequest(client, SUPPORT_GROUP)
+    journal_request.setup_revision_invitation(note.id)
+
+    baseurl = openreview.tools.get_site_url(client)
+
+    client.post_message(
```
celestemartinez, line 13: Could we send an email to support when a new request is submitted? These requests won't show up in moderation and I don't expect we will get many of them, but don't want to miss them when we do (or add them to the moderation page)

melisabok: We will add them to the moderation page.

celestemartinez (approved):
Looks good! Creating the revision invitation for existing journals seems to work.

PR #3168, opened 2026-08-18
Fix/ Journal: keep AE identity hidden from authors until paper is approved for review

celestemartinez:
TMLR asked to keep the AE's name hidden until the paper has been approved for review. This way, the AE can desk-reject a paper anonymously. 

If AE names should be public, they get released once the paper is approved for review and stays public even if the AE is changed.

### DIFF
--- a/openreview/journal/process/ae_assignment_process.py
+++ b/openreview/journal/process/ae_assignment_process.py
@@ -74,14 +74,12 @@ def process_update(client, edge, invitation, existing_edge):
             'assigned_action_editor': { 'value': edge.tail }
         }
 
-        if journal.is_action_editor_anonymous():
-            content['assigned_action_editor']['readers'] = [journal.venue_id, journal.get_action_editors_id(number=note.number), journal.get_reviewers_id(number=note.number)]
-
+        # when AC is assigned, keep hidden until paper is approved for review
+        content['assigned_action_editor']['readers'] = [journal.venue_id, journal.get_action_editors_id(number=note.number), journal.get_reviewers_id(number=note.number)]
 
         if journal.assigning_AE_venue_id == note.content['venueid']['value']:
             content['venueid'] = { 'value': journal.assigned_AE_venue_id }
             content['venue'] = { 'value': f'{journal.short_name} Assigned AE' }
-
 
         client.post_note_edit(invitation= journal.get_meta_invitation_id(),
                             signatures=[journal.venue_id],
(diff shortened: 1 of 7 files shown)


openreview/journal/process/ae_assignment_process.py
```diff
@@ -74,15 +74,13 @@ def process_update(client, edge, invitation, existing_edge):
             'assigned_action_editor': { 'value': edge.tail }
         }
 
-        if journal.is_action_editor_anonymous():
-            content['assigned_action_editor']['readers'] = [journal.venue_id, journal.get_action_editors_id(number=note.number), journal.get_reviewers_id(number=note.number)]
-
+        # when AC is assigned, keep hidden until paper is approved for review
+        content['assigned_action_editor']['readers'] = [journal.venue_id, journal.get_action_editors_id(number=note.number), journal.get_reviewers_id(number=note.number)]
```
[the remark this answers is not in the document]
celestemartinez: fixed.

openreview/journal/process/review_approval_process.py
```diff
@@ -21,7 +21,18 @@ def process(client, edit, invitation):
...
         journal.notify_readers(edit, content_fields=['under_review', 'comment'])
-        
+
+        # update AE group readers to be everyone if AE is not anonymous
+        if not journal.is_action_editor_anonymous():
+            client.post_group_edit(invitation=journal.get_meta_invitation_id(),
+                signatures=[venue_id],
+                group=openreview.api.Group(
+                    id=paper_action_editor_group.id, 
+                    readers=['everyone']
```
melisabok: can you use {delete: true} instead of everyone? submissions may not be public but the AE is shown to the authors,

melisabok: sorry, this is a group, we need to see if the journal as public submissions or not.

celestemartinez: ok, I was going by this: https://github.com/openreview/openreview-py/blob/fa989b0749d87a9567d9e6e1b2e5ed31415f8626/openreview/journal/group.py#L494

so set to everyone only if the AE is not anonymous and the venue has public submissions? I can make this change

melisabok: I'm thinking to use delete: true instead.

celestemartinez: I don't understand why we want to use {"delete": true} here, if we are setting the readers of the group.

melisabok (approved):

PR #3170, opened 2026-08-19
create profiles with DOB field

melisabok:
## Summary

The API now collects a date of birth at registration and uses it to gate profile creation and
activation. This branch teaches the Python clients to send one, updates the test helpers so every
test user has one, and adds coverage for the age rules the API enforces.

## Motivation

`POST /register` now rejects a request without a `dob`, so every test that creates a user failed at
the first call. Fixing it needed more than adding a field: the date of birth is enforced in three
separate places, and each one broke a different set of tests.

- **Registration** rejects a missing `dob`, and rejects anyone under 13 before creating anything.
- **Activation** replaces the stored profile content wholesale, so an activation payload without a
  `dob` erases the one registration just saved. Because `dob` is write-once, it has to be the *same*
  value both times.
- **Profile edits** are validated against a schema that requires `dob`, so a profile that predates
  the field cannot be edited until one is supplied.

## Key changes

- `register_user` takes an optional `dob` on both clients and sends it only when provided. It is
  left optional rather than defaulted, because a client library should not invent a birth date, and
  because the V1 API rejects the field outright.
- `Helpers.dob_for_age(years)` and `Helpers.default_dob()` in `conftest.py` produce epoch
  milliseconds in UTC. `default_dob()` is memoized: registration and activation must send the
  identical value, and a long run can cross UTC midnight between them.
- `Helpers.create_user` takes `dob=None`, defaulting to age 30, and threads it through both
  registration and the activation content. The default is deliberately an adult — anyone under 18 is
  rejected, which would have broken tests in a far more confusing way than a 400.
- Direct `register_user` callers in `test_arr_venue_v2.py`, `test_client.py`, `test_edges.py` and
  `test_profile_management.py` pass a `dob` where they talk to the V2 API, and deliberately do not
  where they talk to V1.

## New coverage: `tests/test_profile_dob.py`

Eleven tests pinning the age rules end to end.

| Area | Behaviour |
| --- | --- |
| Registration | under 13 refused with `MinimumAgeError` and no profile created; exactly 13 accepted |
| Minors | 13–17 rejected at signup and on an existing profile, with the parental consent mail and its upload link |
| Under 13 | rejected on an existing profile with a different mail and no consent link |
| Write-once | the owner cannot change or retract a `dob`; super user and support can do both |
| Visibility | `dob` to owner and privileged; `isMinor` to privileged; `isOver18` to the super user |
| Backward policy | a profile with no `dob` cannot be edited until one is supplied |

Both rejection mails are asserted in full. The two generated credentials in the consent mail — the
upload JWT and the resubmission token — are matched by shape, everything else exactly.

The minor test creates an **adult on the same institutional domain** as a control. Without it the
test passes for a non-institutional domain, where moderation is the ordinary outcome and being a
minor changes nothing.

`isOver18` also reaches an active venue account, which is a group id rather than a person. That tier
needs a venue, so it is covered in `test_journal.py` by impersonating `TMLR`, and the age-flag
visibility rules are also asserted in `test_profile_management.py`.

## Migration off the V1 API

`test_client.py` and `test_tools.py` were moved to the API 2 client, which surfaced five differences
that needed real fixes rather than a client swap:

- V2 has no `get_groups(ids=[...])`, only a single `id`.
- V2 has no `post_group`; groups are created with `post_group_edit`, which returns an edit.
- `first` / `last` are not V2 search filters — they are sent and ignored, returning every profile.
  V2 searches names by `fullname`.
- `emailsConfirmed` is populated at activation, not registration, so `confirmedEmails` finds nothing
  for a registered-but-inactive profile.
- V2 validates that a group member exists before adding it.

Tests whose subject *is* the V1 client are deliberately left on it: `test_create_client`,
`test_login_user`, `test_login_expiration`, `test_retry_strategy`, the `_v1` MFA tests, and the
`get_expertise_*` tests that assert both clients behave identically.

## Companion change

Requires the `openreview-api` branch `worktree-fix+add-dob-to-profile-with-minor-visibility`. The
tests will fail against a `master` api-v2, which still queues minors for moderation instead of
rejecting them.

## Before merging

- [ ] Revert the `openreview-api-v2-branch` default in `.circleci/config.yml` back to `main` once the
      API change is merged. It is currently pinned to the companion branch so CI exercises the new
      behaviour.

## Test plan

Passing:

- [x] `pytest tests/test_profile_dob.py` — 11 passed
- [x] `pytest tests/test_reviewers_only.py` — 23 passed
- [x] `pytest tests/test_client.py tests/test_edges.py tests/test_tools.py` — 60 passed, 2 skipped

Not yet verified:

- [ ] `pytest tests/test_journal.py` — running
- [ ] `pytest tests/test_profile_management.py`
- [ ] `pytest tests/test_arr_venue_v2.py`

Note that `test_client.py`, `test_edges.py` and `test_tools.py` were last run before the most recent
API changes (minor signup rejection, support profile edits) and should be re-run before merging.

## Notes for review

Two API behaviours worth a second opinion, both currently pinned by tests as they are:

1. **api-v1 has no `dob` in its profile content allowlist**, so any V1 consumer doing a
   read-modify-write on a profile is rejected once registration stores one. `test_tools.py` sidesteps
   this by moving to V2, but real V1 consumers would still hit it.
2. **`tools.replace_members_with_ids` loses member order on the V2 path** — it rebuilds members as
   `list(set(...))` where the V1 path preserved the input order, so V2 callers get a nondeterministic
   ordering. The affected assertions were made order-insensitive.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ parameters:
     default: "master"
   openreview-api-v2-branch:
     type: string
-    default: "worktree-fix+add-dob-to-profile-with-minor-visibility"
+    default: "main"
   python-version:
     type: string
     default: "3.13.1"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ parameters:
     default: "master"
   openreview-api-v2-branch:
     type: string
-    default: "main"
+    default: "worktree-fix+add-dob-to-profile-with-minor-visibilit"
   python-version:
     type: string
     default: "3.13.1"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ parameters:
     default: "master"
   openreview-api-v2-branch:
     type: string
-    default: "worktree-fix+add-dob-to-profile-with-minor-visibilit"
+    default: "worktree-fix+add-dob-to-profile-with-minor-visibility"
   python-version:
     type: string
     default: "3.13.1"
--- a/openreview/api/client.py
+++ b/openreview/api/client.py
@@ -337,7 +337,7 @@ class OpenReviewClient(object):
         self.__handle_authorization(json_response)
         return json_response
 
-    def register_user(self, email = None, fullname = None, password = None):
+    def register_user(self, email = None, fullname = None, password = None, dob = None):
         """
         Registers a new user
 
@@ -347,6 +347,8 @@ class OpenReviewClient(object):
         :type fullname: str, optional
         :param password: Password used to log into OpenReview
         :type password: str, optional
+        :param dob: Date of birth as epoch milliseconds. Mandatory when the API is configured with profile.requireDob
+        :type dob: int, optional
 
         :return: Dictionary containing the new user information including his id, username, email(s), readers, writers, etc.
         :rtype: dict
@@ -356,6 +358,8 @@ class OpenReviewClient(object):
             'fullname': fullname,
             'password': password
         }
+        if dob is not None:
+            register_payload['dob'] = dob
         response = self.session.post(self.register_url, json = register_payload, headers = self.headers)
         response = self.__handle_response(response)
         return response.json()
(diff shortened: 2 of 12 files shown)


celestemartinez (approved):

enrubio (approved):
Changes look good

PR #3171, opened 2026-08-19
Require human verification on Bid and Expertise Selection invitations

melisabok:
It requires https://github.com/openreview/openreview-web/pull/2895

## Summary

Sets `humanVerificationRequired` on the edge invitations reviewers post to directly — Bid and
Expertise Selection — now that the API enforces the check on edges and tags, not only on note edits.
Edge invitations get their own, higher limit (100/hour) rather than reusing the submission limit.

## Motivation

`humanVerificationRequired` already protected the invitations a user posts *notes* through:
venue submission, journal submission, the arXiv record invitation and `Archive/-/Direct_Upload`. The
API has since grown the same enforcement for edges and tags (`models/edge.js`, `models/tag.js`), which
leaves bidding and expertise selection — the two edge invitations a committee member posts to in bulk,
by hand, one click per paper — with no rate limit at all.

## Changes

A second constant in `openreview/tools.py`, so the two kinds of invitation can be tuned independently:

```python
# Default rate limit applied to submission (and publication import) invitations
# to prevent a user from posting too many edits/notes in a short window
DEFAULT_HUMAN_VERIFICATION = { 'limit': 15, 'windowMs': 3600000 }

# Edge and tag invitations get a higher limit: bidding and expertise selection are
# one click per paper, so a single legitimate session posts far more than a submission does
DEFAULT_EDGE_TAG_HUMAN_VERIFICATION = { 'limit': 100, 'windowMs': 3600000 }
```

Applied at the three places that build these edge invitations:

| Location | Invitation |
|---|---|
| `openreview/venue/invitation.py` — `set_bid_invitations` | `<venue>/<committee>/-/Bid`, one per bid stage (reviewers, ACs, SACs) |
| `openreview/venue/invitation.py` — `set_expertise_selection_invitations` | `<venue>/<committee>/-/Expertise_Selection` |
| `openreview/journal/invitation.py` — `set_expertise_selection_invitations` | `<journal>/Reviewers/-/Expertise_Selection`, `<journal>/Action_Editors/-/Expertise_Selection` |

ARR inherits both through `venue_invitation_builder`. The four note-invitation call sites are
untouched and keep the 15/hour constant.

## Behaviour

The counter is keyed per user **and** per invitation, and a valid Turnstile token *resets* it rather
than buying a single request — so the limit is "a challenge every N posts", not a hard ceiling. At
100/hour a reviewer bids on 100 papers, solves one challenge, and continues.

Note that these limits are only meaningful for invitations whose invitees are real groups: the API
rejects `humanVerificationRequired` when `invitees` contains `everyone`, the anonymous group, or the
guest group. Bid and Expertise Selection are invited to committee group ids, so they are fine.

## Test plan

- [x] `pytest tests/test_aaai_conference.py` — asserts the config on the two Bid invitations and the
      three Expertise_Selection invitations. (22 passed against the 15/hour value; re-running for the
      100/hour value.)
- [ ] `pytest tests/test_matching_v2.py` — extends `test_setup_matching`: after the six bid edges it
      already posts, drops the reviewer Bid limit to `{'limit': 1, 'windowMs': 300000}` through the
      venue meta invitation, asserts the next bid raises `Human verification required`, then restores
      the default so later tests are not rate limited.
- [ ] `pytest tests/test_journal.py` — asserts the config on the TMLR Reviewers and Action_Editors
      Expertise_Selection invitations, next to the existing submission assertion.

The challenge can only be *reached* from python, not solved — the client has no way to send a
`cf-turnstile-token` header. The solve-and-retry path is covered on the web side.

## Dependencies

Requires the openreview-api branch `worktree-fix-human-verification-edges-tags` (commit "Enforce human
verification on edges and tags"). Without it the field is stored but never enforced on edges.

Pairs with the openreview-web change that renders the Turnstile challenge in the bidding and expertise
consoles — without it, a reviewer who hits the limit sees a raw `HumanVerificationRequiredError` and
has no way to continue.

## Known issue, not fixed here

Each click in the bidding console posts the edge **twice**: `BidWidget.js` puts `onClick` on a
`<label>` wrapping the radio `<input>`, and label activation dispatches a second click that bubbles
back to the same handler. Every bid therefore costs two requests, so 100/hour is really ~50 bids
before the first challenge. Worth fixing separately; it also means bid edges are being written twice.


enrubio:
These changes look good. This wouldn't be an issue for ARR right? Since they have a process function that reposts expertise selection edges, but it uses admin level permissions so it shouldn't get limited?

melisabok:
@enrubio it should be fine, otherwise test must have failed.

celestemartinez (approved):

enrubio (approved):

PR #3181, opened 2026-08-31
Support role overlap restrictions in workflow recruitment

melisabok:
## Motivation

The old venue request form supports an **`allow_role_overlap`** setting that lets venue organizers prevent the same person from serving as both a Reviewer and an Area Chair. The legacy implementation enforces it in two places: at invite time (`invite_committee` skips invitees already invited to or members of another official role) and at response time (the recruitment process silently converts an "Accept" into a decline and emails the user when they already belong to the overlapping committee).

The new workflow configuration (`openreview/workflows/`) had no equivalent — any invited user could accept a recruitment invitation regardless of the roles they already hold.

This PR adds overlap support to the workflow recruitment. Instead of a `true`/`false` flag, the restriction is expressed as a **list of group ids the responding user cannot be a member of**, configured on the `Recruitment_Response` invitation.

## Design decisions

- **Validation happens at response time only.** Program chairs can invite the same person to multiple roles; whichever invitation the user accepts first wins, and to switch roles they decline one and accept the other. Response-time validation is the only authoritative check anyway, since memberships change between invite and response. An invite-time courtesy filter can be layered on later without any schema change (the request process can read the list off the response invitation).
- **Validation happens in the preprocess, not the process.** The legacy flow accepted the note, then silently moved the user to the declined group and emailed them that their accept didn't count. Here the accept is rejected up front with a clear, actionable error shown in the UI, so no invalid state is ever written and no follow-up email is needed. This matches the existing preprocess pattern (rejecting a decline after a paper assignment).
- **Empty/absent list means overlap is allowed** — the default behavior is unchanged for venues that don't configure it.

### `openreview/workflows/edit_invitations.py`

`set_edit_committee_recruitment_invitation` now also creates an **`Overlap_Committees`** edit invitation (`<committee_id>/-/Recruitment_Response/Overlap_Committees`), following the existing `Reduced_Load` pattern. PCs post:

- **`overlap_committee_ids`** — `string[]` of group ids (default `[]`), stamped into the `Recruitment_Response` invitation content.

Since the builder runs from `committee_group_template_process.py` at deployment, every committee of a newly deployed venue gets this invitation automatically.

### `openreview/workflows/process/committee_recruitment_response_pre_process.js`

On a **Yes** response, the preprocess reads `overlap_committee_ids` from the invitation content and checks membership with `getGroups({ id: <group_id>, member: user })` for each listed group. If the user belongs to any of them, the edit is rejected with:

> You have already accepted an invitation to serve as Action Editor for EFGH 2025. If you would like to change your decision and serve as Reviewer, please decline the invitation to be Action Editor and then accept the invitation to be Reviewer.

Role names come from each group's `committee_pretty_name` content (falling back to `Tools.prettyId`); the wording follows the legacy venue request form message. The existing decline-after-assignment check is unchanged.

## Tests

New `test_recruit_reviewers` in `tests/test_acs_and_reviewers.py`:

1. PC configures `overlap_committee_ids: [<venue>/Action_Editors]` on the Reviewers `Recruitment_Response` via the new `Overlap_Committees` invitation and the value is stamped on the invitation.
2. An accepted Action Editor is invited as a Reviewer; their **Accept is rejected** with the error above, and no response note edit or group change is produced.
3. A regular invitee accepts normally.
4. The Action Editor **declines the AC invitation and then successfully accepts** the Reviewer invitation (the "let the user decide" flow).
5. State is restored (AC re-accepts their original role) so the downstream matching/assignment tests in the file are unaffected.

## Not in this PR

- **Invite-time filtering** (legacy option: refuse to even invite users holding another role). Deferred; can be added to `committee_recruitment_request_process.py` by reading the same list.
- **Backfill for already-deployed venues** — existing venues get the `Overlap_Committees` invitation only when their recruitment edit invitations are rebuilt.
- The old venue request form / `venue/` module behavior is untouched.

### DIFF
--- a/openreview/workflows/edit_invitations.py
+++ b/openreview/workflows/edit_invitations.py
@@ -1880,7 +1880,7 @@ class EditInvitationsBuilder(object):
                 'content' :{
                     'overlap_committee_ids': {
                         'order': 1,
-                        'description': 'List of committee group ids that invited users can not be members of when accepting this invitation. Users that are already members of any of these groups will be asked to decline that role first. Leave this field empty to allow users to serve in multiple roles.',
+                        'description': 'List of committee group ids that invited users cannot be members of when accepting this invitation. Users that are already members of any of these groups will be asked to decline that role first. Leave this field empty to allow users to serve in multiple roles.',
                         'value': {
                             'param': {
                                 'type': 'string[]',
--- a/openreview/workflows/edit_invitations.py
+++ b/openreview/workflows/edit_invitations.py
@@ -1864,9 +1864,54 @@ class EditInvitationsBuilder(object):
             invitation.duedate = due_date
 
         self.save_invitation(invitation, replacement=False)
-        
+
+        invitation_id = super_invitation_id + '/Overlap_Committees'
+
+        invitation = Invitation(
+            id = invitation_id,
+            invitees = [venue_id],
+            signatures = [venue_id],
+            readers = [venue_id],
+            writers = [venue_id],
+            edit = {
+                'signatures': [venue_id],
+                'readers': [venue_id],
+                'writers': [venue_id],
+                'content' :{
+                    'overlap_committee_ids': {
+                        'order': 1,
+                        'description': 'List of committee group ids that invited users can not be members of when accepting this invitation. Users that are already members of any of these groups will be asked to decline that role first. Leave this field empty to allow users to serve in multiple roles.',
+                        'value': {
+                            'param': {
+                                'type': 'string[]',
+                                'input': 'text',
+                                'default': []
+                            }
+                        }
+                    }
+                },
+                'invitation': {
+                    'id': super_invitation_id,
+                    'content': {
+                        'overlap_committee_ids': {
+                            'value': '${4/content/overlap_committee_ids/value}'
+                        }
+                    },
+                    'signatures': [venue_id]
+                }
+            }
+        )
+
+        if process_file:
+            invitation.process = self.get_process_content(f'{process_file}')
+
+        if due_date:
+            invitation.duedate = due_date
+
+        self.save_invitation(invitation, replacement=False)
+
         return invitation
-    
+
     def set_edit_committee_recruitment_request_invitation(self, super_invitation_id, process_file=None, due_date=None):
 
         venue_id = self.venue_id
--- a/openreview/workflows/edit_invitations.py
+++ b/openreview/workflows/edit_invitations.py
@@ -1880,12 +1880,12 @@ class EditInvitationsBuilder(object):
                 'content' :{
                     'overlap_committee_ids': {
                         'order': 1,
-                        'description': 'List of committee group ids that invited users cannot be members of when accepting this invitation. Users that are already members of any of these groups will be asked to decline that role first. Leave this field empty to allow users to serve in multiple roles.',
+                        'description': 'List of committee group ids that invited users cannot be members of when accepting this invitation. Users that are already members of any of these groups will be asked to decline that role first. Delete this value to allow users to serve in multiple roles.',
                         'value': {
                             'param': {
                                 'type': 'string[]',
                                 'input': 'text',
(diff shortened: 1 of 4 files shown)


openreview/workflows/edit_invitations.py
```diff
@@ -1864,9 +1864,54 @@ def set_edit_committee_recruitment_invitation(self, super_invitation_id, process
...
+                                'input': 'text',
+                                'default': []
+                            }
+                        }
+                    }
+                },
+                'invitation': {
+                    'id': super_invitation_id,
+                    'content': {
+                        'overlap_committee_ids': {
```
melisabok, line 1896: @celestemartinez should we initialize this valid with AC group for reviewers and reviewers for AC group?

otherwise the PCs would need to remember to set this up.

openreview/workflows/process/committee_recruitment_response_pre_process.js
```diff
@@ -12,6 +12,21 @@ async function process(client, edit, invitation) {
...
+
+    for (const overlapCommitteeId of overlapCommitteeIds) {
+      const { groups: overlapGroups } = await client.getGroups({ id: overlapCommitteeId, member: user })
+      if (overlapGroups.length > 0) {
+        const overlapPrettyName = overlapGroups[0].content?.committee_pretty_name?.value ?? Tools.prettyId(overlapCommitteeId, true)
+        return Promise.reject(new OpenReviewError({ name: 'Error', message: `You have already accepted an invitation to serve as ${overlapPrettyName} for ${shortPhrase}. If you would like to change your decision and serve as ${committeePrettyName}, please decline the invitation to be ${overlapPrettyName} and then accept the invitation to be ${committeePrettyName}.` }))
+      }
+    }
+    return
+  }
```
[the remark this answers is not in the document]
melisabok: getGroups by member is checking all the memberships of that user, ignore this

openreview/workflows/edit_invitations.py
celestemartinez: yes, I think it makes sense to have this as the default.

celestemartinez:
I think it looks good. The only issue I found is PCs cannot delete this value once it has been added. Not sure what the best way to do this is. We can set the field as optional/deletable, but then the path is copied to the invitation content.
image

melisabok:
can they set an empty value? let me check

openreview/workflows/workflow_process/committee_group_template_process.py
```diff
@@ -142,6 +142,26 @@ def process(client, edit, invitation):
 
     content[f'{committee_role}_recruitment_id'] = { 'value': invitation_edit['invitation']['id'] }
 
+    ## By default, do not allow users to serve as both reviewers and area chairs
+    if committee_role == 'area_chairs' and 'reviewers_recruitment_id' in domain.content:
+        client.post_invitation_edit(
```
enrubio, line 147: I guess we would need to parameterize `area_chairs` here. Don't think we need to do the same for `reviewers_recruitment_id` since we don't change the domain fields based on custom roles.

melisabok: I believe this is the role type not the name.

celestemartinez (approved):
Looks good! We still need to support deleting the field from the UI, right?

openreview/workflows/workflow_process/committee_group_template_process.py
enrubio: I see, got it

enrubio, line 147: what about SACs? it's ok by default for a person to be an SAC and a reviewer?

melisabok: this is the default behavior, I don't think we have that in the old venue request. SACs can be reviewers.

enrubio (approved):
Changes look good!

PR #3027, opened 2026-04-17
Fix/ New Venue Request: more descriptive field names for role names

celestemartinez:
This PR updates the fields `reviewers_name` and `area_chairs_name` to the more descriptive field names `reviewer_role_name` and `area_chair_role_name`. It also adds a preprocess so we can check any instances of these two fields having the same string, which could cause problems after deployment. 

Fixes #3022

PR #3183, opened 2026-08-31
Fix/ Venue Request: KeyError on re-deployment when legacy role name fields are missing

melisabok:
## Problem

Re-deploying an already-deployed venue (Conference_Review_Workflow) fails in the deployment process function with:

```
KeyError: 'value'
  File "openreview/venue/venue.py", line 141, in set_main_settings
    self.reviewers_name = request_note.content['reviewers_name']['value']
```

## Root cause

After the first deployment, the process hides request form fields from PCs by posting a note edit that sets `readers` on each field — including the legacy `reviewers_name` and `area_chairs_name` fields. Those fields were removed from the request form in #3027 (renamed to `reviewer_groups_names` / `area_chair_groups_names`), but the hide edit was never updated.

When an edit sets only `readers` on a field the note doesn't have, the API creates a value-less stub field (`{'readers': [...]}` with no `value` key). On re-deployment, `set_main_settings` sees the stub, passes the `'reviewers_name' in request_note.content` check, and crashes reading `['value']`.

Existing re-deployment tests didn't catch this because they all set `reviewer_groups_names`, which takes the first branch of the guard and never reaches the legacy field.

## Changes

- **`conference_review_workflow_deployment.py`**: the hide-fields edit now only includes fields actually present on the request note, so it never creates stub fields. The legacy field names stay in the list because request notes created before #3027 still carry them with real values and should remain hidden. This also prevents stub creation for optional fields the PC left empty (`full_submission_deadline`, `senior_area_chair_groups_names`, etc.).
- **`venue.py` (`set_main_settings`)**: the `reviewers_name` / `area_chairs_name` guards now check for an actual value (`.get(field, {}).get('value')`) instead of key presence, falling back to the defaults (`Reviewers` / `Area_Chairs`). This is needed for venues already deployed in production, whose request notes already have the stub fields.

### DIFF
--- a/openreview/venue/venue.py
+++ b/openreview/venue/venue.py
@@ -137,7 +137,7 @@ class Venue(object):
         if request_note.content.get('reviewer_groups_names', {}).get('value'):
             self.reviewer_roles = request_note.content['reviewer_groups_names']['value']
             self.reviewers_name = self.reviewer_roles[0]
-        elif 'reviewers_name' in request_note.content:
+        elif request_note.content.get('reviewers_name', {}).get('value'):
             self.reviewers_name = request_note.content['reviewers_name']['value']
             self.reviewer_roles = [self.reviewers_name]
         preferred_email_groups = [self.get_reviewers_id(), self.get_authors_id()]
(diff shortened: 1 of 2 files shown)


celestemartinez (approved):

enrubio (approved):
These changes look good

PR #3185, opened 2026-09-01
Feat/Allow PCs to upload affinity scores via a job ID

celestemartinez:
Requires https://github.com/openreview/openreview-api/pull/1243

### Summary
- Adds a `job_id` option to `Matching.setup()` / `Venue.setup_committee_matching()` so program chairs can supply an existing expertise-API job ID instead of triggering a new affinity score computation. `Matching._upload_scores_from_job()` fetches the completed job's results via `get_expertise_results` (CSV) and metadata via `get_expertise_metadata`, then builds scores the same way as a freshly computed job.
- Extends the PC-facing invitation content (`edit_invitations.py`) with an `"I will upload my own affinity scores"` option on `affinity_score_model` and a new optional `expertise_job_id` field, plus an `"I will upload my own conflicts"` option on `conflict_policy`.
- Updates `compute_affinity_scores_process.py` and `compute_conflicts_process.py` to no-op (leaving the invitation active) when PCs choose the manual-upload options, instead of posting an error status to the request form.
- Simplifies the affinity-score process's exception handling to just record `str(e)` rather than special-casing a few known error messages.
- Adds a test in `tests/test_acs_and_reviewers.py` covering the manual affinity-score upload path: selecting "I will upload my own affinity scores", confirming the invitation stays active, and confirming no status note is posted to the request form.

### Test plan
- [x] `tests/test_acs_and_reviewers.py` extended with coverage for the manual-upload flow


openreview/workflows/process/compute_affinity_scores_process.py
```diff
@@ -19,8 +19,17 @@ def process(client, invitation):
 
     support_user = domain.content['request_form_invitation']['value'].split('/Venue_Request')[0]
 
-    affinity_scores_model = invitation.get_content_value('affinity_score_model')
-    if not affinity_scores_model:
+    affinity_scores_model = invitation.get_content_value('affinity_score_model', False)
+    print('affinity_scores_model', affinity_scores_model)
+    expertise_job_id = invitation.get_content_value('expertise_job_id', False)
+    print('expertise_job_id', expertise_job_id)
```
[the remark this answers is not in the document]
celestemartinez: Removed

openreview/venue/matching.py
```diff
@@ -640,6 +640,26 @@ def _compute_scores(self, score_invitation_id, submissions, model='specter2+scin
...
             raise openreview.OpenReviewException('There was an error connecting with the expertise API: ' + str(e))
 
+    def _upload_scores_from_job(self, job_id, score_invitation_id, submissions):
+        client = self.client
+        matching_status = {
+            'no_profiles': [],
+            'no_publications': []
+        }
+
+        result = client.get_expertise_results(job_id, wait_for_complete=True, format='csv')
```
celestemartinez, line 650: Two questions here:
1. should we have wait_for_complete or check somewhere first that the job has completed?
2. should I check somewhere that the job_id is valid?

melisabok: wait for complete is ok. if the job_id is not valid then get_expertise_results will fail.

Another option is to validate the job_id is complete when the user enters it.

celestemartinez: Added a preprocess that checks the job_id is valid and it has status complete.

openreview/venue/venue.py
```diff
@@ -1387,7 +1387,7 @@ def setup_all_committees_matching(self):
             venue_matching = matching.Matching(self, self.client.get_group(self.get_reviewers_id(name=reviewer_role)))
             venue_matching.setup()
 
-    def setup_committee_matching(self, committee_id=None, compute_affinity_scores=False, compute_conflicts=False, compute_conflicts_n_years=None, alternate_matching_group=None, submission_track=None):
+    def setup_committee_matching(self, committee_id=None, compute_affinity_scores=False, compute_conflicts=False, compute_conflicts_n_years=None, alternate_matching_group=None, submission_track=None, job_id=None):
```
melisabok: can job_id be in compute_affinity_scores param?

maybe it can a dict with a key 'job_id', different than 'model' and 'percentile'

or it could be a string different than filename, maybe with a prefix 'job_id:saldkjf2'

celestemartinez: Fixed, `job_id` is now passed as a dict in `compute_affinity_scores`.

melisabok (approved):

openreview/workflows/process/compute_affinity_scores_process.py
```diff
@@ -58,22 +64,19 @@ def process(client, invitation):
 
     matching_status = {}
 
+    if expertise_job_id:
+        affinity_scores_model = {
+            'job_id': expertise_job_id
+        }
```
enrubio, line 70: If a PC computes scores with a job ID first, then re-computes by choosing a model, it looks like this process would still try to upload the old job ID instead of recomputing scores? Since the `expertise_job_id` would stay in the invitation content.

Maybe we can check if the model is "I will upload my own affinity scores", but I wonder if it should go in a preprocess so the user can see the error right away.

openreview/workflows/edit_invitations.py
```diff
@@ -1080,10 +1080,11 @@ def set_edit_conflict_settings_invitation(self, super_invitation_id):
             edit = {
                 'content': {
                     'conflict_policy': {
+                        'description': 'Select the conflict policy to use, or select "I will upload my own conflicts" if you prefer to upload your own.',
                         'value': {
                             'param': {
                                     'type': 'string',
-                                    'enum': ['Default', 'NeurIPS'] ## TODO: Add the authors only policy
+                                    'enum': ['Default', 'NeurIPS', 'I will upload my own conflicts']
```
enrubio: We don't support the Comprehensive conflict policy for new UI venues?

openreview/workflows/process/compute_affinity_scores_process.py
celestemartinez: I think if they want to use a model then they would have to delete the job_id from the invitation. I don't want to save it in the invitation if it is not being used. I can add this to the preprocess.

openreview/workflows/edit_invitations.py
celestemartinez: I didn't realize Yoshitomo had added this to the venue request. I can add it here.

openreview/workflows/process/compute_affinity_scores_process.py
celestemartinez: Fixed.

openreview/workflows/edit_invitations.py
celestemartinez: Fixed.

enrubio (approved):
Looks good!

PR #3193, opened 2026-09-09
Make the submission process robust when the note has no `title`

melisabok:
### Problem

The submission process function crashed in production with:

```
TypeError: 'NoneType' object is not subscriptable
  File "...", line 31, in process_update
    author_subject = f'''{short_phrase} has received your submission titled {note.content['title']['value']}'''
```

`process_update` reads `note.content['title']['value']` unconditionally. Two situations break that assumption:

1. **All edits of a submission are deleted.** The API rebuilds the note from nothing, stamps a `ddate` on it and returns it without a `content` key, so `note.content` is `None` on the Python side. The process ran `process_update` for the deleted Submission edit and raised the `TypeError` above.
2. **Program chairs rename the title field.** The submission form can be edited to replace `title` with another field name. Posting a submission then raised `KeyError: 'title'` at the same line.

### Fix

`openreview/venue/process/submission_process.py`:

- Read the title with `.get()` from the note content, falling back to the previous edit's content (`existing_edit`) when the note has none, then to an empty string. The author ids fall back to the previous edit the same way, so the authors group keeps its members and the co-authors are still emailed when a submission is deleted. When no title is available the notification subjects drop the `titled ...` suffix (`<venue> has received your submission` / `<venue> has received a new submission`).
- When the note is deleted (`note.ddate`), update the paper/authors groups, remove the authors group from the venue Authors group, send the deletion notifications and return. The process no longer tries to create forum invitations, post the `Post_Submission` edit or run group invitations on a deleted note. Notes deleted through the dedicated `Deletion` invitation are not affected, that invitation runs its own process.
- The email sending code is extracted into a local `send_notifications()` helper shared by the deleted branch and the normal path.

The anonymous preprint process keeps its direct `content['title']` access on purpose: that invitation defines its own content and `title` is always present.

### Tests

`tests/test_venue_submission.py`:

- `test_delete_submission_edit`: posts a submission, deletes every edit of the note (the `Post_Submission` edit through the meta invitation, then the Submission edit last), and asserts the note comes back deleted with no content, both process runs succeed, the authors group keeps both authors while being detached from the venue Authors group, and the deleting author and the co-author each receive the deletion email.
- `test_submission_with_renamed_title_field`: replaces `title` with `paper_title` in the submission invitation, posts a submission, and asserts the process succeeds, the note has no `title` field, the author receives the "posted" email without a title, and the author groups are updated.

`tests/test_venue_submission_unified_authors.py` (new file) runs the same two scenarios against a venue configured with `unified_authors=True`, where the submission form uses the `authors` object instead of `authors`/`authorids`. It also checks the program chair notifications for both the deletion and the renamed title field.

Before the fix these tests reproduced `TypeError: 'NoneType' object is not subscriptable` and `KeyError: 'title'` respectively. After the fix `tests/test_venue_submission.py` passes in full (22 passed) and the new unified authors file passes (3 passed).


openreview/venue/process/submission_process.py
```diff
@@ -37,10 +44,53 @@ def process_update(client, edit, invitation, existing_edit):
...
+                client.post_message(
+                    invitation=meta_invitation_id,
+                    subject=author_subject,
+                    message=coauthor_message,
+                    recipients=[authors_group_id],
+                    ignoreRecipients=[edit.tauthor],
+                    replyTo=contact,
+                    signature=venue_id,
+                    sender=sender
+                )
```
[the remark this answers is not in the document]
melisabok: fixed

celestemartinez (approved):

openreview/venue/process/submission_process.py
```diff
@@ -70,13 +123,16 @@ def process_update(client, edit, invitation, existing_edit):
             writers=[venue_id],
             signatures=[venue_id],
             signatories=[venue_id, authors_group_id],
-            members=list(set(note.authorids)) ## always update authors
+            members=list(set(note_authorids)) ## always update authors
```
enrubio: Seems like it's possible now to update the author group with IDs from an old edit and then the wrong authors could get notified. Maybe we check if the authors are coming from the current note?

melisabok: it should be fixed.

enrubio (approved):
Changes look good!

PR #3195, opened 2026-09-09
Apply venue group content updates on redeployment

celestemartinez:
## Summary
- `create_venue_group()` no longer skips the content update wholesale for
  template-related workflows. Previously, once the venue group had any content,
  a redeployment left it untouched, so any configuration change made before
  redeploying never reached the group.
- That guard existed to stop the content diff from deleting keys written to the
  venue group outside `create_venue_group()`. It is no longer needed — no fields
  are added to the content after deployment, and every key written to the group
  elsewhere is either recomputed by `create_venue_group()` or explicitly carried
  forward from the existing content.
- Adds `tests/test_redeploy_add_area_chairs.py`: deploys a reviewers-only venue,
  enables area chairs on the request form, redeploys, and asserts the group
  content picks up the area chair configuration (`area_chairs_id`,
  `area_chairs_name`, and the updated `preferred_emails_groups`).
- `tests/test_reviewers_only.py`: asserts the group content is unchanged when a
  redeployment carries no configuration change.

## Test plan
- [x] `tests/test_redeploy_add_area_chairs.py`
- [x] `tests/test_reviewers_only.py`

### DIFF
--- a/openreview/venue/group.py
+++ b/openreview/venue/group.py
@@ -378,8 +378,6 @@ class GroupBuilder(object):
             content['status_invitation_id'] = { 'value': f'{self.venue.support_user}/Venue_Request/Conference_Review_Workflow/-/Status' }
 
         update_content = self.get_update_content(venue_group.content, content)
-        if self.venue.is_template_related_workflow() and venue_group.content:
-            update_content = False # avoid updating the content on every deployment for template related workflows
         if update_content:
 
             description = f'''Set up and customize the peer review process for your venue. As a program chair, you can define key aspects of the reviewing workflow, including reviewer assignment, review forms, conflict policies, deadlines, and decision criteria. [Learn more in the documentation](https://docs.openreview.net/new-venue-ui-beta/overview).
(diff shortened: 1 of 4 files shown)


melisabok (approved):

melisabok:
the constraint here is that we can not redeploy a venue after PCs made changes in the workflow, otherwise their changes will be overwritten. The ones related to the venue content: venue title, location, etc

celestemartinez:
> the constraint here is that we can not redeploy a venue after PCs made changes in the workflow, otherwise their changes will be overwritten. The ones related to the venue content: venue title, location, etc

Yes, this was happening for us before too and in the past we have just told the PCs this would happen. Not sure how to avoid it unless we read any value that already exists in the group content and keep it.

celestemartinez:
Just as a bit more context, I had added the check not to repost the group content for new UI venues because the committee group template would later on add new values to the content (for example `reviewers_declined_id`). On redeployment, these values were overwritten and not added again. Now that Melisa changed this so that these values are not saved in the content group anymore, then it is fine if content gets recomputed on redeployment.

enrubio (approved):
This looks good to me

PR #3198, opened 2026-09-10
Fix/ DBLP import - accept json instead of xml

xkopenreview:
this pr should change dblp import invitation to accept json instead of xml
this pr should also change post process of dblp import invitation to call updated util function


xkopenreview:
not required anymore

PR #3199, opened 2026-09-11
ARR: Update for October ARR Content

haroldrubio:
This PR mainly contains new wording for the ARR invitations with a few changes to the workflow outside of content fields:

- There is a profile link pre-process attached to each registration stage and the submitted authors form. This is to replace the current field which asks the user to input their ORCID/ACL Anthology and DBLP links - the ground truth will be the the OpenReview profile and we’ll just validate that it actually exists.
- A new field `paper_matching_feedback` is added to the workflow and should be private to only the chairs and reviewer who submitted the review
- A new flag `overall_level` for desk reject verification is added
- The track information content field in the registration forms is being changed to `indicate_your_research_areas` so the setup matching scripts have to be updated to read this field instead of the old field
- The checklist preprocess is once again to validate that there is a justification if any violation flag is raised
- The PC Console webfield also needs to read the new tracks field name


haroldrubio:
Closing in order to make a PR off of an isolated branch while Anna keeps updating that branch

PR #3203, opened 2026-09-15
Tests: Add test review reply tests for AC console UI

enrubio:
This PR adds tests for creating review replies for new UI venues. It will be used to determine how to display this info in the AC console.


tests/test_iclr_conference_with_templates.py
```diff
@@ -1802,6 +1802,124 @@ def test_review_release_stage(client, openreview_client, helpers):
...
+                    'review': { 'value': 'Excellent paper, accept'},
+                    'rating': { 'value': 8},
+                    'confidence': { 'value': 4},
+                }
+            )
+        )
+        helpers.await_queue_edit(openreview_client, edit_id=review_edit['id'])
+
+    venue = openreview.venue.helpers.get_venue(pc_client, 'ICLR.cc/2026/Conference', support_user='openreview.net/Support')
+    venue.review_stage = openreview.stages.ReviewStage(name='Official_Review')
```
enrubio, line 1848: @melisabok I noticed that in order to create review replies we need to set the `venue.review_stage` like this. Looks like it's because `get_venue` doesn't populate the `venue.review_stage` but `get_conference` did. PCs didn't need to do this before, I'm worried they'll be confused and think they're modifying their review stage settings. What do you think?

melisabok: this is because we don't syncronize the Venue class anymore, we only used it to deploy the venue. 

Is the Review Stage needed just to know the name of the invitation when using this param? "reply_to=openreview.stages.CustomStage.ReplyTo.REVIEWS,"

I think "reply_to" can be "Official_Review" directly and you don't need venue.review_stage being set,  could you confirm that?

tests/test_iclr_conference_with_templates.py
```diff
@@ -1802,6 +1802,124 @@ def test_review_release_stage(client, openreview_client, helpers):
...
+    reviews = openreview_client.get_notes(invitation='ICLR.cc/2026/Conference/Submission2/-/Official_Review')
+    assert len(reviews) == 3
+    review_ids = {review.id for review in reviews}
+
+    invitations = openreview_client.get_invitations(invitation='ICLR.cc/2026/Conference/-/AI_Review_Detection')
+    assert len(invitations) == 3
+
+    # post an AI review detection reply to each review
+    detection_data = [
+        { 'score': { 'value': 0.92 }, 'label': { 'value': 'AI' }, 'link': { 'value': 'https://dashboard.example.com/review-detection/1' } },
```
melisabok, line 1915: is this a link to an external site? do you know why?

enrubio: They want to link to a dashboard that has more in depth details on the review, e.g. it highlights which parts are AI generated.

They want to display the score, label, and the link in the AC console.

tests/test_iclr_conference_with_templates.py
enrubio: > Is the Review Stage needed just to know the name of the invitation when using this param? "reply_to=openreview.stages.CustomStage.ReplyTo.REVIEWS,"

Yes exactly.

> I think "reply_to" can be "Official_Review" directly and you don't need venue.review_stage being set, could you confirm that?

Got it, that makes sense. Let me update it, but yes it should work.

melisabok:
@enrubio @xkopenreview here is a screenshot that shows ALL the stage invitation replies, not just the first one (assuming there is one per submission)

Screenshot 2026-09-16 at 12 23 37 PM

melisabok:
Second option to show them next to the official review:

Screenshot 2026-09-16 at 12 43 58 PM

melisabok:
Format fixed:

Screenshot 2026-09-16 at 1 00 54 PM

enrubio:
That last option is if the AI detection info is in the review note right? @melisabok

melisabok:
> That last option is if the AI detection info is in the review note right?

No, it is still a reply to the review. The AC console detects that the custom stage reply is a reply to a review so it renders it below the review data.

melisabok:
it needs https://github.com/openreview/openreview-web/pull/2938

